PR #2, opened 2024-09-06
feat: add initial workspace creation

thedadams:


pkg/api/handlers/agent.go
```diff
@@ -91,7 +95,20 @@ func (a *AgentHandler) Create(ctx context.Context, req api.Request) error {
 		return api.NewErrBadRequest("replace requires \"id\" in the manifest to be set")
 	}
 
-	agent := v2.Agent{
+	var (
+		agent           v2.Agent
+		createWorkspace bool
+		httpError       *api.ErrHTTP
+	)
+	if err = req.Get(&agent, spec.Manifest.ID); errors.As(err, &httpError) && httpError.Code == http.StatusNotFound {
```
thedadams, line 103: It was easier to reason about creating the workspaces only when we need to if we get the agent first to see if it exists.

pkg/invoke/invoker.go
```diff
@@ -63,19 +70,28 @@ func (i *Invoker) getThread(ctx context.Context, agent *v2.Agent, input, threadN
 		}
 	}
 
+	workspaceID, err := i.workspaceClient.Create(ctx, i.workspaceProvider)
```
ibuildthecloud: this should be creating a workspace from the agent workspace, right?

pkg/storage/services/config.go
```diff
@@ -15,7 +15,8 @@ type Config struct {
 	StorageToken      string `usage:"Token for storage access, will be generated if not passed"`
 	//AuditLogPath       string `usage:"Location of where to store audit logs"`
 	//AuditLogPolicyFile string `usage:"Location of audit log policy file"`
-	DSN string `usage:"Database dsn in driver://connection_string format" default:"sqlite://file:otto.db?_journal=WAL&cache=shared&_busy_timeout=30000"`
+	DSN               string `usage:"Database dsn in driver://connection_string format" default:"sqlite://file:otto.db?_journal=WAL&cache=shared&_busy_timeout=30000"`
+	WorkspaceProvider string `usage:"Workspace provider" default:"directory"`
```
ibuildthecloud: No point in adding this config now as it really isn't usable.

pkg/invoke/invoker.go
thedadams: Yes, fixed.

pkg/storage/services/config.go
thedadams: Hardcoded for now.

ibuildthecloud (approved):

PR #90, opened 2024-09-27
Fix: Add proper onedrive sync status to api

StrongMonkey:
This PR adds proper syncing status from onedrive tool to api so that UI can display the status when it is syncing files.

image

### DIFF
--- a/pkg/controller/handlers/uploads/onedrive.go
+++ b/pkg/controller/handlers/uploads/onedrive.go
@@ -207,15 +207,13 @@ func (u *UploadHandler) HandleUploadRun(req router.Request, resp router.Response
 	} else if err != nil {
 		return err
 	}
-	if !run.Status.State.IsTerminal() {
-		var thread v1.Thread
-		if err := req.Get(&thread, oneDriveLinks.Namespace, oneDriveLinks.Status.ThreadName); apierrors.IsNotFound(err) {
-			// Might not be in the cache yet.
-			return nil
-		} else if err != nil {
-			return err
-		}
 
+	var thread v1.Thread
+	if err := req.Get(&thread, oneDriveLinks.Namespace, oneDriveLinks.Status.ThreadName); err != nil {
+		return err
+	}
+
+	if !run.Status.State.IsTerminal() {
 		file, err := u.workspaceClient.OpenFile(req.Ctx, thread.Spec.WorkspaceID, ".metadata.json")
 		if err != nil {
 			return err
@@ -230,11 +228,6 @@ func (u *UploadHandler) HandleUploadRun(req router.Request, resp router.Response
 		oneDriveLinks.Status.Error = output["output"].Error
 		resp.RetryAfter(5 * time.Second)
 		return nil
-	}
-
-	var thread v1.Thread
-	if err := req.Get(&thread, oneDriveLinks.Namespace, oneDriveLinks.Status.ThreadName); err != nil {
-		return err
 	}
 
 	ws, err := knowledgeWorkspaceFromParent(req, oneDriveLinks)


pkg/controller/handlers/uploads/onedrive.go
```diff
@@ -180,11 +182,51 @@ func (u *UploadHandler) RunUpload(req router.Request, _ router.Response) error {
...
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				file, err := u.workspaceClient.OpenFile(req.Ctx, thread.Spec.WorkspaceID, ".metadata.json")
+				if err != nil {
+					log.Errorf("failed to open metadata file: %v", err)
+					continue
+				}
+				defer file.Close()
```
thedadams: Using `defer` inside of a for-loop will leak resources until the for-loop is completely done.

pkg/controller/handlers/uploads/onedrive.go
```diff
@@ -180,11 +182,51 @@ func (u *UploadHandler) RunUpload(req router.Request, _ router.Response) error {
...
+
+				var output map[string]v1.OneDriveLinksConnectorStatus
+				if err = json.NewDecoder(file).Decode(&output); err != nil {
+					log.Errorf("failed to decode metadata file: %v", err)
+					continue
+				}
+				if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
thedadams: If this runs before the controllers finish, then this will mess things up. Specifically, we MUST make sure that `oneDriveLinks.Status.RunName` is set before trying to update the status here. Also, if `oneDriveLinks.Status.RunName` goes blank, then that means the run is done and we don't need to overwrite the status again. After you get the object, can you check this field and `return nil` if the RuName isn't set?

Lastly, we don't want to update the status of the OneDriveLinks object unnecessarily because that will cause the controllers to run again. Can you check that the `Status` and `Error` fields are set before overwriting them?

pkg/controller/handlers/uploads/onedrive.go
```diff
@@ -180,11 +182,57 @@ func (u *UploadHandler) RunUpload(req router.Request, _ router.Response) error {
...
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				file, err := u.workspaceClient.OpenFile(req.Ctx, thread.Spec.WorkspaceID, ".metadata.json")
+				if err != nil {
+					log.Errorf("failed to open metadata file: %v", err)
+					continue
+				}
+				func() {
```
thedadams: I wasn't expecting this approach, but I like it.

pkg/controller/handlers/uploads/onedrive.go
```diff
@@ -180,11 +182,57 @@ func (u *UploadHandler) RunUpload(req router.Request, _ router.Response) error {
...
 
+	ctx, cancel := context.WithCancel(context.Background())
+
 	go func() {
 		// Don't care about the events here, but we need to pull them out
 		r.Wait()
+		cancel()
 	}()
 
+	go func(ctx context.Context) {
```
thedadams: I realized something this morning (not sure why it popped in my head). The controller below essentially does the same thing except that it waits for the run status to be in a terminal state before it does it.

Instead of having this go-routine, you could incorporate this logic into the next handler. Read the metadata file and put the information on the status of the oneDriveLinks object, but stop there unless the run is in a terminal state. This would get rid of the weirdness of the conflicts, retries, and ensuring that the RunName is set.

Feel free to merge as-is, since I already approved. I can incorporate it later.

pkg/controller/handlers/uploads/onedrive.go
StrongMonkey: chatgpt all the way!

pkg/controller/handlers/uploads/onedrive.go
StrongMonkey: Yeah, this is a better way to handle that. Moved 👍

pkg/controller/handlers/uploads/onedrive.go
```diff
@@ -202,11 +203,34 @@ func (u *UploadHandler) HandleUploadRun(req router.Request, resp router.Response
...
 		return err
 	}
+	if !run.Status.State.IsTerminal() {
+		var thread v1.Thread
+		if err := req.Get(&thread, oneDriveLinks.Namespace, oneDriveLinks.Status.ThreadName); apierrors.IsNotFound(err) {
+			// Might not be in the cache yet.
+			return nil
+		} else if err != nil {
+			return err
+		}
```
thedadams: This is all duplicated below. At the very least, can we pull the getting of the thread out of this if-statement and remove the place below where we get the thread.

PR #163, opened 2024-10-10
Enhance: add more states to integration, control ingestion

StrongMonkey:
Adding a couple of new fields for integration we can show these in UI, more specifically, from will/s requirement that we need to show a list of files/pages from notion/website after syncing so that user can choose which one they want to exclude. We also need to add the ability to disable ingestion after syncing, so that we can have user select exclude files before ingestion to avoid ingesting files they don't want.


pkg/api/handlers/remoteknowledgesource.go
```diff
@@ -71,14 +72,17 @@ func updateRemoteKnowledgeSource(req api.Context, linksID, parentName string, pa
...
+	configChanged := checkConfigChanged(input, remoteKnowledgeSource)
 	remoteKnowledgeSource.Spec.RemoteKnowledgeSourceInput = input
 
 	if err := req.Update(&remoteKnowledgeSource); err != nil {
 		return fmt.Errorf("failed to update RemoteKnowledgeSource: %w", err)
 	}
 
-	if err := createSyncRequest(req, remoteKnowledgeSource); err != nil {
-		return fmt.Errorf("failed to create sync request: %w", err)
+	if configChanged {
```
thedadams, line 82: My understanding is that there is no harm in running the sync if nothing has changed. Just curious why we are adding this.

apiclient/types/remoteKnowledgeSource.go
```diff
@@ -28,6 +28,7 @@ type RemoteKnowledgeSourceManifest struct {
 type RemoteKnowledgeSourceList List[RemoteKnowledgeSource]
 
 type RemoteKnowledgeSourceInput struct {
+	DisableIngestion      bool                      `json:"disableIngestion,omitempty"`
```
thedadams: Just thought of something after I approved: can this be `DisableIngestionAfterSync`? Based on the code, it looks like ingestion can still happen when other files are added. Or when the user makes a call to manually ingest.

pkg/api/handlers/remoteknowledgesource.go
StrongMonkey: The problem is that when we delete files, we also add it to the exlcude list of remoteKnowledgeSource, but this would trigger a resync unnecessary.

PR #190, opened 2024-10-15
feat: create and list oauth-apps

ryanhopperlowe:
- Creates a List view for oauth-apps
- Adds a navigation item to the sidebar
- Allows users to create an oauth-app

### DIFF
--- a/ui/admin/app/lib/routers/apiRoutes.ts
+++ b/ui/admin/app/lib/routers/apiRoutes.ts
@@ -11,7 +11,10 @@ const buildUrl = (path: string, params?: object) => {
         ? queryString.stringify(params, { skipNull: true })
         : "";
 
-    if (process.env.NODE_ENV === "production" || import.meta.env.VITE_API_IN_BROWSER === "true") {
+    if (
+        process.env.NODE_ENV === "production" ||
+        import.meta.env.VITE_API_IN_BROWSER === "true"
+    ) {
         return {
             url: prodBaseUrl + path + (query ? "?" + query : ""),
             path,


ui/admin/app/components/oauth-apps/OAuthAppForm.tsx
```diff
@@ -0,0 +1,68 @@
...
+                {fields.map(({ key, label }) => (
+                    <ControlledInput
+                        key={key}
+                        name={key}
+                        label={label}
+                        control={form.control}
+                    />
+                ))}
+
+                <Button type="submit">Submit</Button>
```
tylerslaton, line 64:
```suggestion
                <Button variant="secondary" type="submit">Submit</Button>
```

I've been trying to make button have `secondary` be the default variant. What do you think?

ui/admin/app/lib/service/api/oauthAppService.ts
```diff
@@ -0,0 +1,84 @@
+import { OAuthApp, OAuthAppBase, OAuthAppSpec } from "~/lib/model/oauthApps";
+import { ApiRoutes } from "~/lib/routers/apiRoutes";
+
+import { request } from "./primitives";
+
+const getOauthApps = async () => {
+    const res = await request<{ items: OAuthApp[] }>({
+        url: ApiRoutes.oauthApps.getOauthApps().url,
```
tylerslaton, line 8: Can we add the `errorMessage` property for all of these?

```suggestion
        url: ApiRoutes.oauthApps.getOauthApps().url,
        errorMessage: "Failed to fetch OAuth apps"
```

tylerslaton: Applies to the whole file.

ui/admin/app/components/oauth-apps/OAuthAppForm.tsx
ryanhopperlowe: I'd be fine with it however

ryanhopperlowe: I think that should be handled in a separate PR though, since we're going to have to go through and make sure all buttons that don't specify a `variant` are properly updated

PR #209, opened 2024-10-16
feat: add auth support

tylerslaton:
This PR adds in support for authentication to the admin side of Otto. This includes a sign-in page, user model, me routes, and a AuthContext for rendering the authenticated user state.

This also includes a change to the API to change how the default routing was being handled. This was written by @thedadams.

https://github.com/user-attachments/assets/4366a48f-b30f-4b89-91e0-8dc60bdd4b4e

### DIFF
--- a/ui/admin/app/components/signin/SignIn.tsx
+++ b/ui/admin/app/components/signin/SignIn.tsx
@@ -30,7 +30,7 @@ const SignIn: React.FC<SignInProps> = ({ className }) => {
             </CardHeader>
             <CardFooter className="border-t pt-4">
                 <Button
-                    variant="destructive"
+                    variant="secondary"
                     className="w-full"
                     onClick={() => {
                         window.location.href = "/oauth2/start?rd=/admin/";
--- a/ui/admin/app/routes/_auth.agents.$agent.tsx
+++ b/ui/admin/app/routes/_auth.agents.$agent.tsx
@@ -43,6 +43,10 @@ export const clientLoader = async ({
     const { threadId, from } =
         parseQueryParams(request.url, paramSchema).data || {};
 
+    if (!agentId) {
+        throw redirect("/agents");
+    }
+
     // preload the agent
     const agent = await AgentService.getAgentById(agentId).catch(noop);
 
--- a/ui/admin/app/routes/_auth.thread.$id.tsx
+++ b/ui/admin/app/routes/_auth.thread.$id.tsx
@@ -33,6 +33,10 @@ import {
 export const clientLoader = async ({ params }: ClientLoaderFunctionArgs) => {
     const { id } = $params("/thread/:id", params);
 
+    if (!id) {
+        throw redirect("/threads");
+    }
+
     const thread = await ThreadsService.getThreadById(id);
     if (!thread) throw redirect("/threads");
 


ui/admin/app/components/auth/index.tsx
ryanhopperlowe, line 1: nit: can we either use this file or ditch it?

ui/admin/app/components/me/Me.tsx
```diff
@@ -0,0 +1,63 @@
...
+                    }}
+                >
+                    Sign Out
+                </Button>
+            </PopoverContent>
+        </Popover>
+    );
+};
+
+export default Me;
```
ryanhopperlowe: Can we stick with Named Exports and make this name a bit more descriptive? (i.e. `UserMenu` or something)

ui/admin/app/components/header/HeaderNav.tsx
```diff
@@ -66,6 +68,9 @@ export function HeaderNav() {
                     </div>
 
                     <div className="flex items-center justify-center p-4 mr-4">
+                        {me.username !== "nobody" && (
```
ryanhopperlowe: since `Me` is already pulling from `useAuth` you may as well move  this logic in there and make it completely self-encapsulated:

ui/admin/app/components/me/Me.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { useAuth } from "../auth/AuthContext";
+import { Button } from "../ui/button";
+
+interface MeProps {
+    className?: string;
+}
+
+const Me: React.FC<MeProps> = ({ className }) => {
+    const { me } = useAuth();
+
```
ryanhopperlowe: based on prev comment:
```ts
if (me.username === 'nobody') {
  return null;
}
```

ui/admin/app/lib/model/users.ts
```diff
@@ -0,0 +1,25 @@
...
+    createdAt: Date;
+    username: string;
+    email: string;
+    role: Role;
+};
+
+export enum Role {
+    Admin = 1,
+    Default = 10,
+}
```
ryanhopperlowe: Can you use a const enum here? It will behave exactly the same and is actually a lot more ergonomic to use

```ts
export const Role = {
  Admin: 1,
  Default: 2
} as const;
export type Role = (typeof Role)[keyof typeof Role];
```

ryanhopperlowe: This isn't that big of a deal for `string` based enums, but numeric enums do some really gross stuff after transpilation (and are technically not even typesafe), so fixing this is actually pretty important

For context, this is the output of a numeric enum, which has actually caused quite a few bugs for me in the past.

Screenshot 2024-10-16 at 12 45 48 PM
Screenshot 2024-10-16 at 12 46 37 PM

If you're interested in reading more, you can check this out :)

https://dev.to/ivanzm123/dont-use-enums-in-typescript-they-are-very-dangerous-57bh#:~:text=Excessive%20use%20of%20regular%20ENUMs,new%20values%20can%20be%20added.

ui/admin/app/components/auth/index.tsx
tylerslaton: This was a mistake! Removed.

ui/admin/app/components/me/Me.tsx
tylerslaton: Yep, this was just a boilerplate thing. Fixed.

ui/admin/app/components/header/HeaderNav.tsx
tylerslaton: Good point, done.

ui/admin/app/components/me/Me.tsx
tylerslaton: Done!

ui/admin/app/lib/model/users.ts
tylerslaton: Fixed.

PR #215, opened 2024-10-16
feat: add better linting targets and CI

tylerslaton:
This PR adds in a couple of new make targets:
- `make lint-api` - Lints the API
- `make lint-admin` - Lints the admin UI
- `make lint` - The previous two combined

Along with this I've split out the CI to run the `api` lint on Go changes and `admin` lint on changes to the admin UI. This helps linting locally and saves some time on PRs so everything is not linted together.


Makefile
```diff
@@ -23,4 +23,24 @@ dev: ui
 	@echo "Starting dev otto server and admin UI..."
 	./dev.sh
 
-.PHONY: ui build all clean dev
+# Lint the project
+lint: lint-admin lint-api
+
+lint-api:
+	go fmt ./... && \
+	go vet ./...
```
thedadams: This should probably be the `golangci-lint` that we have used in other places.

PR #221, opened 2024-10-17
feat: add support for listing users

tylerslaton:
I'll be adding a follow-up PR that will add in a delete operation for users that an admin can perform.

Screenshot 2024-10-17 at 11 09 04 AM

### DIFF
--- a/ui/admin/app/components/user/UserMenu.tsx
+++ b/ui/admin/app/components/user/UserMenu.tsx
@@ -4,15 +4,14 @@ import React from "react";
 import { roleToString } from "~/lib/model/users";
 import { cn } from "~/lib/utils";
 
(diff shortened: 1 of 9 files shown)


ui/admin/app/components/user/UserMenu.tsx
```diff
@@ -4,24 +4,23 @@ import React from "react";
...
-
 interface UserMenuProps {
     className?: string;
 }
 
 export const UserMenu: React.FC<UserMenuProps> = ({ className }) => {
     const { me } = useAuth();
 
-    if (me.username === "nobody") {
+    if (me.username === AuthDisabledUsername) {
```
ryanhopperlowe, line 23: instead of doing this, can we return an `isLoggedIn` from `useAuth`?

Issue #223, opened 2024-10-17
Chat responses not presented in the chat window until a hard refresh is done.


sangee2004:
**Steps to reproduce the problem:**
1. Create an agent.
2. Chat with the agent and ask a question.
Notice that after few seconds even the question disappears from the chat window 

**Note** - On hard refresh, the question and response show up.

PR #225, opened 2024-10-17
fix: remove id from event stream messages

tylerslaton:
This is a fix for issue #223. The event stream for messages included a new id entry that was not
expected by the client.

### DIFF
--- a/ui/admin/app/lib/stream.ts
+++ b/ui/admin/app/lib/stream.ts
@@ -39,7 +39,10 @@ export async function readStream<T>({
 
             // Process complete messages
             for (const message of messages) {
-                const dataString = message.replace(/^data: /, "").trim();
+                const dataString = message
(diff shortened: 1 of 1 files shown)


ui/admin/app/lib/stream.ts
```diff
@@ -39,7 +39,10 @@ export async function readStream<T>({
 
             // Process complete messages
             for (const message of messages) {
-                const dataString = message.replace(/^data: /, "").trim();
+                const dataString = message
+                    .replace(/^(.*)\n/, "")
```
thedadams: Let's be a little more careful here. And we don't need the capture.
```suggestion
                    .replace(/^id.*$/, "")
```

tylerslaton: Had to adjust this recommendation a bit but removed the capture.

PR #233, opened 2024-10-18
enhance: webhooks: add password authentication

g-linville:
This adds a new, optional authentication method for invoking webhooks. In addition to the existing HMAC validation, you can also configure a password that the webhook will need to send either in the header `X-Otto8-Webhook-Password` or in the query parameter `webhookPassword`.

We store the bcrypt hash of the password in the status of the webhook API object. The password itself never gets stored. The password can be updated with the `update` API call.

### DIFF
--- a/pkg/api/handlers/webhooks.go
+++ b/pkg/api/handlers/webhooks.go
@@ -56,25 +56,18 @@ func (a *WebhookHandler) Update(req api.Context) error {
 		return err
 	}
 
-	var (
-		hash []byte
-		err  error
-	)
 	if webhookReq.Password != "" {
-		hash, err = bcrypt.GenerateFromPassword([]byte(webhookReq.Password), bcrypt.DefaultCost)
+		hash, err := bcrypt.GenerateFromPassword([]byte(webhookReq.Password), bcrypt.DefaultCost)
 		if err != nil {
 			return fmt.Errorf("failed to hash password: %w", err)
 		}
+		wh.Spec.PasswordHash = hash
 		webhookReq.Password = ""
 	}
 
 	wh.Spec.WebhookManifest = webhookReq.WebhookManifest
 	for i, h := range wh.Spec.Headers {
 		wh.Spec.Headers[i] = textproto.CanonicalMIMEHeaderKey(h)
-	}
-
-	if hash != nil {
-		wh.Spec.PasswordHash = hash
 	}
 
 	if err := req.Update(&wh); err != nil {
@@ -107,18 +100,6 @@ func (a *WebhookHandler) Create(req api.Context) error {
 		return err
 	}
 
-	var (
-		hash []byte
-		err  error
-	)
-	if webhookReq.Password != "" {
-		hash, err = bcrypt.GenerateFromPassword([]byte(webhookReq.Password), bcrypt.DefaultCost)
-		if err != nil {
-			return fmt.Errorf("failed to hash password: %w", err)
-		}
-		webhookReq.Password = ""
-	}
-
 	wh := v1.Webhook{
 		ObjectMeta: metav1.ObjectMeta{
 			GenerateName: system.WebhookPrefix,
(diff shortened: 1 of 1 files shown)


pkg/api/handlers/webhooks.go
```diff
@@ -15,17 +15,28 @@ import (
...
 type WebhookHandler struct{}
 
 func NewWebhookHandler() *WebhookHandler {
 	return new(WebhookHandler)
 }
 
+type webhookRequest struct {
+	types.WebhookManifest `json:",inline"`
+	Password              string `json:"password"`
+}
```
g-linville, line 38: I designed it this way so that `Password` did not need to be a field on `types.WebhookManifest`. I didn't want to risk accidentally storing the password somehow.

pkg/storage/apis/otto.gptscript.ai/v1/webhook.go
```diff
@@ -49,6 +49,7 @@ type WebhookStatus struct {
 	Conditions                 []metav1.Condition          `json:"conditions,omitempty"`
 	External                   types.WebhookExternalStatus `json:"external,omitempty"`
 	LastSuccessfulRunCompleted *metav1.Time                `json:"lastSuccessfulRunCompleted,omitempty"`
+	PasswordHash               []byte                      `json:"passwordHash,omitempty"`
```
thedadams: Let's put this field on the spec. We can run into issues trying to update the status in the API if creating/updating causes controllers to kick off and also update the status.

pkg/api/handlers/webhooks.go
g-linville: Done

pkg/api/handlers/webhooks.go
```diff
@@ -54,6 +77,13 @@ func (a *WebhookHandler) Update(req api.Context) error {
 		return err
 	}
 
+	if hash != nil {
+		wh.Spec.PasswordHash = hash
+		if err := req.Update(&wh); err != nil {
+			return err
+		}
+	}
```
thedadams: Let's make it so there is only one update call in this code path.

pkg/api/handlers/webhooks.go
```diff
@@ -98,6 +140,13 @@ func (a *WebhookHandler) Create(req api.Context) error {
...
```
thedadams: Move this into the create call.

pkg/api/handlers/webhooks.go
g-linville: Fixed now. Sorry about that lol. I refactored this poorly when moving the hash to the spec

pkg/api/handlers/webhooks.go
g-linville: Fixed

pkg/api/handlers/webhooks.go
```diff
@@ -36,20 +47,36 @@ func (a *WebhookHandler) Update(req api.Context) error {
...
+
+	wh.Spec.WebhookManifest = webhookReq.WebhookManifest
 	for i, h := range wh.Spec.Headers {
 		wh.Spec.Headers[i] = textproto.CanonicalMIMEHeaderKey(h)
 	}
 
+	if hash != nil {
+		wh.Spec.PasswordHash = hash
+	}
+
```
thedadams:
```suggestion
	if webhookReq.Password != "" {
		hash, err := bcrypt.GenerateFromPassword([]byte(webhookReq.Password), bcrypt.DefaultCost)
		if err != nil {
			return fmt.Errorf("failed to hash password: %w", err)
		}
		
		wh.Spec.PasswordHash = hash
		webhookReq.Password = ""
	}

	wh.Spec.WebhookManifest = webhookReq.WebhookManifest
	for i, h := range wh.Spec.Headers {
		wh.Spec.Headers[i] = textproto.CanonicalMIMEHeaderKey(h)
	}
```

pkg/api/handlers/webhooks.go
```diff
@@ -71,29 +98,45 @@ func (a *WebhookHandler) Delete(req api.Context) error {
...
+	if err := req.Read(&webhookReq); err != nil {
 		return err
 	}
 
-	if err := validateManifest(req, manifest); err != nil {
+	if err := validateManifest(req, webhookReq.WebhookManifest); err != nil {
 		return err
 	}
 
+	var (
```
thedadams: Similar nit here.

PR #234, opened 2024-10-18
feat: add random initial names for agents

tylerslaton:
Agents will now get a randomly selected name by default when created. This is designed to alleviate the issues described in https://github.com/otto8-ai/otto8/issues/74.


ui/admin/app/lib/nameGenerator.ts
ryanhopperlowe, line 1: Can you move this file into `~/lib/service`?

tylerslaton: Done!

Issue #76, opened 2024-09-26
UI - Not able to present details of threads associated with deleted agents/workflows.


sangee2004:
UI - `0097d07`

**Steps to reproduce the problem:**
1. Create an agent.
2. Chat with this agent.
3. Delete the agent.
4. Try to view the threads related to the deleted agent.

We are presented with agent list view.
<img width="1395" alt="Screenshot 2024-09-26 at 12 15 38 PM" src="https://github.com/user-attachments/assets/f05b3f18-ba8f-46c0-9a79-512d727832da">

Agent look up done gets a 404 which is not handled well .

PR #235, opened 2024-10-18
fix: don't error when agent isn't found for thread inspect

tylerslaton:
This is a fix for issue #76. There will be a follow up that overhauls how errors are handled and displayed but this is a quick fix for the original issue.


ui/admin/app/routes/_auth.thread.$id.tsx
```diff
@@ -39,7 +40,7 @@ export const clientLoader = async ({ params }: ClientLoaderFunctionArgs) => {
     if (!thread) throw redirect("/threads");
 
     const agent = thread.agentID
-        ? await AgentService.getAgentById(thread.agentID)
+        ? await AgentService.getAgentById(thread.agentID).catch(noop)
```
ryanhopperlowe, line 43: Did you want to make the error boundary changes here, or in another PR?

tylerslaton: Another PR was my plan

PR #242, opened 2024-10-21
Ui/feat/overhaul-oauth-page

ryanhopperlowe:
This PR Overhauls the OAuth Apps page in a few ways:
- Switches to a tile view instead of a table view
- Users can only "modify" oauth app behavior
- Each OAuth app form will have it's own set of instructions on how to set them up in their respective applications and in Otto
- complete above steps for GitHub OAuth


ui/admin/app/lib/model/oauthApps/hubspot.ts
```diff
@@ -0,0 +1,11 @@
+// hubspot: {
```
tylerslaton: I think we should just go ahead and delete the files with the commented out code since we're not using them.

ui/admin/app/lib/model/oauthApps/oauth-helpers.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import { apiBaseUrl } from "~/lib/routers/apiRoutes";
+
+export const OAuthProvider = {
+    GitHub: "github",
+    // Google: "google",
+    // HubSpot: "hubspot",
+    // Microsoft365: "microsoft365",
+    // Notion: "notion",
+    // Slack: "slack",
```
tylerslaton: If we do delete the other files, we should delete this as well.

ui/admin/app/components/ui/tooltip.tsx
```diff
@@ -7,7 +7,7 @@ const TooltipProvider = TooltipPrimitive.Provider;
 
 // const Tooltip = TooltipPrimitive.Root;
 const Tooltip = ({
-    delayDuration = 0,
+    delayDuration = 200,
```
tylerslaton, line 10: Are you just wanting the default delay to go up?  This changes all tooltips in the app afaik.

ui/admin/app/components/composed/CopyText.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+        return () => clearTimeout(timeout);
+    }, [isCopied]);
+
+    return (
+        <div className={cn("flex items-center gap-2 w-full", className)}>
+            <TooltipProvider>
+                <Tooltip>
+                    <TooltipTrigger
+                        type="button"
```
tylerslaton, line 39: Could we make this be a button with the secondary color? Looks a little strange as black/white.

ui/admin/app/components/ui/tooltip.tsx
ryanhopperlowe: Yeah, I removed this a while back, but now I see the value in debouncing tooltips. I think adding a delay for all of them will be better

ui/admin/app/components/composed/CopyText.tsx
ryanhopperlowe: done!

ui/admin/app/lib/model/oauthApps/oauth-helpers.ts
ryanhopperlowe: done!

ui/admin/app/lib/model/oauthApps/hubspot.ts
ryanhopperlowe: yeah I was leaving it there for myself, but probably better to just put it in my notes

PR #283, opened 2024-10-23
Fix: Add error status handling for ingestion and sync source

StrongMonkey:
This PR addes the following things
1. It add knowledgeSetStatus to agent and propogate error from ingestion run.
2. Fixed a bug where ingestion is not running when it errors
3. UI components to render error status

https://github.com/otto8-ai/otto8/issues/186

### DIFF
--- a/pkg/controller/handlers/knowledge/knowledge.go
+++ b/pkg/controller/handlers/knowledge/knowledge.go
@@ -79,7 +79,7 @@ func (a *Handler) IngestKnowledge(req router.Request, resp router.Response) erro
 	}
 
 	// The status handler will clean this up
-	if ws.Status.IngestionRunName != "" {
+	if ws.Status.CurrentIngestionRunName != "" {
 		return nil
 	}
 
@@ -205,7 +205,7 @@ func (a *Handler) IngestKnowledge(req router.Request, resp router.Response) erro
 		}
 
 		ws.Status.IngestionRunHash = hash
-		ws.Status.IngestionRunName = run.Run.Name
+		ws.Status.CurrentIngestionRunName = run.Run.Name
 		ws.Status.IngestionGeneration++
 		return req.Client.Status().Update(req.Ctx, ws)
 	}
@@ -234,28 +234,19 @@ func toStream(events <-chan types.Progress) io.ReadCloser {
 func (a *Handler) UpdateFileStatus(req router.Request, _ router.Response) error {
 	ws := req.Object.(*v1.Workspace)
 
-	if ws.Status.IngestionRunName == "" {
+	if ws.Status.CurrentIngestionRunName == "" {
 		return nil
 	}
 
 	var run v1.Run
-	if err := req.Get(&run, ws.Namespace, ws.Status.IngestionRunName); apierrors.IsNotFound(err) {
-		if err := req.Get(uncached.Get(&run), ws.Namespace, ws.Status.IngestionRunName); apierrors.IsNotFound(err) {
+	if err := req.Get(&run, ws.Namespace, ws.Status.CurrentIngestionRunName); apierrors.IsNotFound(err) {
+		if err := req.Get(uncached.Get(&run), ws.Namespace, ws.Status.CurrentIngestionRunName); apierrors.IsNotFound(err) {
 			// Orphaned? User deleted the run? Solar flare?
-			ws.Status.IngestionRunName = ""
+			ws.Status.CurrentIngestionRunName = ""
 		}
 		return nil
 	} else if err != nil {
 		return err
-	}
-
-	if run.Status.State.IsTerminal() {
-		if err := updateIngestionError(req, ws, &run); err != nil {
-			return err
-		}
-
-		ws.Status.IngestionRunName = ""
-		return nil
 	}
 
 	_, progress, err := a.events.Watch(req.Ctx, ws.Namespace, events.WatchOptions{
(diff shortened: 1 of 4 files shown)


pkg/api/handlers/agent.go
```diff
@@ -96,10 +98,21 @@ func convertAgent(agent v1.Agent, prefix string) *types.Agent {
 		}
 		links = []string{"invoke", prefix + "/invoke/" + refName}
 	}
+
+	var knowledgeSetsStatus types.AgentKnowledgeSetStatus
+
```
thedadams: nit: remove line.

pkg/controller/handlers/agents/agents.go
```diff
@@ -27,6 +28,7 @@ func WorkspaceObjects(req router.Request, _ router.Response) error {
 	}
 
 	if len(agent.Status.KnowledgeSetNames) == 0 {
+		logrus.Info("Created default knowledge set")
```
thedadams: Can this be removed?

pkg/controller/handlers/knowledge/knowledge.go
```diff
@@ -248,19 +260,59 @@ func (a *Handler) UpdateFileStatus(req router.Request, _ router.Response) error
...
+				logger.Errorf("failed to update ingestion error: %s", err)
+				break
+			}
+
+			if stop {
+				break
+			}
+			time.Sleep(time.Second * 5)
+		}
+	}()
```
thedadams: The above call to `compileFileStatuses` will only return once the run is complete (either success or failure). This would mean that this coroutine would only run once.

Can it be removed?

pkg/controller/handlers/knowledge/knowledge.go
```diff
@@ -191,10 +191,22 @@ func (a *Handler) IngestKnowledge(req router.Request, resp router.Response) erro
...
+			if err := req.Client.Status().Update(req.Ctx, &knowledgeSet); err != nil {
+				return err
+			}
+		}
+
 		ws.Status.IngestionRunHash = hash
 		ws.Status.IngestionRunName = run.Run.Name
 		ws.Status.IngestionGeneration++
-		return req.Client.Status().Update(req.Ctx, ws)
+		return nil
```
thedadams: This should be removed. If the next handler fails for some reason, we want to make sure that the status here is saved.

pkg/controller/handlers/knowledge/knowledge.go
```diff
@@ -248,19 +260,59 @@ func (a *Handler) UpdateFileStatus(req router.Request, _ router.Response) error
...
+	return false, nil
+}
+
+func compileFileStatuses(ctx context.Context, client kclient.Client, ws *v1.Workspace, progress <-chan types.Progress) (map[string]types.Item, error) {
 	input := toStream(progress)
 	defer input.Close()
 	scanner := bufio.NewScanner(input)
 
-	final := map[string]string{}
+	final := map[string]types.Item{}
```
thedadams: Previously, this would indicate the status of the files when they didn't finish. Do we want to remove that information?

pkg/controller/handlers/threads/threads.go
```diff
@@ -60,7 +60,7 @@ func CreateWorkspaces(req router.Request, resp router.Response) error {
 
 func CreateKnowledgeSet(req router.Request, _ router.Response) error {
 	thread := req.Object.(*v1.Thread)
-	if len(thread.Status.KnowledgeSetNames) == 0 {
+	if len(thread.Status.KnowledgeSetNames) == 0 && thread.Labels[invoke.SystemThreadLabel] != "true" {
```
thedadams, line 63: Will system threads ever have `KnowledgeSetNames` set?

pkg/controller/handlers/knowledge/knowledge.go
StrongMonkey: The problem here is that even when run is finished, the run.status.error is not propagated from other controller(run controller) here so it needs to wait for a bit for the error to be propagated.

pkg/controller/handlers/knowledge/knowledge.go
StrongMonkey: Hmm ok, I forgot how baahh works. I thought the status will get updated regardless.

pkg/controller/handlers/knowledge/knowledge.go
StrongMonkey: I think only the key is used, not the value. So value can be simplified as struct{}

pkg/controller/handlers/threads/threads.go
StrongMonkey: Not for now I think. The problem here is that we are creating multiple empty knowledgeset, even for ingestion and sync file run.

pkg/controller/handlers/knowledge/knowledge.go
thedadams: If that's the case, then I think this should be a controller handler.

pkg/controller/handlers/knowledge/knowledge.go
thedadams: This is being put on the workspace status in the `NotFinished` field. Previously, the intention was that this be a map from files to the last known status. I think that extra bit of information is valuable so the we or the user can debug when necessary.

pkg/controller/handlers/knowledge/knowledge.go
StrongMonkey: we are wiping out `ws.Status.IngestionRunName` in the end, so other controller won't have information about the latest run. So it might be easy to just add logic to to retrieve the error.

pkg/controller/handlers/knowledge/knowledge.go
StrongMonkey: Ok, that's fine. I can add it back.

pkg/controller/handlers/knowledge/knowledge.go
thedadams: I pushed a commit with how I would understand this to work if we didn't use a goroutine. Let me know what you think about it. If it doesn't work like you expect, then we can just drop the commit I added.

pkg/controller/handlers/knowledge/knowledge.go
```diff
@@ -249,63 +249,44 @@ func (a *Handler) UpdateFileStatus(req router.Request, _ router.Response) error
 		return err
 	}
 
+	if run.Status.State.IsTerminal() {
```
StrongMonkey: @thedadams There is a chance that when run is not in terminal state and we are not immediately setting `ws.Status.IngestionRunName = ''`, the controller will reprocess the logs from the run events, which could be a lot of computes.

So I think the best way to address this without a goroutine is to add a new field `lastIngestionRunName` to keep track it the run and update error based on that.

Issue #292, opened 2024-10-23
UI - "TypeError: Cannot read properties of undefined (reading 'map')" seen when using "Add Tool" option in Agent. This error is also seen when listing tools.


sangee2004:
**Steps to reproduce the problem:**
1. Create an agent and add tools using `Add Tool` option.

User is presented with the following error message:
```
TypeError: Cannot read properties of undefined (reading 'map')
    at http://localhost:8080/admin/assets/_auth.agents._agent-BQSDvW1-.js:54:24781
    at Array.map (<anonymous>)
    at Ea (http://localhost:8080/admin/assets/_auth.agents._agent-BQSDvW1-.js:54:24669)
    at eu (http://localhost:8080/admin/assets/index-oz6ujnPe.js:22:16958)
    at ya (http://localhost:8080/admin/assets/index-oz6ujnPe.js:24:43694)
    at va (http://localhost:8080/admin/assets/index-oz6ujnPe.js:24:39499)
    at kf (http://localhost:8080/admin/assets/index-oz6ujnPe.js:24:39430)
    at Qr (http://localhost:8080/admin/assets/index-oz6ujnPe.js:24:39289)
    at xi (http://localhost:8080/admin/assets/index-oz6ujnPe.js:24:35710)
    at wo (http://localhost:8080/admin/assets/index-oz6ujnPe.js:24:36511)
```

I am able to get around the problem by doing a manual refresh after which using `Add Tool` option presents the tool list as expected.

PR #299, opened 2024-10-24
fix: use different key for tool reference getters

tylerslaton:
This is a workaround to make the SWR caching work nicely with the getToolReferences function. The reason we do this is because otherwise the SWR cache will not be invalidated between the two function calls and thus break type safety since types in TypeScript are erased at runtime. as-map=true does not exist in the API.

In addition, this PR also updates the ToolGrid to use the same getter as the ToolCatalog.

#292

### DIFF
--- a/ui/admin/app/lib/service/api/toolreferenceService.ts
+++ b/ui/admin/app/lib/service/api/toolreferenceService.ts
@@ -54,13 +54,8 @@ async function getToolReferencesCategoryMap(type?: ToolReferenceType) {
 }
 getToolReferencesCategoryMap.key = (type?: ToolReferenceType) =>
     ({
-        // tylerslaton: This is a workaround to make the SWR caching work
-        // nicely with the getToolReferences function. The reason we do this
-        // is because otherwise the SWR cache will not be invalidated between
-        // the two function calls and thus break type safety since types in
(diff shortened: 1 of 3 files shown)


ui/admin/app/lib/service/api/toolreferenceService.ts
```diff
@@ -54,7 +54,13 @@ async function getToolReferencesCategoryMap(type?: ToolReferenceType) {
...
 getToolReferencesCategoryMap.key = (type?: ToolReferenceType) =>
     ({
-        url: ApiRoutes.toolReferences.base({ type }).path,
+        // tylerslaton: This is a workaround to make the SWR caching work
+        // nicely with the getToolReferences function. The reason we do this
+        // is because otherwise the SWR cache will not be invalidated between
+        // the two function calls and thus break type safety since types in
+        // TypeScript are erased at runtime. as-map=true does not exist in
+        // the API.
+        url: ApiRoutes.toolReferences.base({ type }).path + "?as-map=true",
```
ryanhopperlowe: instead of modifying the url, can you just add another property (i.e. `responseType: 'map'` or `modifier: 'as-map'`)

tylerslaton: Did not even think about that haha, good call.

PR #313, opened 2024-10-24
Fix: fix ui to render error from data source properly

StrongMonkey:
https://github.com/otto8-ai/otto8/issues/186

Fixed an issue to render the error properly in ui

### DIFF
--- a/ui/admin/app/components/knowledge/RemoteKnowledgeSourceStatus.tsx
+++ b/ui/admin/app/components/knowledge/RemoteKnowledgeSourceStatus.tsx
@@ -30,9 +30,15 @@ const RemoteKnowledgeSourceStatus: React.FC<
                     />
                 )}
                 <span
-                    className={`text-sm mr-2 ${source?.error ? "text-destructive" : "text-gray-500"}`}
+                    className={`text-sm mr-2 ${
+                        !source.runID && source.error
+                            ? "text-destructive"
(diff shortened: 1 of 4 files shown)


ui/admin/app/components/knowledge/RemoteKnowledgeSourceStatus.tsx
```diff
@@ -30,9 +30,15 @@ const RemoteKnowledgeSourceStatus: React.FC<
                     />
                 )}
                 <span
-                    className={`text-sm mr-2 ${source?.error ? "text-destructive" : "text-gray-500"}`}
+                    className={`text-sm mr-2 ${
+                        !source.runID && source.error
+                            ? "text-destructive"
+                            : "text-gray-500"
+                    }`}
```
ryanhopperlowe, line 37: nit: we have a helper (`import { cn } from "~/lib/utils"`) that will clean up any duplicates in your styles + clean up the code a bit.

Also can you use a preset color from the `tailwind.config.ts` instead of `text-gray-500` (I'm guessing you want something like `text-muted` or `text-secondary-foreground` but use whichever color makes most sense to you)

```ts
                    className={cn(`text-sm mr-2`, {
                        "text-destructive": !source.runID && source.error,
                        "text-muted": source.runID || !source.error,
                    })}
```

PR #316, opened 2024-10-25
feat: custom messages for oauth prompts

ryanhopperlowe:
Screenshot 2024-10-24 at 7 48 33 PM

### DIFF
--- a/ui/admin/app/components/chat/Message.tsx
+++ b/ui/admin/app/components/chat/Message.tsx
@@ -78,7 +78,7 @@ export const Message = React.memo(({ message }: MessageProps) => {
                                 />
                             )}
 
-                            {message.prompt ? (
+                            {message.prompt?.metadata ? (
                                 <PromptMessage prompt={message.prompt} />
                             ) : (
                                 <Markdown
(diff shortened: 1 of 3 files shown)


ui/admin/app/components/chat/Message.tsx
```diff
@@ -124,3 +135,37 @@ export const Message = React.memo(({ message }: MessageProps) => {
...
+                    className="flex items-center gap-2 w-fit"
+                >
+                    <ToolIcon
+                        icon={prompt.metadata?.icon}
+                        category={prompt.metadata?.category}
+                        name={prompt.name}
+                        className="w-5 h-5 invert dark:invert-0"
+                        disableTooltip
+                    />
+                    Authenticate with {prompt.metadata?.category}
```
cjellick: I just realized: this UX relies on metadata.category and metadata.icon existing. That could potentially be an issue (though it should not be for the tools we have now).

Can you either address by having a fallback to a simple link (like darren's `Authentication is required <span class="underline">click here</span> to log-in using OAuth"`) or at least add a TODO to fix this in the future?

ryanhopperlowe: yeah definitely

ryanhopperlowe: Unfortunately that will mean we have to display the long, ugly url, because the `authUrl` property only exists in the metadata

PR #319, opened 2024-10-25
feat: add error boundary

tylerslaton:
This adds an error boundary for the _auth routes. By adding this, errors are handled in a more consistent manner in a way that is interactable for the user.

In addition this removes the sonners that would be shown when API errors occurred. If they are blocking errors, the user will now be directed to the error boundary. We'll want things to be more specific in the future, but this is a great start.

https://github.com/user-attachments/assets/880756a1-1fa4-4846-99ba-ac01e7a3df00

### DIFF
--- a/ui/admin/app/components/errors/RouteError.tsx
+++ b/ui/admin/app/components/errors/RouteError.tsx
@@ -1,0 +1,40 @@
+import { ErrorResponse } from "@remix-run/react";
+
+import { OttoLogo } from "~/components/branding/OttoLogo";
+import { Button } from "~/components/ui/button";
+import {
+    Card,
+    CardContent,
+    CardDescription,
+    CardFooter,
+    CardHeader,
+    CardTitle,
+} from "~/components/ui/card";
+
+export function RouteError({ error }: { error: ErrorResponse }) {
+    return (
+        <div className="flex min-h-screen w-full items-center justify-center p-4">
+            <Card className="w-96">
+                <CardHeader className="mx-4">
+                    <OttoLogo classNames={{ image: "h-10 w-10" }} />
+                </CardHeader>
+                <CardContent className="space-y-2 text-center border-b mb-4">
+                    <CardTitle>Oops! {error.status}</CardTitle>
+                    <CardDescription>{error.statusText}</CardDescription>
+                    <p className="text-sm text-muted-foreground">
+                        {error.data}
+                    </p>
+                </CardContent>
+                <CardFooter>
+                    <Button
+                        className="w-full"
+                        variant="secondary"
+                        onClick={() => window.location.reload()}
+                    >
+                        Try Again
+                    </Button>
+                </CardFooter>
+            </Card>
+        </div>
+    );
+}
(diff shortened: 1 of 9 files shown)


ui/admin/app/components/errors/RouteError.tsx
```diff
@@ -0,0 +1,40 @@
...
+                    <CardDescription>{error.statusText}</CardDescription>
+                    <p className="text-sm text-muted-foreground">
+                        {error.data}
+                    </p>
+                </CardContent>
+                <CardFooter>
+                    <Button
+                        className="w-full"
+                        variant="secondary"
+                        onClick={() => window.location.reload()}
```
ryanhopperlowe, line 32: we should be using the `useNavigate` hook to perform routing options. It will handle some other stuff that Remix needs to manage it's router

ui/admin/app/components/errors/Unauthorized.tsx
```diff
@@ -0,0 +1,40 @@
...
+                        in with an authorized account or contact your
+                        administrator.
+                    </CardDescription>
+                </CardContent>
+                <CardFooter>
+                    <Button
+                        className="w-full"
+                        variant="secondary"
+                        onClick={() => {
+                            window.location.href =
```
ryanhopperlowe, line 30: same here. The `useNavigate` hook is preferred over setting the window.location

ui/admin/app/routes/_auth.tsx
```diff
@@ -1,22 +1,32 @@
...
+                me.username &&
+                me.username !== AuthDisabledUsername
+            ) {
+                return <Unauthorized />;
+            }
+            return <SignIn />;
+        case isRouteErrorResponse(error):
+            return <RouteError error={error} />;
+        default:
+            return <Error error={error as Error} />;
```
ryanhopperlowe, line 27: I didn't know this was possible with JS switches 😮

ui/admin/app/components/signin/SignIn.tsx
```diff
@@ -17,31 +16,33 @@ interface SignInProps {
...
+                    <CardDescription className="text-center w-3/4 mx-auto pt-4">
+                        Please sign in using the button below.
+                    </CardDescription>
+                </CardHeader>
+                <CardFooter className="border-t pt-4">
+                    <Button
+                        variant="secondary"
+                        className="w-full"
+                        onClick={() => {
+                            window.location.href = "/oauth2/start?rd=/admin/";
```
ryanhopperlowe, line 38: same here

ui/admin/app/components/errors/Unauthorized.tsx
tylerslaton: This is intentional since the Remix router will route to this with `/admin` prefixed. We do not want this here.

ui/admin/app/routes/_auth.tsx
tylerslaton: Yeah classic switch(true)!

ui/admin/app/components/signin/SignIn.tsx
tylerslaton: Same as above ^ intentional since it won't work with navigate or link.

ui/admin/app/components/errors/RouteError.tsx
tylerslaton: How would you recommend using useNavigate here? I tried a couple of things but nothing worked quite like this. I Know it reloads the whole page but that was desired by me.

PR #328, opened 2024-10-25
feat: add button to a user's thread in table view

tylerslaton:
https://github.com/user-attachments/assets/654cf87b-836e-4a61-9169-d108be0481a8

### DIFF
--- a/ui/admin/app/routes/_auth.users.tsx
+++ b/ui/admin/app/routes/_auth.users.tsx
@@ -1,7 +1,9 @@
+import { Link } from "@remix-run/react";
 import { ColumnDef, createColumnHelper } from "@tanstack/react-table";
 import useSWR from "swr";
 
 import { User, roleToString } from "~/lib/model/users";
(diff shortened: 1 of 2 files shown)


ui/admin/app/routes/_auth.users.tsx
```diff
@@ -10,29 +13,71 @@ import { DataTable } from "~/components/composed/DataTable";
 
 export default function Users() {
     const getUsers = useSWR(UserService.getUsers.key(), UserService.getUsers);
-
     const users = getUsers.data || [];
 
+    // Fetch threads for all users
```
ryanhopperlowe: If there is a `/users` api that can fetch all users, it might make more sense to call 1 api for the users and 1 api for all threads. Is that possible?

tylerslaton: Ah yeah, that's true. I updated it to make a map of threadIds and then use those in the rows. What do you think?

PR #340, opened 2024-10-28
UI fix: Update apiBaseUrl to use http instead of https

ryanhopperlowe:
- this will need a better fix in the future
- I'm thinking Env variable from the backend

### DIFF
--- a/ui/admin/app/lib/routers/apiRoutes.ts
+++ b/ui/admin/app/lib/routers/apiRoutes.ts
@@ -4,7 +4,7 @@ import { mutate } from "swr";
 import { ToolReferenceType } from "~/lib/model/toolReferences";
 
 // todo: We need to have a discussion on using https in dev
-export const apiBaseUrl = "https://localhost:8080/api" as const;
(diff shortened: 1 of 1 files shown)


ui/admin/app/lib/routers/apiRoutes.ts
```diff
@@ -4,7 +4,7 @@ import { mutate } from "swr";
 import { ToolReferenceType } from "~/lib/model/toolReferences";
 
 // todo: We need to have a discussion on using https in dev
-export const apiBaseUrl = "https://localhost:8080/api" as const;
+export const apiBaseUrl = "http://localhost:8080/api" as const;
```
tylerslaton, line 7: Why is this only a change we need now, I wonder?

ryanhopperlowe: I accidentally merged this change when I was testing the slack OAuth

ryanhopperlowe: we do need to implement a way at some point to not hardcode this URL though, but that'll likely be a backend change

Issue #157, opened 2024-10-10
Not able to use otto credentials delete to delete gateway credential


sangee2004:
**Steps to reproduce the problem:**

1. Have `github.com/gptscript-ai/gateway` credentials created.
2. Try to delete this credential using otto credentials delete command
404 page not found error is seen
```
% otto credentials delete github.com/gptscript-ai/gateway
2024/10/10 10:27:01 error code 404 (Not Found): 404 page not found
 % otto credentials delete "github.com/gptscript-ai/gateway"
2024/10/10 10:27:09 error code 404 (Not Found): 404 page not found
% otto credentials                                         
NAME                              ENV                         EXPIRES
github.com/gptscript-ai/gateway   GPTSCRIPT_GATEWAY_API_KEY   10 hours from now
```

**Note** - This issue is seen only for gateway credential

PR #342, opened 2024-10-28
fix: query escape credential name to avoid path issues

drpebcak:
relates to [#157](https://github.com/otto8-ai/otto8/issues/157)

### DIFF
--- a/apiclient/credentials.go
+++ b/apiclient/credentials.go
@@ -3,6 +3,7 @@ package apiclient
 import (
 	"context"
 	"net/http"
+	"net/url"
 
 	"github.com/otto8-ai/otto8/apiclient/types"
 )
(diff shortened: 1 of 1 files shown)


apiclient/credentials.go
```diff
@@ -40,7 +41,7 @@ type DeleteCredentialsOptions struct {
 }
 
 func (c *Client) DeleteCredential(ctx context.Context, name string, opts DeleteCredentialsOptions) (err error) {
-	path := "/credentials/" + name
+	path := "/credentials/" + url.QueryEscape(name)
```
thedadams: This seems more like a `url.PathEscape`.

drpebcak: I thought about this after I did it, but was too lazy to look at what the exact difference was. I just pushed the change. It looks like there are some subtle differences that would make this worthwhile to do correctly for sure.

PR #411, opened 2024-11-01
enhance: add tool registry revision to version api

njhale:
Add the tool registry (index) and workspace provider revision to the version API.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,6 +42,11 @@ set -e
 mkdir -p /run/sshd
 /usr/sbin/sshd -D &
 mkdir -p /data/cache
+export OTTO_SERVER_VERSIONS="$(cat <<VERSIONS
+Tools: $(cd /otto8-tools && git rev-parse HEAD)
+WorkspaceProvider: $(cd /otto8-tools/workspace-provider && git rev-parse HEAD)
+VERSIONS
+)"
 exec tini -- otto8 server
 EOF
 
--- a/pkg/api/handlers/version.go
+++ b/pkg/api/handlers/version.go
@@ -1,10 +1,40 @@
 package handlers
 
 import (
+	"os"
+	"strings"
+
 	"github.com/otto8-ai/otto8/pkg/api"
 	"github.com/otto8-ai/otto8/pkg/version"
 )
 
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -42,6 +42,7 @@ set -e
 mkdir -p /run/sshd
 /usr/sbin/sshd -D &
 mkdir -p /data/cache
+export OTTO_SERVER_TOOL_REGISTRY_REVISION=$(cd /otto8-tools && git rev-parse HEAD)
```
njhale: did it this way since Docker's `ENV` doesn't support running commands

pkg/api/handlers/version.go
```diff
@@ -1,10 +1,56 @@
...
+		return "", err
+	}
+
+	for _, toolRef := range toolRefList.Items {
+		if strings.Contains(toolRef.Spec.Reference, toolRegistryURL) && len(toolRef.Spec.Revision) >= 7 {
+			return toolRef.Spec.Revision[:7], nil
+		}
+	}
+
+	return "", nil
```
njhale: On the fence about dropping this part. Kind of kludgy and not sure how much value it's adding other than for running this locally.

Dockerfile
drpebcak: would be nice to pass this in as a build arg instead. we should stop putting things like .git dirs into our docker images.

pkg/api/handlers/version.go
```diff
@@ -1,10 +1,40 @@
...
+			values[parts[0]] = parts[1]
+		}
+	}
+
+	// Populate the versionResponse struct using the map
+	return &versionResponse{
+		Otto:              version.Get().String(),
+		Tools:             strings.TrimSpace(values["Tools"]),
+		WorkspaceProvider: strings.TrimSpace(values["WorkspaceProvider"]),
+	}
```
njhale, line 39: I could just do this in an init function so it happens only once. LMK if you want me to change it.

pkg/api/handlers/version.go
njhale: Removed

PR #444, opened 2024-11-05
Ui/feat/add-refresh-flow-for-unauthorized-requests

ryanhopperlowe:


ui/admin/app/lib/service/api/primitives.ts
```diff
@@ -26,6 +35,32 @@ export async function request<T, R = AxiosResponse<T>, D = unknown>({
     } catch (error) {
         console.error(errorMessage);
 
+        if (isAxiosError(error) && error.response?.status === 401) {
```
tylerslaton, line 38: Oh nice I really like this approach.

ryanhopperlowe: Yeah main thing is being able to use `instanceof` 🙌

PR #448, opened 2024-11-05
UI chore: add typesafe parameter parsing

ryanhopperlowe:


ui/admin/app/lib/model/messages.ts
```diff
@@ -88,6 +88,7 @@ export const chatEventsToMessages = (events: ChatEvent[]) => {
                 text: content,
                 runId: runID,
             });
+            continue;
```
tylerslaton: What's rundown here, was this causing a bug or something?

ryanhopperlowe: Oh no good catch. Must have slipped through the cracks while rebasing

PR #453, opened 2024-11-06
Fix: Record run tasks for clean and load

cjellick:
If clean or load fails, we weren't recording the run ids.


pkg/controller/handlers/knowledgefile/knowledgefile.go
```diff
@@ -185,13 +188,16 @@ func (h *Handler) ingest(ctx context.Context, client kclient.Client, file *v1.Kn
 	}
 	defer loadTask.Close()
 
+	file.Status.RunNames = append(file.Status.RunNames, loadTask.Run.Name)
+	if err := client.Status().Update(ctx, file); err != nil {
+		return err
+	}
+
```
cjellick, line 195: im copying the paradigm used below for the ingestTask

PR #462, opened 2024-11-06
feat: add temporary rendering for workflows and their threads

tylerslaton:
We're adding in a temporary solution for rendering workflows and their threads.

The approach to rendering workflow threads will likely be around for quite some time, it renders it just like it would a chat interaction which makes sense for the architecture of workflows.

Rendering the manifest of the workflow itself, however, will very likely change in the immediate future. Doing this was annoyingly tricky, I had to convert the JSON from the API into YAML and then render that in a code editor with 90% of its functionality disabled. However, this will be a good stop-gap until we intentional render workflows.

### DIFF
--- a/ui/admin/app/lib/service/routeQueryParams.ts
+++ b/ui/admin/app/lib/service/routeQueryParams.ts
@@ -18,6 +18,8 @@ const QueryParamSchemaMap = {
         agentId: z.string().optional(),
         workflowId: z.string().optional(),
     }),
+    "/workflows": z.undefined(),
+    "/workflows/:workflow": z.undefined(),
     "/tools": z.undefined(),
     "/users": z.undefined(),
 } satisfies Record<keyof Routes, ZodSchema | null>;
@@ -55,6 +57,17 @@ function getUnknownQueryParams(pathname: string, search: string) {
             path: "/threads",
             query: parseSearchParams("/threads", search),
         } satisfies QueryParamInfo<"/threads">;
+    }
+
+    if (
+        new RegExp($path("/workflows/:workflow", { workflow: "(.*)" })).test(
+            pathname
+        )
+    ) {
+        return {
+            path: "/workflows/:workflow",
+            query: parseSearchParams("/workflows/:workflow", search),
+        } satisfies QueryParamInfo<"/workflows/:workflow">;
     }
 
     return {};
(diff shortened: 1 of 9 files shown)


ui/admin/app/routes/_auth.workflows.$workflow.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+export default function ChatAgent() {
+    const { isDark } = useTheme();
+    const { workflow } = useLoaderData<typeof clientLoader>();
+
+    return (
+        <div className="flex flex-col h-full">
+            <div className="flex-1">
+                <div className="h-full overflow-hidden">
+                    <Editor
```
tylerslaton, line 39: This whole editor is a bit heavy but its the best solution I could think of to render YAML.

ui/admin/package.json
```diff
@@ -41,6 +42,7 @@
         "cronstrue": "^2.51.0",
         "date-fns": "^4.1.0",
         "isbot": "^4",
+        "js-yaml": "^4.1.0",
```
tylerslaton, line 45: This is older than dirt, it hasn't been updated in 2 years but has 88 million weekly downloads. This is required, I couldn't think of another way to convert JSON to YAML without writing my own library for it.

ui/admin/app/routes/_auth.thread.$id.tsx
```diff
@@ -90,7 +96,7 @@ export default function ChatAgent() {
                             <ThreadMeta
                                 className="rounded-none border-none"
                                 thread={thread}
-                                agent={agent ?? ({} as Agent)}
+                                for={agent ?? workflow ?? ({} as Agent)}
                                 files={files}
```
ryanhopperlowe, line 105: instead of lying to typescript and potentially obscuring an error, let's throw the user to the error boundary

```ts
function getEntity() {
    if (agent) return agent;
    if (workflow) return workflow;
    throw new Error();
}
```

ui/admin/app/lib/service/routeQueryParams.ts
```diff
@@ -18,6 +18,8 @@ const QueryParamSchemaMap = {
         agentId: z.string().optional(),
         workflowId: z.string().optional(),
     }),
+    "/workflows": z.undefined(),
+    "/workflows/:workflow": z.undefined(),
```
ryanhopperlowe, line 22: can you also add a case for this in `getUnknownPathParams`?

ui/admin/package.json
ryanhopperlowe: tried to think of a fun joke regarding the xz backdoor... took half a second too long to load an idea in my head so I'm just gonna move on

ryanhopperlowe: All jokes aside. We discussed this offline and it will not be a problem for now

PR #470, opened 2024-11-07
Chore: Knowledge UX improvement

StrongMonkey:
A couple of knowledge UX improvement based on will/craig feedback, includes

1. Ingestion policy word tweaking
2. Show `Checking credential` modal when auth is being checked
3. Don't sort by state on default, add sort button for state
4. wrap error in separate modals.

https://github.com/otto8-ai/otto8/issues/464
https://github.com/otto8-ai/otto8/issues/463
https://github.com/otto8-ai/otto8/issues/451
https://github.com/otto8-ai/otto8/issues/442

### DIFF
--- a/ui/admin/app/components/knowledge/AddSourceModal.tsx
+++ b/ui/admin/app/components/knowledge/AddSourceModal.tsx
@@ -238,7 +238,8 @@ const AddSourceModal: FC<AddSourceModalProps> = ({
                                                 "0 * * * *",
                                                 "0 0 * * *",
                                                 "0 0 * * 0",
-                                            ].includes(syncSchedule ?? "")
+                                            ].includes(syncSchedule ?? "") &&
+                                            syncSchedule !== ""
                                         }
                                     />
                                 </DropdownMenuItem>
--- a/ui/admin/app/components/knowledge/AgentKnowledgePanel.tsx
+++ b/ui/admin/app/components/knowledge/AgentKnowledgePanel.tsx
@@ -78,7 +78,6 @@ export default function AgentKnowledgePanel({
     const [isEditKnowledgeSourceModalOpen, setIsEditKnowledgeSourceModalOpen] =
         useState(false);
 
-    const [errorDialogOpen, setErrorDialogOpen] = useState(false);
     const [errorDialogError, setErrorDialogError] = useState("");
 
     const getLocalFiles: SWRResponse<KnowledgeFile[], Error> = useSWR(
@@ -371,9 +370,6 @@ export default function AgentKnowledgePanel({
                                                             setErrorDialogError(
                                                                 file.error ?? ""
                                                             );
-                                                            setErrorDialogOpen(
-                                                                true
-                                                            );
                                                         }}
                                                     >
                                                         <EyeIcon className="w-4 h-4 text-destructive" />
@@ -517,8 +513,8 @@ export default function AgentKnowledgePanel({
             />
             <ErrorDialog
                 error={errorDialogError}
-                isOpen={errorDialogOpen}
-                onClose={() => setErrorDialogOpen(false)}
+                isOpen={errorDialogError !== ""}
+                onClose={() => setErrorDialogError("")}
             />
             {selectedKnowledgeSourceId && selectedKnowledgeSource && (
                 <KnowledgeSourceDetail
--- a/ui/admin/app/components/knowledge/KnowledgeSourceDetail.tsx
+++ b/ui/admin/app/components/knowledge/KnowledgeSourceDetail.tsx
@@ -95,7 +95,6 @@ const KnowledgeSourceDetail: FC<KnowledgeSourceDetailProps> = ({
     const [isCronDialogOpen, setIsCronDialogOpen] = useState(false);
     const [cronDescription, setCronDescription] = useState("");
 
-    const [errorDialogOpen, setErrorDialogOpen] = useState(false);
     const [errorDialogError, setErrorDialogError] = useState("");
     const [enableSorting, setEnableSorting] = useState(false);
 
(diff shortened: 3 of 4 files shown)


ui/admin/app/components/knowledge/AddSourceModal.tsx
```diff
@@ -200,21 +200,54 @@ const AddSourceModal: FC<AddSourceModalProps> = ({
...
+                                    Daily
+                                    <DropdownMenuCheckboxItem
+                                        checked={syncSchedule === "0 0 * * *"}
+                                    />
+                                </DropdownMenuItem>
+                                <DropdownMenuItem
+                                    className="cursor-pointer"
+                                    onClick={() => setSyncSchedule("0 0 * * 0")}
+                                >
+                                    Weekly
```
iwilltry42, line 225: I wonder if at least for weekly it should note that it's Sunday 00:00 - My expectation for any option would probably be "in this interval, starting **now**" :thinking:

ui/admin/app/components/knowledge/OAuthSignDialog.tsx
```diff
@@ -1,25 +1,48 @@
...
 
+    useEffect(() => {
+        if (
+            !authChecked &&
+            knowledgeSource.state === KnowledgeFileState.Pending
+        ) {
+            setIsAuthCheckingDialogOpen(true);
+        } else {
+            setIsAuthCheckingDialogOpen(false);
+        }
```
iwilltry42: Nit: Could save some lines :)
```suggestion
            setIsAuthCheckingDialogOpen(!authChecked && knowledgeSource.state === KnowledgeFileState.Pending);
```

pkg/api/handlers/agent.go
```diff
@@ -470,6 +470,11 @@ func (a *AgentHandler) EnsureCredentialForKnowledgeSource(req api.Context) error
 		return req.Write(convertAgent(agent, server.GetURLPrefix(req)))
 	}
 
+	// if auth is already authenticated, then don't continue.
+	if authStatus.Authenticated {
+		return req.Write(convertAgent(agent, server.GetURLPrefix(req)))
+	}
```
thedadams, line 476: What if I want to change the credentials? This would make it impossible? Also, if the user is already authenticated and the existing credentials are still valid, then running the code below should produce an "authenticated" result with no user interaction.

ui/admin/app/components/knowledge/AddSourceModal.tsx
```diff
@@ -200,21 +200,54 @@ const AddSourceModal: FC<AddSourceModalProps> = ({
                                         checked={syncSchedule === ""}
                                     />
                                 </DropdownMenuItem>
+                                <DropdownMenuItem
+                                    className="cursor-pointer"
+                                    onClick={() => setSyncSchedule("0 * * * *")}
```
thedadams, line 205: The cron package we are using also allows `@hourly`, `@daily`, `@weekly`, etc, if we want to use that.

ui/admin/app/components/knowledge/KnowledgeSourceDetail.tsx
```diff
@@ -394,24 +427,85 @@ const KnowledgeSourceDetail: FC<KnowledgeSourceDetailProps> = ({
                                             }
                                         />
                                     </DropdownMenuItem>
+                                    <DropdownMenuItem
+                                        className="cursor-pointer"
+                                        onClick={() => {
+                                            setSyncSchedule("0 * * * *");
```
thedadams, line 432: Same comment here about `@hourly`.

ui/admin/app/components/knowledge/AgentKnowledgePanel.tsx
```diff
@@ -471,6 +515,11 @@ export default function AgentKnowledgePanel({
                     setIsEditKnowledgeSourceModalOpen(true);
                 }}
             />
+            <ErrorDialog
+                error={errorDialogError}
+                isOpen={errorDialogOpen}
+                onClose={() => setErrorDialogOpen(false)}
+            />
```
ryanhopperlowe: nit:

ditch the extra `errorDialogOpen` state

```suggestion
            <ErrorDialog
                error={errorDialogError}
                isOpen={!!errorDialogError}
                onClose={() => setErrorDialogError("")}
            />
```

ui/admin/app/components/knowledge/KnowledgeSourceDetail.tsx
```diff
@@ -746,6 +941,12 @@ const KnowledgeSourceDetail: FC<KnowledgeSourceDetailProps> = ({
                     <OauthSignDialog
                         agentId={agentId}
                         sourceType={sourceType}
+                        knowledgeSource={knowledgeSource}
+                    />
+                    <ErrorDialog
+                        error={errorDialogError}
+                        isOpen={errorDialogOpen}
```
ryanhopperlowe: same nit here

pkg/api/handlers/agent.go
StrongMonkey: Haven't thought about changing credentials. This is just trying to make the second request faster if user is already autheticated. Right now it still produces like 10 seconds delay after user have autheticated, which is not ideal for frontend to check authetication status.

ui/admin/app/components/knowledge/AddSourceModal.tsx
StrongMonkey: Yeah we decided to just support client side for now. I will probably do add api support later.

ui/admin/app/components/knowledge/AddSourceModal.tsx
StrongMonkey: Yeah we discussed and it should just be `Sunday 00:00` but shows as weeklt. If you select this option it will show as `Sunday at 12 am`
image

ui/admin/app/components/knowledge/AddSourceModal.tsx
thedadams: What does "client-side cron" mean here?

StrongMonkey: It is just manually translating hourly, daily and weekly into cron syntax into cron syntax from client, instead have a real API support for doing hourly, daily and weekly sync.

Actually when I was typing I realized you were saying we can set `@hourly, @daily, @weekly` in client.

PR #474, opened 2024-11-07
fix: update workflow if and while apply logic

cloudnautique:
The apply logic was updated to mimic the old resp.Object() behavior The condition steps had checks and validations that assumed it was going to be created.

### DIFF
--- a/pkg/controller/handlers/workflowstep/while.go
+++ b/pkg/controller/handlers/workflowstep/while.go
@@ -18,13 +18,13 @@ func (h *Handler) RunWhile(req router.Request, _ router.Response) (err error) {
 	}
 
 	var completeResponse bool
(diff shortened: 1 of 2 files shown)


pkg/controller/handlers/workflowstep/while.go
```diff
@@ -103,7 +113,7 @@ func (h *Handler) RunWhile(req router.Request, _ router.Response) error {
```
thedadams, line 114: Based on the old logic, we need `completeResponse = true` here and also on line 77 (under the `if !conditionResult {`).

cloudnautique: added based on original resp.Object handling version.

PR #528, opened 2024-11-11
Fix: remove loading file section, fix dark mode

StrongMonkey:
- Drop `loading file` section
- Fix the knowledge details page in dark mode

### DIFF
--- a/ui/admin/app/components/knowledge/KnowledgeSourceDetail.tsx
+++ b/ui/admin/app/components/knowledge/KnowledgeSourceDetail.tsx
@@ -1,4 +1,3 @@
-import { DialogDescription } from "@radix-ui/react-dialog";
 import cronstrue from "cronstrue";
 import {
     ArrowDownUp,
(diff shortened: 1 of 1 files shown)


ryanhopperlowe (changes requested):
sorry to do this, but we can't use tailwind builtin colors. Please only use colors listed in `tailwind.config.js` and `tailwind.css`

ui/admin/app/components/knowledge/KnowledgeSourceDetail.tsx
```diff
@@ -313,14 +313,14 @@ const KnowledgeSourceDetail: FC<KnowledgeSourceDetailProps> = ({
 
     return (
         <Dialog open={isOpen} onOpenChange={onOpenChange}>
-            <DialogContent className=" h-[80vh] w-[80vw] max-w-none max-h-none">
+            <DialogContent className="h-[80vh] w-[80vw] max-w-none max-h-none dark:bg-gray-900">
                 <div
-                    className=" sticky top-0 bg-white z-10 overflow-y-auto mr-2"
+                    className="sticky top-0 bg-white dark:bg-gray-900 z-10 overflow-y-auto mr-2"
```
ryanhopperlowe: I'm having trouble locating this component in the UI , but my blind guess is you could replace this entire PR by removing `bg-white` from this line

StrongMonkey:
yeah @ryanhopperlowe you are right, I was using some inner div which set bd to white. I removed that and everything seems to work fine. Changed to remove the extra div.

ui/admin/app/components/knowledge/KnowledgeSourceDetail.tsx
```diff
@@ -313,760 +313,723 @@ const KnowledgeSourceDetail: FC<KnowledgeSourceDetailProps> = ({
 
     return (
         <Dialog open={isOpen} onOpenChange={onOpenChange}>
-            <DialogContent className=" h-[80vh] w-[80vw] max-w-none max-h-none">
-                <div
```
StrongMonkey, line 317: The only change is the div className moved to DialogContent.

ryanhopperlowe (approved):

ryanhopperlowe:
> yeah @ryanhopperlowe you are right, I was using some inner div which set bd to white. I removed that and everything seems to work fine. Changed to remove the extra div.

Sweet! thanks for fixing it!

ui/admin/app/components/knowledge/KnowledgeSourceDetail.tsx
StrongMonkey: yeah, this is side effect of using too much ai 😄

PR #549, opened 2024-11-12
Ui/feat/workflow-form

ryanhopperlowe:
Revamp workflow page to look like agents page.
- Info
- Tools
- Steps
- Parameters
Screenshot 2024-11-11 at 7 25 41 PM

Screenshot 2024-11-11 at 7 25 48 PM


thedadams:
Two things jump out:

1. I don't have the ability to create workflows in the UI. Seems like something we can add now that we are allowing editing of workflows.
2. The "workflow list" page still says "Inspect Workflow" and has a "page" icon. Seems like that should change to "Edit Workflow" and be the same as the "agent list" page.

Screenshot 2024-11-12 at 07 40 08

ryanhopperlowe:
> Two things jump out:
> 
> 1. I don't have the ability to create workflows in the UI. Seems like something we can add now that we are allowing editing of workflows.
> 2. The "workflow list" page still says "Inspect Workflow" and has a "page" icon. Seems like that should change to "Edit Workflow" and be the same as the "agent list" page.
> 
> Screenshot 2024-11-12 at 07 40 08

Both good call outs. I'll switch out the "inspect" button for an "edit" button for sure, but I'd like to add the "create" functionality in another PR. There are a few other things @cjellick asked me to do regarding workflows that may take higher priority.

cjellick:
@thedadams @ryanhopperlowe - Yeah this aligns with my asks. Some known gaps, but this will get us what we need for Kubecon

thedadams (approved):

PR #580, opened 2024-11-13
docs: remove toc from first tool guide and add unique ids to markdown headers

njhale:
- The default table of contents is incompatible with tabs, so remove it
  from the first-tool guide
- Auto-generated IDs for markdown headers in the first-tool page aren't
  unique, so generate unique IDs and inject them at snippet fetch time
- Re-fetch latest README.md files from `*-hash-tool` repos to fix
  guide doc formatting issues

### DIFF
--- a/docs/snippets/python-hash-tool-readme.md.mdx
+++ b/docs/snippets/python-hash-tool-readme.md.mdx
@@ -1,4 +1,4 @@
-## Writing your first tool in Python
+## Writing your first tool in Python {#writing-your-first-tool-in-python-28572835}
 
 [python-hash-tool](https://github.com/otto8-ai/python-hash-tool) contains a reference `Python` implementation of the `Hash` Tool.
 
@@ -12,7 +12,7 @@ git clone git@github.com:otto8-ai/python-hash-tool
 
 <br/>
 
-## Tool Repo Structure
+## Tool Repo Structure {#tool-repo-structure-28572835}
 
 The directory tree below highlights the files required to implement `Hash` in Python and package it for `Otto8`.
 
@@ -24,7 +24,7 @@ python-hash-tool
 ```
 <br/>
 
-## Defining the `Hash` Tool
+## Defining the `Hash` Tool {#defining-the-hash-tool-28572835}
 
 The `tool.gpt` file contains [GPTScript Tool Definitions](https://docs.gptscript.ai/tools/gpt-file-reference) which describe a set of Tools that can be used by Agents in `Otto8`.
 Every Tool repository must have a `tool.gpt` file in its root directory.
@@ -84,7 +84,7 @@ Param: algo: The algorithm to generate a hash with. Default is "sha256". Support
 
 <br/>
 
-## Tool Metadata
+## Tool Metadata {#tool-metadata-28572835}
 
 The `tool.gpt` file also provides the following metadata for use in `Otto8`:
 
@@ -144,7 +144,7 @@ https://cdn.jsdelivr.net/npm/@phosphor-icons/core@2/assets/duotone/fingerprint-d
 
 <br/>
 
-## Implementing Business Logic
+## Implementing Business Logic {#implementing-business-logic-28572835}
 
 The `hash.py` file executed by the `Tool Body` is the concrete implementation of the Tool's business logic.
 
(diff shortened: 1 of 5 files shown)


docs/snippets/python-hash-tool-readme.md.mdx
```diff
@@ -78,16 +82,20 @@ Param: algo: The algorithm to generate a hash with. Default is "sha256". Support
...
 The `tool.gpt` file also provides the following metadata for use in `Otto8`:
 
 - `!metadata:*:category` which tags Tools with the `Crypto` category to promote organization and discovery
 - `!metadata:*:icon` which assigns `https://cdn.jsdelivr.net/npm/@phosphor-icons/core@2/assets/duotone/fingerprint-duotone.svg` as the Tool icon
 
-Where `*` is a wild card pattern that applies the metadata to all Tools in a `tool.gpt`.
+<br/>
 
-```text
+> **Note:** `*` is a wild card pattern that applies the metadata to all Tools in a `tool.gpt`.
```
thedadams: There is a "callout" for things like this.
```suggestion
:::note

`*` is a wild card pattern that applies the metadata to all Tools in a `tool.gpt`.

:::
```

docs/snippets/python-hash-tool-readme.md.mdx
```diff
@@ -276,36 +296,70 @@ To do this, run through the following steps in the root of your local fork:
...
-   ![Enter Tool Repo Reference](https://raw.githubusercontent.com/otto8-ai/python-hash-tool/refs/heads/main/docs/add-tools-step-2.png "Enter Tool Repo Reference")
-
-Afterwords, the Tool will be available for use in `Otto8`.
-
-You can search for the Tool by category or name on the Tools page to verify:
-
-![Search For Newly Added Tools](https://raw.githubusercontent.com/otto8-ai/python-hash-tool/refs/heads/main/docs/add-tools-step-3.png "Search For Newly Added Tools")
-
-### Using The `Hash` Tool in an Agent
+1. <details>
```
thedadams: Curious why we need HTML instead of Markdown?

docs/src/plugins/fetch-snippets/index.ts
```diff
@@ -1,16 +1,15 @@
 import fs from 'fs';
 import path from 'path';
 import axios from 'axios';
+import crypto from 'crypto';
 import { LoadContext, Plugin } from '@docusaurus/types';
 
-// Array of permalinks to raw files in different repositories
 const FILE_URLS = [
-  'https://raw.githubusercontent.com/otto8-ai/python-hash-tool/main/README.md',
+  'https://raw.githubusercontent.com/njhale/python-hash-tool/main/README.md',
```
thedadams: Is this an expected change?

docs/src/plugins/fetch-snippets/index.ts
```diff
@@ -62,6 +64,25 @@ async function fetchFiles(outputDir: string) {
   );
 }
 
+// Function to add unique IDs to Markdown headers
+function addUniqueHeaderIds(url, markdown: string): string {
```
thedadams, line 68: Why is this necessary? Can we just add the IDs directly?

docs/src/plugins/fetch-snippets/index.ts
njhale: nope! this was leftover from testing, thank you for catching that.

docs/snippets/python-hash-tool-readme.md.mdx
njhale: HTML gives us better control the size and position of the images in this document. In this case I'm also using `<details>` to keep them hidden until the user clicks on the step, greatly reducing visual clutter.

If there's a better way to do this without HTML, I'm all ears.

docs/snippets/python-hash-tool-readme.md.mdx
njhale: Is this valid markdown, or is it `mdx`? These files are automatically fetched at build time from the example repos; in this case https://github.com/otto8-ai/python-hash-tool/blob/main/README.md, so it's nice to stick to GitHub compatible markdown if we can.

docs/src/plugins/fetch-snippets/index.ts
njhale: This makes it easier for folks to fetch and use Markdown from any repo together without having to make changes to those repos to make the IDs unique.

Also, the syntax to add these IDs is `mdx`, so any IDs would render in GitHub's markdown preview:

Screenshot 2024-11-14 at 1 17 22 PM

thedadams (approved):

PR #628, opened 2024-11-19
Feat/admin/webhooks

ryanhopperlowe:
Screenshot 2024-11-20 at 11 14 18 AM
Screenshot 2024-11-20 at 11 14 27 AM

### DIFF
--- a/ui/admin/app/components/webhooks/WebhookConfirmation.tsx
+++ b/ui/admin/app/components/webhooks/WebhookConfirmation.tsx
@@ -27,18 +27,19 @@ export type WebhookConfirmationProps = {
 
 export const WebhookConfirmation = ({
     webhook,
-    original,
+    original: _original,
     token,
     secret,
-    type: _ = "github",
-    tokenRemoved,
+    type: _type = "github",
+    tokenRemoved: _tokenRemoved,
     secretRemoved,
 }: WebhookConfirmationProps) => {
-    const showUrlChange =
-        !original ||
-        original.links?.invoke !== webhook.links?.invoke ||
-        !!token ||
-        tokenRemoved;
+    // note(ryanhopperlowe): commenting out as to not lose the logic here
+    // const showUrlChange =
+    //     !original ||
+    //     original.links?.invoke !== webhook.links?.invoke ||
+    //     !!token ||
+    //     tokenRemoved;
 
     return (
         <Dialog open>
@@ -56,22 +57,12 @@ export const WebhookConfirmation = ({
                     This information will not be shown again.
                 </DialogDescription>
 
-                <div
-                    className={cn("flex flex-col gap-1", {
-                        "flex-row gap-2": !showUrlChange,
-                    })}
-                >
+                <div className={cn("flex flex-col gap-1")}>
                     <TypographyP>Payload URL: </TypographyP>
-                    {showUrlChange ? (
-                        <CopyText
-                            text={getWebhookUrl(webhook, token)}
-                            className="min-w-fit"
-                        />
-                    ) : (
-                        <TypographyP className="text-muted-foreground">
-                            (Unchanged)
-                        </TypographyP>
-                    )}
+                    <CopyText
+                        text={getWebhookUrl(webhook, token)}
+                        className="min-w-fit"
+                    />
                 </div>
 
                 <div
--- a/ui/admin/app/components/webhooks/WebhookConfirmation.tsx
+++ b/ui/admin/app/components/webhooks/WebhookConfirmation.tsx
@@ -21,6 +21,8 @@ export type WebhookConfirmationProps = {
     token?: string;
     secret: string;
     type?: "github";
+    tokenRemoved: boolean;
+    secretRemoved: boolean;
 };
 
 export const WebhookConfirmation = ({
@@ -29,11 +31,14 @@ export const WebhookConfirmation = ({
     token,
     secret,
     type: _ = "github",
+    tokenRemoved,
+    secretRemoved,
 }: WebhookConfirmationProps) => {
     const showUrlChange =
         !original ||
         original.links?.invoke !== webhook.links?.invoke ||
-        !!token;
+        !!token ||
+        tokenRemoved;
 
     return (
         <Dialog open>
@@ -83,7 +88,7 @@ export const WebhookConfirmation = ({
                         />
                     ) : (
                         <TypographyP className="text-muted-foreground">
-                            (Unchanged)
+                            ({secretRemoved ? "None" : "Unchanged"})
                         </TypographyP>
                     )}
                 </div>
--- a/ui/admin/app/components/webhooks/WebhookConfirmation.tsx
+++ b/ui/admin/app/components/webhooks/WebhookConfirmation.tsx
@@ -1,6 +1,7 @@
 import { $path } from "remix-routes";
 
 import { Webhook } from "~/lib/model/webhooks";
+import { cn } from "~/lib/utils";
 
 import { TypographyP } from "~/components/Typography";
 import { CopyText } from "~/components/composed/CopyText";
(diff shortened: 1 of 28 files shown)


ryanhopperlowe:
Sorry for the long PR. I'll do a better job of splitting things up going forward

ui/admin/app/components/webhooks/WebhookConfirmation.tsx
```diff
@@ -0,0 +1,75 @@
...
+    return (
+        <Dialog open>
+            <DialogContent className="max-w-[700px]" hideCloseButton>
+                <DialogHeader>
+                    <DialogTitle>Webhook Saved</DialogTitle>
+                </DialogHeader>
+
+                <DialogDescription>
+                    Your webhook has been saved in Otto. Make sure to copy
+                    payload url and secret to your webhook provider.
```
g-linville: nit:

```suggestion
                    Your webhook has been saved in Otto8. Make sure to copy
                    the payload URL and secret to your webhook provider.
```

ui/admin/app/components/webhooks/WebhookForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+            <FormItem>
+                <FormLabel>Type</FormLabel>
+                <Select value="Github" disabled>
+                    <SelectTrigger>
+                        <SelectValue />
+                    </SelectTrigger>
+
+                    <SelectContent>
+                        <SelectItem value="Github">Github</SelectItem>
```
g-linville: nit:

```suggestion
                <Select value="GitHub" disabled>
                    <SelectTrigger>
                        <SelectValue />
                    </SelectTrigger>

                    <SelectContent>
                        <SelectItem value="GitHub">GitHub</SelectItem>
```

ui/admin/app/components/webhooks/WebhookForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+                label="Secret"
+                description="This secret should match the secret you provide to GitHub."
+                placeholder={hasSecret ? "(unchanged)" : ""}
+            />
+
+            <ControlledInput
+                control={form.control}
+                name="token"
+                label="Token (optional)"
+                description="Optionally provide a token to add an extra layer of security."
```
g-linville: I think we should either add more information here or link to docs (if they exist) to explain what this does. Whatever token is set will have to be provided by the webhook provider (i.e. GitHub) in a particular query param or header.

ui/admin/app/components/webhooks/WebhookFormContext.tsx
```diff
@@ -0,0 +1,142 @@
...
+        });
+    });
+
+    return (
+        <Form {...form}>
+            <Context.Provider
+                value={{
+                    error: updateWebhook.error || createWebhook.error,
+                    isEdit: !!webhookId,
+                    hasSecret: !!webhookId,
```
g-linville: I'm confused about how the webhook ID would be related to whether or not the webhook has a secret set...

ui/admin/app/components/workflow/steps/While.tsx
```diff
@@ -51,7 +51,7 @@ export function WhileComponent({
 
     return (
         <div className={cn("border rounded-md", className)}>
-            <ButtonDiv
+            <ClickableDiv
```
StrongMonkey, line 54: Oh cool, I should have used that

ui/admin/app/components/webhooks/WebhookForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+        return workflows.map((workflow) => (
+            <SelectItem key={workflow.id} value={workflow.id}>
+                {workflow.name}
+            </SelectItem>
+        ));
+    }
+}
+
+const GithubHeaderOptions = [
```
StrongMonkey, line 229: Do we allow any header? these headers seems to be specific to github

ui/admin/app/components/webhooks/WebhookForm.tsx
```diff
@@ -0,0 +1,179 @@
...
+    "X-GitHub-Hook-ID",
+    "X-GitHub-Event",
+    "X-GitHub-Delivery",
+    "X-Hub-Signature",
+    "X-Hub-Signature-Ed25519",
+    "User-Agent",
+    "X-GitHub-Hook-Installation-Target-Type",
+    "X-GitHub-Hook-Installation-Target-ID",
```
thedadams: Glad to see this isn't a dropdown and is just a list of suggestions. Can we remove the signature headers? They don't seem useful here and the user can type them in if they really want it.

thedadams (commented):
I noticed that the webhook create and edit pages take up the full-screen width. It looks a little weird on my larger monitor. Does it make sense to shrink the width down?

ui/admin/app/components/webhooks/WebhookFormContext.tsx
ryanhopperlowe: You can't create a webhook without a secret right? I was thinking that anytime a webhook exists in db (aka has an id), it must have a secret.

That being said it probably makes more sense to check for a secret hash instead

ui/admin/app/components/workflow/steps/While.tsx
ryanhopperlowe: Yeah I got annoyed that shadcn doesn't provide a camouflaged button

ui/admin/app/components/webhooks/WebhookForm.tsx
ryanhopperlowe: This form will be adapted in the future to switch between various providers. For now it's specifically for github

ui/admin/app/components/webhooks/WebhookForm.tsx
ryanhopperlowe: Ah, I mistakenly assumed the token would be provided in the returned `link` meta. I'll update this

ui/admin/app/components/webhooks/WebhookFormContext.tsx
thedadams: A webhook does not have to have a secret.

ui/admin/app/components/webhooks/WebhookForm.tsx
ryanhopperlowe: done

ui/admin/app/components/webhooks/WebhookFormContext.tsx
ryanhopperlowe: discussed with Donnie offline. This has been updated

ui/admin/app/components/webhooks/WebhookForm.tsx
ryanhopperlowe: I discussed this with Will and he decided to hide the token feature entirely until we can talk it out more with Bill/Donnie

ui/admin/app/components/webhooks/WebhookForm.tsx
ryanhopperlowe: done

ui/admin/app/components/webhooks/WebhookConfirmation.tsx
ryanhopperlowe: done

thedadams:
A few thoughts here:
1. Let's remove alias until we get feedback from Craig on it.
2. I am not able to click the "No Secret" checkbox. Also, not sure what it is used for. The user isn't required to enter a secret.
3. `*` shouldn't be the default value for headers. Maybe add some detail text to explain that `*` can be used to include all headers.

Screenshot 2024-11-22 at 06 39 10

njhale (approved):

thedadams (approved):

PR #639, opened 2024-11-20
Feat: Add file tree, add include and exclude prefix for auto ingestion

StrongMonkey:
This PR add file tree to show knowledge files. Demo is here https://www.loom.com/share/0e707405c1194ea1a9a1ffd595347a22 

It also adds a backend piece of functionality to include/exclude folder prefix so that user can add a folder in UI to automatically ingest files inside the folder.

### DIFF
--- a/apiclient/types/knowledgesource.go
+++ b/apiclient/types/knowledgesource.go
@@ -36,12 +36,16 @@ type KnowledgeSource struct {
 	LastSyncStartTime       *Time                `json:"lastSyncStartTime,omitempty"`
 	LastSyncEndTime         *Time                `json:"lastSyncEndTime,omitempty"`
 	LastRunID               string               `json:"lastRunID,omitempty"`
+	FilePathPrefixInclude   []string             `json:"filePathPrefixInclude,omitempty"`
+	FilePathPrefixExclude   []string             `json:"filePathPrefixExclude,omitempty"`
 }
 
 type KnowledgeSourceManifest struct {
-	SyncSchedule         string `json:"syncSchedule,omitempty"`
-	AutoApprove          *bool  `json:"autoApprove,omitempty"`
-	KnowledgeSourceInput `json:",inline"`
+	SyncSchedule          string   `json:"syncSchedule,omitempty"`
+	AutoApprove           *bool    `json:"autoApprove,omitempty"`
+	FilePathPrefixInclude []string `json:"filePathPrefixInclude,omitempty"`
+	FilePathPrefixExclude []string `json:"filePathPrefixExclude,omitempty"`
+	KnowledgeSourceInput  `json:",inline"`
 }
 
 type KnowledgeSourceList List[KnowledgeSource]
--- a/apiclient/types/knowledgesource.go
+++ b/apiclient/types/knowledgesource.go
@@ -36,8 +36,6 @@ type KnowledgeSource struct {
 	LastSyncStartTime       *Time                `json:"lastSyncStartTime,omitempty"`
 	LastSyncEndTime         *Time                `json:"lastSyncEndTime,omitempty"`
 	LastRunID               string               `json:"lastRunID,omitempty"`
-	FilePathPrefixInclude   []string             `json:"filePathPrefixInclude,omitempty"`
-	FilePathPrefixExclude   []string             `json:"filePathPrefixExclude,omitempty"`
 }
 
 type KnowledgeSourceManifest struct {
--- a/pkg/api/handlers/knowledgesource.go
+++ b/pkg/api/handlers/knowledgesource.go
@@ -25,8 +25,6 @@ func convertKnowledgeSource(agentName string, knowledgeSource v1.KnowledgeSource
 		LastSyncStartTime:       types.NewTime(knowledgeSource.Status.LastSyncStartTime.Time),
 		LastSyncEndTime:         types.NewTime(knowledgeSource.Status.LastSyncEndTime.Time),
 		LastRunID:               knowledgeSource.Status.RunName,
-		FilePathPrefixExclude:   knowledgeSource.Spec.Manifest.FilePathPrefixExclude,
-		FilePathPrefixInclude:   knowledgeSource.Spec.Manifest.FilePathPrefixInclude,
 	}
 }
 
--- a/pkg/api/handlers/knowledgesource.go
+++ b/pkg/api/handlers/knowledgesource.go
@@ -25,6 +25,8 @@ func convertKnowledgeSource(agentName string, knowledgeSource v1.KnowledgeSource
 		LastSyncStartTime:       types.NewTime(knowledgeSource.Status.LastSyncStartTime.Time),
 		LastSyncEndTime:         types.NewTime(knowledgeSource.Status.LastSyncEndTime.Time),
 		LastRunID:               knowledgeSource.Status.RunName,
+		FilePathPrefixExclude:   knowledgeSource.Spec.Manifest.FilePathPrefixExclude,
+		FilePathPrefixInclude:   knowledgeSource.Spec.Manifest.FilePathPrefixInclude,
 	}
 }
 
--- a/pkg/controller/handlers/knowledgefile/knowledgefile.go
+++ b/pkg/controller/handlers/knowledgefile/knowledgefile.go
@@ -101,14 +101,12 @@ func (h *Handler) IngestFile(req router.Request, _ router.Response) error {
 
 	// Check approval
 	if file.Spec.Approved == nil {
-		matchInclude := isFileMatchPrefixPattern(file.Spec.FileName, source.Spec.Manifest.FilePathPrefixInclude)
-		matchExclude := isFileMatchPrefixPattern(file.Spec.FileName, source.Spec.Manifest.FilePathPrefixExclude)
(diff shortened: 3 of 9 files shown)


pkg/controller/handlers/knowledgefile/knowledgefile.go
```diff
@@ -402,3 +416,23 @@ func (h *Handler) Cleanup(req router.Request, _ router.Response) error {
...
+
+func isFileMatchExcludePattern(filePath string, filePathPrefixExclude []string) bool {
+	for _, exclude := range filePathPrefixExclude {
+		if strings.HasPrefix(filePath, exclude) {
+			return true
+		}
+	}
+
+	return false
+}
```
njhale:
```suggestion
func hasAnyPrefix(filePath string,  prefixes []string) bool {
	for _, prefix := range prefixes {
		if strings.HasPrefix(filePath, prefix) {
			return true
		}
	}

	return false
}
```

ui/admin/app/components/knowledge/FileTree.tsx
```diff
@@ -0,0 +1,479 @@
...
+                                                    ...source,
+                                                    filePathPrefixExclude: [
+                                                        ...(source.filePathPrefixExclude ??
+                                                            []),
+                                                        file.fileName,
+                                                    ],
+                                                });
+                                            }
+                                        }}
+                                        className="justify-start items-center group ml-2 hidden group-hover:block"
```
ryanhopperlowe: use `visible`/`invisible` instead of `hidden`/`block`

This will prevent the row from resizing when it adds the icon button to the row

```suggestion
                                        className="justify-start items-center group ml-2 invisible group-hover:visible"
```

ui/admin/app/components/knowledge/FileTree.tsx
```diff
@@ -0,0 +1,479 @@
...
+            filePathPrefixInclude,
+            filePathPrefixExclude,
+        });
+    };
+
+    return (
+        <div className={cn("border-l", level > 0 && "ml-4")}>
+            <Collapsible open={isOpen} onOpenChange={setIsOpen}>
+                <CollapsibleTrigger asChild>
+                    <div className="flex flex-row p-2 hover:bg-accent hover:text-accent-foreground justify-between hover:cursor-pointer group">
```
ryanhopperlowe: can you remove `hover:cursor-pointer`? clicking the div doesn't actually do anything

pkg/controller/handlers/knowledgefile/knowledgefile.go
```diff
@@ -99,6 +99,30 @@ func (h *Handler) IngestFile(req router.Request, _ router.Response) error {
...
+			if err := req.Client.Update(req.Ctx, file); err != nil {
+				return err
+			}
+		} else if matchExclude {
+			file.Spec.Approved = typed.Pointer(false)
+			if err := req.Client.Update(req.Ctx, file); err != nil {
+				return err
+			}
+		}
+	}
```
njhale:
```suggestion
	if file.Spec.Approved == nil {
	    var approved *bool
	    
	    switch {
	    case source.Spec.Manifest.AutoApprove != nil && *source.Spec.Manifest.AutoApprove:
	        approved = typed.Pointer(true)
	    case matchExclude:
	        approved = typed.Pointer(false)
	    case matchInclude:
	        approved = typed.Pointer(true)
	    }
	    
		if approved != nil {
		    file.Spec.Approved = approved
		    if err := req.Client.Update(req.Ctx, file); err != nil {
		        return err
		    }
		}
	}
```

njhale (changes requested):

ui/admin/app/components/knowledge/FileTree.tsx
StrongMonkey: Actually it will open and close the folder tree that's why I added this.

ryanhopperlowe: I see, can you remove it when the row has no children then? I'm still seeing it for leaf nodes

StrongMonkey: Oh sure 👍

pkg/controller/handlers/knowledgefile/knowledgefile.go
```diff
@@ -99,6 +99,23 @@ func (h *Handler) IngestFile(req router.Request, _ router.Response) error {
...
+		case source.Spec.Manifest.AutoApprove != nil && *source.Spec.Manifest.AutoApprove:
+			file.Spec.Approved = typed.Pointer(true)
+		case matchInclude && !matchExclude:
+			file.Spec.Approved = typed.Pointer(true)
+		case matchExclude:
+			file.Spec.Approved = typed.Pointer(false)
+		}
+		if err := req.Client.Update(req.Ctx, file); err != nil {
+			return err
+		}
```
njhale: The code from before you addressed my comment didn't perform an update if the value of `file.Spec.Approved` was still `nil` after checking all the approve/deny conditions; in other words, it didn't update when "no decision" about approval was reached.

This updated code __always__ performs an update, even when there's no change to `file.Spec.Approved`.

Was this change in behavior intentional?

njhale (changes requested):

njhale (approved):

apiclient/types/knowledgesource.go
```diff
@@ -36,12 +36,16 @@ type KnowledgeSource struct {
 	LastSyncStartTime       *Time                `json:"lastSyncStartTime,omitempty"`
 	LastSyncEndTime         *Time                `json:"lastSyncEndTime,omitempty"`
 	LastRunID               string               `json:"lastRunID,omitempty"`
+	FilePathPrefixInclude   []string             `json:"filePathPrefixInclude,omitempty"`
+	FilePathPrefixExclude   []string             `json:"filePathPrefixExclude,omitempty"`
```
thedadams: Do we need this both here and `KnowledgeSourceManifest`? Seems like just in `KnowledgeSourceManifest` is enough.

pkg/api/handlers/knowledgesource.go
```diff
@@ -25,6 +25,8 @@ func convertKnowledgeSource(agentName string, knowledgeSource v1.KnowledgeSource
 		LastSyncStartTime:       types.NewTime(knowledgeSource.Status.LastSyncStartTime.Time),
 		LastSyncEndTime:         types.NewTime(knowledgeSource.Status.LastSyncEndTime.Time),
 		LastRunID:               knowledgeSource.Status.RunName,
+		FilePathPrefixExclude:   knowledgeSource.Spec.Manifest.FilePathPrefixExclude,
+		FilePathPrefixInclude:   knowledgeSource.Spec.Manifest.FilePathPrefixInclude,
```
thedadams: Can be removed if we use the fields from `KnowledgeSourceManifest`.

pkg/controller/handlers/knowledgefile/knowledgefile.go
```diff
@@ -99,6 +99,26 @@ func (h *Handler) IngestFile(req router.Request, _ router.Response) error {
...
+		matchInclude := isFileMatchPrefixPattern(file.Spec.FileName, source.Spec.Manifest.FilePathPrefixInclude)
+		matchExclude := isFileMatchPrefixPattern(file.Spec.FileName, source.Spec.Manifest.FilePathPrefixExclude)
+		switch {
+		case source.Spec.Manifest.AutoApprove != nil && *source.Spec.Manifest.AutoApprove:
+			file.Spec.Approved = typed.Pointer(true)
+		case matchExclude:
+			file.Spec.Approved = typed.Pointer(false)
+		case matchInclude:
+			file.Spec.Approved = typed.Pointer(true)
+		}
```
thedadams: nit: only do the processing that's necessary.
```suggestion
		switch {
		case source.Spec.Manifest.AutoApprove != nil && *source.Spec.Manifest.AutoApprove:
			file.Spec.Approved = typed.Pointer(true)
		case isFileMatchPrefixPattern(file.Spec.FileName, source.Spec.Manifest.FilePathPrefixInclude):
			file.Spec.Approved = typed.Pointer(false)
		case isFileMatchPrefixPattern(file.Spec.FileName, source.Spec.Manifest.FilePathPrefixInclude):
			file.Spec.Approved = typed.Pointer(true)
		}
```

thedadams (changes requested):

ryanhopperlowe (approved):

ryanhopperlowe:
Screenshot 2024-11-21 at 9 34 30 AM

Noticing that the add button is getting cutoff when the file name is too long. I think we need to truncate the text and add a tooltip to it

ryanhopperlowe (changes requested):
discovered above issue

pkg/api/handlers/knowledgesource.go
StrongMonkey: yeah, right. I will drop that

ryanhopperlowe:
> @ryanhopperlowe good catch, should be fixed 🙏 image

Beautiful 🙌

ryanhopperlowe (approved):

iwilltry42 (approved):

thedadams (approved):

PR #658, opened 2024-11-21
Chore: Add pprof to otto server

StrongMonkey:
Going to add pprof to be able to troubleshoot cpu usage in testing env.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	_ "net/http/pprof"
+	"os"
 
 	"github.com/otto8-ai/otto8/logger"
 	"github.com/otto8-ai/otto8/pkg/api/router"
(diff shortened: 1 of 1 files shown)


pkg/server/server.go
```diff
@@ -15,6 +16,11 @@ import (
 var log = logger.Package()
 
 func Run(ctx context.Context, c services.Config) error {
+	go func() {
```
drpebcak: Do we want this to be always on? now that we have a released version, we probably want this off by default so that people aren't accidentally exposing debug info to the world.

drpebcak (changes requested):

pkg/server/server.go
```diff
@@ -15,6 +16,11 @@ import (
...
+		log.Infof("Starting pprof on :6060")
+		log.Errorf(http.ListenAndServe("localhost:6060", nil).Error())
```
njhale: pico-nit: The chance of one of us configuring otto8 wrong is basically zero, but if we wanted to be really pedantic, we could check `c.HTTPListenPort` and return an error if it conflicts with the pprof port

njhale (approved):

pkg/server/server.go
njhale: I'm a fan of leaving it on in production, because that's where we're most likely to run into issues that will require profiling info to resolve.

Is there's some other way to ensure we don't accidentally expose the port publicly?

drpebcak: I don't have an issue with us having it on in production per se, but we shouldn't expose it for everybody who might run a copy of otto themselves.

StrongMonkey: it is fine, I can only run it behind an env variable 👍

StrongMonkey:
This is already enabled through /debug/pprof, dropping.

PR #661, opened 2024-11-22
change: use cleanup handler for runstate instead of finalizer on a run

iwilltry42:
Finalizers can be a PITA and leave over dangling refs.
We've also seen RunStates hanging around while their Runs were already gone.
This change switches RunStates over to the same cleanup logic we use for most other objects.


thedadams (commented):
We need to make sure that existing runs have this finalizer removed.

thedadams (approved):

StrongMonkey (approved):

PR #677, opened 2024-11-26
Chore: setup postgres and vector in dockerfile

StrongMonkey:
This PR should configure postgres and pgvector into otto dockerfile. It will run and use the built-in postgres process if no OTTO8_SERVER_DSN is set.

I ran some test against knowledge ingestion and it works fine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,25 @@ RUN --mount=type=cache,id=pnpm,target=/root/.local/share/pnpm/store \
     --mount=type=cache,target=/root/go/pkg/mod \
     UV_LINK_MODE=copy BIN_DIR=/bin make package-tools
 
-FROM cgr.dev/chainguard/wolfi-base AS final
+FROM cgr.dev/chainguard/postgres:latest-dev AS build-pgvector
+RUN apk add build-base git postgresql-dev
+RUN git clone --branch v0.8.0 https://github.com/pgvector/pgvector.git && \
+    cd pgvector && \
+    make clean && \
+    make OPTFLAGS="" && \
+    make install && \
+    cd .. && \
+    rm -rf pgvector
+
+FROM cgr.dev/chainguard/postgres:latest-dev AS final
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -22,35 +22,33 @@ RUN --mount=type=cache,id=pnpm,target=/root/.local/share/pnpm/store \
     UV_LINK_MODE=copy BIN_DIR=/bin make package-tools
 
 FROM cgr.dev/chainguard/wolfi-base AS final
+
+# Setup postgres and postgres-vector, although in production setup we don't really need standalone postgres
+RUN apk add postgresql build-base git postgresql-dev
+RUN git clone --branch v0.8.0 https://github.com/pgvector/pgvector.git && \
```
cjellick, line 26: darren mentioned this fleetingly when the feasibility of this was discussed. i should have passed it along sooner. he said you could just get it via from, like
```
FROM pgvector/pgvector:pg17
```
or whatever. i think that would be a little cleaner and more official feeling?

StrongMonkey: yep, I did detour to avoid dropping wolfi-base image we already uses.

thedadams (approved):

StrongMonkey:
Did the following change based on feedback
1. Move the pgvector build stage into a separate step. And copy the necessary artifacts to the final image after it is done.
2. Use `cgr.dev/chainguard/postgres:latest-dev` as the final image. Noticed that we can't use `cgr.dev/chainguard/postgres:latest` because it is a `distroless` image that doesn't come with `apk`
3. Use `/data/postgresql` as data store for postgresql.

iwilltry42 (approved):

cjellick:
Looks good. I'd like a rereview by @thedadams or a review by @ibuildthecloud

ibuildthecloud (approved):
Assuming it works...

StrongMonkey:
@ibuildthecloud come on where is the trust :)

PR #686, opened 2024-11-26
docs: add knowledge agent tutorial

g-linville:


ryanhopperlowe:
Do you think it would be helpful to let the user know that changes save automatically and they do not need to look for a submit/save button?

g-linville:
> Do you think it would be helpful to let the user know that changes save automatically and they do not need to look for a submit/save button? (Or even start a new thread for that matter)

@ryanhopperlowe good call. I just pushed a commit to add this.

ryanhopperlowe (approved):

PR #693, opened 2024-11-26
fix: Publish URL showing based on aliasAssigned and entityID check

ivyjeong13:
This addresses issue 634.
* checks against entityID instead of id.
* updates AgentContext to return agent from getAgent.data or default to the agent supplied from clientLoader.
* updates agentUpdates with appropriate aliasAssigned when agent is updated.

### DIFF
--- a/ui/admin/app/components/agent/Agent.tsx
+++ b/ui/admin/app/components/agent/Agent.tsx
@@ -28,12 +28,16 @@ export function Agent({ className, onRefresh }: AgentProps) {
     const [agentUpdates, setAgentUpdates] = useState(agent);
 
     useEffect(() => {
-        if (agent.id === agentUpdates.id) {
-            setAgentUpdates({
-                ...agentUpdates,
-                aliasAssigned: agent.aliasAssigned ?? false,
-            });
-        }
+        setAgentUpdates((prev) => {
+            if (agent.id === prev.id) {
+                return {
+                    ...prev,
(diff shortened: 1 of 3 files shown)


ui/admin/app/components/agent/Agent.tsx
```diff
@@ -27,6 +27,15 @@ export function Agent({ className, onRefresh }: AgentProps) {
...
     const [agentUpdates, setAgentUpdates] = useState(agent);
 
+    useEffect(() => {
+        if (agent.id === agentUpdates.id) {
+            setAgentUpdates({
+                ...agentUpdates,
+                aliasAssigned: agent.aliasAssigned ?? false,
+            });
+        }
+    }, [agent]);
```
ivyjeong13, line 41: Oops, missed agentUpdates for dependency, will address.

ryanhopperlowe: can you do this to get rid of the linter warning here?

```suggestion

    useEffect(() => {
        if (agent.id === agentUpdates.id) {
            setAgentUpdates((prev) => ({
                ...prev,
                aliasAssigned: agent.aliasAssigned ?? false,
            }));
        }
    }, [agent]);
```

ivyjeong13: Yes! Made the change

ryanhopperlowe (approved):

Issue #613, opened 2024-11-15
No way to figure out which model provider I am picking when a model is selected from model drop down when creating agent.


sangee2004:
**Steps to reproduce the problem:**
1. Create  a model provider as `Azure OpenAI` with target model as `gpt4-o`

Model List:
<img width="1134" alt="Screenshot 2024-11-15 at 3 11 14 PM" src="https://github.com/user-attachments/assets/d2095033-c875-4303-a866-83abe47ca783">

2. Create an agent and open the Model drop down

Notice that there are 2 entries for `gpt4-o` and no way to tell what the model provider is for these 2 entries

<img width="687" alt="Screenshot 2024-11-15 at 3 07 31 PM" src="https://github.com/user-attachments/assets/29d37b06-7e7f-4985-afdc-7f2288d4e796">

**Expected Behavior:**
Should be able to show the model provider along with model in the drop down.

thedadams:
@sangee2004 Thanks for opening this issue. I noticed this as well, and planned to bring it up with Ryan this week.

ryanhopperlowe:
Assigning this to @ivyjeong13 because it will be a pretty easy fix and will help her get acclimated to the codebase.

@ivyjeong13 The object that is iterated over to create this list **should** already have a `provider` property on it, so once you locate the dropdown it should be a simple fix.

ivyjeong13:
@ryanhopperlowe @thedadams Hey hey, quick question for the provider! 

I see `modelProvider` but wondering if it should match with what's displayed on the `Models` page? ex. `Azure OpenAI` instead of `azure-openai-model-provider`. 

![Image](https://github.com/user-attachments/assets/537517a8-c51a-403a-9ead-f2a9bec08271)

sangee2004:
This issue is addressed when testing with latest build.

![Image](https://github.com/user-attachments/assets/3d6d2ef5-1021-46dd-b404-9313feb1c69e)

PR #699, opened 2024-11-26
fix: show modelProvider in agent's model dropdown

ivyjeong13:
Quick fix to display `modelProvider` in the model dropdown when selecting a model for an agent.

Note: Added quick question about the `modelProvider` in the related issue: #613 

Screenshot 2024-11-26 at 4 27 44 PM


ryanhopperlowe:
do you think you could make the provider text use the `text-muted-foreground` class? Might make it a bit easier to read

ivyjeong13:
> do you think you could make the provider text use the `text-muted-foreground` class? Might make it a bit less busy

Yeah of course, lemme add that in

ivyjeong13:
Screenshot 2024-11-26 at 5 07 45 PM

ryanhopperlowe (approved):
love it

PR #700, opened 2024-11-26
fix: use ToolIcon instead of img with dark:invert for tool tooltip

ivyjeong13:
Uses ToolIcon instead of img with constant dark:invert for the Tool Tooltip.


ryanhopperlowe (approved):
Nice job finding the existing component 🙌

g-linville (approved):

PR #710, opened 2024-11-27
fix: minor ToolCard styling change

ivyjeong13:
* minor update to avoid chip being cut off and icon being shrunk as browser is resized
* instead of margins, removed and used gap-2 so if the chip / "Bundle" is moved to a new line, its left side is still aligned with the above line

Before:

https://github.com/user-attachments/assets/2ca84f83-2b0b-47c6-8108-81d3f470cbab

After:

https://github.com/user-attachments/assets/379393bf-9a51-4078-97a9-8a49902c6df9


ryanhopperlowe (approved):
🙌  Nice find! Love removing 1-off margins/padding

PR #724, opened 2024-12-02
fix: give website-cleaner 3 attempts to succeed before exiting (often remote server errors)

iwilltry42:
I feel like this is a quite "hacky" implementation since at this level we don't have control over the actual model calls - we're just handing over to gptscript which is relying on the chat-completion-client sdk.
For real control over statuscode- or http-client-error-based reasoning, we'd need to move the logic into the website-cleaner tool, which right now is just a simple GPTScript without any code.


iwilltry42:
Closing this as we're going to rewrite the website-cleaning tool to be code-based and handle the retries itself.

PR #737, opened 2024-12-03
chore: update publish with aliasAssigned api change

ivyjeong13:
Synced with Donnie and the changes made to the api for aliasAssigned were made:

* `undefined` means the alias is still resolving
*  `true` is alias has been assigned
* `false` is alias has not been assigned (whether published or not)

It looks like `aliasAssigned` is properly updated on the revalidation call so doesn't look like polling is needed currently.

### DIFF
--- a/ui/admin/app/components/agent/AgentPublishStatus.tsx
+++ b/ui/admin/app/components/agent/AgentPublishStatus.tsx
@@ -15,13 +15,11 @@ import { LoadingSpinner } from "~/components/ui/LoadingSpinner";
 
 type AgentPublishStatusProps = {
     agent: Agent;
-    isUpdating: boolean;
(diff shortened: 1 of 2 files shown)


ui/admin/app/components/agent/AgentPublishStatus.tsx
```diff
@@ -88,7 +86,8 @@ export function AgentPublishStatus({
             );
         }
 
-        if (isUpdating)
+        // if aliasAssigned is undefined, it is still resolving
+        if (typeof agent.aliasAssigned === "undefined")
```
ryanhopperlowe: nit: 
```suggestion
        if (agent.aliasAssigned === undefined)
```

will work too :)

Issue #502, opened 2024-11-09
User is not provided with any error message when register tool fails 


sangee2004:
**Steps to reproduce the problem:**

1. Use` Register New tool` option to register a new tool.

Tool registration fails due to https://github.com/otto8-ai/otto8/issues/501

In this case api fails with `[ToolReference.otto.otto8.ai](http://toolreference.otto.otto8.ai/) "github-com-sangee2004-gptscript-mytool" not found  error.` 

This error message is not presented to the user and the `Register New tool` pop page is still presented.

**Expected Behavior:**
User should be presented with this error message.

ryanhopperlowe:
assigning @ivyjeong13 because it will provide a chance to work with how I setup error handling

ivyjeong13:
I think we'll need to update the backend POST call to wait for if the reference has been successfully loaded:

When creating a tool:
POST Request Payload: /api/tool-references
```
{
    "reference": "github.com/ivyjeong13/gptscript/mytool",
    "toolType": "tool"
}
```

POST Response Payload: /api/tool-references: 200 Call
```
{
    "id": "github-com-ivyjeong13-gptscript-mytool-1",
    "created": "2024-11-27T15:03:54-05:00",
    "type": "toolreference",
    "name": "github-com-ivyjeong13-gptscript-mytool-1",
    "toolType": "tool",
    "reference": "github.com/ivyjeong13/gptscript/mytool",
    "active": true
}
```

Note: `error` is missing from the response.

When I reload the page, I see the tool has been updated with 'error' in GET /api/tool-references?type=tool
```
{
    "id": "github-com-ivyjeong13-gptscript-mytool",
    "created": "2024-11-27T15:01:23-05:00",
    "type": "toolreference",
    "name": "github-com-ivyjeong13-gptscript-mytool",
    "toolType": "tool",
    "reference": "github.com/ivyjeong13/gptscript/mytool",
    "active": true,
    "error": "run encountered an error: run encountered an error: status code 500 with error output: failed to load program: failed to get GitHub commit of ivyjeong13/gptscript at HEAD (fallback error exit status 128: remote: Repository not found.\nfatal: repository 'https://github.com/ivyjeong13/gptscript.git/' not found\n): 404 Not Found {\"message\":\"Not Found\",\"documentation_url\":\"https://docs.github.com/rest/commits/commits#get-a-commit\",\"status\":\"404\"}"
}
```

sangee2004:
[...]

PR #738, opened 2024-12-03
fix: display error if tool creation failed

ivyjeong13:
Addresses #502 

After syncing with Donnie, `resolved` field was added to the ToolReference. Similarly to `aliasAssigned`:
* `undefined`, the creation of tool is still resolving
*  `true/false`, tool has or has not been resolved

Once resolved, `error` should be supplied.

Needed to implement polling to fetch the ToolReference until it is resolved. 

Once resolved, error is shown in ErrorDialog, but lemme know if there is a different preferred error handling!

https://github.com/user-attachments/assets/2601e797-568e-45bc-9e8e-1737263000b9

(Tempted to see if I can make it scroll to the created tool once it's completed so user immediately sees their uploaded tool or maybe sort by recently created so it immediately shows at the top?)

### DIFF
--- a/ui/admin/app/components/tools/CreateTool.tsx
+++ b/ui/admin/app/components/tools/CreateTool.tsx
@@ -1,7 +1,9 @@
 import { PlusCircle } from "lucide-react";
+import { useCallback, useEffect, useState } from "react";
 import { useForm } from "react-hook-form";
+import useSWR, { SWRResponse } from "swr";
 
-import { CreateToolReference } from "~/lib/model/toolReferences";
+import { CreateToolReference, ToolReference } from "~/lib/model/toolReferences";
 import { ToolReferenceService } from "~/lib/service/api/toolreferenceService";
 
 import { Button } from "~/components/ui/button";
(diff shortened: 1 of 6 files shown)


ui/admin/app/routes/_auth.tools._index.tsx
```diff
@@ -5,6 +5,7 @@ import useSWR, { preload } from "swr";
 import { ToolReferenceService } from "~/lib/service/api/toolreferenceService";
 
 import { TypographyH2 } from "~/components/Typography";
+import ErrorDialog from "~/components/knowledge/ErrorDialog";
```
ryanhopperlowe: Unrelated, but can you make this a named export and update usages?

Also since this is being used in multiple packages it should probably be moved to the `~/components/composed/` directory

ui/admin/app/components/tools/CreateTool.tsx
```diff
@@ -1,30 +1,71 @@
...
+    onError: (error: string) => void;
     onSuccess: () => void;
 }
 
-export function CreateTool({ onSuccess }: CreateToolProps) {
+export function CreateTool({ onError, onSuccess }: CreateToolProps) {
     const { register, handleSubmit, reset } = useForm<CreateToolReference>();
 
+    const [loadingToolId, setLoadingToolId] = useState("");
+    const getLoadingTool: SWRResponse<ToolReference, Error> = useSWR(
```
ryanhopperlowe: nit: can probably remove the type annotation here

ui/admin/app/components/tools/CreateTool.tsx
```diff
@@ -1,30 +1,71 @@
...
+
+        if (data?.resolved) {
+            handleCreatedTool(data);
+        }
+    }, [getLoadingTool, handleCreatedTool, loadingToolId]);
+
     const { execute: onSubmit, isLoading } = useAsync(
         async (data: CreateToolReference) => {
-            await ToolReferenceService.createToolReference({
+            const response = await ToolReferenceService.createToolReference({
```
ryanhopperlowe, line 60: does this service method not throw an error when creating a tool fails?

ui/admin/app/components/tools/CreateTool.tsx
```diff
@@ -37,9 +78,9 @@ export function CreateTool({ onSuccess }: CreateToolProps) {
                 />
             </div>
             <div className="flex justify-end">
-                <Button type="submit" disabled={isLoading}>
+                <Button type="submit" disabled={pending}>
                     <PlusCircle className="w-4 h-4 mr-2" />
-                    {isLoading ? "Creating..." : "Register Tool"}
+                    {pending ? "Creating..." : "Register Tool"}
                 </Button>
```
ryanhopperlowe: can you actually replace this with the `Button.loading` prop?


```suggestion
                <Button type="submit" disabled={pending} loading={pending} startContent={<PlusCircle />}>
                    Register Tool
                </Button>
```

ui/admin/app/routes/_auth.tools._index.tsx
ivyjeong13: Yeah can definitely do that!

ryanhopperlowe (changes requested):
Just a couple of small things, but looks good overall!

PR #739, opened 2024-12-03
feat: add the ability to configure model providers dynamically

thedadams:
Prior to this change, all model providers must be configured with environment variables. This change adds the ability to configure model providers dynamically, including replacing any existing credentials.

The OPENAI_API_KEY is special here. If that environment variable is set, then Otto8 will create the necessary credential without user interaction.


pkg/controller/handlers/toolreference/toolreference.go
```diff
@@ -253,3 +296,16 @@ func (h *Handler) Populate(req router.Request, resp router.Response) error {
 
 	return nil
 }
+
+func (h *Handler) RemoveCredential(req router.Request, _ router.Response) error {
```
g-linville: nit: I think this should be renamed for greater clarity.

```suggestion
func (h *Handler) RemoveModelProviderCredential(req router.Request, _ router.Response) error {
```

g-linville (approved):

pkg/controller/handlers/toolreference/toolreference.go
thedadams: Done.

PR #742, opened 2024-12-03
Fix: run pnpm build in package script

StrongMonkey:
We are missing the build step in packaging script that are used to build docker images


ryanhopperlowe (approved):
Nice find

PR #755, opened 2024-12-04
feat: RunSpec.Timeout to override default 10m run timeout

iwilltry42:
... also set default knowledge data-source sync timeout to ~30m~ 1h, as it can easily happen that website-scraping exceeds 10 minutes.

Ref https://github.com/otto8-ai/otto8/issues/743

### DIFF
--- a/pkg/controller/handlers/knowledgesource/knowledgesource.go
+++ b/pkg/controller/handlers/knowledgesource/knowledgesource.go
@@ -199,7 +199,7 @@ func (k *Handler) Sync(req router.Request, _ router.Response) error {
 
 	invokeOpts := invoke.SystemTaskOptions{
 		CredentialContextIDs: []string{credentialContextID},
-		Timeout:              30 * time.Minute,
+		Timeout:              1 * time.Hour,
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/controller/handlers/knowledgesource/knowledgesource.go
```diff
@@ -199,6 +199,7 @@ func (k *Handler) Sync(req router.Request, _ router.Response) error {
 
 	invokeOpts := invoke.SystemTaskOptions{
 		CredentialContextIDs: []string{credentialContextID},
+		Timeout:              30 * time.Minute,
```
cjellick: Let's go to an hour

cjellick (changes requested):
just one small tweak

StrongMonkey (approved):
LGTM with craig's comment addressed

pkg/controller/handlers/knowledgesource/knowledgesource.go
iwilltry42: done

cjellick (approved):

PR #762, opened 2024-12-04
Fix: fix delete knowledge source hanging in UI

StrongMonkey:
Don't list knowledge source with `deleted` as it could be waiting to be cleaned by finalizer. Also validate in cache after deletion.

https://github.com/otto8-ai/otto8/issues/746

### DIFF
--- a/ui/admin/app/components/knowledge/AgentKnowledgePanel.tsx
+++ b/ui/admin/app/components/knowledge/AgentKnowledgePanel.tsx
@@ -83,41 +83,37 @@ export default function AgentKnowledgePanel({
     const getLocalFiles: SWRResponse<KnowledgeFile[], Error> = useSWR(
         KnowledgeService.getLocalKnowledgeFilesForAgent.key(agentId),
         ({ agentId }) =>
-            KnowledgeService.getLocalKnowledgeFilesForAgent(agentId).then(
-                (items) =>
-                    items
-                        .sort((a, b) => a.fileName.localeCompare(b.fileName))
-                        .map(
-                            (item) =>
-                                ({
-                                    ...item,
-                                }) as KnowledgeFile
-                        )
-                        .filter((item) => !item.deleted)
-            ),
+            KnowledgeService.getLocalKnowledgeFilesForAgent(agentId),
         {
             revalidateOnFocus: false,
             refreshInterval: blockPollingLocalFiles ? undefined : 5000,
         }
     );
     const localFiles = useMemo(
-        () => getLocalFiles.data || [],
+        () =>
+            getLocalFiles.data
+                ?.sort((a, b) => a.fileName.localeCompare(b.fileName))
+                .map(
+                    (item) =>
+                        ({
+                            ...item,
+                        }) as KnowledgeFile
+                )
+                .filter((item) => !item.deleted) || [],
         [getLocalFiles.data]
     );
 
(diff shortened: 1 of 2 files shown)


ui/admin/app/components/knowledge/AgentKnowledgePanel.tsx
```diff
@@ -107,7 +107,10 @@ export default function AgentKnowledgePanel({
 
     const getKnowledgeSources = useSWR(
         KnowledgeService.getKnowledgeSourcesForAgent.key(agentId),
-        ({ agentId }) => KnowledgeService.getKnowledgeSourcesForAgent(agentId),
+        ({ agentId }) =>
```
ryanhopperlowe: when performing mutations on swr fetchers, please pull out the mutation into it's own useMemo.

If I were to try to use this key again with the generic fetcher method, my code would break because the data from the response would not align with the data in the cache

ryanhopperlowe (changes requested):

ui/admin/app/components/knowledge/AgentKnowledgePanel.tsx
ryanhopperlowe: example

```ts
// wrong

const { data }  = useSWR(
    Service.method.key(),
    () => Service.method().then(() => someMutation()) // cache has mutated data
)

// in another file

const { data }  = useSWR(
    Service.method.key(),
    () => Service.method() // cache has data directly from response
)

// an error may be thrown because the data in the cache does not match the structure of the data from the response
```

instead do this:

```ts
const { data }  = useSWR(
    Service.method.key(),
    () => Service.method() // use data directly from response
)

const mutatedData = useMemo(() => someMutation(data), [data])

// in another file

const { data }  = useSWR(
    Service.method.key(),
    () => Service.method() // also use data directly from response
)


No error now, because the data in the cache will ALWAYS have the same structure
```

ryanhopperlowe: Happy to hop on a huddle and explain in more depth if necessary :)

StrongMonkey: Good to know! changed 👍

ui/admin/app/components/knowledge/AgentKnowledgePanel.tsx
```diff
@@ -83,41 +83,37 @@ export default function AgentKnowledgePanel({
...
-        () => getLocalFiles.data || [],
+        () =>
+            getLocalFiles.data
+                ?.sort((a, b) => a.fileName.localeCompare(b.fileName))
+                .map(
+                    (item) =>
+                        ({
+                            ...item,
+                        }) as KnowledgeFile
+                )
```
ryanhopperlowe, line 101: nit: 

```suggestion
(getLocalFiles.data as KnowledgeFile[])
                ?.sort((a, b) => a.fileName.localeCompare(b.fileName))
```

ryanhopperlowe (approved):
thanks for fixing that!

ui/admin/app/components/knowledge/AgentKnowledgePanel.tsx
ryanhopperlowe: It's ugly, but it's important... If I ever get the time to replace SWR with react-query, this will be super trivial

PR #764, opened 2024-12-04
feat: add ModelProviders page & banner

ivyjeong13:
[Doing some cleanup/rearranging atm but putting up as draft!]

https://github.com/user-attachments/assets/df6c2e05-301b-42f4-a52f-93371b9410ec

### DIFF
--- a/pkg/api/handlers/modelprovider.go
+++ b/pkg/api/handlers/modelprovider.go
@@ -116,10 +116,15 @@ func convertToolReferenceToModelProvider(ctx context.Context, gClient *gptscript
 		return types.ModelProvider{}, err
 	}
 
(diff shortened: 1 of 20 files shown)


pkg/api/handlers/modelprovider.go
```diff
@@ -116,10 +116,15 @@ func convertToolReferenceToModelProvider(ctx context.Context, gClient *gptscript
 		return types.ModelProvider{}, err
 	}
 
+	name := ref.Status.Tool.Name
```
ivyjeong13, line 119: Looking to get the user friendly name of the model provider here, otherwise use the id equivalent (ex. `Open AI` instead of `openai-model-provider`.

PR #774, opened 2024-12-04
feat: Model Providers page

ivyjeong13:
Screenshot 2024-12-04 at 6 47 07 PM

Models Provider Page!


ui/admin/app/components/model-providers/ModelProviderConfigure.tsx
```diff
@@ -0,0 +1,107 @@
...
+            <DialogTrigger asChild>
+                <Button
+                    size="icon"
+                    variant="ghost"
+                    className="mt-0"
+                    onClick={() => setDialogIsOpen(true)}
+                >
+                    <SettingsIcon />
+                </Button>
+            </DialogTrigger>
```
ryanhopperlowe: nit: `DialogTrigger` handles opening/closing the dialog automatically
```suggestion
    const [dialogIsOpen, setDialogIsOpen] = useState(false);
    return (
        <Dialog open={dialogIsOpen} onOpenChange={setDialogIsOpen}>
            <DialogTrigger asChild>
                <Button
                    size="icon"
                    variant="ghost"
                    className="mt-0"
                >
                    <SettingsIcon />
                </Button>
            </DialogTrigger>
```

ui/admin/app/components/model-providers/ModelProviderConfigure.tsx
```diff
@@ -0,0 +1,107 @@
...
+}) {
+    const revealModelProvider = useSWR(
+        ModelProviderApiService.revealModelProviderById.key(modelProvider.id),
+        ({ modelProviderId }) =>
+            ModelProviderApiService.revealModelProviderById(modelProviderId),
+        { keepPreviousData: true }
+    );
+
+    const handleSuccess = (config: ModelProviderConfig) => {
+        revealModelProvider.mutate(config, false);
```
ryanhopperlowe, line 66: nice optomistic update here

ui/admin/app/components/model-providers/ModelProviderContext.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+    const handleConfigureModelProvider = useCallback(
+        (modelProviderId: string, values: Record<string, string>) =>
+            ModelProviderApiService.configureModelProviderById(
+                modelProviderId,
+                values
+            )
+                .then(() => {
+                    getModelProviders.mutate();
+                    mutate(ModelProviderApiService.getModelProviders.key());
```
ryanhopperlowe: these two lines do the same thing

ui/admin/app/components/model-providers/ModelProviderContext.tsx
```diff
@@ -0,0 +1,83 @@
...
+    modelProviders: ModelProvider[];
+    configured: boolean;
+    configureModelProvider: (
+        modelProviderId: string,
+        value: Record<string, string>
+    ) => void;
+    isUpdating: boolean;
+    error?: unknown;
+    lastUpdated?: Date;
+}
```
ryanhopperlowe: so it doesn't really look like much of this is reused (or really even used at all) so I'd argue that this doesn't really need to be put into a Context. My preference would be to just call `useSWR` in each place you need access to the model providers data and let each component.

doubling down on this because `configureModelProvider` is only used in 1 place as well

ui/admin/app/components/model-providers/ModelProviderForm.tsx
```diff
@@ -0,0 +1,167 @@
...
+    modelProviderId: string;
+    onSuccess: (config: ModelProviderConfig) => void;
+    parameters: ModelProviderConfig;
+    requiredParameters: string[];
+}) {
+    const { configureModelProvider } = useModelProviders();
+
+    const form = useForm<ModelProviderFormValues>({
+        resolver: zodResolver(formSchema),
+        mode: "onChange",
```
ryanhopperlowe: nit: this form is pretty simple at it's core (just checks for required fields). I'd argue this prop can be removed (triggers validation only on submit)

ui/admin/app/components/model-providers/ModelProviderForm.tsx
```diff
@@ -0,0 +1,167 @@
...
+            <NameDescriptionForm
+                defaultValues={form.watch("additionalConfirmParams")}
+                onChange={(values) =>
+                    form.setValue("additionalConfirmParams", values)
+                }
+            />
+
+            <div className="flex justify-end">
+                <Button
+                    onClick={form.handleSubmit(onSubmit)}
```
ryanhopperlowe: instead of duplicating submission logic, you can use native HTML APIs to remotely trigger a form

```ts
const FORM_ID = "model-provider-form"

<form id={FORM_ID}>...</form>

// notice that button is outside of the form
<Button type="submit" form={FORM_ID} /> // clicking this button will now trigger submit on the above form
```

ui/admin/app/lib/service/api/modelProviderApiService.ts
```diff
@@ -1,16 +1,90 @@
...
+configureModelProviderById.key = (modelProviderId?: string) => {
+    if (!modelProviderId) return null;
+
+    return {
+        url: ApiRoutes.modelProviders.configureModelProviderById(
+            modelProviderId
+        ).path,
+        modelProviderId,
+    };
+};
```
ryanhopperlowe: nit: most apis that mutate data should not be triggered via SWR and therefore don't require a key

ivyjeong13: Thank you! Will remove that

ryanhopperlowe (changes requested):
couple of things to fix here, but good overall!

ui/admin/app/components/model-providers/ModelProviderForm.tsx
ivyjeong13: Awesome, thank you!

ryanhopperlowe (approved):
Thank you 🙌

PR #775, opened 2024-12-05
Make text-embedding-3-large the default embedding model

sanjay920:
text-embedding-3-large is much better than text-embedding-3-small. It should be the default embedding model. 

pgvector does have a max dim limit of 2000 so the following changes are necessary before merging this PR in: https://github.com/otto8-ai/tools/pull/258


pkg/controller/data/default-model-aliases.yaml
```diff
@@ -25,7 +25,7 @@ items:
     spec:
       manifest:
         alias: text-embedding
-        model: text-embedding-3-small
+        model: text-embedding-3-large
```
thedadams, line 28: This won't update currently deployed environments, but this is the right change.

thedadams (approved):

iwilltry42 (approved):

thedadams:
If I understand correctly, for OpenAI, we were still going to hard-code the suggested default models. This would need to be included there.

I guess the point is: after today, this change will no longer achieve its goal.

PR #784, opened 2024-12-05
enhance: agent name generation in admin ui

ryanhopperlowe:
This PR improves the agent name generation in the admin UI by selecting a pool of nouns to select from.

This should reduce the number of nouns that could also be used as adjectives and increases the number of nouns likely to be familiar to a user.

The end goal: improve the likelihood of generating a "fun" agent name.

Screenshot 2024-12-05 at 11 47 58 AM


ryanhopperlowe:
That makes a ton of sense, I'll try to use commit bodies more often

PR #795, opened 2024-12-06
feat: view models for a model provider & configure helper text for inputs

ivyjeong13:
https://github.com/user-attachments/assets/ff2f6bfd-6fcc-4606-a906-9937e3eeb3ab

* view models from a Model Provider. After configuring a Model Provider, gear shows on top right
* added helper text for Azure and Ollama for some inputs that need info to help fill out

Screenshot 2024-12-06 at 8 37 36 AM

* added helper text for required inputs that may require more context (ex. Ollama & Azure OpenAI Model Provider required configurations)

For context, after discussing with Will & Craig, it was decided that it made more sense to view the models from a Model Provider. This PR adds that in but there may be a follow-up of:

* removing the Models page
* turning the delete into a disable toggle switch. disabling means: If a model is disabled, then that means the model is not usable period (Add a warning that the agents/etc could break with this being broken)
* disable instead of deleting a model
* add changing usage as a dropdown in the table row for the models

### DIFF
--- a/ui/admin/app/components/model-providers/ModelProviderForm.tsx
+++ b/ui/admin/app/components/model-providers/ModelProviderForm.tsx
@@ -14,12 +14,20 @@ import {
     ParamFormValues,
 } from "~/components/composed/NameDescriptionForm";
 import { ControlledInput } from "~/components/form/controlledInputs";
-import { ModelProviderConfigurationLinks } from "~/components/model-providers/constants";
+import {
+    ModelProviderConfigurationLinks,
+    ModelProviderRequiredTooltips,
+} from "~/components/model-providers/constants";
 import { Button } from "~/components/ui/button";
 import { Form } from "~/components/ui/form";
 import { Link } from "~/components/ui/link";
 import { ScrollArea } from "~/components/ui/scroll-area";
 import { Separator } from "~/components/ui/separator";
+import {
+    Tooltip,
+    TooltipContent,
+    TooltipTrigger,
+} from "~/components/ui/tooltip";
 import { useAsync } from "~/hooks/useAsync";
 
 const formSchema = z.object({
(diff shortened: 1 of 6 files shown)


cjellick (approved):

ui/admin/app/components/model-providers/ModelProviderForm.tsx
```diff
@@ -238,4 +246,55 @@ export function ModelProviderForm({
...
+        const link = ModelProviderConfigurationLinks[modelProviderId];
+        return (
+            <Tooltip>
+                <TooltipTrigger asChild>
+                    <Link to={link} size="icon" variant="ghost" as="button">
+                        <CircleHelpIcon className="text-muted-foreground" />
+                    </Link>
+                </TooltipTrigger>
+
+                <TooltipContent
```
ryanhopperlowe, line 260: can you add a cva to `Tooltip` with `default` and `secondary variants` so we can make our styles more reusable?

ui/admin/app/components/model-providers/ModelProviderForm.tsx
```diff
@@ -238,4 +246,55 @@ export function ModelProviderForm({
...
+                            </Button>
+                        </TooltipTrigger>
+
+                        <TooltipContent
+                            side="right"
+                            className="bg-secondary text-foreground max-w-80"
+                        >
+                            {tooltip}
+                        </TooltipContent>
+                    </Tooltip>
```
ryanhopperlowe, line 295: I have lots of places I'd like to repeat this pattern, can you extract this into a `composed` component?

ryanhopperlowe: It'll also help to have these wrapped in custom components if (big if) we decide to add mobile view. Tooltips aren't mobile/touch friendly, so we'd need to add some behavior to open a drawer or something on tap

ryanhopperlowe:
oh 1 more thing... we might want to make the table head row sticky if possible:
image

ui/admin/app/components/model-providers/ModelProviderForm.tsx
ivyjeong13: Yeah of course! Will move these bits to composed

PR #796, opened 2024-12-06
fix: waiting for modelProvider models to load before showing default model selection

ivyjeong13:
https://github.com/user-attachments/assets/a1c84366-7ecc-4576-b44f-ca48ceac68de

* adds polling to wait for models to be back populated after configuring

Also included small fixes:
* No Models Available when no options to select in default model input dropdown
* do not show empty badge when usage isn't set for back populated models
* show name instead of id in table when viewing model provider's models

### DIFF
--- a/ui/admin/app/components/model-providers/ModelProviderConfigure.tsx
+++ b/ui/admin/app/components/model-providers/ModelProviderConfigure.tsx
@@ -32,11 +32,9 @@ export function ModelProviderConfigure({
     const [loadingModelProviderId, setLoadingModelProviderId] = useState("");
 
     const getLoadingModelProviderModels = useSWR(
-        loadingModelProviderId
-            ? ModelProviderApiService.getModelProviderById.key(
-                  loadingModelProviderId
-              )
-            : null,
+        ModelProviderApiService.getModelProviderById.key(
+            loadingModelProviderId
+        ),
         ({ modelProviderId }) =>
             ModelProviderApiService.getModelProviderById(modelProviderId),
         {
--- a/ui/admin/app/components/model-providers/ModelProviderConfigure.tsx
+++ b/ui/admin/app/components/model-providers/ModelProviderConfigure.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import useSWR from "swr";
 
 import { ModelProvider, ModelProviderConfig } from "~/lib/model/modelProviders";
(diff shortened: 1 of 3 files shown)


ui/admin/app/components/model-providers/ModelProviderConfigure.tsx
```diff
@@ -29,6 +29,34 @@ export function ModelProviderConfigure({
...
         useState(false);
 
+    const [loadingModelProviderId, setLoadingModelProviderId] = useState("");
+
+    const getLoadingModelProviderModels = useSWR(
+        loadingModelProviderId
+            ? ModelProviderApiService.getModelProviderById.key(
+                  loadingModelProviderId
+              )
+            : null,
```
ryanhopperlowe: nit: The `getModelProviderById` already returns null if `modelProviderId` is null, so this ternary is redundant

ui/admin/app/components/model-providers/ModelProviderModels.tsx
```diff
@@ -84,18 +84,19 @@ export function ModelProvidersModels({ modelProvider }: ModelsConfigureProps) {
...
-                    <Badge variant="outline">
-                        {getModelUsageLabel(row.original.usage)}
-                    </Badge>
-                ),
+                cell: ({ row }) =>
+                    getModelUsageLabel(row.original.usage) ? (
+                        <Badge variant="outline">
+                            {getModelUsageLabel(row.original.usage)}
+                        </Badge>
+                    ) : null,
```
ryanhopperlowe: nit: should this be sortable on the usage label?

```suggestion
            columnHelper.accessor((model) => getModelUsageLabel(model.usage) || "", {
                id: "usage",
                header: "Usage",
                cell: ({ row, getValue /* or something */ }) =>
                    getValue() ? (
                        <Badge variant="outline">
                            {getValue()}
                        </Badge>
                    ) : null,
```

ryanhopperlowe: I realize we haven't built out dynamic sorting into our data tables, but I have a feeling we definitely will at some point :)

ui/admin/app/components/model/shared/DefaultModelAliasForm.tsx
```diff
@@ -206,7 +206,12 @@ export function DefaultModelAliasForm({
         defaultModel: string
     ) {
         if (!modelOptions) {
-            if (!defaultModel) return null;
+            if (!defaultModel)
+                return (
+                    <div className="p-2 text-sm text-muted-foreground">
```
ryanhopperlowe: TypographyP

ryanhopperlowe (changes requested):
only non-nit is the Typography thing. Looks good!

ui/admin/app/components/model-providers/ModelProviderModels.tsx
ivyjeong13: Yeah especially with a large list of models, I see that happening and a search to filter as well!

ryanhopperlowe (approved):
thanks!

PR #797, opened 2024-12-06
enhance: helper tooltip and minor table ui fix

ivyjeong13:
* create TooltipContent variant
* move tooltips in ModelProviderForms to composed HelperTooltip
* fix sticky table header (needed to remove some wrapping divs, checked the others that used DataTable but lemme know if anything looks off)


ryanhopperlowe (approved):
all set

PR #804, opened 2024-12-06
Fix: Add agent.status.textEmbedding and set it in controller

StrongMonkey:
We are making multiple requests in agent list call because it needs to fetch the knowledge set for each agent to get textEmbeddingModel. The call becomes extremely slow when database is on load or ingestion is running. So add the the embedding field in agent.status directly and set it in controller to avoid extra api calls when listing agent.


g-linville (approved):

thedadams:
Darren asked us to go a different way with this and essentially asked me to get it done. So this issue has been adressed differently and this can be closed.

PR #805, opened 2024-12-06
feat: sort models by name in default model dropdowns

ryanhopperlowe:
Screenshot 2024-12-06 at 4 00 35 PM

Screenshot 2024-12-06 at 4 01 16 PM

Screenshot 2024-12-06 at 5 27 20 PM

 - sorts models by name in default model dropdowns
- adds an "Other" group to dropdowns for models that don't have a usage
- ensures that backend always returns a value for model usage
- apply suggested models to default models form after configuration

### DIFF
--- a/apiclient/types/model.go
+++ b/apiclient/types/model.go
@@ -12,7 +12,7 @@ type ModelManifest struct {
 	ModelProvider string     `json:"modelProvider,omitempty"`
 	Alias         string     `json:"alias,omitempty"`
 	Active        bool       `json:"active"`
-	Usage         ModelUsage `json:"usage,omitempty"`
+	Usage         ModelUsage `json:"usage"`
 }
 
 type ModelList List[Model]
--- a/ui/admin/app/components/model/DefaultModelAliasForm.tsx
+++ b/ui/admin/app/components/model/DefaultModelAliasForm.tsx
@@ -296,7 +296,7 @@ export function DefaultModelAliasFormDialog({
                 <Button disabled={disabled}>Set Default Models</Button>
             </DialogTrigger>
 
-            <DialogContent>
+            <DialogContent className="max-w-2xl">
                 <DialogHeader>
                     <DialogTitle>Default Model Aliases</DialogTitle>
                 </DialogHeader>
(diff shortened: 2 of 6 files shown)


apiclient/types/model.go
```diff
@@ -12,7 +12,7 @@ type ModelManifest struct {
 	ModelProvider string     `json:"modelProvider,omitempty"`
 	Alias         string     `json:"alias,omitempty"`
 	Active        bool       `json:"active"`
-	Usage         ModelUsage `json:"usage,omitempty"`
+	Usage         ModelUsage `json:"usage"`
```
ryanhopperlowe, line 15: Made this change so the frontend can map to the backend response via enum

ivyjeong13 (approved):

apiclient/types/model.go
cjellick: im explicitly  OKing this. We'll see if a real dev yells at me later.

cjellick (approved):

ui/admin/app/components/model/DefaultModelAliasForm.tsx
```diff
@@ -135,21 +155,21 @@ export function DefaultModelAliasForm({
...
-                                      usage === ModelUsage.Vision
-                                          ? ModelUsage.LLM
-                                          : usage
-                                  )
-                                : [];
+                            const usage =
+                                getModelUsageFromAlias(alias) ??
+                                ModelUsage.Unknown;
+
+                            const modelOptions = filterModelsByUsage(
```
ivyjeong13, line 169: nit: Just a thought but wondering if a usage is set, should the options show both ones marked as `usage` and the unknowns? In case a type is set for one, but maybe a user wanted to find one of the unknowns?

ivyjeong13 (approved):

ui/admin/app/components/model/DefaultModelAliasForm.tsx
ryanhopperlowe: check the `renderSelectModelContent` method :)

PR #808, opened 2024-12-06
fix: flicker on models dialog due to dialog animation & select

ivyjeong13:
Before:

https://github.com/user-attachments/assets/57573d8e-77f9-4687-ba98-845864bb4ecd

After:

https://github.com/user-attachments/assets/8ccd67af-daf9-4a54-8409-bf8178f48ad9

* Super weird when there's a select in a dialog, there's a flicker that happens. Seems to be a weird rendering interaction with the select rendering its value and the animation of the dialog happening at the same time?


ivyjeong13:
closing for now, to look more into root issue

PR #811, opened 2024-12-07
fix: link open new tab fix & minor table styling fixes

ivyjeong13:
* adds padding to align model provider header with start of table row
* opens new tab for anthropic link
* removes border-b from table rows


cjellick (approved):

ryanhopperlowe (approved):
JFYI this is going to remove borders from every table app-wide. Just want to make sure you checked that they still look decent :)

ivyjeong13:
> JFYI this is going to remove borders from every table app-wide. Just want to make sure you checked that they still look decent :)

After input from Craig, reverted for now!

PR #818, opened 2024-12-07
bug: remove apply logic for better performance

ibuildthecloud:


pkg/controller/handlers/knowledgesource/knowledgesource.go
```diff
@@ -64,17 +63,77 @@ func safeStatusSave(ctx context.Context, c kclient.Client, source *v1.KnowledgeS
...
+	newNames := map[string]v1.KnowledgeFile{}
+	for _, file := range newFiles {
+		newNames[file.Name] = file
+	}
+
+	for newName, newFile := range newNames {
+		existingFile, ok := existingNames[newName]
+		if !ok {
+			if err := c.Create(ctx, &newFile); apierror.IsAlreadyExists(err) {
+				if err := c.Get(ctx, kclient.ObjectKeyFromObject(&existingFile), &existingFile); err != nil {
```
StrongMonkey: I think if there is a conflict, existingFile will be empty and this will return error to fail the sync. 

should be `kclient.ObjectKeyFromObject(&newFile)`?

cjellick (approved):

pkg/controller/handlers/knowledgesource/knowledgesource.go
ibuildthecloud: let me look

ibuildthecloud: youre right, good catch

PR #833, opened 2024-12-13
enhance: if no default text-embedding set, disable knowledge for agent/workflows

ivyjeong13:
Screenshot 2024-12-13 at 9 44 45 AM

* disables Knowledge description & Add Knowledge button if default text-embedding isn't set
* adds Alert to notify user why it's disabled

### DIFF
--- a/ui/admin/app/components/composed/WarningAlert.tsx
+++ b/ui/admin/app/components/composed/WarningAlert.tsx
@@ -7,7 +7,7 @@ export function WarningAlert({
     description,
 }: {
     title: string;
-    description: string | React.ReactNode;
+    description: React.ReactNode;
 }) {
     return (
         <Alert variant="default">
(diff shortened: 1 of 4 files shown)


ui/admin/app/components/composed/WarningAlert.tsx
```diff
@@ -0,0 +1,19 @@
+import { CircleAlertIcon } from "lucide-react";
+
+import { Alert, AlertDescription, AlertTitle } from "~/components/ui/alert";
+
+export function WarningAlert({
+    title,
+    description,
+}: {
+    title: string;
+    description: string | React.ReactNode;
```
ryanhopperlowe: nit: `ReactNode` already encompasses `string`

ryanhopperlowe (approved):
nice touch with the Alert

Issue #481, opened 2024-11-08
UI - Provide some confirmation when registering tool succeeds.


sangee2004:
**Steps to reproduce the problem:**
1. Register a tool.
Notice that once it is registered , the `Create New Tool Reference` pop up page dismisses and we are in the tools list view.
The tool gets added at the end of the tool list. User needs to scroll all the way down to be able to see it.

It would be good to provide some confirmation to the user that the tool was added successfully. Or we should be able to take him to the newly registered tool in the tool list.

cjellick:
Id like user added tools to be at the top.

sangee2004:
Registered tool now shows up at the top of the tool list.
<img width="434" alt="Image" src="https://github.com/user-attachments/assets/246d833c-4bcb-4d09-a588-acbc87c11cfb" />

PR #839, opened 2024-12-13
fix: user tools to the top & toast on successful tool addition

ivyjeong13:
Addresses #481.

Screenshot 2024-12-13 at 11 51 35 AM

* creates toast when adding tool is successful
* user created tools is at top

Update:
* green checkmark for toast.success
Screenshot 2024-12-13 at 12 45 31 PM


ivyjeong13:
> TIL: This is called a "toast"
> 
> nit: might be nice for the checkmark to be green to intensify the "pat on the back" effect.

Will get that checkmark green!

PR #849, opened 2024-12-13
chore: correct otto8 release version info

njhale:
- **chore: add git tag to ldflags on build**
- **chore: use correct package path for version ldflag**

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -13,7 +13,7 @@ builds:
       - amd64
       - arm64
     ldflags:
-      - -X "github.com/gptscript-ai/otto8/pkg/version.Tag=v{{ .Version }}"
+      - -X "github.com/otto8-ai/otto8/pkg/version.Tag=v{{ .Version }}"
 
 archives:
   - format: tar.gz
--- a/Makefile
+++ b/Makefile
@@ -28,9 +28,11 @@ serve-docs:
 	npm run start
 
 # Build the project
-build:
-	go build -ldflags="-s -w" -o bin/otto8 .
 
+GIT_TAG := $(shell git describe --tags --exact-match 2>/dev/null | xargs -I {} echo -X 'github.com/otto8-ai/otto8/pkg/version.Tag={}')
+GO_LD_FLAGS := "-s -w $(GIT_TAG)"
+build:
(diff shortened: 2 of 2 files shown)


.goreleaser.yaml
```diff
@@ -13,7 +13,7 @@ builds:
       - amd64
       - arm64
     ldflags:
-      - -X "github.com/gptscript-ai/otto8/pkg/version.Tag=v{{ .Version }}"
+      - -X "github.com/otto8-ai/otto8/pkg/version.Tag=v{{ .Version }}"
```
thedadams, line 16: Thanks. This was my fault.

thedadams (approved):

Makefile
```diff
@@ -28,9 +28,11 @@ serve-docs:
 	npm run start
 
 # Build the project
-build:
-	go build -ldflags="-s -w" -o bin/otto8 .
 
+GIT_TAG := $(shell git describe --tags --exact-match 2>/dev/null | xargs -I {} echo -X 'github.com/otto8-ai/otto8/pkg/version.Tag={}')
+GO_LD_FLAGS := "-s -w $(GIT_TAG)"
```
drpebcak, line 33: Is there a reason we would need a tagged version number in a build produced by the makefile? Ideally we could just not have the extra complexity in here.

drpebcak (changes requested):

Makefile
njhale: The logic used to construct the version info w/ the correct git tag -- what we've been using from all the way back in acorn runtime days -- depends on the tag field being set via ldflags during the build. I've already tried grabbing the `Version` from the `Main.Version` field of https://pkg.go.dev/runtime/debug#BuildInfo instead, but that doesn't actually return the correct tag (always has `(devel)`, even for builds on "clean" checkouts of a pushed tag).

I'm open to alternatives if you have any suggestions.

drpebcak: I think it's fine for GoReleaser, but we have never set it in the Makefile before - that's the part I'm asking about.

drpebcak: Are we using the binary produced by the Makefile in an actual release somewhere?

drpebcak: Gross, but it looks like the answer is yes.. in the docker build.

drpebcak (approved):

PR #850, opened 2024-12-13
fix: update make dev user ui ping url

ryanhopperlowe:

### DIFF
--- a/tools/dev.sh
+++ b/tools/dev.sh
@@ -133,6 +133,6 @@ user_ui_ready_pid=$!
 wait "${server_ready_pid}" "${admin_ui_ready_pid}" "${user_ui_ready_pid}"
 print_section_header 120 "All components ready!"
 
-open_browser_tabs http://localhost:8080/admin/ http://localhost:8080/
+open_browser_tabs http://localhost:8080/
 
 wait "${server_pid}" "${admin_ui_pid}" "${user_ui_pid}"
--- a/tools/dev.sh
+++ b/tools/dev.sh
@@ -118,7 +118,7 @@ user_ui_pid=$!
 
 (
   for _ in {1..60}; do # ~1 minute timeout
-    if curl -s -H 'User-Agent: Mozilla/5.0' --head http://localhost:8080/unknown | head -n 1 | grep "200 OK" > /dev/null; then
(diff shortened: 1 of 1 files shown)


tools/dev.sh
```diff
@@ -118,7 +118,7 @@ user_ui_pid=$!
 
 (
   for _ in {1..60}; do # ~1 minute timeout
-    if curl -s --head http://localhost:8080/ | head -n 1 | grep "200 OK" > /dev/null; then
+    if curl -s -H 'User-Agent: Mozilla/5.0' --head http://localhost:8080/unknown | head -n 1 | grep "200 OK" > /dev/null; then
```
njhale: Just checked and you don't actually need the header for this to work, since we have `-s --head` specified.

Let's also switch this to check for the .ico file served by the user UI instead since I think it's probably a bad idea to depend on `/<agent>` returning a 200 when you try to GET an agent that doesn't exist; I suspect that behavior is actually a bug, we probably want some splash page there that tells you that an admin needs to configure some agents.

```suggestion
    if curl -s --head http://localhost:8080/favicon.ico | head -n 1 | grep "200 OK" > /dev/null; then
```

njhale (changes requested):

njhale (changes requested):
You also want to drop the `http://localhost:8080` from the `open` command a few lines below

ryanhopperlowe:
> You also want to drop the `http://localhost:8080/admin/` from the `open` command a few lines below

I checked and this url is actually still valid. It doesn't look like admin is served from `localhost:8080/` yet

njhale (changes requested):
> I checked and this url is actually still valid. It doesn't look like admin is served from localhost:8080/ yet

Your browser should be redirected from `http://localhost:8080` to `http://localhost:8080/admin/` automatically, so it doesn't make sense to open both IMO.

ryanhopperlowe:
> > I checked and this url is actually still valid. It doesn't look like admin is served from localhost:8080/ yet
> 
> Your browser should be redirected from `http://localhost:8080` to `http://localhost:8080/admin/` automatically, so it doesn't make sense to open both IMO.

Ah I think I see what you're saying now. Updated

PR #851, opened 2024-12-13
fix: adds model provider in dropdown options for default models

ivyjeong13:
* adds model provider in dropdown like in agent models dropdown 
Screenshot 2024-12-13 at 3 32 50 PM

* only thing about this is a little busy and drowns out (Suggested), not sure if a smaller font size for the provider would help


ryanhopperlowe:
what if format was 

**<Model>** [(suggested)] - <Model Provider>

ryanhopperlowe (commented):
1 idea for ux

ryanhopperlowe (approved):
Looks good!

PR #863, opened 2024-12-16
chore: change from otto icon to acorn

ivyjeong13:
* switch from otto to acorn icons 
Screenshot 2024-12-16 at 11 21 22 AM

Screenshot 2024-12-16 at 11 21 46 AM

* does this mean we're allowed to use the acorn purple too? 😄


ryanhopperlowe (approved):
This looks amazing 🙌

Issue #838, opened 2024-12-13
Order models by name, model provider


sanjay920:
The current order is random in the UI.
![image](https://github.com/user-attachments/assets/900017ba-cefe-4ad7-928c-e00bf65ea24e)

Ideally, I want each models to sorted alphabetically and grouped by model provider. See how openai does it below:
![image](https://github.com/user-attachments/assets/2f37f921-19dc-45de-8838-99e5b9aedf0b)

ryanhopperlowe:
Additionally let's make it a combo box to allow users to filter results via their keyboard

sangee2004:
When testing with version - `  "obot": "v0.0.0-dev+264972e6"`

Model dropdown in now ordered by name and model providers.

<img width="637" alt="Screenshot 2024-12-17 at 10 01 38 AM" src="https://github.com/user-attachments/assets/809040e0-9301-41fc-9ef4-59e88373e9bf" />

<img width="629" alt="Screenshot 2024-12-17 at 10 01 22 AM" src="https://github.com/user-attachments/assets/3bd71aab-2753-44f4-9a98-aa49ed2ecf8e" />

PR #892, opened 2024-12-16
fix: sort by name, model provider & filter for agent's model dropdown

ivyjeong13:
* Addresses #838 

update: added checkmark for selected to match with select
Screenshot 2024-12-16 at 5 39 27 PM

https://github.com/user-attachments/assets/a10c98ad-a680-4ca9-8979-caa1ddd9dc6f

Screenshot 2024-12-16 at 4 31 17 PM

* add `ComboBox` for dropdown with filter -- also comes with mobile responsive drawer!

### DIFF
--- a/ui/admin/app/components/composed/ComboBox.tsx
+++ b/ui/admin/app/components/composed/ComboBox.tsx
@@ -1,5 +1,3 @@
-"use client";
-
 import { CheckIcon, ChevronsUpDownIcon } from "lucide-react";
 import { ReactNode, useState } from "react";
 
--- a/ui/admin/app/components/ui/drawer.tsx
+++ b/ui/admin/app/components/ui/drawer.tsx
@@ -1,5 +1,3 @@
-"use client";
-
 import * as React from "react";
 import { Drawer as DrawerPrimitive } from "vaul";
 


ui/admin/app/components/ui/drawer.tsx
```diff
@@ -0,0 +1,118 @@
+"use client";
+
+import * as React from "react";
+import { Drawer as DrawerPrimitive } from "vaul";
+
+import { cn } from "~/lib/utils";
+
+const Drawer = ({
```
ivyjeong13, line 6: Pulled Combobox from shadcn example, came with a mobile responsive but can strip it out if needed!

ivyjeong13:
If we're alright with this, I'll update the default models dropdowns to use this as well!

ui/admin/app/components/composed/ComboBox.tsx
```diff
@@ -0,0 +1,178 @@
+"use client";
```
ryanhopperlowe: nit: can you remove the "use client" directives?

ui/admin/app/components/ui/drawer.tsx
```diff
@@ -0,0 +1,118 @@
+"use client";
```
ryanhopperlowe: same here

ryanhopperlowe (approved):

PR #900, opened 2024-12-17
enhance: add gptscript module version to /api/version

njhale:
- Add the module version for `github.com/gptscript-ai/gptscript` to the `/api/version` API response
- Change `/api/version` API response field name from `otto` to `acorn`

Addresses https://github.com/acorn-io/acorn/issues/899

### DIFF
--- a/pkg/api/handlers/version.go
+++ b/pkg/api/handlers/version.go
@@ -2,19 +2,24 @@ package handlers
 
 import (
 	"os"
+	"runtime/debug"
+	"strings"
 
 	"github.com/obot-platform/obot/pkg/api"
 	"github.com/obot-platform/obot/pkg/version"
+	"golang.org/x/mod/module"
 	"sigs.k8s.io/yaml"
 )
 
 type VersionHandler struct {
-	emailDomain string
+	gptscriptVersion string
+	emailDomain      string
 }
 
 func NewVersionHandler(emailDomain string) *VersionHandler {
 	return &VersionHandler{
-		emailDomain: emailDomain,
+		emailDomain:      emailDomain,
+		gptscriptVersion: getGPTScriptVersion(),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/api/handlers/version.go
```diff
@@ -30,7 +35,53 @@ func (v *VersionHandler) getVersionResponse() map[string]string {
...
+func simplifyModuleVersion(version string) string {
+	if version == "" || !module.IsPseudoVersion(version) {
+		return version
+	}
+
+	// Extract the base version (tag) and revision (commit hash)
+	// Panic on failure. This should never happen for well-formed pseudo-versions.
+	base, err := module.PseudoVersionBase(version)
+	if err != nil {
+		panic(fmt.Errorf("error extracting base version: %w", err))
```
StrongMonkey: why are we panicing? would be ideal if version api fails we can make the whole server crash

njhale: this happens once on startup and would only panic with a bad module version, which IIUC, shouldn't be possible at runtime. If we're scared of the panic statement, I can remove it.

pkg/api/handlers/version.go
```diff
@@ -30,7 +35,52 @@ func (v *VersionHandler) getVersionResponse() map[string]string {
 			values["error"] = err.Error()
 		}
 	}
-	values["otto"] = version.Get().String()
+	values["acorn"] = version.Get().String()
```
njhale: This is going to change to `obot` soon, but I figured updating this to follow the current branding will make that easier.

pkg/api/handlers/version.go
njhale: I removed the panic statements to reduce reviewer stress 🙂

StrongMonkey: haha ok. Make sense. 👍

StrongMonkey (approved):

PR #916, opened 2024-12-17
Add Groq model provider

sanjay920:
This PR introduces [groq](https://groq.com/) as a model provider

Relies on https://github.com/acorn-io/tools/pull/283


ui/admin/app/components/model-providers/constants.ts
```diff
@@ -33,6 +35,9 @@ export const ModelProviderRequiredTooltips: {
     [CommonModelProviderIds.OLLAMA]: {
         Host: "IP Address for the ollama server (eg. 127.0.0.1:1234)",
     },
+    [CommonModelProviderIds.GROQ]: {
+        "API Key": "Groq API Key. Can be created and fetched from https://console.groq.com/keys",
```
njhale: Should this be set for all the other providers that require keys as well; e.g. `Anthropic`, `Voyage`, etc? Not that it has to change in this PR. I'm just drying to suss out whether this is correct or not since we don't have any entry for any other model provider API key here.

cc @ryanhopperlowe

njhale (changes requested):
Looks like you need to run `make validate-go-code`.

sanjay920:
updated with naming changes: `ACORN` -> `OBOT`

ui/admin/app/components/model-providers/constants.ts
ryanhopperlowe: Good catch, this should actually read as such

```suggestion
        "Api Key":
            "Groq API Key. Can be created and fetched from https://console.groq.com/keys",
```

ryanhopperlowe: you'll also need to run `pnpm run format` from `/ui/admin` for the build to pass

ryanhopperlowe (changes requested):

StrongMonkey (approved):

ryanhopperlowe (approved):
lgtm

njhale (approved):

Issue #843, opened 2024-12-13
Model Configuration page reloads automatically clearing existing entries


sangee2004:
**Steps to reproduce the problem:**
1. Configure Azure Open AI model provider.
2. After entering few parameters , I see the page reloads automatically clearing existing entries.

Video showing page reload - https://acorn-io.slack.com/archives/C07P0EZR917/p1734112160336639

sangee2004:
This issue is not seen anymore when testing with latest build.

PR #919, opened 2024-12-17
fix: do not retry for reveal 404, will clear existing form entries

ivyjeong13:
* Addresses #843 
* useSWR continues to retry if it receives an error response but 404 response is valid for /reveal (means no credential found for that model provider)
* avoid retry for 404 (causing clear out of form and retry is unnecessary, we revalidate reveal after user submits form)

### DIFF
--- a/ui/admin/app/components/model-providers/ModelProviderConfigure.tsx
+++ b/ui/admin/app/components/model-providers/ModelProviderConfigure.tsx
@@ -130,16 +130,12 @@ export function ModelProviderConfigureContent({
             ModelProviderApiService.revealModelProviderById(modelProviderId),
         {
             keepPreviousData: true,
-            // 404 -> no credential found, so don't retry
-            shouldRetryOnError: (error) => {
-                if (
-                    error instanceof NotFoundError &&
-                    error.message.toLowerCase().includes("no credential found")
-                ) {
-                    return false;
-                }
-                return true;
-            },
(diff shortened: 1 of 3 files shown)


ui/admin/app/components/model-providers/ModelProviderConfigure.tsx
```diff
@@ -127,7 +127,11 @@ export function ModelProviderConfigureContent({
         ModelProviderApiService.revealModelProviderById.key(modelProvider.id),
         ({ modelProviderId }) =>
             ModelProviderApiService.revealModelProviderById(modelProviderId),
-        { keepPreviousData: true }
+        {
+            keepPreviousData: true,
+            // 404 -> no credential found, so don't retry
+            shouldRetryOnError: (error) => error.status !== 404,
```
ryanhopperlowe: can you do 3 things:

- add a `NotFoundError` to `apiErrors.ts` (does not need to be a BoundaryError
- add respective catch clause to `request()` in `~/lib/service/api/primitives.ts`
- replace with this:
```suggestion
            shouldRetryOnError: (error) => error instanceof NotFoundError,
```

ivyjeong13: Oh yeah, can do! :+1:

ryanhopperlowe (approved):
thank you!

Issue #358, opened 2024-10-29
Provide a way to interrupt a response when it is being streamed.


sangee2004:
Currently there is no way to interrupt a response when it is being streamed.

cjellick:
API now exists: `threads/{id}/abort`

@ivyjeong13 - im going to assign this to you so that you can get some exposure to the chat functionality

cjellick:
functionality is already in the consumption ui, so this just needs done in admin ui

sangee2004:
When testing with latest version we are provided with ability to interrupt chat responses.

PR #929, opened 2024-12-17
enhance: allow interruption of running thread

ivyjeong13:
Addresses #358 
* adds stop button to stop running thread process

https://github.com/user-attachments/assets/7868c04b-5bf7-4bf5-abd8-024a8eb98b3d

### DIFF
--- a/ui/admin/app/components/chat/ChatContext.tsx
+++ b/ui/admin/app/components/chat/ChatContext.tsx
@@ -21,6 +21,7 @@ interface ChatContextType {
     messages: Message[];
     mode: Mode;
     processUserMessage: (text: string) => void;
+    abortRunningThread: () => void;
     id: string;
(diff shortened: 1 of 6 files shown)


ui/admin/app/components/chat/ChatContext.tsx
```diff
@@ -133,6 +142,15 @@ function useMessageSource(threadId?: Nullish<string>) {
             }
 
             if (error) {
+                if (error.includes("thread was aborted, cancelling run")) {
```
ryanhopperlowe, line 145: nit: The backend should probably return some type of discriminator so we don't need to do string parsing to know how to display the UI. Might be worth bringing up to Craig.

That being said it looks like Darren is doing the same thing in the User UI for now, so it's probably fine haha

ryanhopperlowe (approved):

Issue #934, opened 2024-12-18
Better user experience when chatting with agents when no model provider is configured.


sangee2004:
I am allowed to chat with the default agent `obot` even when no model provider is configured.

I see the following error presented to the user

<img width="521" alt="Screenshot 2024-12-17 at 4 43 55 PM" src="https://github.com/user-attachments/assets/0370d884-5ca2-4c17-9448-c9fc552b9fea" />

We can provide a better user experience by providing a banner about having to configure the  model provider. 
Related slack conversation - https://acorn-io.slack.com/archives/C03FU91U4SX/p1734475788930659?thread_ts=1734475033.849729&cid=C03FU91U4SX

sangee2004:
We are now provided with this banner in all pages when no model providers are configured.
<img width="991" alt="Image" src="https://github.com/user-attachments/assets/1dd6f6c4-4c09-4cdf-bc9a-60dc15ce5d42" />

Chat is also disabled at this time

PR #949, opened 2024-12-18
fix: model provider disable flows if not set & banner updates

ivyjeong13:
Addresses #934 

https://github.com/user-attachments/assets/b2b1edc7-71bf-4d52-bdf3-7caff44e69a5

* banner on every page

Screenshot 2024-12-18 at 12 05 19 PM

Screenshot 2024-12-18 at 12 04 51 PM

* disabling buttons if a model provider is not configured and notifying user that they need to configure


ui/admin/app/routes/_auth.tsx
```diff
@@ -2,18 +2,26 @@ import { Outlet, isRouteErrorResponse, useRouteError } from "@remix-run/react";
...
-        UserService.getMe()
-    );
+    const promises = await Promise.all([
+        preload(UserService.getMe.key(), () => UserService.getMe()),
+        preload(
+            ModelProviderApiService.getModelProviders.key(),
+            ModelProviderApiService.getModelProviders
+        ),
+    ]);
+    const me = promises[0];
```
ivyjeong13, line 23: Wondering if I can drop this `return { me }`? Thought I'd see a 'useLoaderData' reference but maybe I missed it!

ryanhopperlowe: Yes you should be able to drop it

ryanhopperlowe: just need to return null

ui/admin/app/components/model-providers/ModelProviderContext.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export function useModelProviders() {
+    const context = useContext(ModelProviderContext);
+    if (context === undefined) {
+        throw new Error(
+            "useModelProvider must be used within a ModelProviderProvider"
+        );
+    }
+    return context;
+}
```
ryanhopperlowe: Essentially what this looks like is we are pulling from 1 global store (api cache) just to put it into another global store (Context). If you don't mind, I think a custom hook will get you all of the benefits here without adding the Context Dependency

ryanhopperlowe:
```ts
function useModelProviders() {
    const { data: modelProviders } = useSWR(...);

    const configured = ...;

    return { modelProviders, configured }
}
```

This will drastically reduce the code, give you the same amount of reusability, and now we don't have to wrap every `<Chat />` component with another context provider

ui/admin/app/components/ui/button.tsx
```diff
@@ -21,6 +21,7 @@ const buttonVariants = cva(
                 ghost: "hover:bg-secondary hover:text-secondary-foreground",
                 accent: "bg-accent text-accent-foreground shadow-sm hover:bg-accent/80",
                 link: "text-primary hover:text-primary/70 underline-offset-4 hover:underline shadow-none hover:shadow-none",
+                warning: "bg-warning text-black shadow-sm hover:bg-warning/80",
```
ryanhopperlowe:
```suggestion
                warning: "bg-warning text-foreground shadow-sm hover:bg-warning/80",
```

ryanhopperlowe (changes requested):

ryanhopperlowe (changes requested):
Couple of comments but LOVE the new look

ui/admin/app/components/model-providers/ModelProviderContext.tsx
ivyjeong13: Will do this, thank you!

ryanhopperlowe (approved):
all set. Thanks!

PR #952, opened 2024-12-18
fix: kill current pid instead of 0

thedadams:


njhale (approved):
Works for me! Thanks!

PR #960, opened 2024-12-18
chore: add support for salesforce oauth app

drpebcak:
Relies on changes in https://github.com/obot-platform/tools/pull/290 for salesforce integration to work


pkg/gateway/server/oauth_apps.go
```diff
@@ -445,6 +446,30 @@ func (s *Server) callbackOAuthApp(apiContext api.Context) error {
...
+			State:        state,
+			TokenType:    salesforceTokenResp.TokenType,
+			Scope:        salesforceTokenResp.Scope,
+			AccessToken:  salesforceTokenResp.AccessToken,
+			ExpiresIn:    7200, // Relies on Salesforce admin not overriding the default 2 hours
+			Ok:           true, // Assuming true if no error is present
+			CreatedAt:    createdAt,
+			RefreshToken: salesforceTokenResp.RefreshToken,
+			Extras: map[string]string{
+				"GPTSCRIPT_SALESFORCE_URL": salesforceTokenResp.InstanceURL,
```
njhale, line 470: This approach is definitely more flexible, but it feels like we're losing the self-documenting property of the credentials framework; i.e. we could put the `token` here too and entirely remove the `with GPTSCRIPT_SALESFORCE_TOKEN as token` clause from `salesforce/credential/tool.gpt`

thedadams: The difference between putting the token here and using the credential framework is that a user won't be prompted if the token doesn't exist. Additionally, the credential framework will handle refreshes. So I think the token should stay the way it is.

Another possibility for this `GPTSCRIPT_SALESFORCE_URL` is for it to have its own credential. That is the way environment variables are handled for workflows: each environment variable is turned into a credential on the tool.

However, I'm not sure that is better for this situation. I think I like this approach.

thedadams (approved):
Looks good after linting issues are fixed.

pkg/gateway/server/oauth_apps.go
njhale: > that a user won't be prompted if the token doesn't exist.

That's great point. OK, you've convinced me.

njhale (approved):

PR #987, opened 2024-12-19
chore: remove useForm dependency from BasicToolForm

ryanhopperlowe:
TLDR;
useForm returns `useRef().current` under the hood as a hack to optimize rendering. For the most part, this is totally fine, but it CAN cause some erratic behavior when certain conditions are met (I think it has to do with resetting the form's default values...????)

Either way, it added way more complexity than it was worth (at least for this component) so removing it was the best solution

More in-depth:
After mount, the first rerender (change event) of this specific usage of `BasicToolForm` was causing the `form.watch` method to change references (not only that but it was being treated as a NOOP??) as a result, the `onChange` prop was not getting triggered, but only for the first change. Subsequent rerenders **did not** reinitialize the `watch` method and it behaved as expected.

Looking into the code of `react-hook-form` there is __no reasonable explanation__ for the reference of `form.watch` to change between renders. 

Essentially what happens is `useForm` initializes a `useRef` object with the `watch` method and **never updates it again**. The only way that re-initialization could happen is if the parent component is remounted (tested and this is not the case). 

The only thing that seemed off from within the `useForm` code is that it returns the `current` property of the previously mentioned `useRef` object. This technically breaks the rules of `useRef` ([this one](https://arc.net/l/quote/chttzirx) specifically) which "makes your component's behavior unpredictable"

In the interest of spending another 3+ hours on this topic, I'm sticking with this as the working diagnosis

### DIFF
--- a/ui/admin/app/components/agent/ToolForm.tsx
+++ b/ui/admin/app/components/agent/ToolForm.tsx
@@ -168,13 +168,6 @@ export function ToolForm({
                 <div className="flex justify-end">
                     <ToolCatalogDialog
                         tools={allTools}
-                        onAddTool={(tool) =>
-                            toolFields.append({
-                                tool,
-                                variant: ToolVariant.FIXED,
-                            })
-                        }
-                        onRemoveTools={removeTools}
                         onUpdateTools={(tools) =>
                             updateTools(tools, ToolVariant.FIXED)
                         }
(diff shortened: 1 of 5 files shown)


ui/admin/app/components/agent/ToolForm.tsx
```diff
@@ -159,6 +175,9 @@ export function ToolForm({
                             })
                         }
                         onRemoveTools={removeTools}
+                        onUpdateTools={(tools) =>
```
ivyjeong13, line 171: nit: Silly question, if onUpdateTools handles both addition and removal of tools, could onAddTool and onRemoveTools be axed?

ivyjeong13 (approved):

ui/admin/app/components/agent/ToolForm.tsx
ryanhopperlowe: Not silly at all. You're totally right haha. Will update

ryanhopperlowe: updated

ivyjeong13 (approved):

PR #1008, opened 2024-12-20
chore: refactor knowledge code in admin ui

ryanhopperlowe:
This should have no effect on logic or UI/UX. it just extracts some subcomponents and abstracts data logic into reusable hooks.

Doing this in preparation for thread level knowledge implementations

Refactored largely using cursor


StrongMonkey (approved):
LGTM. Walk down through the code with @ryanhopperlowe and it is working as expected.

Thanks for picking it up and refactor the code! Look much better

ui/admin/app/hooks/knowledge/useKnowledgeSourceFiles.ts
```diff
@@ -15,6 +16,14 @@ export function useKnowledgeSourceFiles(
 ) {
     const [blockPollingFiles, setBlockPollingFiles] = useState(true);
 
+    const startPolling = () => {
+        if (blockPollingFiles) setBlockPollingFiles(false);
+    };
+
+    if (knowledgeSource.state === KnowledgeSourceStatus.Syncing) {
```
StrongMonkey: you should also check for pending state. Something if sync happen to fast it will never fetch state from syncing given we are pulling every 5 seconds

StrongMonkey (approved):

ui/admin/app/hooks/knowledge/useKnowledgeSourceFiles.ts
ryanhopperlowe: fixed! thanks!

PR #1012, opened 2024-12-20
Fix: remove knowledge tool when knowledge set has no content

StrongMonkey:
We should remove knowledge tool when invoking if knowledge set has no content. This avoid LLM to call knowledge when no data is in it.

### DIFF
--- a/pkg/render/render.go
+++ b/pkg/render/render.go
@@ -101,9 +101,19 @@ func Agent(ctx context.Context, db kclient.Client, agent *v1.Agent, oauthServerU
 		return nil, nil, err
 	}
 
-	extraEnv, err = addKnowledgeTools(ctx, db, agent, opts.Thread, extraEnv)
+	extraEnv, added, err := addKnowledgeTools(ctx, db, agent, opts.Thread, extraEnv)
 	if err != nil {
 		return nil, nil, err
+	}
+
+	if !added {
+		filteredTools := make([]string, 0)
+		for _, tool := range mainTool.Tools {
+			if !strings.HasSuffix(tool, "as knowledge") {
+				filteredTools = append(filteredTools, tool)
+			}
+		}
+		mainTool.Tools = filteredTools
 	}
 
 	oauthEnv, err := OAuthAppEnv(ctx, db, agent.Spec.Manifest.OAuthApps, agent.Namespace, oauthServerURL)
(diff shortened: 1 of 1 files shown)


pkg/render/render.go
```diff
@@ -101,11 +101,21 @@ func Agent(ctx context.Context, db kclient.Client, agent *v1.Agent, oauthServerU
...
-	extraEnv, err = addKnowledgeTools(ctx, db, agent, opts.Thread, extraEnv)
+	extraEnv, added, err := addKnowledgeTools(ctx, db, agent, opts.Thread, extraEnv)
 	if err != nil {
 		return nil, nil, err
 	}
 
+	if !added {
+		filteredTools := make([]string, 0)
+		for _, tool := range mainTool.Tools {
+			if !strings.HasSuffix(tool, "as knowledge") {
```
StrongMonkey: This is a bit hacky. Not sure if there is a great way to identify knowledge tool

pkg/render/render.go
```diff
@@ -159,25 +169,25 @@ func OAuthAppEnv(ctx context.Context, db kclient.Client, oauthAppNames []string,
 	return extraEnv, nil
 }
 
-func addKnowledgeTools(ctx context.Context, db kclient.Client, agent *v1.Agent, thread *v1.Thread, extraEnv []string) ([]string, error) {
+func addKnowledgeTools(ctx context.Context, db kclient.Client, agent *v1.Agent, thread *v1.Thread, extraEnv []string) ([]string, bool, error) {
```
g-linville: This function seems improperly named. It sounds like it doesn't actually add the knowledge tools, and instead just gives you an env var, and returns whether or not the knowledge tool is actually needed. I think it should be renamed to something more descriptive.

StrongMonkey: sound good. I will take the chance to rename

pkg/render/render.go
thedadams: Is the part before the "as" knowledge-retrieval? I think that is the name of the tool in the tools repo.

StrongMonkey: no, tool definition is something like this 
```
github.com/obot-platform/tools/knowledge as knowledge
```

So before `as` it is the location of the tool and last part is the tool name in index.yaml I guess

thedadams: OK. Looking at how this tool gets added to the agent, tell me if this makes sense:
1. Move the call to `configureKnowledgeEnvs` up above line 68.
2. In the loop on line 78, the `tool` that gets processed there is just called `knowledge`. That is easier to detect, and we can skip adding it to the tool if `added` is false there.
3. We would have to do that same thing for the thread loop on line 68.

The issue with this that I can think of is that we would have to do the same filtering twice now that we are adding knowledge to threads. But maybe checking for a tool called `knowledge` is better than checking if this name ends in `as knowledge`?

So, overall, I think your approach is fine and will only break if we change how the knowledge tool gets resolved.

thedadams (approved):
One more possible suggestion. If you don't think my suggestion is worth it, then I am definitely fine with this going in as is. I just wanted to throw out an option that might not be as brittle.

pkg/render/render.go
iwilltry42: I second your approach @thedadams - one loop less and even less code in total :+1:

iwilltry42 (approved):
LGTM, however, I think [what Donnie proposed here](https://github.com/obot-platform/obot/pull/1012/files#r1894623766) might be even cleaner :+1:

pkg/render/render.go
StrongMonkey: Sounds good 👍  addressed

PR #1013, opened 2024-12-20
enhance: combobox with nested groups & use for default alias dropdown

ivyjeong13:
* get ComboBox working with nested groups and filtering with nested groups/items
* change default model alias selects to comboboxes

Screenshot 2024-12-20 at 2 23 03 PM

Screenshot 2024-12-20 at 2 26 25 PM

### DIFF
--- a/ui/admin/app/components/composed/ComboBox.tsx
+++ b/ui/admin/app/components/composed/ComboBox.tsx
@@ -35,7 +35,7 @@ type ComboBoxProps<T extends BaseOption> = {
     onChange: (option: T | null) => void;
     options: (T | GroupedOption<T>)[];
     placeholder?: string;
-    suggested?: string[];
+    renderOption?: (option: T) => ReactNode;
     value?: T | null;
 };
 
@@ -43,7 +43,7 @@ export function ComboBox<T extends BaseOption>({
     disabled,
     placeholder,
     value,
-    suggested,
+    renderOption,
     ...props
 }: {
     disabled?: boolean;
@@ -58,7 +58,7 @@ export function ComboBox<T extends BaseOption>({
                 <PopoverContent className="w-full p-0" align="start">
                     <ComboBoxList
                         setOpen={setOpen}
-                        suggested={suggested}
+                        renderOption={renderOption}
                         value={value}
                         {...props}
                     />
@@ -74,7 +74,7 @@ export function ComboBox<T extends BaseOption>({
                 <div className="mt-4 border-t">
                     <ComboBoxList
                         setOpen={setOpen}
-                        suggested={suggested}
+                        renderOption={renderOption}
                         value={value}
                         {...props}
                     />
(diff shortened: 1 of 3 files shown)


ui/admin/app/components/composed/ComboBox.tsx
```diff
@@ -25,22 +25,25 @@ type BaseOption = {
...
 
 type ComboBoxProps<T extends BaseOption> = {
     allowClear?: boolean;
     clearLabel?: ReactNode;
+    emptyLabel?: ReactNode;
     onChange: (option: T | null) => void;
-    options: T[] | GroupedOption<T>[];
+    options: (T | GroupedOption<T>)[];
     placeholder?: string;
+    suggested?: string[];
```
ryanhopperlowe: `suggested` feels like an implementation detail that should be handled by the parent. Could we instead expose a `renderItem` prop that allows the parent to implement each row as it pleases?


```suggestion
    renderItem?: (item: T) => ReactNode;
```

ui/admin/app/components/composed/ComboBox.tsx
```diff
@@ -94,16 +112,85 @@ function ComboBoxList<T extends BaseOption>({
...
+
+        // If both or neither are suggested, maintain original order
+        if (aIsSuggested === bIsSuggested) return 0;
+        // Sort suggested items first
+        return aIsSuggested ? -1 : 1;
+    };
+
+    const handleValueChange = (value: string) => {
+        setFilteredOptions(filterOptions(options, value));
+    };
```
ryanhopperlowe: same here. this should be handled by the parent. Especially if we ever need a combobox that displays data paginated by the server

ui/admin/app/components/composed/ComboBox.tsx
```diff
@@ -94,16 +112,85 @@ function ComboBoxList<T extends BaseOption>({
...
+                        .includes(searchValue.toLowerCase());
+                const isGroupHeadingMatch =
+                    "heading" in option &&
+                    option.heading
+                        .toLowerCase()
+                        .includes(searchValue.toLowerCase());
+
+                if (isGroupHeadingMatch || isSingleValueMatch) {
+                    return [...acc, option];
+                }
```
ryanhopperlowe, line 143: Absolutely no change necessary here! But a fun alternative...

```ts
                const hasMatch = [
                    "name" in option && option.name,
                    "heading" in option && option.heading,
                ]
                    .filter((x) => !!x)
                    .join("|")
                    .toLowerCase()
                    .includes(searchValue.toLowerCase());

                if (hasMatch) {
                    return [...acc, option];
                }

```

ui/admin/app/components/composed/ComboBox.tsx
ivyjeong13: 🙌 Good call out, moved the suggestions out of combobox! Thanks!

ryanhopperlowe (commented):
looking good

ryanhopperlowe (approved):

Issue #993, opened 2024-12-19
Voyage - When voyage model provider is configured with invalid api key, there is no error message presented to the user. Also voyage models are made available to set default aliases.


sangee2004:
**Steps to reproduce the problem:**
1. Configure Voyage  provider  with invalid api keys.

User is presented with no error messages and is taken to the` Default Model Aliases` page were all the voyage models are made available.

<img width="792" alt="Screenshot 2024-12-19 at 12 06 20 PM" src="https://github.com/user-attachments/assets/7fa56db4-27a1-4563-80ad-347ce2e617a6" />

**Expected Behavior:**
User should be presented with an error message when  invalid api keys are provided when configuring voyage provider.

cjellick:
@iwilltry42 is introducing a `/validate` endpoint. @sanjay920, for any model providers you've added, please add this validate endpoint. Also, let's hold off on adding any more model providers. I don't want to unnecessarily increase the surface area for testing and development for second-tier, lower quality models. Maybe we can figure out a system where these go into an "experimental" catalog or something.

cjellick:
Oh - nevermind - this is the embedding model that "goes along with" anthropic. My mistake. I do want to hold off on additional model providers in general, until we figure out a support matrix of some type.

sangee2004:
This issue is addressed when testing with latest build.

Issue #991, opened 2024-12-19
Anthropic provider - When Anthropic provider is configured with invalid api key , there is no error message presented to the user. Also anthropic models are made available to set default aliases.


sangee2004:
**Steps to reproduce the problem:**
1. Configure Anthropic provider  with invalid api keys.

User is presented with no error messages and is taken to the` Default Model Aliases` page were all the anthropic models are made available.

<img width="860" alt="Screenshot 2024-12-19 at 12 07 01 PM" src="https://github.com/user-attachments/assets/0ef7636a-9f7a-41fe-9341-1f262acb233d" />

**Expected Behavior:**
User should be presented with an error message when  invalid api keys are provided when configuring Anthropic provider.

cjellick:
Should be addressed by current `validate` endpoint changes.

sangee2004:
Following error message is presented to the user when Anthropic provider is configured with invalid keys.
<img width="735" alt="Image" src="https://github.com/user-attachments/assets/e6145210-d1aa-4852-a99b-bf29653dcce5" />

Issue #974, opened 2024-12-19
When model is configured with incorrect api keys we do not show the banner relating to model provider setup or restrict chatting.


sangee2004:
**Steps to reproduce the problem:**
1. Start otto server (with out passing openai api key environment variable)
2. When accessing the otto server url , user will be presented with following banner

<img width="991" alt="Screenshot 2024-12-18 at 3 57 19 PM" src="https://github.com/user-attachments/assets/dbc52f19-0b4f-4670-a7e6-ba923f401330" />

3. Configure OpenAI model provider with incorrect keys. 
4. When error message is presented , do not correct the key values so that  OpenAI model provider is misconfigured.
5. Go to agents page (or any other page).Notice that the banner from step 2 is not present anymore.
User will also be allowed to chat with `obot` agent which will result in error relating to default model not configured.

This issue is related to the fact that the OpenAI model provider is shown as being in configured state even when it was not successfully configured. This issue is tracked in  https://github.com/obot-platform/obot/issues/819

cjellick:
@sangee2004 please fix the formatting in this issue. the numbered list runs together

cjellick:
@iwilltry42 is introducing a `validate` endpoint to model providers. Once that is in, we will be able to more cleanly handle this. 

@iwilltry42  @thedadams - where did we land on the idea of a "validate and save" piece of functionality that will only save the config if validation passes?

iwilltry42:
> @iwilltry42 is introducing a `validate` endpoint to model providers. Once that is in, we will be able to more cleanly handle this. 
> 
> @iwilltry42  @thedadams - where did we land on the idea of a "validate and save" piece of functionality that will only save the config if validation passes?

@cjellick I discussed my implementation with @thedadams and he liked it, so I'll put up the PR once I'm back 👍

iwilltry42:
I think this may be a duplicate of #819 - getting some feedback for when an already configured provider is being invalidated.
We're now populating `toolReference.Status.Error` when a provider goes bad.
Currently that's still based on the available models API, but in 90% of the cases that includes validation logic.
We may want to reconsider if we want to have validation as an extra step here.

sangee2004:
This issue is addressed in latest build.

Issue #842, opened 2024-12-13
Need consistent way to validate a model-provider is properly configured


cjellick:
Right now, to verify a model-provider was properly configured,  the model provider modal calls an api for listing available models, which directly calls the model provider. This does not work for all our model providers. I forget which ones @thedadams said it does work for.

Id like a consistent way to verify a model-provider is properly configured. Maybe model-provider needs to have a /validate endpoint?

cjellick:
Tangentially related to https://github.com/otto8-ai/otto8/issues/819. That one makes me think the backend needs to call a validate endpoint and mark the provider as being in an error state.

thedadams:
Based on a conversation about this issue, I say we do the following:
1. Implement the validation as we discussed: use the Error field to indicate if there were any issues getting the available models in the controller. We definitely want this change no matter what in case a user configures the model provider incorrectly.
2. After that is done, then we can add the /validate endpoint discussed here. I currently don't have an idea of how this would work. If you have one great. Otherwise we can discuss it.

cjellick:
What is the status of this @iwilltry42? Is it ready for frontend development?

iwilltry42:
@cjellick frontend code is already in my PR (and approved by Ryan) - I just need to address some comments and then we need the validate tool on all providers before merging that PR.

sangee2004:
When model providers are configured with incorrect values, Model providers are shown as being in "Not Configured" and models from the provider are not available as expected.

Error message presented to the user is always `Network Error` in this case which need to be more descriptive error message. This issue is tracked in https://github.com/obot-platform/obot/issues/1385

PR #1019, opened 2024-12-23
feat: model provider configuration validation

iwilltry42:
Important: This draft PR contains a couple of changes.

1. Change to the handler to populate ToolReference.Status.Error with an error message if available models couldn't be fetched
	Screenshot-20241223111504-1128x61
2. Change to the UI when calling `/available-models/{model-provider-id}` so it shows the error message to the user
	Screenshot 2024-12-23 at 10-45-22
3. Change to the UI to call /validate and only if that succeeds, save the config and call /available-models
	- If it fails, don't save the config and display an appropriate error
4. Add the /validate endpoint for model providers which calls the subtool `validate from <model-provider-tool>` to validate the config (if available)

Tested with the "reference implementation" which is the gemini-vertex provider: https://github.com/obot-platform/obot/pull/986 / https://github.com/obot-platform/tools/pull/297
	
	
Ref:
- https://github.com/obot-platform/obot/issues/842
- https://github.com/obot-platform/obot/issues/819
- https://github.com/obot-platform/obot/issues/993
- https://github.com/obot-platform/obot/issues/991
- https://github.com/obot-platform/obot/issues/974

### DIFF
--- a/pkg/api/handlers/modelprovider.go
+++ b/pkg/api/handlers/modelprovider.go
@@ -1,6 +1,7 @@
 package handlers
 
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 
@@ -8,7 +9,10 @@ import (
 	"github.com/obot-platform/obot/apiclient/types"
 	"github.com/obot-platform/obot/pkg/api"
 	"github.com/obot-platform/obot/pkg/gateway/server/dispatcher"
+	"github.com/obot-platform/obot/pkg/invoke"
 	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
+	"github.com/obot-platform/obot/pkg/system"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -16,12 +20,14 @@ import (
 type ModelProviderHandler struct {
 	gptscript  *gptscript.GPTScript
 	dispatcher *dispatcher.Dispatcher
+	invoker    *invoke.Invoker
 }
 
-func NewModelProviderHandler(gClient *gptscript.GPTScript, dispatcher *dispatcher.Dispatcher) *ModelProviderHandler {
+func NewModelProviderHandler(gClient *gptscript.GPTScript, dispatcher *dispatcher.Dispatcher, invoker *invoke.Invoker) *ModelProviderHandler {
 	return &ModelProviderHandler{
 		gptscript:  gClient,
 		dispatcher: dispatcher,
+		invoker:    invoker,
 	}
 }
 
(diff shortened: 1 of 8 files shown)


pkg/api/handlers/modelprovider.go
```diff
@@ -89,6 +97,78 @@ func (mp *ModelProviderHandler) List(req api.Context) error {
 	return req.Write(types.ModelProviderList{Items: resp})
 }
 
+type ValidationError struct {
+	Err string `json:"error"`
+}
+
+func (ve *ValidationError) Error() string {
+	return fmt.Sprintf("model-provider credentials validation failed: {\"error\": \"%s\"}", ve.Err)
```
thedadams, line 104: Do we need to return the error as JSON here? Can we just return the string?

pkg/api/handlers/modelprovider.go
```diff
@@ -89,6 +97,78 @@ func (mp *ModelProviderHandler) List(req api.Context) error {
...
+	var ref v1.ToolReference
+	if err := req.Get(&ref, req.PathValue("id")); err != nil {
+		return err
+	}
+
+	if ref.Spec.Type != types.ToolReferenceTypeModelProvider {
+		return types.NewErrBadRequest("%q is not a model provider", ref.Name)
+	}
+
+	l := logger.Package()
```
thedadams: nit: make this a constant for the package.

pkg/api/handlers/modelprovider.go
```diff
@@ -89,6 +97,78 @@ func (mp *ModelProviderHandler) List(req api.Context) error {
...
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: system.ThreadPrefix + "-" + ref.Name + "-validate-",
+			Namespace:    ref.Namespace,
+		},
+		Spec: v1.ThreadSpec{
+			SystemTask: true,
+		},
+	}
+
+	if err := req.Get(thread, thread.Name); apierrors.IsNotFound(err) {
```
thedadams: Does this work? It looks like `thread.Name` will always be empty here.

pkg/api/handlers/modelprovider.go
```diff
@@ -89,6 +97,78 @@ func (mp *ModelProviderHandler) List(req api.Context) error {
...
+	if err != nil {
+		return err
+	}
+	defer task.Close()
+
+	res, err := task.Result(req.Context())
+	if err != nil {
+		if strings.Contains(err.Error(), "tool not found: validate from "+ref.Spec.Reference) { // there's no simple way to do errors.As/.Is at this point unfortunately
+			l.Infof("Model provider %q does not provide a validate tool. Looking for 'validate from %s'", ref.Name, ref.Spec.Reference)
+			return nil // Do not fail if model provider doesn't provide a validate tool
```
thedadams: Does it make sense for this to be a 400-level error?

ryanhopperlowe (approved):

pkg/controller/handlers/toolreference/toolreference.go
```diff
@@ -450,7 +452,23 @@ func (h *Handler) BackPopulateModels(req router.Request, _ router.Response) erro
 	availableModels, err := availablemodels.ForProvider(req.Ctx, h.dispatcher, req.Namespace, req.Name)
 	if err != nil {
 		// Don't error and retry because it will likely fail again. Log the error, and the user can re-sync manually.
-		log.Errorf("Failed to get available models for model provider %q: %v", toolRef.Name, err)
+		// Also, the toolRef.Status.Error field will bubble up to the user in the UI.
+
+		// Check if the model provider returned a properly formatted error message and set it as status
+		re := regexp.MustCompile(`\{.*"error":.*}`)
```
thedadams: This can be made a constant and only compiled once (on startup).

pkg/api/handlers/modelprovider.go
iwilltry42: Do you want to capture and ignore it higher up (in the frontend) or actually error out?
The latter will have us wait for all providers to provide the validate tool first.

thedadams: I think it makes sense to error out and we can add "validate" tools to the model providers. We should probably do that before merging this.

pkg/api/handlers/modelprovider.go
iwilltry42: We're checking for this contained JSON blob in the frontend code so we only "extract the important part" in this error style - given that it's part of the widely used `useAsync` hook.

pkg/api/handlers/modelprovider.go
iwilltry42: found one in shell.go :+1:

pkg/api/handlers/modelprovider.go
iwilltry42: Likely not - leftover from when I didn't use GenerateName

pkg/controller/handlers/toolreference/toolreference.go
iwilltry42: done

pkg/api/handlers/modelprovider.go
iwilltry42: done - using a not found err

thedadams (approved):

iwilltry42:
Waiting for one last PR to be merged before we can merge this one: https://github.com/obot-platform/tools/pull/332

pkg/api/handlers/modelprovider.go
```diff
@@ -89,6 +95,78 @@ func (mp *ModelProviderHandler) List(req api.Context) error {
...
+	res, err := task.Result(req.Context())
+	if err != nil {
+		if strings.Contains(err.Error(), "tool not found: validate from "+ref.Spec.Reference) { // there's no simple way to do errors.As/.Is at this point unfortunately
+			log.Errorf("Model provider %q does not provide a validate tool. Looking for 'validate from %s'", ref.Name, ref.Spec.Reference)
+			return types.NewErrNotFound(
+				fmt.Sprintf("`validate from %s` tool not found", ref.Spec.Reference),
+				ref.Name,
+			)
+		}
+		return &ValidationError{Err: strings.Trim(err.Error(), "\"'")}
```
thedadams: When we return an error here, the API will return a 500. Can we return a 400-level error for these?

iwilltry42: 401 it is now :+1:

thedadams (approved):

PR #1044, opened 2024-12-23
feat: add support for authenticating on agents and workflows

thedadams:
Agents and workflows can have tools that require credentials. To authenticate these credentials, this change adds the ability to track which tools need which credentials, whether those credentials exist, and the ability to create and delete them.


thedadams:
There is a flaw in this. Please do not merge.

PR #1058, opened 2024-12-26
fix: continue showing tool calls in revisiting thread

ivyjeong13:
* instead of skipping events with toolCall.output, update the previous toolCall without an output with it during event phase
* allow showing of messages with toolCall.output

* tested with while running with a workflow, newly created agent 
* tested with going to Threads and visiting a pre-existing thread where a tool was used

### DIFF
--- a/ui/admin/app/components/chat/ChatContext.tsx
+++ b/ui/admin/app/components/chat/ChatContext.tsx
@@ -171,10 +171,8 @@ function useMessageSource(threadId?: Nullish<string>) {
                 return copy;
             }
 
-            // skip tool call output events
-            if (toolCall && !toolCall.output) {
-                copy.push(toolCallMessage(toolCall));
-                return copy;
+            if (toolCall) {
+                return handleToolCallEvent(copy, event);
             }
 
             if (prompt) {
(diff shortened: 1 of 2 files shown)


ui/admin/app/components/chat/ChatContext.tsx
```diff
@@ -232,3 +230,31 @@ function useMessageSource(threadId?: Nullish<string>) {
 
     return { messages, isRunning };
 }
+
+const findIndexLastPendingToolCall = (messages: Message[]) => {
+    for (let i = messages.length - 1; i >= 0; i--) {
+        const message = messages[i];
+        if (message.tools && !message.tools[0].output) {
```
njhale, line 237: nit: is `.output` the raw output of the tool being called? if so, is it possible for a tool that has finished running to have no output here?

njhale (approved):
One question, but otherwise lgtm

ui/admin/app/components/chat/ChatContext.tsx
ivyjeong13: Will do some verification to make sure! But it looked like `output` was always a summarized response what it was able to do with the tool and always provided at the finishing of the tool. (Ex. `"Created dataset with ID gds://4de0a with 12 issues\n{\"id\":\"gds://4de0a\",\"name\..."` or `"Sent message successfully"`)

Issue #816, opened 2024-12-07
When navigating to Agents list view from thread list specific to agent , thread list view gets loaded once before Agent list is shown.


sangee2004:
**Steps to reproduce the problem:**
**Pre Req** - Have few agents and have few conversations with the agent.
1. From Agent list view, navigate to thread list view for any agent.
2. From this thread list view , navigate to Agent list view using "Agents" breadcrumbs.
Notice that thread list view gets loaded once before Agent list is shown.
When the thread list view is loaded , notice that id of agent is shown in this page.

https://github.com/user-attachments/assets/7b1e93db-d151-4194-ab60-a56e5c1a2e8e

Note - This issue is more noticeable in slower setups (`test` env exhibits this issue).

sangee2004:
This issue is not seen anymore when testing in test env.

Issue #815, opened 2024-12-07
Agent name in Threads view shows as id  for conversation had with a new agent until manually refreshed.


sangee2004:
**Steps to reproduce the problem:**
1. Create an agent.
2. Have a conversation.
3. Go to Threads view
Notice that the thread relating to conversation from step 2 shows up as id instead of agent name
On refresh , we see the agent name as expected.

https://github.com/user-attachments/assets/003056b4-c984-4cb8-9162-f3d2a02e922e

sangee2004:
This issue is addressed when testing with latest build.

PR #1069, opened 2024-12-30
fix: agent id briefly showing on breadcrumb in threads

ivyjeong13:
Addresses #816 and #815

* fixing showing of agent id instead of agent name in Threads page

### DIFF
--- a/ui/admin/app/routes/_auth.agents._index.tsx
+++ b/ui/admin/app/routes/_auth.agents._index.tsx
@@ -24,7 +24,6 @@ import {
 } from "~/components/ui/tooltip";
 
 export async function clientLoader() {
-    mutate(AgentService.getAgents.key(), ThreadsService.getThreads.key());
     await Promise.all([
         preload(AgentService.getAgents.key(), AgentService.getAgents),
         preload(ThreadsService.getThreads.key(), ThreadsService.getThreads),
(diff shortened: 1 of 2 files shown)


ui/admin/app/routes/_auth.agents._index.tsx
```diff
@@ -24,7 +24,6 @@ import {
 } from "~/components/ui/tooltip";
 
 export async function clientLoader() {
-    mutate(AgentService.getAgents.key(), ThreadsService.getThreads.key());
```
ivyjeong13, line 27: Just wanted to make sure this is alright to remove!

What this was causing to happen:
-> Click "Agents" in breadcrumbs from Threads page with an agent filtered
-> `clientLoader` is called before the Agents view is rendered
-> (3) getAgents gets revalidated while Threads page is still visible which causes a temporary empty object. Because of that, it defaults to showing the filters.agentId instead. 
-> `clientLoader` completes and Agents view is rendered

Removing this will make it so (3) doesn't happen. And the useSWR for these two calls are already in Agents, so they should revalidate if stale. Just wanting to make sure there's something I'm not missing!

ryanhopperlowe: I know there was a bug that this line fixed (as a workaround), but go ahead and remove it. If it causes any issues, we may have to revert.

ryanhopperlowe (approved):

ui/admin/app/routes/_auth.agents._index.tsx
ivyjeong13: :+1: I'll keep an eye out and revert if it comes up!

PR #1085, opened 2024-12-31
feat: auth providers

g-linville:
This PR is a substantial refactoring of our auth providers so that they are now brought in as external tools rather than being built into the gateway. We also now allow multiple auth providers to be configured at the same time.

~~There is a known bug in the frontend where you have to refresh the page after configuring the auth provider in order to see the change. It's taken me long enough to get this PR ready, so I figured it would be best to fix that in a follow up (or have Ryan or Ivy figure it out for me, lol).~~  Fixed. Thanks, Ivy!

I will also do another follow-up PR to support deconfiguring auth providers. Currently, this is unimplemented in the backend.


pkg/accesstoken/accesstoken.go
```diff
@@ -0,0 +1,14 @@
...
+type accessTokenKey struct{}
+
+func ContextWithAccessToken(ctx context.Context, accessToken string) context.Context {
+	return context.WithValue(ctx, accessTokenKey{}, accessToken)
+}
+
+func GetAccessToken(ctx context.Context) string {
+	accessToken, _ := ctx.Value(accessTokenKey{}).(string)
+	return accessToken
+}
```
njhale, line 14: nit: compiler optimizations aside, you can avoid extra allocations by creating the key object once, up front:

```go
package accesstoken

import "context"

var accessTokenKey = struct{}{}

func ContextWithAccessToken(ctx context.Context, accessToken string) context.Context {
	return context.WithValue(ctx, accessTokenKey, accessToken)
}

func GetAccessToken(ctx context.Context) string {
	accessToken, _ := ctx.Value(accessTokenKey).(string)
	return accessToken
}
```

pkg/api/handlers/authprovider.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	if err := ap.gptscript.CreateCredential(req.Context(), gptscript.Credential{
+		Context:  string(ref.UID),
+		ToolName: ref.Name,
+		Type:     gptscript.CredentialTypeTool,
+		Env:      envVars,
+	}); err != nil {
+		return fmt.Errorf("failed to create credential for auth provider %q: %w", ref.Name, err)
+	}
```
njhale, line 138: If the env hasn't changed this will still delete and recreate the credential and stop the provider even the env hasn't changed, right? Would it be better to check for changes to the credential first?

ui/admin/app/components/auth-and-model-providers/ProviderForm.tsx
```diff
@@ -0,0 +1,332 @@
...
+			},
+		}
+	);
+
+	const configureAuthProvider = useAsync(
+		AuthProviderApiService.configureAuthProviderById,
+		{
+			onSuccess: async () => {
+				mutate(AuthProviderApiService.revealAuthProviderById.key(provider.id));
+				onSuccess();
```
ivyjeong13, line 129: > There is a known bug in the frontend where you have to refresh the page after configuring the auth provider in order to see the change. It's taken me long enough to get this PR ready, so I figured it would be best to fix that in a follow up (or have Ryan or Ivy figure it out for me, lol).

My initial guess is you may need to add `mutate(AuthProviderApiService.getAuthProviders.key())` here! That should get auth providers updated after a successful configure.

pkg/gateway/server/oauth.go
```diff
@@ -15,20 +15,26 @@ import (
...
+		return types2.NewErrHttp(http.StatusBadRequest, "no name path parameter provided")
+	}
+
+	// Check to make sure this auth provider exists.
+	list, err := s.dispatcher.ListConfiguredAuthProviders(apiContext.Context(), namespace)
+	if err != nil {
+		return fmt.Errorf("could not list configured auth providers: %w", err)
+	} else if !slices.Contains(list, name) {
+		return types2.NewErrHttp(http.StatusNotFound, "auth provider not found")
 	}
```
njhale, line 38: Could you just `Get` the ToolReference for the specific auth provider by name and check if it's configured (and that it's an authprovider type tool)?. Would result in way fewer operations IIUC.

ui/admin/app/components/auth-and-model-providers/ProviderForm.tsx
```diff
@@ -0,0 +1,332 @@
...
+	optionalParameters: string[],
+	parameters: ProviderConfig
+): ProviderFormValues["optionalConfigParams"] =>
+	optionalParameters.map((optionalParameterKey) => ({
+		label: translateUserFriendlyLabel(optionalParameterKey),
+		name: optionalParameterKey,
+		value: parameters[optionalParameterKey] ?? "",
+	}));
+
+export function ProviderForm({
```
ivyjeong13, line 102: nit: Wondering we could break the model provider / auth provider logic into a parent component and pass that in as props to ProviderForm, ProviderConfigure, ProviderDeconfigure. 

Thinking something like: 
```
// called in auth_providers.tsx specific context
<ProviderForm
    onConfigure={configureAuthProvider}
    loading={configureAuthProvider.isLoading || isLoading}
    tooltips={AuthProviderRequiredTooltips}
/>

// called in model_providers.tsx specific context
<ProviderForm
    onConfigure={configureModelProvider}
    loading={configureModelProvider.isLoading || isLoading}
    tooltips={ModelProviderRequiredTooltips}
/>
```

Then when you have to call them, you're not worried about handling the other provider type cases (or if something else uses this in the future).

ui/admin/app/components/auth-and-model-providers/ProviderForm.tsx
g-linville: Thanks, this worked!

ui/admin/app/components/auth-and-model-providers/ProviderForm.tsx
g-linville: If it's alright, I'd like for this to be in a follow-up.

ryanhopperlowe:
Screenshot 2025-01-16 at 3 47 00 PM

I'm being prompted to setup an auth provider, but I don't see any to pick from

ui/admin/app/components/auth-and-model-providers/Bootstrap.tsx
```diff
@@ -0,0 +1,62 @@
...
+				setError("Failed to login: " + result.statusText);
+				return;
+			}
+
+			setError("");
+			window.location.href = "/admin/auth-providers";
+		} catch (e) {
+			setError("Failed to login: " + e);
+		}
+	};
```
ryanhopperlowe, line 39: Let's Create a `BootstrapService` for all of these methods

ui/admin/app/components/auth-and-model-providers/ProviderDeconfigure.tsx
```diff
@@ -0,0 +1,110 @@
...
+							</Button>
+						</DialogClose>
+					</div>
+				</DialogFooter>
+			</DialogContent>
+		</Dialog>
+	);
+}
+
+function warningMessage(t: string | undefined): string | undefined {
```
ryanhopperlowe: nit: 

```suggestion
function warningMessage(t?: string) {
```

ui/admin/app/components/auth-and-model-providers/ProviderIcon.tsx
```diff
@@ -0,0 +1,34 @@
...
+		<img
+			src={provider.icon}
+			alt={provider.name}
+			className={cn({
+				"h-6 w-6": size === "md",
+				"h-16 w-16": size === "lg",
+				"dark:invert":
+					provider.id !== CommonModelProviderIds.AZURE_OPENAI &&
+					provider.id !== CommonAuthProviderIds.GOOGLE &&
+					provider.id !== CommonAuthProviderIds.GITHUB,
```
ryanhopperlowe: nit:
```suggestion
					![CommonModelProviderIds.AZURE_OPENAI, CommonAuthProviderIds.GOOGLE, CommonAuthProviderIds.GITHUB].includes(provider.id)
```

ui/admin/app/lib/model/providers.ts
```diff
@@ -1,18 +1,24 @@
...
 
-export type ModelProviderStatus = {
+export type ProviderStatus = {
 	configured: boolean;
-	modelsBackPopulated?: boolean;
 	icon?: string;
 	requiredConfigurationParameters?: string[];
+	optionalConfigurationParameters?: string[];
 	missingConfigurationParameters?: string[];
 };
```
ryanhopperlowe, line 9: These should follow a more True-to-life pattern:

```ts

type ProviderBase = {...}

type ModelProviderBase = ProviderBase & {...}
type ModelProvider = EntityMeta & ProviderBase & { type: 'modelprovider' }

type AuthProviderBase = ProviderBase & {...}
type AuthProvider = EntityMeta & AuthProviderBase & { type: 'authprovider' }

type Provider = ModelProvider | AuthProvider
```

ui/admin/app/routes/_auth.tsx
```diff
@@ -47,6 +48,8 @@ export function ErrorBoundary() {
 	switch (true) {
 		case error instanceof UnauthorizedError:
 		case error instanceof ForbiddenError:
+		case error instanceof AxiosError &&
+			[401, 403].includes(error.response?.status ?? 0):
```
ryanhopperlowe, line 52: @g-linville can you explain why you needed to add this? I'd like to make sure our first layer of error handling can catch this and convert it to the proper error type

g-linville:
@ryanhopperlowe you need to download and build my branch of the tools repo from this PR: https://github.com/obot-platform/tools/pull/320

And then set the `OBOT_SERVER_TOOL_REGISTRY` environment variable to point to your local copy of it when you run `make dev`. This will make the auth providers show up.

ui/admin/app/components/auth-and-model-providers/Bootstrap.tsx
g-linville: Sounds good. If it's okay, I'd like for that to be a follow-up refinement. For this PR, the UI just needs to work. Non-optimal things about the code can be fixed afterward.

ui/admin/app/lib/model/providers.ts
g-linville: I'd like for this to be a follow-up improvement.

ui/admin/app/routes/_auth.tsx
g-linville: I don't 100% remember but I think it had to do with the error we get back when we hit `/api/me` with my new changes to the backend.

ui/admin/app/components/auth-and-model-providers/ProviderIcon.tsx
g-linville: Fixed

ui/admin/app/components/auth-and-model-providers/ProviderDeconfigure.tsx
g-linville: Fixed

g-linville:
Sorry for force pushing again. There were more conflicts that I needed to fix.

njhale (approved):
Some nits and suggestions, nothing that should hold this up.

njhale:
> ... And then set the OBOT_SERVER_TOOL_REGISTRY environment variable to point to your local copy of it when you run make dev. This will make the auth providers show up.

@g-linville @ryanhopperlowe this has recently changed due to https://github.com/obot-platform/obot/pull/1193 merging.

Now, instead of setting `OBOT_SERVER_TOOL_REGISTRY`, you'll need set the `GPTSCRIPT_TOOL_REMAP` environment variable; e.g.

```
export GPTSCRIPT_TOOL_REMAP='github.com/obot-platform/tools=/Users/nick/projects/obot-platform/tools'
```

pkg/accesstoken/accesstoken.go
njhale: We can refine this whenever.

pkg/gateway/server/oauth.go
njhale: We can refine this later.

pkg/api/handlers/authprovider.go
g-linville: All I did here was copy the logic that we used for model providers. I think @thedadams may have been the one who implemented that, and would know the answer to this.

g-linville:
@njhale @ryanhopperlowe for testing with this branch, the old env var should be used to set the registry, since I have not (yet) brought in those changes from main to this branch.

pkg/api/handlers/authprovider.go
```diff
@@ -0,0 +1,245 @@
...
+		return types.NewErrNotFound(
+			"auth provider %q not found",
+			ref.Name,
+		)
+	}
+
+	var credEnvVars map[string]string
+	if ref.Status.Tool != nil {
+		if envVars := ref.Status.Tool.Metadata["envVars"]; envVars != "" {
+			fmt.Printf("revealing creds for auth provider %q\n", ref.Name)
```
StrongMonkey: Nit: this is debug log? if not should we use our standard log library to print?

pkg/api/handlers/authprovider.go
```diff
@@ -0,0 +1,245 @@
...
+	var ref v1.ToolReference
+	if err := req.Get(&ref, req.PathValue("id")); err != nil {
+		return err
+	}
+
+	if ref.Spec.Type != types.ToolReferenceTypeAuthProvider {
+		return types.NewErrBadRequest("%q is not an auth provider", ref.Name)
+	}
+
+	fmt.Printf("revealing creds for auth provider %q\n", ref.Name)
```
StrongMonkey: Debug log

pkg/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,117 @@
...
+		bytes := make([]byte, 32)
+		_, err := rand.Read(bytes)
+		if err != nil {
+			return nil, fmt.Errorf("failed to generate random token: %w", err)
+		}
+
+		token = fmt.Sprintf("%x", bytes)
+
+		// We deliberately only print the token if it was not provided by the user.
+		fmt.Printf("Bootstrap token: %s\n", token)
```
StrongMonkey: Do we print it so that user can grab it from the log? if so I would suggest the log to be more descriptive.

pkg/api/handlers/authprovider.go
g-linville: Thanks for finding these! I forgot to remove them before. Pushed the fix

pkg/api/handlers/authprovider.go
g-linville: Fixed

pkg/bootstrap/bootstrap.go
g-linville: Fixed

g-linville:
Rebased again to fix conflicts.

apiclient/types/authprovider.go
```diff
@@ -0,0 +1,23 @@
...
+
+type AuthProvider struct {
+	Metadata
+	AuthProviderManifest
+	AuthProviderStatus
+}
+
+type AuthProviderManifest struct {
+	Name          string `json:"name"`
+	Namespace     string `json:"namespace"`
```
thedadams, line 11: This shouldn't be part of the manifest.

pkg/api/router/router.go
```diff
@@ -36,6 +37,7 @@ func Router(services *services.Services) (http.Handler, error) {
 	sendgridWebhookHandler := sendgrid.NewInboundWebhookHandler(services.StorageClient, services.EmailServerName, services.SendgridWebhookUsername, services.SendgridWebhookPassword)
 
 	// Version
+
```
thedadams: nit: remove line

pkg/api/server/server.go
```diff
@@ -57,35 +56,20 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 			http.Error(rw, err.Error(), http.StatusUnauthorized)
 			return
 		}
 
-		isOAuthPath := strings.HasPrefix(req.URL.Path, "/oauth2/")
-		if isOAuthPath || strings.HasPrefix(req.URL.Path, "/api/") && !s.authorizer.Authorize(req, user) {
-			// If this is not a request coming from browser or the proxy is not enabled, then return 403.
-			if !isOAuthPath && (s.proxyServer == nil || req.Method != http.MethodGet || slices.Contains(user.GetGroups(), authz.AuthenticatedGroup) || !strings.Contains(strings.ToLower(req.UserAgent()), "mozilla")) {
+		if strings.HasPrefix(req.URL.Path, "/api/") {
+			if !s.authorizer.Authorize(req, user) {
```
thedadams: This if-statement should not be nested.

pkg/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,117 @@
...
+		MaxAge:   60 * 60 * 24 * 7, // 1 week
+		HttpOnly: true,
+		Secure:   strings.HasPrefix(b.serverURL, "https://"),
+	})
+	http.Redirect(req.ResponseWriter, req.Request, "/admin/auth-providers", http.StatusFound)
+
+	return nil
+}
+
+func (b *Bootstrap) Logout(req api.Context) error {
```
thedadams, line 105: Can we delete the `bootstrap` user from the database in this case? If it doesn't make sense to do here, can we do it at some point?

pkg/gateway/server/dispatcher/dispatcher.go
```diff
@@ -21,15 +21,18 @@ import (
...
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type Dispatcher struct {
 	invoker    *invoke.Invoker
 	gptscript  *gptscript.GPTScript
 	client     kclient.Client
-	lock       *sync.RWMutex
-	urls       map[string]*url.URL
+	modelLock  *sync.RWMutex
```
thedadams, line 32: nit: the dispatcher can be separated from the "transformer" and the dispatcher made generic with the implementation used for both model and auth providers.

pkg/gateway/server/dispatcher/dispatcher.go
```diff
@@ -251,3 +299,111 @@ func readBody(r *http.Request) (map[string]any, error) {
...
+	if err := d.client.Get(ctx, kclient.ObjectKey{Namespace: namespace, Name: authProviderName}, &authProvider); err != nil || authProvider.Spec.Type != types.ToolReferenceTypeAuthProvider {
+		return nil, fmt.Errorf("failed to get auth provider: %w", err)
+	}
+
+	credCtx := []string{string(authProvider.UID)}
+	if authProvider.Status.Tool == nil {
+		return nil, fmt.Errorf("auth provider %q has not been resolved", authProviderName)
+	}
+
+	// Ensure that the auth provider has been configured so that we don't get stuck waiting on a prompt.
```
thedadams, line 332: Can we use the `isAuthProviderConfigured` method here?

pkg/gateway/server/tokenreview.go
```diff
@@ -20,14 +20,15 @@ func (s *Server) AuthenticateRequest(req *http.Request) (*authenticator.Response
...
-	var authProviderID string
+	var namespace, name string
 	if err := s.db.WithContext(req.Context()).Transaction(func(tx *gorm.DB) error {
 		tkn := new(types.AuthToken)
 		if err := tx.Where("id = ? AND hashed_token = ?", id, hashToken(token)).First(tkn).Error; err != nil {
 			return err
 		}
 
-		authProviderID = fmt.Sprint(tkn.AuthProviderID)
+		namespace = fmt.Sprint(tkn.AuthProviderNamespace)
```
thedadams: Is it necessary to use `fmt.Sprintf` here?

pkg/proxy/proxy.go
```diff
@@ -2,151 +2,233 @@ package proxy
...
+				rdParam := r.URL.Query().Get("rd")
+				if rdParam == "" {
+					rdParam = "/"
+				}
+
+				http.Redirect(w, r, rdParam, http.StatusFound)
+				return
+			}
+
+			http.Error(w, "no auth providers configured", http.StatusInternalServerError)
```
thedadams: Should this be a bad request?

pkg/proxy/proxy.go
```diff
@@ -2,151 +2,233 @@ package proxy
...
+	Email             string     `json:"email"`
+}
+
+func (p *Proxy) authenticateRequest(req *http.Request) (*authenticator.Response, bool, error) {
+	sr := SerializableRequest{
+		Method: req.Method,
+		URL:    req.URL.String(),
+		Header: make(map[string][]string),
+	}
+	for k, v := range req.Header {
```
thedadams: sr.Header = req.Header

pkg/services/config.go
```diff
@@ -62,16 +60,18 @@ type Config struct {
...
 	EncryptionConfigFile       string   `usage:"The path to the encryption configuration file" default:"./encryption.yaml"`
-	KnowledgeSetIngestionLimit int      `usage:"The maximum number of files to ingest into a knowledge set" default:"3000" env:"OBOT_KNOWLEDGESET_INGESTION_LIMIT" name:"knowledge-set-ingestion-limit"`
 	EmailServerName            string   `usage:"The name of the email server to display for email receivers"`
 	Docker                     bool     `usage:"Enable Docker support" default:"false" env:"OBOT_DOCKER"`
 	EnvKeys                    []string `usage:"The environment keys to pass through to the GPTScript server" env:"OBOT_ENV_KEYS"`
+	KnowledgeSetIngestionLimit int      `usage:"The maximum number of files to ingest into a knowledge set" default:"3000" env:"OBOT_KNOWLEDGESET_INGESTION_LIMIT" name:"knowledge-set-ingestion-limit"`
+	NoReplyEmailAddress        string   `usage:"The email to use for no-reply emails from obot"`
+	DisableAuthentication      bool     `usage:"Disable authentication" default:"false" env:"OBOT_DISABLE_AUTHENTICATION"`
```
thedadams: If one were to use this as a CLI flag, using `--disable-authentication` seems that it would actually enable authentication.

pkg/services/config.go
```diff
@@ -62,16 +60,18 @@ type Config struct {
...
 	AWSKMSKeyARN               string   `usage:"The ARN of the AWS KMS key to use for encrypting credential storage" env:"OBOT_AWS_KMS_KEY_ARN" name:"aws-kms-key-arn"`
 	EncryptionConfigFile       string   `usage:"The path to the encryption configuration file" default:"./encryption.yaml"`
-	KnowledgeSetIngestionLimit int      `usage:"The maximum number of files to ingest into a knowledge set" default:"3000" env:"OBOT_KNOWLEDGESET_INGESTION_LIMIT" name:"knowledge-set-ingestion-limit"`
 	EmailServerName            string   `usage:"The name of the email server to display for email receivers"`
 	Docker                     bool     `usage:"Enable Docker support" default:"false" env:"OBOT_DOCKER"`
 	EnvKeys                    []string `usage:"The environment keys to pass through to the GPTScript server" env:"OBOT_ENV_KEYS"`
+	KnowledgeSetIngestionLimit int      `usage:"The maximum number of files to ingest into a knowledge set" default:"3000" env:"OBOT_KNOWLEDGESET_INGESTION_LIMIT" name:"knowledge-set-ingestion-limit"`
+	NoReplyEmailAddress        string   `usage:"The email to use for no-reply emails from obot"`
```
thedadams: Is this related to your changes?

pkg/services/config.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/obot-platform/obot/pkg/api/authn"
 	"github.com/obot-platform/obot/pkg/api/authz"
 	"github.com/obot-platform/obot/pkg/api/server"
+	bootstrap2 "github.com/obot-platform/obot/pkg/bootstrap"
```
thedadams: nit: I don't think this import needs to be renamed.

thedadams (changes requested):

apiclient/types/authprovider.go
g-linville: Are you sure? `Name` is part of the ModelProviderManifest. We need a namespace for auth providers, since I was told I can't assume that they will always be in `default`.

pkg/api/server/server.go
g-linville: If we bring it outside of `if strings.HasPrefix(req.URL.Path, "/api/")`, then we end up denying all unauthenticated requests to the user UI, which basically breaks everything. It needs to be nested in here.

pkg/bootstrap/bootstrap.go
g-linville: I'm okay with doing this, but I'm not sure that it's really necessary?

pkg/gateway/server/tokenreview.go
g-linville: Oh, probably not.

pkg/proxy/proxy.go
g-linville: yeah

pkg/services/config.go
g-linville: Good point. I'll invert this

pkg/services/config.go
g-linville: Nope. Might have been an accident when trying to resolve conflicts. Maybe this got added and removed at some point? I can take it away.

pkg/gateway/server/dispatcher/dispatcher.go
g-linville: I'd like for this to be a later change if it is needed.

pkg/api/server/server.go
thedadams: If that is the case, then that should be fixed outside of this. This definitely should not be nested.

pkg/gateway/server/dispatcher/dispatcher.go
thedadams: I agree. Which is why it is a nit.

thedadams (approved):

PR #1086, opened 2024-12-31
enhance: add no-reply email to server config

njhale:
Add a no-reply email address option to the Obot server configuration. This setting is passed down to running tools so that the sendgrid tool can send emails using a pre-configured no-reply address.

### DIFF
--- a/pkg/invoke/invoker.go
+++ b/pkg/invoke/invoker.go
@@ -35,24 +35,26 @@ import (
 var log = logger.Package()
 
 type Invoker struct {
-	gptClient     *gptscript.GPTScript
-	uncached      kclient.WithWatch
-	gatewayClient *client.Client
-	tokenService  *jwt.TokenService
-	events        *events.Emitter
-	serverURL     string
-	serverPort    int
+	gptClient           *gptscript.GPTScript
+	uncached            kclient.WithWatch
+	gatewayClient       *client.Client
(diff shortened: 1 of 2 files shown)


pkg/invoke/invoker.go
```diff
@@ -35,24 +35,26 @@ import (
...
-	tokenService  *jwt.TokenService
-	events        *events.Emitter
-	serverURL     string
-	serverPort    int
+	gptClient           *gptscript.GPTScript
+	uncached            kclient.WithWatch
+	gatewayClient       *client.Client
+	tokenService        *jwt.TokenService
+	events              *events.Emitter
+	noReplyEmailAddress string
```
njhale, line 43: should this configuration name specifically mention SendGrid; e.g. 'sendGridNoReplyAddress`?

Should it even be a full address? Maybe something like `emailDomain` (to match the version API) is better, and we just use a default address like `obot-noreply@<emailDomain>`?

(these questions apply for the environment variable names plumbed through as well)

g-linville: I don't think it should specifically mention SendGrid. It is likely this would be useful for other similar tools we might add in the future.

I do like the idea of just using an `emailDomain` rather than requiring them to specify the full address, but I personally think it is fine either way.

g-linville (approved):

Issue #1071, opened 2024-12-30
99.999 ingestion success rate


cjellick:
This should be the target for all knowledge sources - website, notion, onedrive and includes the whole process end-to-end.

StrongMonkey:
1. We should add some high-level retry logic to avoid external system failure(like openai flakiness). Should be addressed by https://github.com/obot-platform/obot/pull/1089
2. We should fix some bottleneck issues that are prune to failed. Issues like https://github.com/obot-platform/obot/issues/926 are easy to cause failure during ingestion.

shweta8996:
@cjellick Can you please provide the steps to reproduce this issue and specify the environment where it can be replicated?

cjellick:
I can and will, but dont want to focus on it at this minute. I will revisit later. possibly next week

shweta8996:
Hey @cjellick 
Can I focus on this issue this week? If yes, can you please provide the

PR #1089, opened 2024-12-31
Chore: Add ability to retry ingestion up to 3 times

StrongMonkey:
Add ability to re-enqueue the file ingestion process up to 3 times when it fails. 

https://github.com/obot-platform/obot/issues/1071

### DIFF
--- a/pkg/controller/handlers/knowledgefile/knowledgefile.go
+++ b/pkg/controller/handlers/knowledgefile/knowledgefile.go
@@ -46,7 +46,8 @@ func shouldReIngest(file *v1.KnowledgeFile) bool {
 	return file.Spec.IngestGeneration > file.Status.IngestGeneration ||
 		file.Spec.UpdatedAt != file.Status.UpdatedAt ||
 		file.Spec.Checksum != file.Status.Checksum ||
-		file.Spec.URL != file.Status.URL
+		file.Spec.URL != file.Status.URL ||
+		(file.Status.State == types.KnowledgeFileStateError && file.Status.RetryCount < 3)
 }
 
 func cleanInput(filename string) string {
@@ -167,9 +168,11 @@ func (h *Handler) IngestFile(req router.Request, _ router.Response) error {
 			file.Status.State = types.KnowledgeFileStateError
 		}
 		file.Status.Error = err.Error()
+		file.Status.RetryCount++
 	} else {
 		file.Status.State = types.KnowledgeFileStateIngested
 		file.Status.Error = ""
+		file.Status.RetryCount = 0
 	}
 
 	file.Status.LastIngestionEndTime = metav1.Now()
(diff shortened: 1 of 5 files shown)


pkg/controller/handlers/knowledgesource/knowledgesource.go
```diff
@@ -258,7 +258,7 @@ func (k *Handler) Sync(req router.Request, _ router.Response) error {
 
 	invokeOpts := invoke.SystemTaskOptions{
 		CredentialContextIDs: []string{credentialContextID},
-		Timeout:              1 * time.Hour,
+		Timeout:              2 * time.Hour,
```
StrongMonkey, line 261: Increase the time since we want to increase the number of pages our scraper can scrape and sometimes it can take up more than 1 hour to finish.

pkg/services/config.go
```diff
@@ -61,7 +61,7 @@ type Config struct {
 	HelperModel                string `usage:"The model used to generate names and descriptions" default:"gpt-4o-mini"`
 	AWSKMSKeyARN               string `usage:"The ARN of the AWS KMS key to use for encrypting credential storage" env:"OBOT_AWS_KMS_KEY_ARN" name:"aws-kms-key-arn"`
 	EncryptionConfigFile       string `usage:"The path to the encryption configuration file" default:"./encryption.yaml"`
-	KnowledgeSetIngestionLimit int    `usage:"The maximum number of files to ingest into a knowledge set" default:"1000" env:"OBOT_KNOWLEDGESET_INGESTION_LIMIT" name:"knowledge-set-ingestion-limit"`
+	KnowledgeSetIngestionLimit int    `usage:"The maximum number of files to ingest into a knowledge set" default:"3000" env:"OBOT_KNOWLEDGESET_INGESTION_LIMIT" name:"knowledge-set-ingestion-limit"`
```
StrongMonkey, line 64: Increasing the number of files one agent can ingest, since we should be able to ingest 3000 files per agent without any error. 

Note: we haven't really systematically tested whether increasing the number of ingested file will deteriorate the retrival result. This is just the success rate of ingestion.

g-linville (approved):

pkg/controller/handlers/knowledgefile/knowledgefile.go
```diff
@@ -46,7 +46,8 @@ func shouldReIngest(file *v1.KnowledgeFile) bool {
 	return file.Spec.IngestGeneration > file.Status.IngestGeneration ||
 		file.Spec.UpdatedAt != file.Status.UpdatedAt ||
 		file.Spec.Checksum != file.Status.Checksum ||
-		file.Spec.URL != file.Status.URL
+		file.Spec.URL != file.Status.URL ||
+		(file.Status.State == types.KnowledgeFileStateError && file.Status.RetryCount < 3)
```
iwilltry42, line 50: Should we parameterize the number of retries? Maybe even give it an undocumented env for testing?

iwilltry42 (approved):

pkg/controller/handlers/knowledgefile/knowledgefile.go
StrongMonkey: I think let's not complicate this for now. It should be enough to resolve most transient error and have proper retry.

Issue #873, opened 2024-12-16
Add scheduled triggers for workflows to the UI


cjellick:
Backend already exists for this. Frontend already exists in the user UI.

When you do this, I'd like both webhook and scheduled triggers to be on the workflow page itself. So, I should be able to create a workflow and configure it's triggers without ever leaving that page. Additionally, the webhooks page should become "Webhook Triggers" and list both types of triggers.

ivyjeong13:
Oops forgot to rename the Webhooks page to Webhook Triggers and list both, pulling back to complete those!

sangee2004:
All issues tracked in this bug are addressed in latest build.

1. Able to create a workflow and configure it's triggers without ever leaving the page.
2. Able to create workflow and add scheduled triggers.
3. `Webhooks` page is now renamed to `Workflow Triggers` and both scheduled and webhook triggers are listed in this view.

PR #1096, opened 2025-01-02
enhance: add schedule & webhook options into workflows

ivyjeong13:
Addresses #873

* adds scheduling and webhooks into updating/creating workflows
* reuses webhook form

https://www.loom.com/share/d5cba91f5a334a6392bb2d0c1bb5ac01

Screenshot 2025-01-02 at 3 30 09 PM

### DIFF
--- a/ui/admin/app/components/workflow/WorkflowSchedulePanel.tsx
+++ b/ui/admin/app/components/workflow/WorkflowSchedulePanel.tsx
@@ -1,0 +1,189 @@
+import { ClockIcon } from "lucide-react";
+
+import { TypographyH4 } from "~/components/Typography";
+import { CardDescription } from "~/components/ui/card";
+import { Label } from "~/components/ui/label";
+import {
+    Select,
+    SelectContent,
+    SelectItem,
+    SelectTrigger,
+    SelectValue,
+} from "~/components/ui/select";
+import { Switch } from "~/components/ui/switch";
+import { useCronjob } from "~/hooks/cronjob/useCronjob";
+
+export function WorkflowSchedulePanel({ workflowId }: { workflowId: string }) {
+    const { cronJob, createCronJob, deleteCronJob, updateCronJob } =
+        useCronjob(workflowId);
+    const cronFrequency = getCronFrequency(cronJob?.schedule ?? "");
+    const timeOptions = cronFrequency
+        ? getTimeOptionsForInterval(cronFrequency)
+        : [];
+
+    const hasCronJob = !!cronJob;
+
+    const handleCheckedChange = (checked: boolean) => {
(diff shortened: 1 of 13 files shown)


ui/admin/app/components/workflow/WorkflowSchedulePanel.tsx
```diff
@@ -0,0 +1,195 @@
...
+                { label: "Wednesday at Midnight", value: "0 0 * * 3" },
+                { label: "Friday at Midnight", value: "0 0 * * 5" },
+            ];
+        case "monthly":
+            return [
+                { label: "1st at Midnight", value: "0 0 1 * *" },
+                { label: "15th at Midnight", value: "0 0 15 * *" },
+                {
+                    label: "Last Day at Midnight",
+                    value: "0 0 L * *",
```
ivyjeong13, line 183: TBD on this option: After syncing with Donnie, we determined this cron string (though valid) does not work with the existing cron parser. Similarly, it is not working in user ui (error happens during cron run).

ui/admin/app/components/workflow/WorkflowSchedulePanel.tsx
```diff
@@ -0,0 +1,189 @@
...
+}
+
+function getCronFrequency(
+    cronString: string
+): "hourly" | "daily" | "weekly" | "monthly" | null {
+    const patterns = {
+        hourly: /^(0|\*\/\d+)\s+\*\s+\*\s+\*\s+\*$/, // ex. "0 * * * *" or "*/15 * * * *"
+        daily: /^0\s+\d+\s+\*\s+\*\s+\*$/, // ex. "0 6 * * *"
+        weekly: /^0\s+\d+\s+\*\s+\*\s+\d$/, // ex. "0 0 * * 3"
+        monthly: /^0\s+\d+\s+\d+\s+\*\s+\*$/, // "0 0 15 * *"
```
ryanhopperlowe: `\s` is any whitespace character including tab, newline, etc. right? is this expected?

ivyjeong13: Nope, not expected! I think technically cron strings can have tabs / new lines and still be valid but we should be passing them with just standard spacing. I'll go ahead and change them to just validate with spaces. :+1:

ui/admin/app/components/workflow/WorkflowSchedulePanel.tsx
ivyjeong13: Option being kept! New cron package will be switched in next week to allow for 'L' validating.

ryanhopperlowe (approved):
thanks!

PR #1106, opened 2025-01-03
Chore: use SSE to deliver change event for knowledgefiles

StrongMonkey:
We used to pull knowledge at an interval but this is not effiencient. Switching to use SSE events to watch for changes.

### DIFF
--- a/pkg/api/handlers/agent.go
+++ b/pkg/api/handlers/agent.go
@@ -15,7 +15,6 @@ import (
 	"github.com/obot-platform/obot/pkg/api"
 	"github.com/obot-platform/obot/pkg/invoke"
 	"github.com/obot-platform/obot/pkg/render"
-	"github.com/obot-platform/obot/pkg/storage"
 	v1 "github.com/obot-platform/obot/pkg/storage/apis/otto.otto8.ai/v1"
 	"github.com/obot-platform/obot/pkg/storage/selectors"
 	"github.com/obot-platform/obot/pkg/system"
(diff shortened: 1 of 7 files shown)


ui/admin/app/components/knowledge/FileTree.tsx
```diff
@@ -141,9 +141,6 @@ export default function FileTreeNode({
             filePathPrefixInclude,
             filePathPrefixExclude,
         });
-
-        // We should manually approve/unapprove all files in the folder at once so that we don't rely on backend reconciliation logic as it will cause delay in updating the UI.
```
StrongMonkey, line 145: It seems not ideal to approve files in frontend just for fast feedback from UI. I think it is ok to just rely on backend to reconcile all the files to be approved.

ryanhopperlowe (approved):

ui/admin/app/hooks/knowledge/useKnowledgeSourceFiles.ts
```diff
@@ -29,27 +28,55 @@ export function useKnowledgeSourceFiles(
...
-            revalidateOnFocus: false,
-            refreshInterval: blockPollingFiles ? undefined : 5000,
-        }
-    );
+                knowledgeSource.id
+            );
+
+        eventSource.onmessage = (event) => {
+            const payload = JSON.parse(event.data);
+            const { eventType, knowledgeFile } = payload;
```
ryanhopperlowe: can you create a type for this?

pkg/api/handlers/agent.go
```diff
@@ -844,6 +850,71 @@ func (a *AgentHandler) Script(req api.Context) error {
...
+	}()
+
+	req.ResponseWriter.Header().Set("Content-Type", "text/event-stream")
+	defer func() {
+		_ = req.WriteDataEvent(api.EventClose{})
+	}()
+
+	for event := range w.ResultChan() {
+		if knowledgeFile, ok := event.Object.(*v1.KnowledgeFile); ok {
+			payload := map[string]interface{}{
```
thedadams: nit:
```suggestion
			payload := map[string]any{
```

pkg/api/router/router.go
```diff
@@ -11,7 +11,7 @@ import (
 func Router(services *services.Services) (http.Handler, error) {
 	mux := services.APIServer
 
-	agents := handlers.NewAgentHandler(services.GPTClient, services.Invoker, services.ServerURL)
+	agents := handlers.NewAgentHandler(services.GPTClient, services.Invoker, services.ServerURL, services.StorageClient)
```
thedadams: The `api.Context` object passed to each of these handlers already has a client. Can we just use that one instead?

thedadams (changes requested):

pkg/api/router/router.go
StrongMonkey: Good call! addressed

thedadams (approved):

PR #1107, opened 2025-01-03
enhance: plumb tz from client-side uis to tools

njhale:
Make the [tz database timezone ID](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) for the default timezone of a user available to tools at runtime via the `OBOT_USER_TIMEZONE` environment variable.

e.g. `OBOT_USER_TIMEZONE=America/New_York`

TThe timezone ID is captured by the admin and user UIs from the browser and passed using the `X-Obot-User-Timezone` header on requests to Obot's APIs. The header is then unpacked in Obot's Authentication middleware and persisted in the respective `User` object. When a thread is invoked, its value is extracted from the `User` and passed to tool calls via the aforementioned environment variable.

The user's timezone can be customized independently of this defaulting mechanism by issuing a PATCH on `/api/users/{username}` with a body containing the desired timezone. Existing non-empty timezone fields on `User` objects take precedence over the timezone header.

Partially addresses https://github.com/obot-platform/obot/issues/1111

### DIFF
--- a/pkg/api/authn/noauth.go
+++ b/pkg/api/authn/noauth.go
@@ -1,20 +1,44 @@
 package authn
 
 import (
+	"fmt"
 	"net/http"
 
+	types2 "github.com/obot-platform/obot/apiclient/types"
 	"github.com/obot-platform/obot/pkg/api/authz"
+	"github.com/obot-platform/obot/pkg/gateway/client"
+	"github.com/obot-platform/obot/pkg/gateway/types"
 	"k8s.io/apiserver/pkg/authentication/authenticator"
 	"k8s.io/apiserver/pkg/authentication/user"
 )
 
 type NoAuth struct {
+	client *client.Client
 }
 
-func (n NoAuth) AuthenticateRequest(*http.Request) (*authenticator.Response, bool, error) {
+func NewNoAuth(client *client.Client) *NoAuth {
+	return &NoAuth{
+		client: client,
+	}
+}
+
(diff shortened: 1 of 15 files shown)


njhale:
There are some issues with this. Putting into draft until I fix it.

g-linville (approved):

pkg/api/authn/noauth.go
```diff
@@ -1,20 +1,44 @@
...
+		client: client,
+	}
+}
+
+func (n *NoAuth) AuthenticateRequest(req *http.Request) (*authenticator.Response, bool, error) {
+	gatewayUser, err := n.client.EnsureIdentityWithRole(
+		req.Context(),
+		&types.Identity{
+			ProviderUsername: "nobody",
+		},
```
thedadams, line 30: We don't want this user in the database.

thedadams (changes requested):

pkg/api/authn/noauth.go
thedadams: Never mind. This was misguided and Nick has shown me the light.

thedadams (approved):

pkg/api/authn/noauth.go
njhale: I added some cleanup logic when auth is enabled that ensure we drop this user from the DB on startup. This should reduce the potential attack surface.

PR #1137, opened 2025-01-07
chore: switch types from otto to obot

thedadams:


pkg/storage/apis/obot.obot.ai/group.go
```diff
@@ -0,0 +1,3 @@
+package obotobotplatformai
+
+var Group = "obot.obot.ai"
```
thedadams, line 3: How do we feel about this group?

njhale: Don't love the stutter, but I'm not sure if `obot.ai` or `core.obot.ai` are any better.

apiclient/invoke.go
```diff
@@ -39,12 +39,12 @@ func (c *Client) Invoke(ctx context.Context, agentID string, input string, opts
 		close(events)
 		return &types.InvokeResponse{
 			Events:   events,
-			ThreadID: resp.Header.Get("X-Otto-Thread-Id"),
+			ThreadID: resp.Header.Get("X-Obot-Thread-Id"),
```
njhale, line 42: nit: Might make migration easier if we fall back to the old header if the new one isn't set.

njhale (approved):
Looks good!

pkg/storage/apis/obot.obot.ai/group.go
thedadams: The other thing I considered was `obot.obot-platform.ai`. I switched it to this because we actually own `obot.ai`.

StrongMonkey: kind of weird to have two obot....but I have no opinion

pkg/controller/routes.go
```diff
@@ -160,3 +171,29 @@ func (c *Controller) setupRoutes() error {
 	c.toolRefHandler = toolRef
 	return nil
 }
+
+func removeOldFinalizers(req router.Request, _ router.Response) error {
```
StrongMonkey: is this the only migration we need to be doing?

thedadams: We also need to delete the models so they can be recreated with the new owner's reference.

StrongMonkey (approved):

iwilltry42 (approved):
the stutter in the group name is weird, but I see no "real" problem with it.

PR #1139, opened 2025-01-07
feat: add support for credential auth for non-tool-references

thedadams:
Support authentication for agent and workflow tools that use non-tool-references in their definition. That is, a GitHub-hosted tool or a local tool.

During the implementation and testing for non-tool-references, I also spotted a gap: nested tools. That is, if an agent uses a tool that uses a tool that has a credential, then the authentication for that tool would not be processed. After this change, it will.


thedadams:
Shoot. One more change needs to be made. Going to draft for now.

StrongMonkey (approved):

iwilltry42 (approved):

PR #1150, opened 2025-01-08
Chore: add sendgrid webhook endpoints to receive webhook events

StrongMonkey:
Add sendgrid webhook endpoint to be able to receive sendgrid inbound parse event and trigger workflow.

A use case would be getting emails regarding to receipts and have dedicated workflows processing these emails.

### DIFF
--- a/pkg/api/handlers/sendgrid/webhook.go
+++ b/pkg/api/handlers/sendgrid/webhook.go
@@ -1,0 +1,40 @@
+package sendgrid
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/obot-platform/obot/apiclient/types"
+	"github.com/obot-platform/obot/pkg/api"
+	"github.com/obot-platform/obot/pkg/emailtrigger"
+	"github.com/sendgrid/sendgrid-go/helpers/inbound"
+	kclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type InboundWebhookHandler struct {
+	emailTrigger *emailtrigger.EmailHandler
+}
+
+func NewInboundWebhookHandler(c kclient.Client, hostname string) *InboundWebhookHandler {
+	emailTrigger := emailtrigger.EmailTrigger(c, hostname)
+	return &InboundWebhookHandler{emailTrigger: emailTrigger}
+}
+
+func (h *InboundWebhookHandler) InboundWebhookHandler(req api.Context) error {
+	if req.Request.Method != http.MethodPost {
+		return types.NewErrHttp(http.StatusMethodNotAllowed, "Invalid request method")
+	}
+
+	inboundEmail, err := inbound.Parse(req.Request)
+	if err != nil {
+		return types.NewErrHttp(http.StatusBadRequest, fmt.Sprintf("Failed to parse inbound email: %v", err))
+	}
+
+	subject := inboundEmail.Headers["Subject"]
+	if err := h.emailTrigger.Handler(req.Context(), inboundEmail.Envelope.From, inboundEmail.Envelope.To, subject, []byte(inboundEmail.TextBody)); err != nil {
+		return types.NewErrHttp(http.StatusInternalServerError, fmt.Sprintf("Failed to handle inbound email: %v", err))
+	}
+
+	req.WriteHeader(http.StatusOK)
+	return nil
+}
--- a/pkg/api/handlers/sendgrid/webhook.go
+++ b/pkg/api/handlers/sendgrid/webhook.go
@@ -13,14 +13,23 @@ import (
 
 type InboundWebhookHandler struct {
 	emailTrigger *emailtrigger.EmailHandler
+	username     string
+	password     string
 }
 
-func NewInboundWebhookHandler(c kclient.Client, hostname string) *InboundWebhookHandler {
+func NewInboundWebhookHandler(c kclient.Client, hostname string, username, password string) *InboundWebhookHandler {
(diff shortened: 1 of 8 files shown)


pkg/api/handlers/sendgrid/webhook.go
```diff
@@ -0,0 +1,40 @@
...
+	emailTrigger := emailtrigger.EmailTrigger(c, hostname)
+	return &InboundWebhookHandler{emailTrigger: emailTrigger}
+}
+
+func (h *InboundWebhookHandler) InboundWebhookHandler(req api.Context) error {
+	if req.Request.Method != http.MethodPost {
+		return types.NewErrHttp(http.StatusMethodNotAllowed, "Invalid request method")
+	}
+
+	inboundEmail, err := inbound.Parse(req.Request)
```
StrongMonkey, line 33: Sendgrid doesn't have a way to attach secret header so it is a bit tricky to vefify whether requests are coming from sendgrid...this makes our webhook server vulnerable to unverified webhook events from other places. (at least not from our account that you can configure to send a secret key as a header)

so there is no validation right to validate if requests are actually coming from sendgrid.

pkg/api/server/server.go
```diff
@@ -14,8 +14,25 @@ import (
 	"github.com/obot-platform/obot/pkg/proxy"
 	"github.com/obot-platform/obot/pkg/storage"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apiserver/pkg/authentication/user"
 )
 
+// authDisabledPaths is a list of paths that do not require authentication.
+var (
```
StrongMonkey, line 21: We will have to disable authenticate for webhook, as sendgrid doesn't have a way to configure the webhook secrets.

For other providers it might support it, but I think it will be a total different workflow to configure token in obot as supposed to having token dedicated for webhook usage. So, no auth for now.

pkg/emailtrigger/emailtrigger.go
```diff
@@ -0,0 +1,132 @@
+package emailtrigger
```
StrongMonkey, line 1: I refactored the code a bit to be able to share some common trigger code from both smtp and sendgrid.

pkg/api/handlers/sendgrid/webhook.go
StrongMonkey: I added basic auth protection for now. The username and password can be configured at global level to protect webhook endpoint, and sendgrid webhook can be configured to have configured username and password in front of the url.

njhale: Does twilio's [event webhook signing feature](https://www.twilio.com/docs/sendgrid/for-developers/tracking-events/getting-started-event-webhook-security-features) not work for this? IIUC, enabling this will add a `X-Twilio-Email-Event-Webhook-Signature` header containing a verification signature to all webhook requests it makes.

StrongMonkey: yeah, that's for their generic event webhook. The inbound parse webhook doesn't support it.

pkg/api/handlers/sendgrid/webhook.go
```diff
@@ -0,0 +1,49 @@
...
+	if h.username != "" && h.password != "" {
+		username, password, ok := req.Request.BasicAuth()
+		if !ok || username != h.username || password != h.password {
+			return types.NewErrHttp(http.StatusUnauthorized, "Invalid credentials")
+		}
+	}
+
+	if req.Request.Method != http.MethodPost {
+		return types.NewErrHttp(http.StatusMethodNotAllowed, "Invalid request method")
+	}
```
thedadams: nit: this is handled when we set up the route.

pkg/api/router/router.go
```diff
@@ -256,6 +259,9 @@ func Router(services *services.Services) (http.Handler, error) {
 	mux.HandleFunc("POST /api/webhooks/{id}/remove-token", webhooks.RemoveToken)
 	mux.HandleFunc("POST /api/webhooks/{namespace}/{id}", webhooks.Execute)
 
+	// Webhook for third party integration to trigger workflow
+	mux.HandleFunc("POST /webhook/sendgrid", sendgridWebhookHandler.InboundWebhookHandler)
```
thedadams: We have the convention that all of our backend APIs start with `/api`. Can this one be `/api/sendgrid`?

pkg/services/config.go
```diff
@@ -69,6 +69,10 @@ type Config struct {
 	NoReplyEmailAddress        string `usage:"The email to use for no-reply emails from obot"`
 	Docker                     bool   `usage:"Enable Docker support" default:"false" env:"OBOT_DOCKER"`
 
+	// Sendgrid webhook
+	SendgridWebhookUsername string `usage:"The username for the sendgrid webhook to authenticate with" env:"OBOT_SENDGRID_WEBHOOK_USERNAME"`
+	SendgridWebhookPassword string `usage:"The password for the sendgrid webhook to authenticate with" env:"OBOT_SENDGRID_WEBHOOK_PASSWORD"`
```
thedadams: All the other configuration variables we use to configure the server start with `OBOT_SERVER_`. If you remove the `env` tag here, that will be the same for these configuration variables.

pkg/api/router/router.go
StrongMonkey: yeah, I am good with that.

thedadams (approved):

Issue #1141, opened 2025-01-07
"Show more" option for tool bundle does not work on searched tool list.


sangee2004:
**Steps to reproduce the problem:**
1. create an agent.
2. Add a tool by entering a search string like "gm" to get Gmail tool.
3. Use the "Show more" option.
Notice that no fine grain tools are presented.

<img width="541" alt="Screenshot 2025-01-07 at 8 20 34 AM" src="https://github.com/user-attachments/assets/7cf3be4d-fc27-4753-a044-6e9154705f3d" />

**Expected Behavior:**
Fine Grained tools should be presented when "Show more" option is used.

sangee2004:
This issue is not seen anymore when testing with latest. builds.

Issue #1021, opened 2024-12-23
Find better wording for "Your Tools" in tools menu


cjellick:
I'm not sure what I want, but "your tools" doesn't make sense.

cjellick:
yeah, let's just go with that for now

sangee2004:
With latest build "Custom Tool" is used both in tools dropdown in `Agents` and `Workflows` page and in `Tools` page.<img width="515" alt="Image" src="https://github.com/user-attachments/assets/1f91df10-3b27-47d7-9e39-c5d9133e201d" />

PR #1163, opened 2025-01-09
fix: Custom Tools & search tool dropdown fix

ivyjeong13:
Addresses #1141 
Addresses #1021 

* "Custom Tools" to "Your Tools"
* fix search/filtering in tool category dropdown not showing results

* found weird bug with native Command filter where searching with a bundle selected and expanded -- the bundle item gets moved to the bottom and stays that way until you deselect, collapse it, and expand it or search again. Moved filtering logic out and that addressed it.

### DIFF
--- a/ui/admin/app/components/tools/ToolCatalog.tsx
+++ b/ui/admin/app/components/tools/ToolCatalog.tsx
@@ -115,6 +115,7 @@ export function ToolCatalog({
                         tools={categoryTools}
                         selectedTools={tools}
                         onUpdateTools={onUpdateTools}
+                        expandFor={search}
                     />
                 ))}
             </CommandList>
--- a/ui/admin/app/components/tools/ToolCatalog.tsx
+++ b/ui/admin/app/components/tools/ToolCatalog.tsx
@@ -124,7 +124,7 @@ export function ToolCatalog({
     function filterToolCatalogBySearch(
         toolCategories: [string, ToolCategory][]
     ) {
-        return toolCategories.reduce(
+        return toolCategories.reduce<[string, ToolCategory][]>(
             (acc, [category, categoryData]) => {
                 const matchesSearch = (str: string) =>
                     str.toLowerCase().includes(search.toLowerCase());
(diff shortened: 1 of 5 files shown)


ui/admin/app/components/tools/ToolCatalog.tsx
```diff
@@ -116,6 +120,49 @@ export function ToolCatalog({
             </CommandList>
         </Command>
     );
+
+    function filterToolCatalogBySearch(
+        toolCategories: [string, ToolCategory][]
+    ) {
+        return toolCategories.reduce(
```
ryanhopperlowe: something I want to start doing Purely for readability is adding the output type annotation to the reduce function instead of just casting the `initialValue`

```suggestion
        return toolCategories.reduce<[string, ToolCategory][]>(
```

ryanhopperlowe (approved):
Man this works so much better! Thanks for making this fix!

ivyjeong13:
> Had an idea. What if we were able to automatically expand categories that had matches in their subtools? So if I type in Create I see this: Screenshot 2025-01-09 at 9 57 21 AM
> 
> This is what drove me crazy when I first tried to implement it so I wouldn't sweat too hard over it 😆
> 
> Edit: You could maybe just auto expand all categories whenever there's a search value... ??

Oh yeah the auto-expand sounds like a good idea! Haha I'll see what I can do, if takes too much time, I'll punt it off as a later improvement!

ivyjeong13:
> > Had an idea. What if we were able to automatically expand categories that had matches in their subtools? So if I type in Create I see this: Screenshot 2025-01-09 at 9 57 21 AM
> > This is what drove me crazy when I first tried to implement it so I wouldn't sweat too hard over it 😆
> > Edit: You could maybe just auto expand all categories whenever there's a search value... ??
> 
> Oh yeah the auto-expand sounds like a good idea! Haha I'll see what I can do, if takes too much time, I'll punt it off as a later improvement!

@ryanhopperlowe Added the auto-expand for if a sub tool matches in! : )

Issue #1039, opened 2024-12-23
Don't show tools that require oauth if oauth is not configured


cjellick:
We should not show tools in the tool dropdown on an agent/workflow if their corresponding oauth app has not been configured.

ivyjeong13:
Updating to show the tool as greyed out rather than filtering it out.

sangee2004:
Jira and salesforce tools bundles which do not have the oauth providers configured by default are greyed out in the tool dropdown on an agent/workflow.

<img width="645" alt="Image" src="https://github.com/user-attachments/assets/c4dbf539-ae16-4023-b436-6cf899411eee" />

Once the oauth providers are configured, then they are shown as active in the  tool dropdown on an agent/workflow.

PR #1164, opened 2025-01-09
enhance: greyed out tool if oauth is unconfigured

ivyjeong13:
Addresses #1039
Addresses #1183 

Screenshot 2025-01-09 at 9 55 32 AM

Screenshot 2025-01-09 at 9 54 56 AM

* per thread with Craig, changed from filtering out to showing greyed out and why it's inaccessible
* noticed small bug with step deletion, fixed here as well

### DIFF
--- a/ui/admin/app/components/tools/ToolItem.tsx
+++ b/ui/admin/app/components/tools/ToolItem.tsx
@@ -1,3 +1,5 @@
+import { TriangleAlertIcon } from "lucide-react";
+
 import { ToolReference } from "~/lib/model/toolReferences";
 import { cn } from "~/lib/utils";
 
@@ -9,6 +11,7 @@ import { CommandItem } from "~/components/ui/command";
 
 type ToolItemProps = {
 	tool: ToolReference;
+	configured: boolean;
 	isSelected: boolean;
 	isBundleSelected: boolean;
 	onSelect: () => void;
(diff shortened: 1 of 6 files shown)


ui/admin/app/components/tools/ToolItem.tsx
```diff
@@ -53,9 +57,20 @@ export function ToolItem({
...
+                            <TriangleAlertIcon className="w-4 h-4 text-warning opacity-50" />
+                        ) : null}
 
-                        <span className={cn("flex items-center")}>
+                        <span
+                            className={cn(
+                                "flex items-center",
+                                !configured && "opacity-20"
+                            )}
+                        >
```
ryanhopperlowe: could some of this styling/opacity be handled by setting the `<CommandItem disabled={!configured />`?

ivyjeong13: I had that initially but found out it also disabled the tooltip. But hmm I can take a look again and see if I can move things around so the tooltip still comes through. edit: Also setting disabled on CommandItem disables the See Less/See More as well which I think would be nice to still have available so user can see what specific actions are unavailable.

ivyjeong13: Actually, don't think I can 😞 Tried moving the ToolTooltip outside the CommandItem and doesn't seem to work.

njhale (approved):

Issue #426, opened 2024-11-04
Allow user role assignment to make users admin


cjellick:
Right now, a user can only add/subtract admin via the "admin emails" environment variable. It should be possible to for an admin user to do this via the admin UI.

thedadams:
The backend has merged, including the change to error when changing the "explicit admins" set via environment variable.

Ready for frontend.

cjellick:
@ryanhopperlowe This is the issue related to what you're working on. I'm moving it to in-progress.

sangee2004:
Admin is allowed to change roles (Admin/Default) for an users.

<img width="1009" alt="Image" src="https://github.com/user-attachments/assets/8f56d081-b244-4e3b-9d19-2662c42d9c4d" />

PR #1195, opened 2025-01-09
feat: update user role in admin ui

ryanhopperlowe:
addresses #426 

https://www.loom.com/share/d95c7f25ac314b25af34045db587e227?sid=6ba07574-a3aa-423d-9745-51b8be47be3a


ivyjeong13 (approved):

ui/admin/app/lib/utils/pluralize.ts
```diff
@@ -0,0 +1,2 @@
+export const pluralize = (count: number, singular: string, plural: string) =>
+	`${count === 1 ? singular : plural}`;
```
ivyjeong13: Minor thought, if most of the time plural is adding an 's', can `plural` be an optional parameter?

```
${count === 1 ? singular : plural ? plural : `${singular}s`;
```

ryanhopperlowe: I like that. Adding now

ivyjeong13 (approved):

Issue #1186, opened 2025-01-09
Authorize agents to specific users


cjellick:
This is a fundamental change to how we publish agents. Here is the UX we want:

We want to drop the word/concept of publish. Instead, we want an `Access Control` section in the agent.
In this section you should be able to:
- add/remove users via a lookup against our user api or by adding "raw" email addresses
- make the agent available to everyone

Here's how publishing should change:
First, even if you don't publish, an agent will always be available in the user UI to admin users. If the user has not set an alias (the name you currently set in the publish modal), it will be available via the agent id. So, we should still yet the user set an alias, but if they don't set it, we should show the user-ui URL using the agent ID.

Darrens put the backend in:
```
Three new APIs are added to handle authorization of agents
GET /api/agents/{id}/authorizations
POST /api/agents/{id}/authorizations/add
POST /api/agents/{id}/authorizations/remove
```
To add/remove you just post `{ userID: "1" }`. The value of the userID is either the ID, which is a number as a string, `*` which means everyone, or email like `darren@acorn.io` which is for that email address of a user that might not exist in obot yet.

Once the UI updates to this new API I will change the backend such that agents are not public based on the alias field. The alias field will just be an alias for the URL and not for authorization. So making an agent public for everyone will require userID: * to be added.

cjellick:
@ivyjeong13 make sure i know when the changes are in so that @ibuildthecloud can do the final step of changing the backend such that agents are not public based on the alias field.

ivyjeong13:
> @ivyjeong13 make sure i know when the changes are in so that @ibuildthecloud can do the final step of changing the backend such that agents are not public based on the alias field.

Will do!

sangee2004:
Able to authorize agents to specific users now.

Related issues seen when testing this feature

https://github.com/obot-platform/obot/issues/1251
https://github.com/obot-platform/obot/issues/1254

PR #1209, opened 2025-01-10
feat: add agent user authorizations and publish ux change

ivyjeong13:
Addresses #1186 

https://www.loom.com/share/896ebc3b3f184f99b1cde408ec3f2614

* changes publish to be edit alias option
* add access control in place of former publish


ui/admin/app/components/agent/UserAuthorizationSelect.tsx
```diff
@@ -0,0 +1,174 @@
...
+		const checked = selectedUsers.has(option.value);
+		if (checked) {
+			removeAuthorizationFromAgent.executeAsync(agentId, option.value);
+		} else {
+			addAuthorizationToAgent.executeAsync(agentId, option.value);
+		}
+	};
+
+	const handleCreate = (email: string) => {
+		addAuthorizationToAgent.executeAsync(agentId, email);
```
ivyjeong13, line 66: One thing I'm considering for fast follow-on is validating that if wants to add someone by email that isn't a user, it does some validation that it's a valid email before allowing them to add, or they add, and it shows an error.

ryanhopperlowe (approved):
one thing to maybe add, is to add a prop to ComboBox that prevents it from closing when an item is selected. That way a user can select multiple entries at a time.

PR #1224, opened 2025-01-11
Fix grammar and prompt structure

oskapt:
Clean up prompt for clarity and fix typographical errors.

Since this is the first impression that someone has of Obot after installing, it's important that the prompt not have obvious typographical errors.


cjellick (approved):

cjellick:
Merging this fell through the cracks and then two thirds of the chnages got made anyway by another dev. I fixed the conflicts and will merge once the build is done.

Issue #1202, opened 2025-01-10
Let me download thread files


cjellick:
You can see in this screenshot that this thread has a bunch of files attached. I would like to be able to access the content of those files.

The simplest version of this is just let me download any of them. The more complicated version would be to have an in-app memory viewer. I think for this initial implementation, I just need/want download.

There is already an API for it:
```
mux.HandleFunc("GET /api/threads/{id}/file/{file...}", threads.GetFile)
```

![Screenshot 2025-01-10 at 9 45 14 AM](https://github.com/user-attachments/assets/4e641c25-cb2a-465d-830b-3f0011213c88)

ryanhopperlowe:
@cjellick So I'm noticing that there's currently no way to upload workspace files in the UI... Do we already have an issue for that?

cjellick:
Im unaware of one. if you cant find on search, go ahead and create. Download is not dependent on upload as the main usecase for download is for files created by the agent, but it would be nice to have both.

sangee2004:
We have the option to now download files from thread inspect view.

PR #1232, opened 2025-01-13
feat: allow admins to download files in thread workspaces

ryanhopperlowe:
addresses #1202 

image

### DIFF
--- a/pkg/api/router/router.go
+++ b/pkg/api/router/router.go
@@ -210,7 +210,7 @@ func Router(services *services.Services) (http.Handler, error) {
 
 	// Thread files
 	mux.HandleFunc("GET /api/threads/{id}/files", threads.Files)
-	mux.HandleFunc("GET /api/threads/{id}/file/{file...}", threads.GetFile)
+	mux.HandleFunc("GET /api/threads/{id}/files/{file...}", threads.GetFile)
 	mux.HandleFunc("POST /api/threads/{id}/files/{file...}", threads.UploadFile)
 	mux.HandleFunc("DELETE /api/threads/{id}/files/{file...}", threads.DeleteFile)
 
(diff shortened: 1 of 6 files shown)


pkg/api/router/router.go
```diff
@@ -210,7 +210,7 @@ func Router(services *services.Services) (http.Handler, error) {
 
 	// Thread files
 	mux.HandleFunc("GET /api/threads/{id}/files", threads.Files)
-	mux.HandleFunc("GET /api/threads/{id}/file/{file...}", threads.GetFile)
+	mux.HandleFunc("GET /api/threads/{id}/files/{file...}", threads.GetFile)
```
ryanhopperlowe, line 213: Updated this to match naming conventions of similar apis. Let me know if this is acceptable

ui/admin/app/lib/utils/downloadFile.ts
```diff
@@ -0,0 +1,14 @@
+export function downloadFile(data: Blob, fileName: string) {
```
ivyjeong13, line 1: nit: is utils/downloadFile being used? Wondering if it can be removed!

ivyjeong13 (approved):

ui/admin/app/lib/utils/downloadFile.ts
ryanhopperlowe: yes, it can. I'll remove it in upcoming PR

pkg/api/router/router.go
StrongMonkey: Seems fine, if there is a reason it needs to be singular then `UploadFile` and `DeleteFile` will be the same way too. So probably better to stick with plural.

thedadams (approved):

PR #1246, opened 2025-01-14
fix: adjust relative credential paths for tool refs

thedadams:


pkg/controller/creds/creds.go
```diff
@@ -149,17 +150,26 @@ func toolRefsFromTools(parentTool gptscript.Tool, parentRef toolRef, tools []str
 
 func fullToolPathName(parentRef toolRef, name string) string {
 	toolName, subTool := gtypes.SplitToolRef(name)
+
+	// If this tool's path is relative to its parent.
 	if strings.HasPrefix(toolName, ".") {
```
njhale, line 155: nit: looks like you can de-indent the entire block here by negating this predicate and returning early

```go
if !strings.HasPrefix(toolName, ".") {
    return name
}

// ...
```

njhale (changes requested):
Noticed some issues with this PR and the paths for credentials on certain model providers; e.g. x-ai.

Spoke with @thedadams about it offline and he's fixing it now.

njhale (approved):
The ToolReference `status.tools.credentials` fields look much better for both relative and absolute paths with this change.

Issue #1038, opened 2024-12-23
Rework oauth app and tool UX


cjellick:
Oauth apps should be more tightly couple to the tools the support in the UI. I am thinking, perhaps, that we should eliminate the oauth app page and just incorporate this into the tools page. However, this UX could get a bit overwhelming. We should consider reworking the tools page in general to just show the top level tool category and the user having to click into it to see fine grained tools

sangee2004:
When testing with latest build there is no  `oauth app` page which is now integrated with tools page.

<img width="1209" alt="Image" src="https://github.com/user-attachments/assets/100ed354-52b0-47b3-af68-aaf6540af642" />

Only tool bundles are listed in `Tools` page now and we are able to into it to see fine grained tools.

Capabilities tools is listed in between build-in tools .It would be better to have all the capabilities tools listed at the end of the built-in tools which is tracked in https://github.com/obot-platform/obot/issues/1273

PR #1257, opened 2025-01-14
Fix: hide builtin tool actions

ryanhopperlowe:
- hides builtin tool actions from the tools page
- prevents stale data from preemptively stopping tool polling after refresh

### DIFF
--- a/ui/admin/app/components/tools/toolGrid/ToolCardActions.tsx
+++ b/ui/admin/app/components/tools/toolGrid/ToolCardActions.tsx
@@ -28,6 +28,8 @@ export function ToolCardActions({ tool }: { tool: ToolReference }) {
 		}
 	);
 
+	if (tool.builtin) return null;
+
 	return (
 		<div className="flex items-center gap-2">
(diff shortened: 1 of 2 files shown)


ui/admin/app/components/tools/toolGrid/ToolCardActions.tsx
```diff
@@ -28,6 +28,8 @@ export function ToolCardActions({ tool }: { tool: ToolReference }) {
 		}
 	);
 
+	if (tool.builtin) return null;
```
ivyjeong13, line 31: Making self note to tweak and account for with #1038 Guessing we don't want to show "Refresh Tool" option for built-in tools?

ivyjeong13 (approved):

ui/admin/app/components/tools/toolGrid/ToolCardActions.tsx
ryanhopperlowe: correct!

Issue #1256, opened 2025-01-14
"Authorized" state shown from Knowledge tool when there is no authentication required for it.


sangee2004:
**Steps to reproduce the problem:**
1. Create an agent and any knowledge file.
Notice that knowledge tool shows  "Authorized" state.
We are also presented with "Remove Authorization" option which when used  does not result in the "Authorized" state being cleared.

<img width="751" alt="Screenshot 2025-01-14 at 1 16 34 PM" src="https://github.com/user-attachments/assets/bc0cfac1-6c25-43f0-acd2-1a39eadafc94" />

**Expected Behavior:**
Knowledge tool should not show any Authorization state.

sangee2004:
This issue is not seen in latest builds.

<img width="545" alt="Image" src="https://github.com/user-attachments/assets/f55b093b-0780-49ac-9b46-1bf2c42fcd90" />

PR #1266, opened 2025-01-15
enhance: allow tools to indicate which credentials don't need auth

thedadams:
Certain tools have credentials that don't require the use to authenticate them. For example, the sys.model.provider.credential is handled by the system and never needs user interaction.

This change introduces the processing of the noAuth metadata field to indicate which credentials don't need user auth. If a credential name is included in the noAuth metadata for a tool, then that credential name is disregarded for user auth. The sys.model.provider.credential is included by default for every tool.

This is more of a proposal. There have already been a couple of instances of us needing to exclude certain credentials from authentication. This will allow tools to tell us that a credential doesn't need authentication instead of us playing whack-a-mole as we go along.

Issue: https://github.com/obot-platform/obot/issues/1256

This PR should be merged first: https://github.com/obot-platform/tools/pull/346

### DIFF
--- a/apiclient/types/toolreference.go
+++ b/apiclient/types/toolreference.go
@@ -25,7 +25,7 @@ type ToolReference struct {
 	Error       string            `json:"error,omitempty"`
 	Builtin     bool              `json:"builtin,omitempty"`
 	Description string            `json:"description,omitempty"`
-	Credentials []string          `json:"credential,omitempty"`
+	Credentials []string          `json:"credentials,omitempty"`
 	Params      map[string]string `json:"params,omitempty"`
 }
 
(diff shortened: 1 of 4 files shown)


apiclient/types/toolreference.go
```diff
@@ -25,7 +25,7 @@ type ToolReference struct {
 	Error       string            `json:"error,omitempty"`
 	Builtin     bool              `json:"builtin,omitempty"`
 	Description string            `json:"description,omitempty"`
-	Credentials []string          `json:"credential,omitempty"`
+	Credentials []string          `json:"credentials,omitempty"`
```
StrongMonkey, line 28: was this causing problems before?

thedadams: No, it wasn't causing any problems, just figured I should fix it since I saw it (and I created it).

iwilltry42 (approved):
LGTM - I don't know how to feel about having `noAuth` as metadata instead of _somehow_ incorporating it directly into the `Credentials` field.
Feels weird, but I have no other idea :man_shrugging:

PR #1285, opened 2025-01-16
Fix: don't delete knowledge dataset when it is not being created

StrongMonkey:
We are issuing a  run to delete knowledge dataset when a thread and knowledge set has been created first but with no content on it. Seems unnecessary to do so that. What is happening today is that when user chats with an agent, a new thread with knowledge set created and we immediately run a `knowledge-delete` run with it even there is no dataset created with it.

### DIFF
--- a/pkg/controller/handlers/knowledgeset/knowledgeset.go
+++ b/pkg/controller/handlers/knowledgeset/knowledgeset.go
@@ -108,7 +108,7 @@ func (h *Handler) CheckHasContent(req router.Request, _ router.Response) error {
 		ks.Status.ExistingFile = ""
 	} else {
 		ks.Status.ExistingFile = files.Items[0].Name
-		ks.Status.EmptyDatasetDeleted = false
+		ks.Status.DatasetCreated = true
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


pkg/controller/handlers/knowledgeset/knowledgeset.go
```diff
@@ -149,7 +150,7 @@ func (h *Handler) CreateWorkspace(req router.Request, _ router.Response) error {
 
 func (h *Handler) Cleanup(req router.Request, _ router.Response) error {
 	ks := req.Object.(*v1.KnowledgeSet)
-	if ks.Status.ThreadName == "" || ks.Status.EmptyDatasetDeleted || (ks.DeletionTimestamp.IsZero() && ks.Status.HasContent) {
+	if ks.Status.ThreadName == "" || !ks.Status.DatasetCreated || ks.Status.EmptyDatasetDeleted || (ks.DeletionTimestamp.IsZero() && ks.Status.HasContent) {
```
StrongMonkey: @thedadams I am not sure if `EmptyDatasetDeleted` is still needed here. is that a way to make sure we don't keep deleting dataset when it is already deleted? I think with the new field `DatasetCreated` it should prevent you from doing that.

thedadams: It seems like DatasetDeleted and DatasetCreated are the same thing, and `DatasetCreated` is a friendlier name, so I am good with getting rid of the other one.

StrongMonkey: sounds good. The EmptyDatasetDeleted was not really preventing it from being deleted when it is first created so I will just remove that and use the new field.

thedadams (approved):

iwilltry42 (approved):

Issue #1290, opened 2025-01-16
Pre Authentication for SendGrid/Bluesky tools times out.


sangee2004:
**Steps to reproduce the probelm:**
1. Create an agent with SendGrid tool
2. Pre Authenticate this tool.

Notice that we are stuck in this state for a very long time (about 5 minutes)
<img width="543" alt="Image" src="https://github.com/user-attachments/assets/d38afae3-1eae-4118-8dee-f21fdc2bf885" />

After this user is presented with the following error message:

<img width="460" alt="Image" src="https://github.com/user-attachments/assets/6a32d291-807a-4420-bf4d-7381978e0ecb" />

**Note** - This issue is also seen when pre authenticating BlueSky tool.

sangee2004:
This issue is not seen anymore when testing with latest builds.

PR #1308, opened 2025-01-17
fix: check live state events first when watching events

thedadams:
Checking the live state events initially will ensure that the events watcher will see any prompts that happen before other call events.

Issue: https://github.com/obot-platform/obot/issues/1290


StrongMonkey (approved):
Nice found!

Issue #1174, opened 2025-01-09
Tweak model dropdown


cjellick:
In the model selector dropdown on agents and workflows, we want to make the following tweaks:
-The model-provider header. Right now it is smaller and slightly indented (at least to my eye). It should feel more like a header and the models under it, it's children.
- Map the ugly provider names, like `openai-model-provider` to user friendlier ones like OpenAI

sangee2004:
This is now addressed in latest builds

<img width="546" alt="Image" src="https://github.com/user-attachments/assets/ed4fb37d-913f-487d-af5d-256de92f499e" />

PR #1311, opened 2025-01-17
enhance: tweak model dropdown & combobox header

ivyjeong13:
Addresses #1174 
* user friendly label for model provider id
* combobox header tweak 

Screenshot 2025-01-21 at 1 09 10 PM
Screenshot 2025-01-21 at 1 08 33 PM

### DIFF
--- a/pkg/api/handlers/model.go
+++ b/pkg/api/handlers/model.go
@@ -165,15 +165,13 @@ func convertModel(ctx context.Context, c kclient.Client, model v1.Model) (types.
 	}
 
 	modelManifest := model.Spec.Manifest
-	if toolRef.Status.Tool != nil {
-		modelManifest.ModelProviderName = toolRef.Status.Tool.Name
-	}
 
 	return types.Model{
 		Metadata:      MetadataFrom(&model),
 		ModelManifest: modelManifest,
 		ModelStatus: types.ModelStatus{
-			AliasAssigned: aliasAssigned,
+			AliasAssigned:     aliasAssigned,
+			ModelProviderName: toolRef.Status.Tool.Name,
 		},
 	}, nil
 }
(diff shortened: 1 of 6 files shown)


pkg/api/handlers/model.go
```diff
@@ -164,9 +164,14 @@ func convertModel(ctx context.Context, c kclient.Client, model v1.Model) (types.
 		aliasAssigned = &model.Status.AliasAssigned
 	}
 
+	modelManifest := model.Spec.Manifest
+	if toolRef.Status.Tool != nil {
```
ivyjeong13, line 169: Checking if it's alright to add ModelProviderName to get the user-friendly label?

ryanhopperlowe (approved):
UI lgtm

pkg/api/handlers/model.go
```diff
@@ -164,11 +164,14 @@ func convertModel(ctx context.Context, c kclient.Client, model v1.Model) (types.
...
+	modelManifest := model.Spec.Manifest
+
 	return types.Model{
 		Metadata:      MetadataFrom(&model),
-		ModelManifest: model.Spec.Manifest,
+		ModelManifest: modelManifest,
 		ModelStatus: types.ModelStatus{
-			AliasAssigned: aliasAssigned,
+			AliasAssigned:     aliasAssigned,
+			ModelProviderName: toolRef.Status.Tool.Name,
```
thedadams: You would have to do something similar to the `aliasAssigned` above to ensure we don't panic if `toolRef.Status.Tool == nil`. I sugest lines 162 to 165 above be changed to:
```go
	var (
		aliasAssigned *bool
		toolName string
	)
	if model.Generation == model.Status.ObservedGeneration {
 		aliasAssigned = &model.Status.AliasAssigned
 	}
	if toolRef.Status.Tool != nil {
		toolName = toolRef.Status.Tool.Name
	}
```

And then use `toolName` here instead.

ivyjeong13: Thank you!

ivyjeong13: And updated! edit: Oop wait, I failed a linter line 🤦‍♀️

thedadams (approved):

PR #1317, opened 2025-01-17
feat: move model and env variables to advanced section

ryanhopperlowe:
image


ivyjeong13 (approved):
LGTM! Only super minor nit I might have is it looks like there is a lot of padding in the advanced dropdown content. Not sure if would look better if Model and Environment variable lined up with the tool capabability icons?

ivyjeong13 (approved):

PR #1320, opened 2025-01-17
Chore: Add sendgrid docs

StrongMonkey:
Add sendgrid inbound webhook docs on how to use sendgrid. I also refactored our current workflow docs given that we have UI for workflow triggers now.

Note: This doc and sendgrid implemetation are currently designed to be matched with darren's smtp approach. But for sendgrid or other providers it can be flexaible as configuring multiple email server address and have all of them forward to obot. Currently Obot only supports one email server address that is configured at start up.


pkg/services/config.go
```diff
@@ -356,7 +357,7 @@ func New(ctx context.Context, config Config) (*Services, error) {
 		authenticators = union.New(authenticators, authn.NewNoAuth(gatewayClient))
 	}
 
-	if config.EmailServerName != "" {
+	if config.EmailServerName != "" && config.EnableSMTPServer {
```
StrongMonkey, line 360: we are likely going to change the implementation for how we should match the email server name for sendgrid, but right now it follows the same pattern with smtp approach. So we need to set the email server name without having to setup smtp server. This is a way to do it specifically through a different envariable.

pkg/api/authz/authz.go
```diff
@@ -46,7 +46,7 @@ var staticRules = map[string][]string{
 		"GET /api/app-oauth/callback/{id}",
 		"GET /api/app-oauth/get-token",
 
-		"POST /webhook/sendgrid",
+		"POST /api/sendgrid",
```
StrongMonkey, line 49: forgot to update the endpoint that is addressed from last PR comment.

pkg/services/config.go
thedadams, line 360: I don't think this works because of this line: https://github.com/obot-platform/obot/blob/6d035afc3b0b112c2aa8b58b2209f25224876eaa/pkg/emailtrigger/emailtrigger.go#L46

Essentially, the email will never be sent if `config.EnableSMTPServer` is set, but `config.EmailServerName` is not.

StrongMonkey: Yeah, my intention is that both `EmailServerName` and `EnableSMTPServer` have to be set to use smtp. But `EmailServerName` will need to be set to enable sendgrid webhook separately without having to enable smtp server.

thedadams: Ah, sorry. I misunderstood.

thedadams (approved):

docs/docs/05-configuration/05-email-webhook.md
```diff
@@ -0,0 +1,40 @@
...
+4. Set the domain to the email server name you configured in Obot.
+5. Set the webhook URL to `https://{obot_server_url}/api/sendgrid`. Replace `{obot_server_url}` with the actual URL of your Obot server.
+
+To see more details, follow the detailed instructions provided by SendGrid [here](https://www.twilio.com/docs/sendgrid/for-developers/parsing-email/setting-up-the-inbound-parse-webhook).
+
+:::note
+By default, Sendgrid inbound webhook does not provide a way to verify the signature and payload coming from SendGrid. In order to verify requests that are coming from sendgird, Obot provides a way to configure both `OBOT_SENDGRID_WEBHOOK_USERNAME` and `OBOT_SENDGRID_WEBHOOK_PASSWORD` to enable basic authentication for the webhook endpoint. So you can configure url with basic auth `https://{username}:{password}@{obot_server_url}/api/sendgrid` to ensure that only requests from SendGrid are received in Obot. For more details, refer to this example: [Handling SendGrid Inbound Parse](https://www.twilio.com/en-us/blog/microservice-template-handle-sendgrid-inbound-parse). It is recommended to configure these credentials to secure the endpoint and protect it from unverified payloads.
```
g-linville:
```suggestion
By default, SendGrid inbound webhook does not provide a way to verify the signature and payload coming from SendGrid. In order to verify requests that are coming from SendGrid, Obot provides a way to configure both `OBOT_SENDGRID_WEBHOOK_USERNAME` and `OBOT_SENDGRID_WEBHOOK_PASSWORD` to enable basic authentication for the webhook endpoint. So you can configure a URL with basic auth `https://{username}:{password}@{obot_server_url}/api/sendgrid` to ensure that only requests from SendGrid are received in Obot. For more details, refer to this example: [Handling SendGrid Inbound Parse](https://www.twilio.com/en-us/blog/microservice-template-handle-sendgrid-inbound-parse). It is recommended to configure these credentials to secure the endpoint and protect it from unverified payloads.
```

g-linville (approved):

PR #1322, opened 2025-01-17
chore: basic helm chart

drpebcak:


chart/templates/deployment.yaml
```diff
@@ -0,0 +1,151 @@
...
+              containerPort: 8080
+              protocol: TCP
+          livenessProbe:
+            exec:
+              command:
+                - "true"
+          readinessProbe:
+            exec:
+              command:
+                - "true"
```
g-linville: Doesn't need to be fixed right away, but I think there is probably some route we could hit to check server readiness. And if not, that's something we could add to the backend.

chart/templates/deployment.yaml
```diff
@@ -0,0 +1,151 @@
...
+                  name: {{ include "obot.config.secretName" . }}
+                  key: obotServerAuthClientSecret
+            {{- end }}
+            {{- if .Values.config.obotServerAuthCookieSecret }}
+            - name: "OBOT_SERVER_AUTH_COOKIE_SECRET"
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "obot.config.secretName" . }}
+                  key: obotServerAuthCookieSecret
+            {{- end }}
```
g-linville: These three secrets can be removed. With my auth provider refactoring, these are no longer provided to the obot server, and instead are provided when configuring an auth provider.

chart/templates/secret.yaml
```diff
@@ -0,0 +1,55 @@
...
+    {{- end }}
+  {{- if .Values.config.obotServerAuthClientID }}
+  obotServerAuthClientID: {{ .Values.config.obotServerAuthClientID | b64enc }}
+    {{- end }}
+  {{- if .Values.config.obotServerAuthClientSecret }}
+  obotServerAuthClientSecret: {{ .Values.config.obotServerAuthClientSecret | b64enc }}
+    {{- end }}
+  {{- if .Values.config.obotServerAuthCookieSecret }}
+  obotServerAuthCookieSecret: {{ .Values.config.obotServerAuthCookieSecret | b64enc }}
+    {{- end }}
```
g-linville: Should be removed here too.

chart/values.yaml
```diff
@@ -0,0 +1,83 @@
...
+
+  awsAccessKeyID: ""
+  awsRegion: ""
+  awsSecretAccessKey: ""
+  baaahThreadiness: "20"
+  githubAuthToken: ""
+  obotServerAuthAdminEmails: ""
+  obotServerAuthClientID: ""
+  obotServerAuthClientSecret: ""
+  obotServerAuthCookieSecret: ""
```
g-linville: Should be removed here too.

chart/values.yaml
```diff
@@ -0,0 +1,83 @@
+# replicaCount -- The number of Obot server instances to run
+replicaCount: 1
```
g-linville, line 2: Has it been tested with more than one replica? Just curious.

chart/templates/deployment.yaml
drpebcak: We would still need the cookie secret, wouldn't we? That's not directly related to the auth provider

chart/values.yaml
drpebcak: no, but im told it would probably work

chart/templates/deployment.yaml
drpebcak: @thedadams said he would add an endpoint for me

chart/templates/deployment.yaml
g-linville: Nope, those get automatically generated and stored for each auth provider internally. They are no longer able to be provided by the admin.

thedadams (approved):

PR #1338, opened 2025-01-20
fix: markdown embedded html & agent intro not firing onChange

ivyjeong13:
* shared Markdown component that takes in repeated props
* tweak to AgentForm.reset (if one of the fields is not supplied from `agent`, it fails the safeSchema check)
* add rehype-raw & rehype-sanitize to allow embedded html rendering and sanitize html

Screenshot 2025-01-20 at 1 47 24 PM


ryanhopperlowe (approved):
🙌

PR #1347, opened 2025-01-20
fix: hide error toast for reveal api during initial model provider configuration

ryanhopperlowe:
- also add ability to refresh files/knowledge in thread sidebar

### DIFF
--- a/ui/admin/app/lib/service/api/modelProviderApiService.ts
+++ b/ui/admin/app/lib/service/api/modelProviderApiService.ts
@@ -68,7 +68,8 @@ const revealModelProviderById = async (providerKey: string) => {
 		url: ApiRoutes.modelProviders.revealModelProviderById(providerKey).url,
 		method: "POST",
 		errorMessage:
-			"Failed to reveal configuration values on the requested model provider.",
+			"Failed to reveal configuration values on the requested modal provider.",
+		toastError: false,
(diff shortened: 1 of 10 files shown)


ui/admin/app/lib/service/api/modelProviderApiService.ts
```diff
@@ -68,7 +68,8 @@ const revealModelProviderById = async (providerKey: string) => {
 		url: ApiRoutes.modelProviders.revealModelProviderById(providerKey).url,
 		method: "POST",
 		errorMessage:
-			"Failed to reveal configuration values on the requested model provider.",
+			"Failed to reveal configuration values on the requested modal provider.",
+		toastError: false,
```
ivyjeong13, line 72: Just wanted to make sure, the PR title sounded like the error should only hide for the initial configuration but would this make sure the error toast always gets hidden?

ryanhopperlowe: Hmmm... That's a good point. I'll add that back

ryanhopperlowe:
> 👍 Had one lil question otherwise LGTM. I see the refresh knowledge/files is in here too, may be worth to mention in the merged commit message as well.

Yeah I accidentally added that commit to the wrong branch 😆

PR #1348, opened 2025-01-20
feat: open thread meta files accordion by default

ryanhopperlowe:
image

### DIFF
--- a/ui/admin/app/components/thread/ThreadMeta.tsx
+++ b/ui/admin/app/components/thread/ThreadMeta.tsx
@@ -127,7 +127,7 @@ export function ThreadMeta({
 					</table>
 				</div>
 
-				<Accordion type="multiple" className="mx-2">
+				<Accordion type="multiple" className="mx-2" defaultValue={["files"]}>
 					{files.length > 0 && (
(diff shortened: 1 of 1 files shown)


ivyjeong13:
Not related to this issue but wondering if that section should live update, if a user ends up with a file while interacting with the thread?

Example:
* Had a thread where I asked to create a csv file
* Went to check thread, saw the files section with the previously created csv
* Asked to create a different csv file
* I see the message the csv file was created but the files section isn't updated with the new csv. I can refresh to see it!
Screenshot 2025-01-21 at 9 57 18 AM

ryanhopperlowe:
> Not related to this issue but wondering if that section should live update, if a user ends up with a file while interacting with the thread?
> 
> Example:
> 
> * Had a thread where I asked to create a csv file
> * Went to check thread, saw the files section with the previously created csv
> * Asked to create a different csv file
> * I see the message the csv file was created but the files section isn't updated with the new csv. I can refresh to see it!
> 
> Screenshot 2025-01-21 at 9 57 18 AM

Yeah a refresh button here would be pretty useful

PR #1356, opened 2025-01-21
Fix: Don't ignore obot frames

cjellick:
In setting up the debug, we for some reason ignore a frame where the tool is named obot, but that causes us to actually show no output for user threads from the obot agent.


cjellick:
...did this on a foreign laptop from github's vscode editor. only hitch was that it pushed my branch to obot-platform rather than my fork. whoops. 😅

PR #1357, opened 2025-01-21
Fix: Don't ignore obot frame in debug view

cjellick:
When contructing the debug view, we would ignore certain frames, but it
seems we should not ignore frames where the tool is named obot, because
that is the name of the root frame for threads from the obot agent.

### DIFF
--- a/ui/admin/app/components/chat/CallFrames.tsx
+++ b/ui/admin/app/components/chat/CallFrames.tsx
@@ -56,8 +56,9 @@ const CallFrames = ({ calls }: { calls: CallsType }) => {
 
 		sortedCalls.forEach(([id, call]) => {
 			if (
-				call.tool?.name === "GPTScript Gateway Provider" ||
-				call.tool?.name === "Obot"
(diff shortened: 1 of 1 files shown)


ryanhopperlowe (approved):

thedadams:
I believe the reason this was added was to ignore the Obot model provider tool. My guess is that removing this will cause that tool call to show up in the debug view. Perhaps there is a better way to ignore that.

cjellick:
had a feeling that might be the case

cjellick:
Properly fixed and actually tested it @thedadams.

Arguably, we could _just_ rely on the model provider check, but i dont really know if that shows up anywhere else where we'd want to show it. Do you?

thedadams:
I don't believe there are any situations where model providers are shown there. The model providers are started by the obot proxy, which essentially is the obot model provider.

So, I think it's safe to just use the modelProvider field here.

I'm fine with the change either way.

thedadams (approved):

ivyjeong13 (approved):

Issue #1367, opened 2025-01-21
Auth - add configuration option to disable the bootstrap user


g-linville:
Some Obot administrators will likely want to disable bootstrap login as long as there is at least one auth provider configured. We should add an environment variable to enable or disable the bootstrap user.

g-linville:
Bootstrap user is now automatically disabled as soon as there is another admin user. I'm going to close this.

PR #1368, opened 2025-01-21
enhance: add option to disable bootstrap user

g-linville:
for #1367

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -79,13 +79,6 @@ spec:
                   name: {{ include "obot.config.secretName" . }}
                   key: githubAuthToken
             {{- end }}
-            {{- if .Values.config.bootstrapToken }}
-            - name: "OBOT_BOOTSTRAP_TOKEN"
-              valueFrom:
-                secretKeyRef:
-                  name: {{ include "obot.config.secretName" . }}
-                  key: bootstrapToken
-            {{- end }}
             - name: "OBOT_SERVER_ENABLE_AUTHENTICATION"
               value: "{{ .Values.config.enableAuthentication }}"
             - name: "OBOT_SERVER_ENABLE_BOOTSTRAP_USER"
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -79,6 +79,17 @@ spec:
                   name: {{ include "obot.config.secretName" . }}
                   key: githubAuthToken
             {{- end }}
+            {{- if .Values.config.bootstrapToken }}
+            - name: "OBOT_BOOTSTRAP_TOKEN"
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "obot.config.secretName" . }}
+                  key: bootstrapToken
+            {{- end }}
+            - name: "OBOT_SERVER_ENABLE_AUTHENTICATION"
+              value: "{{ .Values.config.enableAuthentication }}"
+            - name: "OBOT_SERVER_ENABLE_BOOTSTRAP_USER"
+              value: "{{ .Values.config.enableBootstrapUser }}"
             {{- if .Values.config.obotServerAuthAdminEmails }}
             - name: "OBOT_SERVER_AUTH_ADMIN_EMAILS"
               valueFrom:
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -79,10 +79,8 @@ spec:
                   name: {{ include "obot.config.secretName" . }}
                   key: githubAuthToken
             {{- end }}
-            - name: "OBOT_SERVER_ENABLE_AUTHENTICATION"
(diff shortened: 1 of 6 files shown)


chart/values.yaml
```diff
@@ -58,6 +58,9 @@ config:
   awsRegion: ""
   awsSecretAccessKey: ""
   baaahThreadiness: "20"
+  bootstrapToken: "" # Randomly generated at runtime if not set
+  enableAuthentication: false
+  enableBootstrapUser: true
```
g-linville: These are the defaults for Obot itself, but I'm not sure if we want them to be the Helm defaults or not.

g-linville: Heard back from Craig. We want auth enabled on the chart by default, so I flipped that one.

chart/values.yaml
```diff
@@ -58,6 +58,9 @@ config:
   awsRegion: ""
   awsSecretAccessKey: ""
   baaahThreadiness: "20"
+  bootstrapToken: "" # Randomly generated at runtime if not set
+  enableAuthentication: true
+  enableBootstrapUser: true
```
StrongMonkey: what happen if you don't define `enableBootstrapUser`? will that be false and disable the user? 

I feel like it is more intuitive to use `disableBoottrapUser` and default to false so that user don't accidently disable that in helm chart

g-linville: If it's undefined, it's true. That's the default that we have it set to in the code.

Our general practice is to avoid negatively stated boolean arguments at all costs, as they generally tend to be more confusing.

pkg/services/config.go
```diff
@@ -66,6 +66,7 @@ type Config struct {
 	EnvKeys                    []string `usage:"The environment keys to pass through to the GPTScript server" env:"OBOT_ENV_KEYS"`
 	KnowledgeSetIngestionLimit int      `usage:"The maximum number of files to ingest into a knowledge set" default:"3000" env:"OBOT_KNOWLEDGESET_INGESTION_LIMIT" name:"knowledge-set-ingestion-limit"`
 	EnableAuthentication       bool     `usage:"Enable authentication" default:"false"`
+	EnableBootstrapUser        bool     `usage:"Enables the bootstrap user, regardless of configured auth providers" default:"true"`
```
StrongMonkey, line 69: Same for CLI flag, I feel like disableBootstrapUser is more intuitive.

chart/values.yaml
StrongMonkey: ok, that's fine then.

StrongMonkey (approved):

chart/templates/secret.yaml
```diff
@@ -13,6 +13,9 @@ data:
   {{- if .Values.config.baaahThreadiness }}
   baaahThreadiness: {{ .Values.config.baaahThreadiness | b64enc }}
     {{- end }}
+  {{- if .Values.config.bootstrapToken }}
+  bootstrapToken: {{ .Values.config.bootstrapToken | b64enc }}
```
drpebcak: `obotBootstrapToken` is already in here.. this config is either a dupe, or you meant to add config for disabling the bootstrap user here.

drpebcak (changes requested):

chart/templates/secret.yaml
g-linville: Yep, that was an accident. Thanks. I just pushed the fix.

g-linville:
I re-requested review from everyone because I changed this to be a true disable, from Craig's feedback.

thedadams (approved):

StrongMonkey (approved):

chart/values.yaml
```diff
@@ -58,6 +58,8 @@ config:
   awsRegion: ""
   awsSecretAccessKey: ""
   baaahThreadiness: "20"
+  enableAuthentication: true
```
drpebcak: This is also a dupe

drpebcak (changes requested):

chart/templates/deployment.yaml
```diff
@@ -79,6 +79,10 @@ spec:
                   name: {{ include "obot.config.secretName" . }}
                   key: githubAuthToken
             {{- end }}
+            - name: "OBOT_SERVER_ENABLE_AUTHENTICATION"
+              value: "{{ .Values.config.enableAuthentication }}"
```
drpebcak: just as a note, most of these keys are just the camelcase version of the env var, so this one doesn't match the rest right now

chart/values.yaml
g-linville: Thanks, fixed

chart/templates/deployment.yaml
g-linville: I updated the bootstrap user one to put `obotServer` in front to make it consistent

drpebcak (approved):

Issue #1370, opened 2025-01-21
Knowledge - Not able to work with certain thread level knowledge files.


sangee2004:
**Steps to reproduce the problem:**
1. Create an agent with [acorn.io](http://acorn.io/) knowledge. Ingest all files.
2. Chat with the agent , add  "alphabet earnings" knowledge file for this thread - [2023q4-alphabet-earnings-release.pdf](https://github.com/user-attachments/files/18494564/2023q4-alphabet-earnings-release.pdf)

3. Ask about alphabet earnings , it returns results as expected.
4. Add  "Reunion under the starts" knowledge file  for this thread- [Reunion-Under-The-Stars.pdf](https://github.com/user-attachments/files/18494565/Reunion-Under-The-Stars.pdf)

5. Chat about information in this knowledge file and it fails to return any information.
There is call to knowledge tool and workspace_read file in this case

<img width="484" alt="Image" src="https://github.com/user-attachments/assets/7e203df5-0230-4751-9b7e-ed1f3732432b" />

Debug logs - [calls - 2025-01-21T084111.635.json](https://github.com/user-attachments/files/18494584/calls.-.2025-01-21T084111.635.json)

sangee2004:
This issue is not seen anymore when testing with latest build.

PR #1373, opened 2025-01-21
Fix: fix thread knowledge dataset not being included

StrongMonkey:
https://github.com/obot-platform/obot/issues/1370

We are not include thread knowledge dataset during the retrival process if there are multiple datasets.


iwilltry42 (approved):
LGTM, but we will likely end up with `No data description available,No data description available`, right? 😅

StrongMonkey:
@iwilltry42 shoot, yeah. But I think it should be fine.

thedadams (approved):

Issue #1175, opened 2025-01-09
Change behavior/relationship between workflow and triggers


cjellick:
Now that we have triggers on the workflow page, I think it is logical to delete triggers when we delete the workflow. However, as discussed, there are some cases where you might want to leave the trigger around to be used with a different workflow. So, when a user deletes a workflow, if it has triggers, can we prompt the user to ask if they want the associated triggers deleted as well?

thedadams:
For the backend change, it would be best to have a new endpoint: `DELETE /api/workflows/{id}/triggers` that will delete all triggers (email, cron, and webhooks) associated to a workflow ID. These objects should all have field selectors for this type of thing. Also, this should work for workflows that are already deleted.

ivyjeong13:
> [@ivyjeong13](https://github.com/ivyjeong13) The backend change has been merged.
> 
> To clean up the triggers that belong to a workflow, you can add this query parameter `delete-triggers=true`. The UX workflow will be the same as craig described, when user deletes a workflow, they get prompted with a dialog to confirm whether they want to clean up the triggers that belong the workflow, and if so, make the delete call with the parameters above.

Thanks @StrongMonkey !

sangee2004:
@ivyjeong13 When I delete a workflow (that has triggers) and use the "Keep" option when the following prompt is presented , I still see the schedule triggers relating to the workflow being deleted. 

<img width="664" alt="Image" src="https://github.com/user-attachments/assets/935ef309-d422-43a8-94e0-05292ff521d6" />


I do see the email/webhook related triggers not being deleted as expected.

Wanted to confirm if  schedule triggers being deleted in this case is expected behavior?

cjellick:
This isn't expected from my perspective.

sangee2004:
Scheduled triggers are now not deleted when the webhook associated with it is deleted with "Keep" triggers option.

PR #1381, opened 2025-01-21
Add delete-triggers parameters to clean up triggers

StrongMonkey:
https://github.com/obot-platform/obot/issues/1175

This PR adds a parameter `delete-triggers` to clean up existing triggers.

### DIFF
--- a/pkg/storage/apis/obot.obot.ai/v1/cronjob.go
+++ b/pkg/storage/apis/obot.obot.ai/v1/cronjob.go
@@ -28,12 +28,14 @@ func (c *CronJob) Get(field string) (value string) {
 	switch field {
 	case "spec.threadName":
 		return c.Spec.ThreadName
+	case "spec.workflow":
+		return c.Spec.Workflow
 	}
 	return ""
 }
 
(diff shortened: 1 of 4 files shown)


pkg/storage/apis/obot.obot.ai/v1/cronjob.go
```diff
@@ -28,12 +28,14 @@ func (c *CronJob) Get(field string) (value string) {
 	switch field {
 	case "spec.threadName":
 		return c.Spec.ThreadName
+	case "spec.workflow":
```
thedadams, line 31: Oh, I thought these already existed!

thedadams (approved):
It is possible for the `spec.workflow` fields here to be the alias and not the workflow ID directly. However, this functionality is not available in the UI so we are not concerned with cleaning such triggers up here.

pkg/storage/apis/obot.obot.ai/v1/cronjob.go
StrongMonkey: haha, yeah I though the same too until i ran into an error

Issue #1372, opened 2025-01-21
Allow credentials to support OAuth and personal access tokens


thedadams:
This feature encompases changes to the Obot OAuth credential that will allow tools to specify that it supports OAuth and PATs.

A tool will be able to specify that PATs are acceptable via a metadata field. If PATs are acceptable by the tool, then, when the user adds the tool, they will have the option of using OAuth or a personal access token.

### Backend changes
1. Add a new metadata field to tools that support PATs. I am going with `oauthPATAllowed`.
2. Change the OAuth credential in Obot tools to prompt for a PAT if no OAuth URLs are sent.
3. Change two things in Obot:
a. Make the `Global` field on OAuth apps default to false.
b. Send some configuration to the oauth tool to indicate that it can/should prompt for a PAT if not OAuth URLs are sent.

### Frontend changes
The flow for adding a tool that uses OAuth should be the following:
1. User adds tool to agent that requires some OAuth.
2. If the tool also indicates that is supports PAT, then ask the user which they would like to do.
a. If they choose PAT, then add the tool to the agent and you're done. Otherwise, continue with these steps.
3. If the OAuth app in question has not been configured, then prompt them to go configure the OAuth app and come back (or do a configuration modal right there if that is easy to do). Additionally, the OAuth app should not set the `global` field anymore.
4. If the OAuth app is configured, then add it to the `oauthApps` field on the agent at the same time the tool is added to the `tools` field.
a. One sidenote here is that we also need to remove the OAuth app from the `oauthApps` field on the agent when removing a tool, but only when removing the last tool that uses that OAuth app.

The UI also maintains "default OAuth apps" that are assumed to be configured, but a user can "replace" their configuration. An example of such an app is GitHub or Slack (and a non-example is Jira). The concept of a preconfigured or default OAuth app is going away. If a user wants to use an OAuth app, then it must be configured in Obot.

sangee2004:
Tools are now allowed to support OAUTH as well as PAT

1. When adding github tool, user is promoted to choose one of the following 2 options
<img width="661" alt="Image" src="https://github.com/user-attachments/assets/3095bf33-faff-4f7b-b19f-8fe803593d83" />

2. If PAT is chosen , then user is asked to enter AccessToken when chatting with the agent / when Pre authenticating the tool.

3. If Oauth is chosen, then user is presented with option to configure the oauth . Up on successful completion, oauth gets set as credential provider.
[...]

PR #1387, opened 2025-01-22
feat: add support for OAuth PATs

thedadams:
Instead of configuring and using OAuth, an app can specify that it supports using personal access tokens. If this is the case, then Obot will pass an extra environment variable to the oauth credential tool to indicate which integrations support tokens. If the oauth2 credential tool should prompt the user for a token instead of using OAuth, then Obot will not pass the environment variables that feed the URLs to the tool.

A side effect of this change is that OAuth apps no longer default to global.

Issue: https://github.com/obot-platform/obot/issues/1372
Requires: https://github.com/obot-platform/tools/pull/361


StrongMonkey (approved):

StrongMonkey:
>A side effect of this change is that OAuth apps no longer default to global.

What does this mean?

thedadams:
> A side effect of this change is that OAuth apps no longer default to global.

Right now, every OAuth app that is created is available to every agent/workflow. After this change, that will no longer be the case. Agents and workflows must explicitly designate which OAuth apps they want to use on their `oauthApps` field.

iwilltry42 (approved):

ui/admin/app/components/oauth-apps/CustomOAuthAppForm.tsx
```diff
@@ -172,10 +171,7 @@ export function CustomOAuthAppForm({
 							control={form.control}
 							onChange={(e) => {
 								if (deriveIntegrationFromName) {
-									form.setValue(
-										"integration",
-										convertToIntegration(e.target.value)
-									);
+									form.setValue("alias", convertToIntegration(e.target.value));
```
ivyjeong13: nit: With integration removed, prolly better to also convert this function name to something like `convertToAlias`!

ivyjeong13 (approved):
:+1: Checked integration -> alias changes and tested!

Issue #1131, opened 2025-01-06
o1 doesnt work


sanjay920:
I want to use OpenAI o1 to power my agent but it fails with the following errors in obot server:

```
[server] time="2025-01-06T15:01:42-08:00" level=info msg="Handled request: method POST, path /run" id=dcba5071-273b-44e4-b237-93ee4dc39a07
[server] 2025/01/06 15:01:42 ERROR sendRequestStream failed due to non-retriable statuscode code=400 status="400 Bad Request" tries=1 maxTries=6 failures="#1/6 error response received: error, status code: 400, message: Unsupported parameter: 'temperature' is not supported with this model.; exiting due to non-retriable error in try #1/6: 400 400 Bad Request"
[server] time="2025-01-06T15:01:42-08:00" level=info msg="Handled request: method POST, path /evaluate" id=81e0d164-fd56-496f-a3ba-00720245509b
[server] time="2025-01-06T15:01:42-08:00" level=error msg="failed to save state: failed to run: failed calling model for completion: request failed on non-retriable status-code: 400 400 Bad Request
failed to run: failed calling model for completion: request failed on non-retriable status-code: 400 400 Bad Request" logger=8/pkg/invoke
[server] time="2025-01-06T15:01:42-08:00" level=error msg="run failed: failed to run: failed calling model for completion: request failed on non-retriable status-code: 400 400 Bad Request
failed to run: failed calling model for completion: request failed on non-retriable status-code: 400 400 Bad Request" logger=8/pkg/invoke
```

<img width="2335" alt="image" src="https://github.com/user-attachments/assets/4f5fe667-326f-432a-99f4-2ef1de320da1" />

sangee2004:
With latest build   `obot:v0.0.0-dev+285d246c` , able to chat with agent using "o1" model.

PR #1398, opened 2025-01-22
Chore: support OPENAI O1 model

StrongMonkey:
Support openai o1 model. This is done by setting a special env in gptscript so that it will be recognized and tweak the api properly to accommate O1 model.

https://github.com/obot-platform/obot/issues/1131

require: https://github.com/gptscript-ai/gptscript/pull/937

### DIFF
--- a/pkg/render/render.go
+++ b/pkg/render/render.go
@@ -45,6 +45,20 @@ func Agent(ctx context.Context, db kclient.Client, agent *v1.Agent, oauthServerU
 		Credentials:  agent.Spec.Credentials,
 	}
 
+	if agent.Spec.Manifest.Model != "" {
+		var model v1.Model
+		if err := db.Get(ctx, kclient.ObjectKey{
+			Namespace: agent.Namespace,
+			Name:      agent.Spec.Manifest.Model,
+		}, &model); err != nil {
+			return nil, nil, err
+		}
+
+		if model.Spec.Manifest.Name == "o1" && model.Spec.Manifest.ModelProvider == "openai-model-provider" {
+			extraEnv = append(extraEnv, "OPENAI_MODEL_NAME=o1")
+		}
+	}
+
 	extraEnv = append(extraEnv, agent.Spec.Env...)
 
 	for _, env := range agent.Spec.Manifest.Env {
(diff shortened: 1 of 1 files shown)


pkg/render/render.go
```diff
@@ -45,6 +45,20 @@ func Agent(ctx context.Context, db kclient.Client, agent *v1.Agent, oauthServerU
...
+			Namespace: agent.Namespace,
+			Name:      agent.Spec.Manifest.Model,
+		}, &model); err != nil {
+			return nil, nil, err
+		}
+
+		if model.Spec.Manifest.Name == "o1" && model.Spec.Manifest.ModelProvider == "openai-model-provider" {
+			extraEnv = append(extraEnv, "GPTSCRIPT_INTERNAL_OPENAI_STREAMING=false")
+		}
+	}
```
thedadams, line 60: It seems like a decent compromise to do this, but I would prefer to not do this. A few reasons why:

- You would also have to check the default model alias for `llm` to see if the user has set that.
- Then you would also have to check if they are using `o1-mini` or have set the `llm-mini` default model alias to `o1-mini`. However, we probably don't want to disable all streaming in GPTScript just because the use has the `llm-mini` set to `o1-mini`.
- A user could have some other configuration or tool where the agent model and `llm` default model alias aren't using `o1`, but some other tool is trying to use `o1` directly.

I prefer that we find a way to add a "rewrite response" in the openai-model-provider to send back a streaming response for `o1` models when streaming is turned off.

StrongMonkey: - Yeah, need to account for default model. 
- So far o1-mini is not included in the PR, because it doesn't support system message so I would like it to be a seperate effort to support that.
- Yeah, this usecase is complicated. How do we do that in Obot? Does user have to register a custom tool that use o1 model in their gptscript script? I will definitely need to change the current approach to account for that.

> I prefer that we find a way to add a "rewrite response" in the openai-model-provider to send back a streaming response for o1 models when streaming is turned off.

This seems quite complicated, are you suggesting that we should modify the response from  openai-model-provider in order to support SSE event that our currrent chat library is using to make it compatiable? I feel like it is going to be a bit hacky and not really worth the effort.

StrongMonkey:
Closing in favor of adding support to the proxy.

PR #1400, opened 2025-01-22
Fix: Update docs

StrongMonkey:
More updates to workflow on email triggers.

### DIFF
--- a/docs/docs/02-concepts/06-workflows.md
+++ b/docs/docs/02-concepts/06-workflows.md
@@ -70,8 +70,8 @@ To create an email trigger
 2. Click **Create Trigger**.
 3. Select **Email** as the trigger type.
 4. Fill in the required fields:
-   - **Name**: The name of the email trigger. 
-   - **Alias**: The alias name will match the name of the email address that you want to receive emails from. For example, if the email address is `john@{email_server_name}`, the name will be `john`. If you leave this field blank, the alias name will be generated.
+   - **Name**: The name of the email trigger.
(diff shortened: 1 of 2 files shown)


iwilltry42 (approved):

docs/docs/02-concepts/06-workflows.md
```diff
@@ -70,7 +70,8 @@ To create an email trigger
 2. Click **Create Trigger**.
 3. Select **Email** as the trigger type.
 4. Fill in the required fields:
-   - **Name**: The name of the email trigger. The name will match the name of the email address that you want to receive emails from. For example, if the email address is `john@{email_server_name}`, the name will be `john`.
+   - **Name**: The name of the email trigger. 
+   - **Alias**: The alias name will match the name of the email address that you want to receive emails from. For example, if the email address is `john@{email_server_name}`, the name will be `john`. If you leave this field blank, the alias name will be generated.
```
sangee2004: _Alias: The alias name will match the name of the email address that you want to receive emails from.
This can be changed to:_
"This needs to be  altered as the alias name will match the name of the email address that you want to receive emails to."

_For example, if the email address is `john@{email_server_name}`, the name will be `john`. If you leave this field blank, the alias name will be generated._
This can be changed to:
 For example, if the recipient email address is `john@{email_server_name}`, the alias value should be set as `john`. If you leave this field blank,  alias  will be generated.

thedadams (approved):

PR #1411, opened 2025-01-23
Fix: Fix alias api when email receiver is first created

StrongMonkey:
When email receiver is first created in ui, the emailAddress will always return with random string at the start even if alias is set. This is problematic as it will change to use alias email on refresh once controller has set the alias. 

https://github.com/obot-platform/obot/issues/1412

### DIFF
--- a/pkg/api/handlers/emailreceiver.go
+++ b/pkg/api/handlers/emailreceiver.go
@@ -90,14 +90,24 @@ func convertEmailReceiver(emailReceiver v1.EmailReceiver, hostname string) *type
 	er := &types.EmailReceiver{
 		Metadata:              MetadataFrom(&emailReceiver),
 		EmailReceiverManifest: manifest,
-		AddressAssigned:       aliasAssigned,
+		AliasAssigned:         aliasAssigned,
 	}
+
 	if hostname != "" {
 		name := emailReceiver.Name
-		if er.AddressAssigned != nil && *er.AddressAssigned {
-			name = er.Alias
+		// If alias Name is set, we should return the alias email address if AliasAssigned is true, otherwise return the original email address
+		// return empty email address if AliasAssigned is not set because UI will need to poll until AliasAssigned is set
+		if emailReceiver.Spec.Alias != "" {
+			if er.AliasAssigned == nil {
+				er.EmailAddress = ""
+			} else if *er.AliasAssigned {
+				er.EmailAddress = fmt.Sprintf("%s@%s", emailReceiver.Spec.Alias, hostname)
+			} else {
+				er.EmailAddress = fmt.Sprintf("%s@%s", name, hostname)
+			}
+		} else {
+			er.EmailAddress = fmt.Sprintf("%s@%s", name, hostname)
 		}
-		er.EmailAddress = fmt.Sprintf("%s@%s", name, hostname)
 	}
 	return er
 }
--- a/ui/admin/app/components/webhooks/WebhookConfirmation.tsx
+++ b/ui/admin/app/components/webhooks/WebhookConfirmation.tsx
@@ -1,4 +1,4 @@
-import { useEffect } from "react";
+import { useEffect, useState } from "react";
 import { $path } from "safe-routes";
 import useSWR from "swr";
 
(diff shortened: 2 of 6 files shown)


pkg/api/handlers/emailreceiver.go
```diff
@@ -84,17 +84,20 @@ func convertEmailReceiver(emailReceiver v1.EmailReceiver, hostname string) *type
 	manifest := emailReceiver.Spec.EmailReceiverManifest
 
 	var aliasAssigned *bool
-	if emailReceiver.Generation == emailReceiver.Status.ObservedGeneration {
+	if emailReceiver.Generation == emailReceiver.Status.ObservedGeneration && emailReceiver.Status.ObservedGeneration != 0 {
```
StrongMonkey, line 87: Don't set the AliasAssigned when email receiver is first received.

pkg/api/handlers/emailreceiver.go
```diff
@@ -85,16 +85,19 @@ func convertEmailReceiver(emailReceiver v1.EmailReceiver, hostname string) *type
...
-		AddressAssigned:       aliasAssigned,
+		AliasAssigned:         aliasAssigned,
 	}
 	if hostname != "" {
 		name := emailReceiver.Name
-		if er.AddressAssigned != nil && *er.AddressAssigned {
+		// We should use the alias if the following is true
+		// 1. Alias is assigned and set to true
+		// 2. Alias is not assigned but Alias is set at the initial creation.
+		if (er.AliasAssigned != nil && *er.AliasAssigned) || (er.AliasAssigned == nil && er.Alias != "") {
```
thedadams: We don't want to do it this way because if `er.AliasAssigned == nil` then we may be sending back bad data.

The way this works for agent, for example, is the UI polls to make sure it gets back a value for `AliasAssigned` (which should remain a pointer. If that value is not set, then the UI knows it should keep polling until it gets true or false.

thedadams (changes requested):
I think this is actually a UI issue. The UI polls for aliases for agents. It should do the same thing for email aliases.

pkg/api/handlers/emailreceiver.go
StrongMonkey: Ok. I think the UI only poll once when receiver is created. The problem is that the current logic returns undetermistic result from API when AliasAssigned is set, like the address will change from `name@address.com` to `alias@address.com` when controller is settled. I think we should return empty email address when er.AliasAssigned == nil and have UI poll until the alias is set. what do you think?

thedadams: Yep. That sounds good to me!

StrongMonkey:
Closing in favor of adding the tweak in proxy

pkg/api/handlers/emailreceiver.go
StrongMonkey: Changed the code so that when webhook/email receiver's alias is set, it will return empty address until alias is being set. Otherwise it will return the right address depends on whether alias has been set successfully or not.

The change UI will need to make is that it should start polling for alias if `alias` is set until AliasAssigned is either false or true.

g-linville (approved):

thedadams (approved):

ui/admin/app/hooks/workflow-triggers/useWorkflowTriggers.ts
```diff
@@ -22,13 +23,24 @@ export function useWorkflowTriggers(props?: UseWorkflowTriggersProps) {
...
+	useEffect(() => {
+		if (
+			emailReceivers &&
+			emailReceivers.some(
+				(receiver) => receiver.aliasAssigned == null && receiver.alias
+			)
+		) {
+			mutateEmailReceivers();
+		}
+	}, [emailReceivers, mutateEmailReceivers]);
```
ryanhopperlowe: this seems like it would be better handled with polling

ryanhopperlowe: see how that is handled in `~/lib/hooks/knowledge/useKnowledgeFiles.ts` as an example

ui/admin/app/components/webhooks/WebhookConfirmation.tsx
```diff
@@ -43,6 +47,23 @@ export const WebhookConfirmation = ({
...
+
+	useEffect(() => {
+		if (webhookData?.alias && webhookData.aliasAssigned === undefined) {
+			const interval = setInterval(async () => {
+				await getWebhookData.mutate();
+			}, 1000);
+
+			return () => clearInterval(interval);
+		}
+	}, [webhookData?.alias, webhookData?.aliasAssigned]);
```
ryanhopperlowe: same here, let's use the polling capability built into useSWR

StrongMonkey:
@ryanhopperlowe @ivyjeong13 would appreciate you taking a look at the frontend code piece.

ryanhopperlowe (changes requested):
Sorry had this review started, but never submitted

ryanhopperlowe (approved):

PR #1414, opened 2025-01-23
fix: bootstrap: disable bootstrap when another admin user exists

g-linville:
This makes the following changes to how initial auth and bootstrap work:
- The bootstrap token is now stored as a credential, based on the very first value set for it: either randomly generated, or from OBOT_BOOTSTRAP_TOKEN
  - OBOT_BOOTSTRAP_TOKEN can always be used to override it
- The bootstrap user is only available until there is another admin user in the database, after which it is no longer available
  - It can be made available again by setting OBOT_SERVER_FORCE_ENABLE_BOOTSTRAP=true
- `/api/bootstrap` returns a JSON string `{"enabled":true}` if bootstrap login is enabled (auth is enabled and bootstrap is enabled), or `{"enabled":false}` if auth is disabled or bootstrap is disabled

### DIFF
--- a/pkg/bootstrap/bootstrap.go
+++ b/pkg/bootstrap/bootstrap.go
@@ -3,61 +3,134 @@ package bootstrap
 import (
 	"context"
 	"crypto/rand"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
 	"strings"
 
+	"github.com/gptscript-ai/go-gptscript"
 	types2 "github.com/obot-platform/obot/apiclient/types"
 	"github.com/obot-platform/obot/pkg/api"
 	"github.com/obot-platform/obot/pkg/api/authz"
 	"github.com/obot-platform/obot/pkg/gateway/client"
-	"github.com/obot-platform/obot/pkg/gateway/server/dispatcher"
 	"github.com/obot-platform/obot/pkg/gateway/types"
-	"github.com/obot-platform/obot/pkg/system"
 	"k8s.io/apiserver/pkg/authentication/authenticator"
 	"k8s.io/apiserver/pkg/authentication/user"
 )
 
-const bootstrapCookie = "obot-bootstrap"
+const (
+	bootstrapCookie            = "obot-bootstrap"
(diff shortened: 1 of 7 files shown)


pkg/bootstrap/bootstrap.go
```diff
@@ -3,61 +3,134 @@ package bootstrap
...
-	} else if !enableBootstrapUser {
-		configuredAuthProviders, err := d.ListConfiguredAuthProviders(ctx, system.DefaultNamespace)
-		if err == nil && len(configuredAuthProviders) == 0 {
-			fmt.Printf("WARNING: Bootstrap user is disabled, and no auth providers are configured. You will be unable to log in to Obot.\n")
+			printToken(token)
 		}
 	}
 
+	if len(token) < 6 {
+		return nil, errors.New("error: bootstrap token must be at least 6 characters")
```
g-linville, line 77: figured it would be good to have something like this

pkg/bootstrap/bootstrap.go
```diff
@@ -3,61 +3,134 @@ package bootstrap
...
-	"github.com/obot-platform/obot/pkg/system"
 	"k8s.io/apiserver/pkg/authentication/authenticator"
 	"k8s.io/apiserver/pkg/authentication/user"
 )
 
-const bootstrapCookie = "obot-bootstrap"
+const (
+	bootstrapCookie            = "obot-bootstrap"
+	bootstrapCredential        = "obot-bootstrap"
+	bootstrapCredentialContext = "obot-bootstrap"
```
thedadams: Having three constants with the same name seems strange. Can we have one constant with a reasonable name to use for all three?

thedadams (approved):

pkg/bootstrap/bootstrap.go
g-linville: Yeah, we can do that. I defined three separate constants in case we might want to change one of them for some reason, but I doubt we will.

ryanhopperlowe (approved):

PR #1421, opened 2025-01-23
fix: store auth provider in the same cookie as the token

g-linville:
Previously, after the user logs in, we would store which auth provider they used in a cookie. This was causing problems with otto8. This PR changes it so that the auth provider gets stored within the same cookie. The logic is a bit messy, but we basically just intercept and read/modify the cookie before we send it to the oauth2proxy, and when it sets the cookie in the browser, we also intercept it there to insert the info about which auth provider it is.

Because it needed to be taken care of, this PR also filters the bootstrap user out of `/api/users` and update the "last admin" error logic to exclude the bootstrap user there as well.

### DIFF
--- a/pkg/gateway/client/user.go
+++ b/pkg/gateway/client/user.go
@@ -39,7 +39,7 @@ func (c *Client) DeleteUser(ctx context.Context, username string) (*types.User, 
 		if existingUser.Role.HasRole(types2.RoleAdmin) {
 			var adminCount int64
 			// We filter out empty email users here, because that is the bootstrap user.
-			if err := tx.Model(new(types.User)).Where("role = ? and email != '' and ", types2.RoleAdmin).Count(&adminCount).Error; err != nil {
+			if err := tx.Model(new(types.User)).Where("role = ? and email != ''", types2.RoleAdmin).Count(&adminCount).Error; err != nil {
 				return err
 			}
 
--- a/pkg/gateway/client/user.go
+++ b/pkg/gateway/client/user.go
@@ -38,7 +38,8 @@ func (c *Client) DeleteUser(ctx context.Context, username string) (*types.User, 
 
 		if existingUser.Role.HasRole(types2.RoleAdmin) {
 			var adminCount int64
-			if err := tx.Model(new(types.User)).Where("role = ?", types2.RoleAdmin).Count(&adminCount).Error; err != nil {
+			// We filter out empty email users here, because that is the bootstrap user.
+			if err := tx.Model(new(types.User)).Where("role = ? and email != '' and ", types2.RoleAdmin).Count(&adminCount).Error; err != nil {
 				return err
 			}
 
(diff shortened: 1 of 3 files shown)


pkg/gateway/client/user.go
```diff
@@ -38,7 +38,8 @@ func (c *Client) DeleteUser(ctx context.Context, username string) (*types.User,
 
 		if existingUser.Role.HasRole(types2.RoleAdmin) {
 			var adminCount int64
-			if err := tx.Model(new(types.User)).Where("role = ?", types2.RoleAdmin).Count(&adminCount).Error; err != nil {
+			// We filter out empty email users here, because that is the bootstrap user.
+			if err := tx.Model(new(types.User)).Where("role = ? and email != '' and ", types2.RoleAdmin).Count(&adminCount).Error; err != nil {
```
thedadams: Is there an extra `and` at the end here?

pkg/proxy/proxy.go
```diff
@@ -34,16 +42,39 @@ func NewProxyManager(dispatcher *dispatcher.Dispatcher) *Manager {
...
+
+	cookieOriginalValue := cookie.Value
+	cookieValue, err := url.QueryUnescape(cookie.Value)
 	if err != nil {
 		return nil, false, nil
 	}
 
-	proxy, err := pm.createProxy(req.Context(), c.Value)
+	var contents CookieContents
+	if err = json.Unmarshal([]byte(cookieValue), &contents); err != nil {
```
StrongMonkey, line 57: This should return error?

pkg/proxy/proxy.go
```diff
@@ -34,16 +42,39 @@ func NewProxyManager(dispatcher *dispatcher.Dispatcher) *Manager {
...
 
 func (pm *Manager) AuthenticateRequest(req *http.Request) (*authenticator.Response, bool, error) {
-	c, err := req.Cookie(AuthProviderCookie)
+	cookie, err := req.Cookie(ObotAccessTokenCookie)
+	if err != nil {
+		return nil, false, nil
+	}
+
+	cookieOriginalValue := cookie.Value
+	cookieValue, err := url.QueryUnescape(cookie.Value)
```
StrongMonkey, line 51: are we ignoring error here?

pkg/proxy/proxy.go
```diff
@@ -62,19 +95,37 @@ func (pm *Manager) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			cookie.Value = contents.Token
+			allCookies := r.Cookies()
+			r.Header.Del("Cookie")
+			for _, c := range allCookies {
+				if c.Name != ObotAccessTokenCookie {
+					r.AddCookie(c)
+				}
+			}
+			r.AddCookie(cookie)
+		}
```
StrongMonkey, line 125: Looks like error is not handled when cookieValue is malformed.

pkg/proxy/proxy.go
g-linville: Yeah if we were unable to unescape the cookie value, then we just pass on to the next authenticator in the chain. There's not really anything else for us to do, or any reason to log an error.

pkg/proxy/proxy.go
g-linville: No, that will break authentication for existing users who still have the cookie in their old format. Instead we just want to pass them on to the next authenticator.

pkg/proxy/proxy.go
g-linville: Yep, this also is intentional, so that we can have a smooth migration from the old way to the new way. We just leave the cookie alone if it is bad, and the user will have to reauthenticate anyway, which will replace it with a new, good cookie.

pkg/gateway/client/user.go
g-linville: Thank you for catching this!

StrongMonkey (approved):

njhale (approved):

PR #1428, opened 2025-01-23
Fix: email trigger is not returning back to trigger page after creation

StrongMonkey:
Fixed an issue where email trigger is not returning to trigger page after creation.


ryanhopperlowe (approved):
LGTM

PR #1435, opened 2025-01-24
fix: auth: don't delete the csrf cookie

g-linville:
There was an issue with how I was trying to save the existing cookies when replacing the value of the access token cookie, that was causing all other cookies to get deleted. This fixes that by properly storing the values of the other cookies and then re-adding them to the request.

### DIFF
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -64,15 +64,15 @@ func (pm *Manager) AuthenticateRequest(req *http.Request) (*authenticator.Respon
 	}
 
 	// Overwrite the cookie with just the token.
-	cookie.Value = contents.Token
-	req.Header.Del("Cookie")
-	req.AddCookie(cookie)
+	if err := replaceTokenCookie(contents.Token, req); err != nil {
+		return nil, false, err
+	}
 
 	// Reset the cookie value after authenticating.
 	defer func() {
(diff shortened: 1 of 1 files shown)


cjellick (approved):

pkg/proxy/proxy.go
```diff
@@ -327,3 +321,28 @@ func (p *Proxy) authenticateRequest(req *http.Request) (*authenticator.Response,
...
+		if c.Name != ObotAccessTokenCookie {
+			otherCookies = append(otherCookies, *c)
+		}
+	}
+
+	req.Header.Del("Cookie")
+
+	for _, c := range otherCookies {
+		req.AddCookie(&c)
+	}
```
thedadams, line 344: Could you do something like this so you only have to loop once?
```suggestion
	cookies := req.Cookie()
	req.Header.Del("Cookie")
	for _, c := range cookies {
		if c.Name != ObotAccessTokenCookie {
			req.AddCookie(c)
		}
	}
```

thedadams (approved):
Approving in case my suggestion doesn't work.

njhale (approved):

pkg/proxy/proxy.go
g-linville: No, this is what I was doing before and the `cookies` array is full of pointers that end up getting wiped

PR #1439, opened 2025-01-24
enhance: loop over cookies once in the proxy

thedadams:


g-linville:
Tested and this does work.

g-linville (approved):

StrongMonkey (approved):

PR #1440, opened 2025-01-24
Fix: fix Docs build, add doc build CI

StrongMonkey:


docs/docs/02-concepts/06-workflows.md
```diff
@@ -61,8 +61,8 @@ Refer to your service's webhook documentation to find the values to set for thes
 This will require configuration as described in the [Email Webhook Configuration](/configuration/email-webhook#configure-obot) documentation.
 :::
 
-You can trigger a workflow by sending an email to an email address configured in Obot. The email address should be in the format of `{name}@{email_server_name}`.
```
drpebcak, line 64: wont`{name}` break this in the same way?

drpebcak (changes requested):

docs/docs/02-concepts/06-workflows.md
StrongMonkey: Hmm it doesn't, looks like it has to be `{name}`. But I can update it if you want.

drpebcak (approved):

PR #1445, opened 2025-01-24
fix: return error when trying to use an unconfigured auth provider 

g-linville:
If the user had an old access token with a now-deconfigured auth provider and went to `/oauth2/start`, they would end up getting a `credential not found` error. This makes the error message a bit clearer, and deletes their cookie, so that if they try again, they will just end up using whichever auth provider is available.

This is a minor fix and does not need to go in before the release.

### DIFF
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"net/url"
+	"slices"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/proxy/proxy.go
```diff
@@ -142,10 +142,43 @@ func (pm *Manager) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		configuredProviders, err := pm.dispatcher.ListConfiguredAuthProviders(r.Context(), namespace)
+		if err != nil {
+			http.Error(w, fmt.Sprintf("failed to list configured auth providers: %v", err), http.StatusInternalServerError)
+			return
+		}
+
+		if !slices.Contains(configuredProviders, name) {
+			// The requested auth provider isn't configured. Return an error, unless the user is signing out, in which case, just redirect.
+			if r.URL.Path == "/oauth2/sign_out" {
+				http.Redirect(w, r, rdParam, http.StatusFound)
```
thedadams, line 179: Should we clear the cookie in this case, too?

thedadams (approved):

pkg/proxy/proxy.go
g-linville: Good call

thedadams (approved):

PR #1451, opened 2025-01-25
Fix: fix object validator to allow null

StrongMonkey:
validator from zod is throwing error complaining about object being null.


ivyjeong13:
@StrongMonkey Will close this one out! Since the feat got reverted, re-opening with your fix also included! : )

PR #1458, opened 2025-01-27
chore: switch to async agent/workflow invocations

ryanhopperlowe:
run agent invocations asynchronously to stop opening unnecessary event streams in the UI.


cjellick (approved):

sangee2004:
We now call agent and workflow with `async=true` when chatting with agent/workflow. We do not see "context cancelled" error anymore when chat responses take more than 30 secs to stream

Screenshot 2025-01-28 at 10 29 15 AM


We still have the issue with chat message (and tool calls ) being presented to the user multiple times when message response is being streamed . This issue is seen in both admin and user ui and tracked in https://github.com/obot-platform/obot/issues/1470

Issue #1254, opened 2025-01-14
"Failed to fetch agent" error seen when creating a handle for the agent


sangee2004:
**Steps to reproduce the problem:**

1. Create an agent.
2. Using `Edit Agent url alias` option, set a handle

Notice that "Failed to fetch agent" error message is seen which gets autodismissed.

https://www.loom.com/share/e1b7bf19ce694922b25cf7f7288b83ad?sid=d3df22ef-7064-45be-a8d2-9e2b84f09ee4

There is call to the alias that fails with "404" and then eventually succeeds.
<img width="1198" alt="Screenshot 2025-01-14 at 11 56 28 AM" src="https://github.com/user-attachments/assets/e1d0159b-5bdb-4f75-a4c5-882775227724" />

**Expected Behavior:**
There should be no error messages presented to the user when alias is set for agents.

sangee2004:
This issue is not seen anymore when testing with latest build.

PR #1478, opened 2025-01-28
Fix: make sure only to load alias when aliaAssigned is true

StrongMonkey:
https://github.com/obot-platform/obot/issues/1254

Changed the alias logic in the frontend to match the behavior in backend. The idea is that when agent.AliasAssigned is null it means that the agent needs controller reconcilation and UI should start polling until it reaches a final state.

### DIFF
--- a/ui/admin/app/components/agent/Agent.tsx
+++ b/ui/admin/app/components/agent/Agent.tsx
@@ -42,19 +42,6 @@ export function Agent({ className, currentThreadId, onRefresh }: AgentProps) {
 	const [agentUpdates, setAgentUpdates] = useState(agent);
 
 	useEffect(() => {
-		if (agent.aliasAssigned === undefined && agent.alias) {
-			const intervalId = setInterval(() => {
-				refreshAgent();
-				if (agent.aliasAssigned !== undefined) {
-					clearInterval(intervalId);
-				}
-			}, 1000);
-
-			return () => clearInterval(intervalId);
-		}
-	}, [agent, refreshAgent]);
-
-	useEffect(() => {
 		setAgentUpdates((prev) => {
 			return {
 				...agent,
--- a/ui/admin/app/components/agent/Agent.tsx
+++ b/ui/admin/app/components/agent/Agent.tsx
@@ -1,11 +1,9 @@
 import { GearIcon } from "@radix-ui/react-icons";
 import { BlocksIcon, LibraryIcon, PlusIcon, WrenchIcon } from "lucide-react";
 import { useCallback, useEffect, useState } from "react";
-import useSWR from "swr";
 
 import { Agent as AgentType } from "~/lib/model/agents";
 import { AssistantNamespace } from "~/lib/model/assistants";
-import { AgentService } from "~/lib/service/api/agentService";
 import { cn } from "~/lib/utils";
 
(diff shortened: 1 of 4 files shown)


ui/admin/app/components/agent/Agent.tsx
```diff
@@ -41,46 +39,31 @@ export function Agent({ className, currentThreadId, onRefresh }: AgentProps) {
...
-		});
-	}, [agent]);
-
-	const getLoadingAgent = useSWR(
-		AgentService.getAgentById.key(loadingAgentId),
-		({ agentId }) => AgentService.getAgentById(agentId),
-		{
-			revalidateOnFocus: false,
-			refreshInterval: 2000,
+		if (agent.aliasAssigned === undefined && agent.alias) {
```
StrongMonkey: I refactored the logic a bit so that

1. When agent.aliasAssigned is not set and agent.alias is set, we should poll agent until it is being set so we know the final state.
2. Once UI makes a update for alias, set aliasAssigned to null and have it repoll agent until it is set.

StrongMonkey:
@ryanhopperlowe would appreciate some eyes to catch some corner cases

ui/admin/app/components/agent/Agent.tsx
```diff
@@ -41,46 +39,31 @@ export function Agent({ className, currentThreadId, onRefresh }: AgentProps) {
...
-			setAgentUpdates((prev) => {
-				return {
-					...prev,
-					aliasAssigned: data.aliasAssigned,
-				};
-			});
-			setLoadingAgentId("");
-		}
-	}, [getLoadingAgent, loadingAgentId]);
+		setAgentUpdates((prev) => {
```
StrongMonkey, line 45: This is a tricky case because when you update an agent, it would put agent into reconcilation state where it will wipe out aliasAssigned, but it doesn't indicate where alias needs reconcilation or not. So the workaround in UI is that it should never reset agent.aliasAssigned unless it is specifically trying to modify the alias. Otherwise every random update will cause `aliasAssigned` to be null and make alias url spinning.

ui/admin/app/components/agent/Agent.tsx
```diff
@@ -41,46 +39,31 @@ export function Agent({ className, currentThreadId, onRefresh }: AgentProps) {
...
+				refreshAgent();
+				if (agent.aliasAssigned !== undefined) {
+					clearInterval(intervalId);
+				}
+			}, 1000);
+
+			return () => clearInterval(intervalId);
 		}
-	);
+	}, [agent, refreshAgent]);
```
ryanhopperlowe: instead of adding a useEffect, can you implement polling the same way we do in the `useKnowledgeFiles` hook?

This would mean you would have to move this logic into the AgentContext component.

ryanhopperlowe: (optional) to add increased traceability, you could expose some sort of method like `pollAgentUntil` to show exactly where polling is started.

```tsx
const { ..., pollAgentUntil } = useAgent();

return (
  ...
  <AliasSection agent={agent} onChange={(changes) => {
    partialSetAgent(changes);
    pollAgentUntil((a) => a.aliasAssigned !== undefined || !a.alias);
  }} />
  ...
)
```

ryanhopperlowe (approved):

PR #1485, opened 2025-01-28
Chore: Add lint check for user ui

cjellick:
This PR made me realize that user ui only PRs will never pass the lint check
https://github.com/obot-platform/obot/pull/1484

### DIFF
--- a/.github/workflows/user-ui.yml
+++ b/.github/workflows/user-ui.yml
@@ -1,4 +1,4 @@
-name: admin
+name: user
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


.github/workflows/user-ui.yml
```diff
@@ -0,0 +1,39 @@
+name: admin
```
njhale:
```suggestion
name: user
```

njhale (changes requested):

njhale (approved):

thedadams (approved):

PR #1487, opened 2025-01-28
chore: backend: support / for the oauth2 callback URL

g-linville:
for #1488

This is a small change to support just `/` as the OAuth2 callback URL for auth providers. There will be some slight changes needed for the frontend and the auth provider tools. But that will be pretty easy to do. Merging this now will not affect anything.

### DIFF
--- a/ui/handler.go
+++ b/ui/handler.go
@@ -24,28 +24,43 @@ func Handler(devPort int, client kclient.Client) http.Handler {
 		lock:   new(sync.RWMutex),
 	}
 
-	if devPort == 0 {
-		return server
+	if devPort != 0 {
+		server.rp = &httputil.ReverseProxy{
+			Director: func(r *http.Request) {
+				r.URL.Scheme = "http"
+				if strings.HasPrefix(r.URL.Path, "/admin") {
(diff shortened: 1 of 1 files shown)


ui/handler.go
```diff
@@ -24,28 +24,43 @@ func Handler(devPort int, client kclient.Client) http.Handler {
...
+		server.rp = &httputil.ReverseProxy{
+			Director: func(r *http.Request) {
+				r.URL.Scheme = "http"
+				if strings.HasPrefix(r.URL.Path, "/admin") {
+					r.URL.Host = fmt.Sprintf("localhost:%d", devPort)
+				} else {
+					r.URL.Host = fmt.Sprintf("localhost:%d", devPort+1)
+				}
+			},
+		}
```
g-linville, line 37: I moved the reverse proxy into the uiServer object, rather than returning it as an alternative to the uiServer. That way we can have one unified check for the oauth-related query parameters, before serving the request with either the reverse proxy (for the `make dev` setup) or the normal uiServer.

thedadams (approved):

njhale (approved):

Issue #1133, opened 2025-01-07
Authorization status is being evaluated constantly on the tools when changes are made to other fields in Agent/workflow


sangee2004:
**Steps to reproduce the problem:**
Scenario 1: 
1. Create a workflow with tools that requires authentication. 
2. Add a step and start typing the contents for the step.
Notice that Authorization status is being evaluated constantly when entering the steps for a workflow.

https://www.loom.com/share/567ecba6e0364c13b299e7b49410beea?sid=7dc383df-734e-4c97-a3c4-07ef368e83c4

**Expected Behavior:**
Ideally we should not have to be constantly evaluating authorization status for the tools when entering the steps for a workflow.

**Scenario2:**
1. Create an agent with tools that require authentication.
2. Try to add/edit contents in instruction filed.

Notice that authorization status for the tools is being constantly evaluated when instruction field is being entered.

**Expected Behavior:**
Ideally we should not have to evaluate the tool's authorization status at this time since it is already determined.

ryanhopperlowe:
@cjellick @thedadams  - this has to do with the backend response. I think this should start as a backend task.

thedadams:
This isn't a UI issue.

sangee2004:
This issue is not seen anymore when testing with latest build.

Authorization status for tools is not being evaluated constantly on existing tools when other tools are added/other changes are made to the agent/workflow.

PR #1496, opened 2025-01-28
Fix: always return toolInfo and poll for it when necessary

StrongMonkey:
The API will return empty toolInfo when it is trying to reconcile. This causes client to reconcile the toolStatus everytime agent is changed, even when no tool is being updated. Instead API can always return the toolInfo and client should be polling when a missing toolInfo is not present.

https://github.com/obot-platform/obot/issues/1133


pkg/api/handlers/agent.go
```diff
@@ -212,19 +212,17 @@ func convertAgent(agent v1.Agent, textEmbeddingModel, baseURL string) (*types.Ag
...
-		toolInfos = &agent.Status.ToolInfo
 	}
 
 	return &types.Agent{
 		Metadata:           MetadataFrom(&agent, links...),
 		AgentManifest:      agent.Spec.Manifest,
 		AliasAssigned:      aliasAssigned,
 		AuthStatus:         agent.Status.AuthStatus,
-		ToolInfo:           toolInfos,
+		ToolInfo:           &agent.Status.ToolInfo,
```
StrongMonkey: @thedadams I think this should be fine but let me know if you think it will break other places.

The downside I can think of is that if toolInfo is bad and never set, we will still return the old data instead of empty.(Which should is fine for client though)

thedadams (changes requested):
This does break unauthenticating a tool in the UI. If I unauthenticate a tool, it will look like it is still authetnicated until I refresh the page.

StrongMonkey:
@thedadams yeah, you are right. This is a bit tricky to deal with in backend.

@ryanhopperlowe I end up having a different fix in frontend. The problem in the backend is that right now if you make any updates in agent, it would put agent into reconcilation state whether it is tool changes or not. So I refactored the tool code a little to accommodate that.

ui/admin/app/components/agent/shared/ToolAuthenticationStatus.tsx
```diff
@@ -56,7 +52,12 @@ export function ToolAuthenticationStatus({
 
 		if (error) return;
 
-		onUpdate({ ...toolInfo, authorized: false });
+		setToolInfo({
```
StrongMonkey, line 58: When tool is being authorized or deauthorized, we should just remove the tool from toolInfo and repoll the agent until toolInfo is set.

ui/admin/app/hooks/toolAuth/useToolAuthPolling.ts
```diff
@@ -28,29 +32,32 @@ export function useToolAuthPolling(
...
-
-	const { tools, toolInfo } = getInfo();
-
-	// when tool credentials are processing the api will respond with { toolInfo: null }
-	// we need to poll until the toolInfo is not null or there are no tools
-	const shouldPoll = !!tools?.length && !toolInfo;
-	if (shouldPoll !== isPolling) setIsPolling(shouldPoll);
-
-	return { toolInfo, isPolling };
+	useEffect(() => {
```
ryanhopperlowe, line 39: why did we need to wrap this in a useEffect?

ryanhopperlowe: I might be wrong, but I'm 90% sure this will be more performant, and safer to run this logic outside of a useEffect.

ui/admin/app/hooks/toolAuth/useToolAuthPolling.ts
```diff
@@ -28,29 +32,32 @@ export function useToolAuthPolling(
...
+					return { tools: agentTools, toolInfo: agent?.toolInfo };
+				case AssistantNamespace.Workflows:
+					return { tools: workflow?.tools, toolInfo: workflow?.toolInfo };
+				default:
+					return {};
+			}
+		};
+
+		const { tools, toolInfo: toolInfoFromAgent } = getInfo();
+		if (toolInfoFromAgent) setToolInfo(toolInfoFromAgent);
```
ryanhopperlowe, line 58: might need to add a check here though:

```suggestion
		if (toolInfoFromAgent && toolInfoFromAgent !== toolInfo) setToolInfo(toolInfoFromAgent);
```

ui/admin/app/components/agent/shared/ToolAuthenticationStatus.tsx
```diff
@@ -107,7 +108,7 @@ export function ToolAuthenticationStatus({
 				</TooltipTrigger>
 
 				<TooltipContent>
-					This tool does not require authentication.
+					<>This tool does not require authentication.</>
```
ryanhopperlowe: nit: you can remove this fragment

ui/admin/app/components/agent/shared/ToolAuthenticationStatus.tsx
```diff
@@ -116,12 +117,16 @@ export function ToolAuthenticationStatus({
 		<>
 			<Tooltip>
 				<TooltipContent className="max-w-xs">
-					{authorized ? (
+					{loading ? (
+						<>
```
ryanhopperlowe: nit: don't need this fragment either

ryanhopperlowe (changes requested):
couple of things to fix. Plus the 1 issue we discussed offline

ui/admin/app/hooks/toolAuth/useToolAuthPolling.ts
StrongMonkey: It will keep rendering the component forever if not done in useEffect. Because now toolInfo will be set everytime agent is polled.

ui/admin/app/hooks/toolAuth/useToolAuthPolling.ts
StrongMonkey: I think it should be fine to always set it though. It doesn't do deepequal right? anyway agent. toolInfo should be source of truth and always set on the toolInfo variable.

StrongMonkey:
@ryanhopperlowe I pushed the fix which should address your comment and fix the cancel issue.

ryanhopperlowe (approved):

thedadams (approved):

PR #1522, opened 2025-01-29
feat: add vitest & react-testing-library

ivyjeong13:
* install vitest & react-testing-library
* add some base test cases for agent

Screenshot 2025-01-30 at 12 06 05 PM


ui/admin/app/lib/model/agents.ts
```diff
@@ -9,25 +9,25 @@ export const KNOWLEDGE_TOOL = "knowledge";
 export type AgentBase = {
 	name: string;
 	description: string;
-	temperature?: number;
-	cache?: boolean;
+	temperature?: number | null;
```
ivyjeong13, line 12: While going through and making mocks, noticed some properties that had null that didn't match with your types so updated them as well. Lemme know if that's not needed and can revert!

ui/admin/app/lib/model/__mocks__/agents.ts
```diff
@@ -0,0 +1,44 @@
+import { Agent } from "~/lib/model/agents";
+
+export const mockedAgent: Agent = {
```
ivyjeong13: Thinking maybe just throwing all the mocks into the test directory as well instead?

ui/admin/vite.config.ts
```diff
@@ -4,6 +4,25 @@ import { defineConfig } from "vite";
...
+	plugins: [
+		!process.env.VITEST && reactRouter(),
+		tsconfigPaths(),
+		safeRoutes(),
+	],
 	base: "/admin/",
+	server: {
+		watch: {
+			// Exclude test files from HMR
+			ignored: !process.env.VITEST
```
ivyjeong13, line 16: Added changes here to avoid hot reloading React application when changing test files while running `make dev`. But when watching tests, changes in files will still trigger a re-run of the test suites.

ui/admin/vite.config.ts
```diff
@@ -4,6 +4,25 @@ import { defineConfig } from "vite";
 import tsconfigPaths from "vite-tsconfig-paths";
 
 export default defineConfig({
-	plugins: [reactRouter(), tsconfigPaths(), safeRoutes()],
+	plugins: [
+		!process.env.VITEST && reactRouter(),
```
ryanhopperlowe, line 8: why don't we want react router when testing?

ui/admin/app/components/agent/__tests__/Agent.test.tsx
```diff
@@ -0,0 +1,209 @@
...
+			"description",
+			mockedAgent.description || "Add a description...",
+			"placeholder",
+		],
+		["prompt", "Instructions", "textbox", 2],
+		["introductionMessage", "Introductions", "textbox"],
+	])("Updating %s triggers save", async (field, searchFor, as, index = 0) => {
+		render(
+			<AgentProvider agent={mockedAgent}>
+				<Agent />
```
ryanhopperlowe, line 92: should we add a way to change the debounce value for updates? (maybe a prop with default value) Waiting 1 second for a test is pretty crazy

ivyjeong13:
Also if we're interested in applying gherkins/BDD, I can add on vitest-cucumber. It will require a little tweaking of the tests but can be done.

ui/admin/vite.config.ts
ivyjeong13: Still looking for a solution, but seems like it doesn't work in the testing environment?

With it in, the following error occurs:
```
Error: React Router Vite plugin can't detect preamble. Something is wrong.
```

Removed it based on mentions from:
https://akoskm.com/react-router-vitest-example/

https://remix.run/docs/en/main/guides/vite#plugin-usage-with-other-vite-based-tools-eg-vitest-storybook

ryanhopperlowe: Oh interesting. Just curious

ryanhopperlowe (approved):

PR #1533, opened 2025-01-30
Fix: Set Obot dark mode icon to blue

cjellick:
We use invert logic when an agent doesn't have an explicit dark icon
set. Generally, this should be fine, but for obot, we want the dark mode
to be blue, not the orange, which is the inversion of blue.


g-linville (approved):

StrongMonkey (approved):

sangee2004:
This issue is addressed when testing with latest builds.

User ui
Screenshot 2025-01-30 at 11 42 39 AM
Admin ui
Screenshot 2025-01-30 at 11 42 10 AM

PR #1539, opened 2025-01-30
fix: select & delete tools in agent fix

ivyjeong13:
* do not use useEffect form watch to trigger `onChange` along with reset (can potentially cause an infinite loop of onChange firing and saving)
* fix for deleting selected tool (using wrong key tool.tool instead of tool.id)
* `prev` instead of `agent`, looks like change was for making sure aliasAssigned was appropriately updated but doesn't mean agentUpdates should be constantly matched to agent especially while user is mid-modifying

the infinite loop from bullet point 2:

https://github.com/user-attachments/assets/261b0da1-bb56-4f18-8a67-92dc4e38e164

Tested on 3G Network to emulate slow network behavior.
Screenshot 2025-01-30 at 3 50 24 PM

### DIFF
--- a/ui/admin/app/components/agent/Agent.tsx
+++ b/ui/admin/app/components/agent/Agent.tsx
@@ -45,7 +45,7 @@ export function Agent({ className, currentThreadId, onRefresh }: AgentProps) {
 	useEffect(() => {
 		setAgentUpdates((prev) => {
 			return {
-				...agent,
+				...prev,
 				aliasAssigned:
 					agent.aliasAssigned !== undefined
 						? agent.aliasAssigned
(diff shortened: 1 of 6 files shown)


ui/admin/app/components/agent/Agent.tsx
```diff
@@ -62,8 +62,6 @@ export function Agent({ className, currentThreadId, onRefresh }: AgentProps) {
 
 			debouncedUpdateAgent(updatedAgent);
 
-			setAgentUpdates(updatedAgent);
```
ivyjeong13, line 65: setAgentUpdates happens below this so axed this one

ryanhopperlowe (approved):
lgtm!

Issue #880, opened 2024-12-16
Microsoft Entra for authentication


cjellick:
The platform formerly known as AzureAD

sangee2004:
Able to configure Microsoft Entra and work with it successfully in "Single Tenant" mode.

Seeing issue when testing with MultiTenant mode which is tracked in https://github.com/obot-platform/obot/issues/1780

PR #1541, opened 2025-01-30
enhance: add support for entra auth provider

g-linville:
for #880


pkg/proxy/proxy.go
```diff
@@ -199,14 +240,6 @@ func (pm *Manager) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	// If the legacy auth provider cookie exists, delete it.
-	http.SetCookie(w, &http.Cookie{
-		Name:   ObotAuthProviderQueryParam,
-		Value:  "",
-		Path:   "/",
-		MaxAge: -1,
-	})
```
g-linville: It's been long enough, so there should no longer be any of these out there. They won't do or interfere with anything anyway.

pkg/proxy/proxy.go
```diff
@@ -24,6 +24,8 @@ var log = logger.Package()
 
 const (
 	ObotAccessTokenCookie      = "obot_access_token"
+	ObotAccessTokenCookieZero  = "obot_access_token_0"
+	ObotAccessTokenCookieOne   = "obot_access_token_1"
```
thedadams: Can this be made more generic so we can seamlessly handle an arbitrary number of cookies? I would hate for a random edge case to overflow the second cookie, too.

g-linville: This will change based on my other PR to refactor how we handle the cookies.

g-linville:
Moving to draft for now, as this will need to change, based on other stuff that I am going to merge.

pkg/proxy/proxy.go
```diff
@@ -75,6 +77,11 @@ func (pm *Manager) cacheCleanup(ctx context.Context) {
 
 func getTokenHash(req *http.Request) (string, error) {
 	c, err := req.Cookie(ObotAccessTokenCookie)
+	if errors.Is(err, http.ErrNoCookie) {
+		// Check the zero cookie. This one is present when the token is too large to fit in one cookie and
+		// must be split into two.
+		c, err = req.Cookie(ObotAccessTokenCookieZero)
+	}
```
thedadams, line 84: This seems a little strange to me, so I just want to double-check: we want to look for the `_0` cookie only if the non-`_0` cookie doesn't exist? If we have split the cookie into two parts, then I would expect that both cookies should exist.

g-linville: Yes, this is intentional. If the token is in one cookie, it will be called `obot_access_token`. If it is in more than one cookie, it will be in `obot_access_token_0`, `obot_access_token_1`, and so on. So the presence of `_0` indicates a multi-part token. In this case we only bother to hash the first piece of it, since there's really no need to hash the whole thing, as all we need is a unique identifier.

g-linville: ^ Let me know if you think any of that info should go into the code comments.

thedadams (approved):

iwilltry42 (approved):

PR #1549, opened 2025-01-31
chore: update service method primitives

ryanhopperlowe:
Addresses #1581 

Refactor work performed only on agentService.ts (subsequent work can be done incrementally)

Improvements:

- provides a programatic way to create service methods with configurable capabilities
- abstracts a lot of the mess  useSWR
- loosens coupling with useSWR as a library (foreshadowing..?)
- provides builtin support for AbortControllers
- switches SWR keys to be arrays instead of objects. This allows us to meaningfully invalidate the api cache using partial keys

Things to Note:
- Going forward all service methods will only be able to take 1 parameter which must always be an object. It's a bit more restrictive, but it shouldn't prevent us from doing anything and I think it's worth it for the DX

Future work:
- we may want to improve the way keys are passed to `createFetcher`. Right now there's no way to know that a key is unique and won't collide with another service method.

### DIFF
--- a/ui/admin/app/components/agent/AgentContext.tsx
+++ b/ui/admin/app/components/agent/AgentContext.tsx
@@ -6,7 +6,7 @@ import {
 	useEffect,
 	useState,
 } from "react";
-import useSWR, { mutate } from "swr";
+import useSWR from "swr";
 
 import { Agent } from "~/lib/model/agents";
 import { AgentService } from "~/lib/service/api/agentService";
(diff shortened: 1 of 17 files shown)


ivyjeong13 (commented):
I like the changes! 🙌 Especially with the abort controller additions! Only thing is if the signal does get cancelled, it does toast the errorMessage and console.error the canceled request.

Screenshot 2025-01-31 at 10 53 15 AM

ryanhopperlowe:
> I like the changes! 🙌 Especially with the abort controller additions! Only thing is if the signal does get cancelled, it does toast the errorMessage and console.error the canceled request.
> 
> Screenshot 2025-01-31 at 10 53 15 AM

Good catch, will fix!

ivyjeong13 (approved):

PR #1561, opened 2025-01-31
fix: store auth provider info in cookie name

g-linville:
This PR refactors the way we store auth provider information with our access token cookies. Before, we were intercepting the cookies and inserting/removing auth provider information inside of them. This was a bit error-prone and unreliable. A much more simple solution is to just expect the provider info to be present in the name of the cookie.

In the case that there are multiple access token cookies present, we take the one that expires the farthest from now, under the assumption that that is what the user most recently logged in with. This shouldn't happen under ordinary circumstances, but could happen during development.

This PR also fixes a bug around token refresh that I accidentally introduced, where a whole ton of cookies would get created due to a malformed `Set-Cookie` header.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,6 @@ module github.com/obot-platform/obot
 go 1.23.4
 
 replace (
-	github.com/oauth2-proxy/oauth2-proxy/v7 => github.com/obot-platform/oauth2-proxy/v7 v7.0.0-20241008204315-265dabe17f43
 	github.com/obot-platform/obot/apiclient => ./apiclient
 	github.com/obot-platform/obot/logger => ./logger
 )
--- a/pkg/api/server/server.go
+++ b/pkg/api/server/server.go
@@ -96,11 +96,3 @@ func (s *Server) wrap(f api.HandlerFunc) http.HandlerFunc {
 		}
 	}
 }
-
-func firstValue(m map[string][]string, key string) string {
-	values := m[key]
-	if len(values) == 0 {
-		return ""
-	}
-	return values[0]
-}
--- a/pkg/api/server/server.go
+++ b/pkg/api/server/server.go
@@ -62,8 +62,10 @@ func (s *Server) wrap(f api.HandlerFunc) http.HandlerFunc {
 			return
 		}
 
-		if setCookie := firstValue(user.GetExtra(), "set-cookie"); setCookie != "" {
-			rw.Header().Set("Set-Cookie", setCookie)
+		if user.GetExtra()["set-cookies"] != nil {
+			for _, setCookie := range user.GetExtra()["set-cookies"] {
+				rw.Header().Add("Set-Cookie", setCookie)
+			}
 		}
 
 		if !s.authorizer.Authorize(req, user) {
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -133,8 +133,7 @@ func (pm *Manager) ServeHTTP(user user.Info, w http.ResponseWriter, r *http.Requ
 		fromCookie bool
 		err        error
 	)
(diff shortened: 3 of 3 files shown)


go.mod
```diff
@@ -3,7 +3,6 @@ module github.com/obot-platform/obot
 go 1.23.4
 
 replace (
-	github.com/oauth2-proxy/oauth2-proxy/v7 => github.com/obot-platform/oauth2-proxy/v7 v7.0.0-20241008204315-265dabe17f43
```
g-linville, line 6: We removed this dependency but never removed this `replace` line.

pkg/api/server/server.go
```diff
@@ -62,8 +62,10 @@ func (s *Server) wrap(f api.HandlerFunc) http.HandlerFunc {
 			return
 		}
 
-		if setCookie := firstValue(user.GetExtra(), "set-cookie"); setCookie != "" {
-			rw.Header().Set("Set-Cookie", setCookie)
+		if user.GetExtra()["set-cookies"] != nil {
+			for _, setCookie := range user.GetExtra()["set-cookies"] {
+				rw.Header().Add("Set-Cookie", setCookie)
+			}
```
g-linville, line 68: The auth providers may return more than one cookie that needs to be set.

pkg/proxy/proxy.go
```diff
@@ -100,28 +131,34 @@ func (pm *Manager) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-	if param := r.URL.Query().Get(ObotAuthProviderQueryParam); param != "" {
-		// If the provider is set in the query params, use that.
-		provider = param
-	} else if cookie, err := r.Cookie(ObotAccessTokenCookie); err == nil {
-		// Extract the provider from the cookie, if it's there.
-		cookieValue, err := url.QueryUnescape(cookie.Value)
-		if err != nil {
-			http.Error(w, fmt.Sprintf("failed to unescape cookie value: %v", err), http.StatusBadRequest)
+	if user.GetExtra()["auth_provider_name"] != nil && len(user.GetExtra()["auth_provider_name"]) > 0 &&
+		user.GetExtra()["auth_provider_namespace"] != nil && len(user.GetExtra()["auth_provider_namespace"]) > 0 {
```
thedadams:
```suggestion
	if len(user.GetExtra()["auth_provider_name"]) > 0 && len(user.GetExtra()["auth_provider_namespace"]) > 0 {
```

pkg/proxy/proxy.go
```diff
@@ -358,9 +370,9 @@ func (p *Proxy) authenticateRequest(req *http.Request) (*authenticator.Response,
 		},
 	}
 
-	if ss.SetCookie != "" {
+	if ss.SetCookies != nil {
```
thedadams: nit:
```suggestion
	if len(ss.SetCookies) != 0 {
```

thedadams (approved):

pkg/proxy/proxy.go
```diff
@@ -42,49 +39,83 @@ func NewProxyManager(dispatcher *dispatcher.Dispatcher) *Manager {
...
+	}
+
+	_, provider, exists := strings.Cut(cookieName, ObotAccessTokenCookiePrefix)
+	if !exists {
+		// This should be impossible, but we'll account for it anyway.
+		return "", fmt.Errorf("failed to find provider in cookie name %s", cookieName)
+	}
+
+	// The provider namespace and name should be sparated by a double underscore.
+	namespace, name, exists := strings.Cut(provider, "__")
```
StrongMonkey, line 78: Double underscore? 😄

StrongMonkey (approved):

pkg/proxy/proxy.go
g-linville: My options were pretty limited, lol

PR #1568, opened 2025-02-01
feat: implement pagination for thread files

ryanhopperlowe:
Addresses #1580 
- creates some rough abstractions for client side pagination (used to mock server side pagination until we suport it in backend)
- hooks up pagination to thread files view
- updates skeletons to take children

https://www.loom.com/share/781c47583175404cbef1f5a15adafacd?sid=e0d3edb6-a5d9-4567-97af-f7ff228a70e5


ivyjeong13 (approved):
:+1: Will be nice to get the pagination from the backend with this!

PR #1623, opened 2025-02-03
enhance: use just one auth cookie for Obot

g-linville:
for https://github.com/obot-platform/obot/issues/1630

This is (another) refactoring of how we handle auth cookies. Now, instead of doing one cookie per auth provider, we are just doing one cookie period. We figure out which auth provider is associated with it by storing its hash in a cache, but that only lasts for 24h before it gets removed. If it's not in the cache, we try it with each configured auth provider until we find one that recognizes it, and then we cache that auth provider in association with that token hash.

### DIFF
--- a/pkg/gateway/server/dispatcher/dispatcher.go
+++ b/pkg/gateway/server/dispatcher/dispatcher.go
@@ -26,26 +26,34 @@ import (
 )
 
 type Dispatcher struct {
-	invoker    *invoke.Invoker
-	gptscript  *gptscript.GPTScript
-	client     kclient.Client
-	modelLock  *sync.RWMutex
-	modelUrls  map[string]*url.URL
-	authLock   *sync.RWMutex
-	authUrls   map[string]*url.URL
-	openAICred string
+	invoker                     *invoke.Invoker
+	gptscript                   *gptscript.GPTScript
+	client                      kclient.Client
+	modelLock                   *sync.RWMutex
+	modelUrls                   map[string]*url.URL
+	authLock                    *sync.RWMutex
+	authUrls                    map[string]*url.URL
+	configuredAuthProvidersLock *sync.RWMutex
+	configuredAuthProviders     []string
+	openAICred                  string
 }
 
 func New(invoker *invoke.Invoker, c kclient.Client, gClient *gptscript.GPTScript) *Dispatcher {
-	return &Dispatcher{
-		invoker:   invoker,
-		gptscript: gClient,
-		client:    c,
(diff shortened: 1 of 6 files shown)


njhale (approved):
I think we'll need to move away from in-memory mutexes for managing critical regions when/if we want to start horizontally scaling the server, but IMO we shouldn't be concerned with that for now since there's a lot of them scattered throughout the codebase.

g-linville:
@njhale agreed. We will probably wanna support Redis/Valkey or something similar at some point.

pkg/gateway/server/dispatcher/dispatcher.go
```diff
@@ -358,25 +366,41 @@ func (d *Dispatcher) startAuthProvider(ctx context.Context, namespace, authProvi
...
+}
+
+func (d *Dispatcher) UpdateConfiguredAuthProviders() {
+	d.configuredAuthProvidersLock.Lock()
+	defer d.configuredAuthProvidersLock.Unlock()
+
 	var authProviders v1.ToolReferenceList
-	if err := d.client.List(ctx, &authProviders, &kclient.ListOptions{
-		Namespace: namespace,
+	if err := d.client.List(context.Background(), &authProviders, &kclient.ListOptions{
```
thedadams: I think this function should still take a `context.Context`. We could always pass `context.Background()` to it when we need to.

pkg/gateway/server/oauth.go
```diff
@@ -30,10 +30,7 @@ func (s *Server) oauth(apiContext api.Context) error {
 	}
 
 	// Check to make sure this auth provider exists.
-	list, err := s.dispatcher.ListConfiguredAuthProviders(apiContext.Context(), namespace)
-	if err != nil {
-		return fmt.Errorf("could not list configured auth providers: %w", err)
-	} else if !slices.Contains(list, name) {
+	if providerList := s.dispatcher.ListConfiguredAuthProviders(namespace); !slices.Contains(providerList, name) {
```
thedadams, line 33: For example, it would be find to continue to pass `apiContext.Context()` here.

g-linville: This is a different function. The `List` function just returns the stored list. The `Update` function is what refreshes the list.

thedadams: Sorry for my confusion here.

g-linville: No worries. I do think it is better to pass the context in like you suggested, so I made that change

Issue #1527, opened 2025-01-30
Registering custom tool with multiple tools , results in no tool being made available


sangee2004:
**Steps to reproduce the problem:**
1. Register a custom tool with multiple tools - github.com/sangee2004/tools/testtool3
2. Once this tool is registered , notice that no bundles is shown in `tools` view

<img width="456" alt="Image" src="https://github.com/user-attachments/assets/12e1f896-7a5f-4f98-ad3d-6f2dc016422c" />

Also notice that no tools are listed in tools drop down option in Agent create/edit page

<img width="670" alt="Image" src="https://github.com/user-attachments/assets/6f4fea14-f8ce-4696-9d47-ad9b9dad297c" />

3. Add this tool bundle 
4. Chat with agent 

Following error is presented

<img width="497" alt="Image" src="https://github.com/user-attachments/assets/a1d937f8-1126-4eee-b2f8-6d4d616c59e2" />

Debug logs - [calls - 2025-01-29T173753.329.json](https://github.com/user-attachments/files/18596242/calls.-.2025-01-29T173753.329.json)

cjellick:
I dont believe the python error is related to the previous steps. If you can consistently reproduce the python error, dig deeper to indentify the exact steps to reproduce. I can create a custom tool with multiple tools (https://github.com/cjellick/multi-tool-test) and do not get that error.

StrongMonkey:
The expected behavior is that when you register bundle tools, you should be seeing child tools in UI. The tool call error is likely unrelated to the original issue.

sangee2004:
When testing with latest build, able to see all the child tools being listed in the UI.
<img width="577" alt="Image" src="https://github.com/user-attachments/assets/653e2f7f-31b3-4a6b-a3f4-9feb6316eabc" />

Issue #1526, opened 2025-01-30
Registering a custom tool with category results in tool being shown along with built in tools in tool drop down list.


sangee2004:
**Steps to reproduce the problem:**

1. Register a custom tool with category set (MyBluesky in my case)- github.com/sangee2004/tools/testtool3

Notice that in `Tools` page this is shown in `Custom Tools` section as expected

<img width="1112" alt="Image" src="https://github.com/user-attachments/assets/5aa44191-5ad0-4111-a4b5-be2b36ad90be" />

In agent tool drop down in Agent create/edit page , it is listed along with built in tools 

<img width="670" alt="Image" src="https://github.com/user-attachments/assets/bf1c3389-d739-4651-a6cc-c1d7f0febbe9" />

**Expected Behavior:**
We should list the custom tool in its own section at the Built-in list.

cjellick:
These steps to reproduce are not good

cjellick:
@StrongMonkey -  ill assign to you as part of the tool import refactor.

sangee2004:
When registering the tool using - github.com/sangee2004/tools/testtool3 , It gets listed under its own section as expected

<img width="690" alt="Image" src="https://github.com/user-attachments/assets/2f095e6e-2f39-469c-8a08-ed1df2e9de7c" />

@StrongMonkey Do we expect the section title to be "Custom" in this case?

Issue #1525, opened 2025-01-30
Registering a custom tool with a category name that is associated with a existing built-in tool  results in this built-in tool not listed in `tools` page/ in agent tools drop down list anymore.


sangee2004:
**Steps to reproduce the problem:**
1. Register a custom tool with category set to a category associated with some built-in tool (Bluesky in my case)- github.com/sangee2004/tools/testtool1

Category set as follows in tool.gpt.
```
!metadata:*:category
Bluesky
```
Once the tool is registered , notice that the built-in  `Bluesky` tool is missing from `tools` page/ in agent tools drop down list .
<img width="1151" alt="Image" src="https://github.com/user-attachments/assets/77ec4405-c844-4c6a-b562-29be41d88c82" />

<img width="580" alt="Image" src="https://github.com/user-attachments/assets/6c30b845-5133-4f22-aa1b-2a281928b4d7" />

**Expected Behavior:**

We should be able to show custom tool - "MyBlueSky" and built in tool -"BlueSky" in this case.

cjellick:
how do this and https://github.com/obot-platform/obot/issues/1526 relate to each other?

cjellick:
This title is impossible to understand

cjellick:
@StrongMonkey i think/hope when you finish refactoring category stuff, this will be addressed.

Issue #1026, opened 2024-12-23
Bug: Only the first tool registers in tool.gpt


cjellick:
**Description:**
When attempting to register multiple tools, only the first tool is successfully registered in tool.gpt. All tools should be registered.

~**Steps to Reproduce:**~
~1. Attempt to register multiple tools in tool.gpt.~
~2. Observe that only the first tool is registered.~

~**Expected Behavior:**~
~All tools should be registered, not just the first one.~

This is ballooning into a bigger refactor. see belowf

cjellick:
Requirements:

We want to simplify our indexing logic when building our built-in catalogs and use the same logic for importing tools via the UI and API. The main motivation here is to make it so that an admin user can import a tool.gpt hosted on github and get the same end-result that we do when we add a new tool to [github.com/obot-platform/tools](http://github.com/obot-platform/tools).

Secondary motivations are:
* Make how we index tools more consistent and less confusing when building our images
* Get rid of the need to use both categories and bundles in the UI to organize tools. The two get conflated and are dependent on each other. People writing their own tools will have a hard time getting it right. Note: I still want a category system, but it will be optional and for broader groups like “Social Media Tools” rather a category for just LinkedIn. Right now, our tool count is still low enough where it doesn’t really matter, but eventually the organization will help.

Regarding indexing our built-in tool catalogs, (indexing here specifically means turning a tool found in a .gpt file into an entry in our tool-references api):
* We will no longer support `all: true` in index.yaml. Instead, when we read the tool.gpt file, we will index based on the following rules:
    1. The first tool will be indexed. If that tools is a bundle, any tools referenced in its “Share Tools” directive should also be indexed. If you want to exclude a shared tool from being indexed individually, that tool must opt out with `index: false`.
    2. Indexing a tool.gpt can only support a single bundle and it must be the first tool
    3. We won’t “walk” shared-tool trees while indexing. So if you have a bundle that shares tool X that shares tool Y, only tool X will be indexed

To accommodate changes to how we want to display tools in the UI, we’re adding some new API fields:
* A first class field on tool-reference to indicate whether a tool is a bundle. This should still be metadata inthe tool.gpt file, but get translated to the first-class tool-reference field upon indexing.
* A first class field on tool-reference to indicate what bundle (if any) a tool is part of. The value should be a reference to another tool-reference entry.

Regarding registering a tool from our UI/API:
* When importing in this way, we’ll follow the exact same logic as we do when indexing our built-in tools and respect the same metadata

[...]

PR #1633, opened 2025-02-04
Chore: Refactor tool indexing

StrongMonkey:
This PR implements the tool index refactoring logic defined at https://github.com/obot-platform/obot/issues/1026. It has both frontend and backend changes to support new indexing logic.

Backend:
- Dropping `all: true` and always import the first tool and its shared tools from gptfile. 
- Dropping bundle metadata and always relied on share tools
- Add api to support bundle tool and bundle child tools
- Apply same code to import both custom and built-in tools.
- Add tool commit

Frontend:
- Stop using category as a way to group bundle tool
- Group tool and bundle using new api fields.
- Add tool commit

Should fix
- https://github.com/obot-platform/obot/issues/1525
- https://github.com/obot-platform/obot/issues/1526
- https://github.com/obot-platform/obot/issues/1527

### DIFF
--- a/pkg/api/handlers/toolreferences.go
+++ b/pkg/api/handlers/toolreferences.go
@@ -112,6 +112,8 @@ func (a *ToolReferenceHandler) Create(req api.Context) (err error) {
 		return apierrors.NewBadRequest("reference is required")
 	}
 
+	newToolReference.Reference = strings.TrimPrefix(strings.TrimPrefix(newToolReference.Reference, "https://"), "http://")
+
 	if newToolReference.Name == "" {
 		newToolReference.Name, err = a.pickNameForReference(req, newToolReference.Reference)
 		if err != nil {
@@ -138,11 +140,12 @@ func (a *ToolReferenceHandler) Create(req api.Context) (err error) {
 
 	for _, toolRef := range toolRefs {
 		if err := req.Create(toolRef); err != nil && !apierrors.IsAlreadyExists(err) {
-			return apierrors.NewInternalError(fmt.Errorf("failed to create tool reference %s: %w", toolRef.Name, err))
+			return apierrors.NewInternalError(fmt.Errorf("failed to create tool reference %s: %w", toolRef.GetName(), err))
 		}
 	}
 
-	return req.Write(convertToolReference(*toolRefs[0]))
+	toolRef := toolRefs[0].(*v1.ToolReference)
+	return req.Write(convertToolReference(*toolRef))
 }
 
 func (a *ToolReferenceHandler) Delete(req api.Context) error {
--- a/pkg/api/handlers/toolreferences.go
+++ b/pkg/api/handlers/toolreferences.go
@@ -12,6 +12,7 @@ import (
 	"github.com/obot-platform/obot/apiclient/types"
 	"github.com/obot-platform/obot/pkg/api"
 	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
+	"github.com/obot-platform/obot/pkg/tools"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -33,10 +34,13 @@ func convertToolReference(toolRef v1.ToolReference) types.ToolReference {
 			Name:      toolRef.Name,
 			ToolType:  toolRef.Spec.Type,
 			Reference: toolRef.Spec.Reference,
+			Commit:    toolRef.Status.Commit,
 		},
-		Builtin:  toolRef.Spec.Builtin,
-		Error:    toolRef.Status.Error,
-		Resolved: toolRef.Generation == toolRef.Status.ObservedGeneration,
+		Builtin:        toolRef.Spec.Builtin,
+		Bundle:         toolRef.Spec.Bundle,
+		BundleToolName: toolRef.Spec.BundleToolName,
+		Error:          toolRef.Status.Error,
+		Resolved:       toolRef.Generation == toolRef.Status.ObservedGeneration,
 	}
 	if toolRef.Spec.Active == nil {
 		tf.Active = true
@@ -123,22 +127,22 @@ func (a *ToolReferenceHandler) Create(req api.Context) (err error) {
 		return apierrors.NewBadRequest(fmt.Sprintf("invalid tool type %s", newToolReference.ToolType))
 	}
 
-	toolRef := &v1.ToolReference{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      newToolReference.Name,
-			Namespace: req.Namespace(),
-		},
-		Spec: v1.ToolReferenceSpec{
-			Type:      newToolReference.ToolType,
-			Reference: newToolReference.Reference,
-		},
+	toolRefs, err := tools.ResolveToolReferences(req.Context(), a.gptscript, newToolReference.Name, newToolReference.Reference, false, newToolReference.ToolType)
+	if err != nil {
+		return apierrors.NewBadRequest(fmt.Sprintf("failed to resolve tool references for %s: %v", newToolReference.Reference, err))
 	}
 
-	if err = req.Create(toolRef); err != nil {
-		return err
+	if len(toolRefs) == 0 {
+		return apierrors.NewBadRequest(fmt.Sprintf("no tool references found for %s", newToolReference.Reference))
 	}
 
-	return req.Write(convertToolReference(*toolRef))
+	for _, toolRef := range toolRefs {
+		if err := req.Create(toolRef); err != nil && !apierrors.IsAlreadyExists(err) {
+			return apierrors.NewInternalError(fmt.Errorf("failed to create tool reference %s: %w", toolRef.Name, err))
+		}
+	}
+
+	return req.Write(convertToolReference(*toolRefs[0]))
 }
 
 func (a *ToolReferenceHandler) Delete(req api.Context) error {
@@ -164,6 +168,10 @@ func (a *ToolReferenceHandler) Delete(req api.Context) error {
 
 	if toolRef.Spec.Builtin {
 		return types.NewErrBadRequest("cannot delete builtin tool reference %s", id)
+	}
+
+	if !toolRef.Spec.Bundle {
+		return types.NewErrBadRequest("cannot delete child tool that belongs to a bundle tool")
 	}
 
 	return req.Delete(&v1.ToolReference{
--- a/pkg/api/handlers/toolreferences.go
+++ b/pkg/api/handlers/toolreferences.go
@@ -112,8 +112,6 @@ func (a *ToolReferenceHandler) Create(req api.Context) (err error) {
 		return apierrors.NewBadRequest("reference is required")
 	}
 
-	newToolReference.Reference = strings.TrimPrefix(strings.TrimPrefix(newToolReference.Reference, "https://"), "http://")
-
 	if newToolReference.Name == "" {
 		newToolReference.Name, err = a.pickNameForReference(req, newToolReference.Reference)
 		if err != nil {
--- a/pkg/api/handlers/toolreferences.go
+++ b/pkg/api/handlers/toolreferences.go
@@ -129,7 +129,7 @@ func (a *ToolReferenceHandler) Create(req api.Context) (err error) {
 		return apierrors.NewBadRequest(fmt.Sprintf("invalid tool type %s", newToolReference.ToolType))
 	}
 
-	toolRefs, err := tools.ResolveToolReferences(req.Context(), a.gptscript, newToolReference.Name, newToolReference.Reference, false, newToolReference.ToolType)
+	toolRefs, err := tools.ResolveToolReferences(req.Context(), a.gptscript, newToolReference.Name, newToolReference.Reference, "", false, newToolReference.ToolType)
 	if err != nil {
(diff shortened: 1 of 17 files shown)


pkg/api/handlers/toolreferences.go
```diff
@@ -166,6 +170,10 @@ func (a *ToolReferenceHandler) Delete(req api.Context) error {
 		return types.NewErrBadRequest("cannot delete builtin tool reference %s", id)
 	}
 
+	if !toolRef.Spec.Bundle {
```
thedadams: I'm not seeing anywhere where this is getting set to `true`.

pkg/controller/handlers/toolreference/toolreference.go
```diff
@@ -68,93 +71,27 @@ func New(gptClient *gptscript.GPTScript,
...
-					Reference: entry.Reference,
-					Builtin:   true,
-				},
-			})
+		if entry.NameOverride != "" {
+			toolRefs[0].Name = entry.NameOverride
+		}
+
+		for _, toolRef := range toolRefs {
+			result = append(result, toolRef)
```
thedadams, line 87:
```suggestion
		result = append(result, toolRefs...)
```

pkg/api/handlers/toolreferences.go
StrongMonkey: missing some files when pushing... it is set here https://github.com/obot-platform/obot/pull/1633/files#diff-fc4ca6da48915751ba6a956a3d3912415270a3c9374355c8ea3d6f788b6e222dR47

ui/admin/app/lib/model/toolReferences.ts
```diff
@@ -55,36 +55,44 @@ export function isCapabilityTool(toolReference: ToolReference) {
...
+			// Handle tool that belongs to a bundle
+			const bundleTool = toolMap.get(toolReference.bundleToolName);
+			if (bundleTool && !isCapabilityTool(bundleTool)) {
+				if (!result[toolReference.bundleToolName]) {
+					result[toolReference.bundleToolName] = {
+						...bundleTool,
+						tools: [toolReference],
+					};
+				} else {
+					result[toolReference.bundleToolName].tools!.push(toolReference);
```
ryanhopperlowe: nit:
```suggestion
                    if (!result[toolReference.bundleToolName].tools) throw new Error("This should never happen")
					result[toolReference.bundleToolName].tools.push(toolReference);
```

ui/admin/app/components/tools/ToolCatalog.tsx
```diff
@@ -138,23 +132,30 @@ export function ToolCatalog({
...
+								? UncategorizedToolCategory
+								: CustomToolsToolCategory;
+
+						if (!acc[category]) {
+							acc[category] = [];
+						}
+						acc[category].push(tool);
+						return acc;
+					}, {})
+				).map(([category, tools]) => (
```
ryanhopperlowe, line 155: nit: can you extract this to a constant above?

ui/admin/app/components/tools/ToolCatalog.tsx
```diff
@@ -138,23 +132,30 @@ export function ToolCatalog({
...
-								: true
-						}
-						tools={categoryTools}
+						tools={tools}
 						selectedTools={selectedTools}
 						onAddTool={handleAddTool}
 						onRemoveTool={handleRemoveTool}
-						expandFor={search}
 						oauths={oauths}
+						configuredOauthApps={configuredOauthApps ?? new Set()}
```
ryanhopperlowe: nit: this is unnecessary

```suggestion
						configuredOauthApps={configuredOauthApps}
```

ui/admin/app/components/tools/toolGrid/ToolCard.tsx
```diff
@@ -82,6 +89,48 @@ export function ToolCard({
...
+													const [org, repo, ...rest] = parts.slice(1);
+													const path = rest.join("/");
+													const pathWithGpt = path.endsWith(".gpt")
+														? path
+														: `${path}/tool.gpt`;
+
+													return `github.com/${org}/${repo}/blob/${tool.commit}/${pathWithGpt}`;
+												}
+												return tool.reference;
+											})()}`}
```
ryanhopperlowe: can you extract this to a function outside of the render clause?

ryanhopperlowe (approved):
lgtm - just a few nits

StrongMonkey:
>Let me know if I'm way off here, but:

Does this need to have a dialog? Couldn't it just have a tooltip that says "View Commit"? I'd imagine that most users don't actually need to view the commit hash of the tool, right?

Yeah open to UX changes. I think I like your suggestion, original I did this so that sangeetha can see the actual commit. But probably View Commit is straightforward. I will change to that.

ivyjeong13 (approved):
:+1: Looks good to me!

My only nit might be the category labels are a little repetitive in the tool selection if the bundle and the category name are the same 

Screenshot 2025-02-06 at 12 41 12 PM

StrongMonkey:
@ivyjeong13 yes, we are going to remove all the existing catalog for these tools now. They will have a bigger scope instead of just fine-grain catalogs for each own tools. It is done in a different pr https://github.com/obot-platform/tools/pull/403

pkg/tools/resolve.go
```diff
@@ -0,0 +1,112 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	tool := prg.ToolSet[prg.EntryToolID]
+	isCapability := tool.MetaData["category"] == "Capability"
+	isBundleTool := false
+	if len(tool.LocalTools) > 1 {
+		isBundleTool = true
+	}
```
thedadams: nit:
```suggestion
	isBundleTool := len(tool.LocalTools) > 1
```

pkg/controller/handlers/toolreference/toolreference.go
```diff
@@ -68,94 +71,26 @@ func New(gptClient *gptscript.GPTScript,
...
-					Finalizers:  []string{v1.ToolReferenceFinalizer},
-					Annotations: annotations,
-				},
-				Spec: v1.ToolReferenceSpec{
-					Type:      toolType,
-					Reference: entry.Reference,
-					Builtin:   true,
-				},
-			})
+		if entry.NameOverride != "" {
```
thedadams: If I understand this correctly, I can override the top-level name, but not the name of the other tools. So, if I override the name of the time bundle to `the-time`, then all the "sub" tools would have names like `time-sub-tool`, but the top-level tool would be called `the-time`?

Is this right? Does it make sense to override the names of the sub tools too?

StrongMonkey: Hmm, good point. i didn't think of that. Will change it to match the same logic

thedadams (approved):

ui/admin/app/components/tools/toolGrid/ToolCard.tsx
```diff
@@ -91,44 +108,28 @@ export function ToolCard({
...
-											{tool.commit}
-										</a>
-									) : (
-										tool.commit
-									)}
-								</p>
-							</div>
-						</PopoverContent>
-					</Popover>
+					<TooltipProvider>
```
ryanhopperlowe: can you remove this? It prevents the tooltips from using the global configuration

ryanhopperlowe (changes requested):

ui/admin/app/components/tools/toolGrid/ToolCard.tsx
StrongMonkey: yes

StrongMonkey: removed

ryanhopperlowe (approved):

pkg/controller/handlers/toolreference/migrate.go
```diff
@@ -21,6 +21,11 @@ func (h *Handler) MigrateToolNames(req router.Request, _ router.Response) error
 		tools = o.Spec.Manifest.Tools
 	case *v1.Workflow:
 		tools = o.Spec.Manifest.Tools
+	case *v1.Thread:
```
StrongMonkey: Processing threads will likely have an impact on controller, if so we can store the hash of the migration map into each object and compare to only run migration once map is changed. Although I feel like the speed impact shouldn't be that bad.

pkg/api/handlers/toolreferences.go
```diff
@@ -108,6 +112,8 @@ func (a *ToolReferenceHandler) Create(req api.Context) (err error) {
 		return apierrors.NewBadRequest("reference is required")
 	}
 
+	newToolReference.Reference = strings.TrimPrefix(strings.TrimPrefix(newToolReference.Reference, "https://"), "http://")
```
thedadams: @StrongMonkey Sorry, I just noticed that we probably don't want to do this. It is valid for tool references to start with `http(s)://`. For example, non-GitHub-based tools.

StrongMonkey: yeah, that's right. Let me change this

StrongMonkey: Should be fixed.

pkg/controller/handlers/toolreference/migrate.go
```diff
@@ -0,0 +1,57 @@
...
+		tools = o.Spec.Manifest.Tools
+	case *v1.Workflow:
+		tools = o.Spec.Manifest.Tools
+	case *v1.Thread:
+		tools = o.Spec.Manifest.Tools
+	case *v1.WorkflowStep:
+		tools = o.Spec.Step.Tools
+
+	default:
+		return nil
```
thedadams: Can we add a warning log message here to indicate that the type isn't being migrated?

pkg/controller/handlers/toolreference/migrate.go
```diff
@@ -0,0 +1,57 @@
...
+	switch o := req.Object.(type) {
+	case *v1.Agent:
+		o.Spec.Manifest.Tools = tools
+	case *v1.Workflow:
+		o.Spec.Manifest.Tools = tools
+	case *v1.Thread:
+		o.Spec.Manifest.Tools = tools
+	case *v1.WorkflowStep:
+		o.Spec.Step.Tools = tools
+	}
```
thedadams: I don't think this part is necessary: we aren't changing the slice, just one of its elements.

pkg/tools/resolve.go
```diff
@@ -0,0 +1,110 @@
...
+
+	"github.com/gptscript-ai/go-gptscript"
+	"github.com/obot-platform/obot/apiclient/types"
+	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
+	"github.com/obot-platform/obot/pkg/system"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func ResolveToolReferences(ctx context.Context, gptClient *gptscript.GPTScript, name, reference string, builtin bool, toolType types.ToolReferenceType) ([]client.Object, error) {
```
thedadams: I'd rather this return the concrete type. Converting to `[]client.Object` is a pain, but is less likely to happen.
```suggestion
func ResolveToolReferences(ctx context.Context, gptClient *gptscript.GPTScript, name, reference string, builtin bool, toolType types.ToolReferenceType) ([]*v1.ToolReference, error) {
```

pkg/controller/handlers/toolreference/migrate.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"github.com/obot-platform/nah/pkg/router"
+	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
+)
+
+var toolMigrations = map[string]string{
+	"file-summarizer-file-summarizer": "file-summarizer",
+}
+
+func (h *Handler) MigrateToolNames(req router.Request, _ router.Response) error {
```
thedadams: Instead of this being a controller, how do you feel about it being in the controller `PreStart` [here](https://github.com/obot-platform/obot/blob/main/pkg/controller/controller.go#L39)? That way, it only runs on startup.

pkg/controller/handlers/toolreference/migrate.go
StrongMonkey: Technicailly this handler doesn't apply to other objects... so this actually will never happen

pkg/controller/handlers/toolreference/migrate.go
StrongMonkey: I feel like it is straightforward to do it in a handler. Prestart feel like it is mostly initializing data, also not sure if it is utilizing cache so it might slow things down.

pkg/controller/handlers/toolreference/migrate.go
StrongMonkey: oh yes, you are right

pkg/controller/handlers/toolreference/migrate.go
thedadams: Doing it as a pre-start won't slow things down too much because the new container isn't healthy until the controllers start up. If that is a concern, then you could always add it to the post-start.

I don't really like doing this as a handler because the migration really only needs to run once and not every time there are changes. Additionally, there is long-term work that leads to not handling objects on startup. This won't work anymore when that change goes in.

StrongMonkey: ok, I can try to move it.

StrongMonkey: Move to prestart 👍

thedadams (approved):

PR #1635, opened 2025-02-04
fix: threads page, users initially showing empty

ivyjeong13:
What causes a rerender of `DataTable` is when `data` changes. Because `userMap` is not tied to that data, a rerender does not fire when getUsers.data is updated. Two solutions: Either preload to make sure the data is available before rendering or update `data` so it accounts for that information and does a rerender.


ui/admin/app/routes/_auth.threads._index.tsx
```diff
@@ -45,6 +45,7 @@ export async function clientLoader({
 		preload(...AgentService.getAgents.swr({})),
 		preload(WorkflowService.getWorkflows.key(), WorkflowService.getWorkflows),
 		preload(ThreadsService.getThreads.key(), ThreadsService.getThreads),
+		preload(UserService.getUsers.key(), UserService.getUsers),
```
ivyjeong13, line 48: Just throwing at top to not forget!

TODO: test case

ryanhopperlowe (approved):

PR #1652, opened 2025-02-05
fix: build: upgrade all packages to fix curl incompatibility

g-linville:
`curl` was not working in the container, causing the image build to fail. We need to upgrade all the packages in the image in order to fix this.

More information: https://github.com/orgs/wolfi-dev/discussions/41391


njhale (approved):

StrongMonkey (approved):
dId they already fix it upstream? My pr seems to be passing the check this morning

g-linville:
@StrongMonkey you're right. Looks like a new version of the image has been published, and this is fixed now. I'm gonna close this.

Issue #1463, opened 2025-01-27
support viewing database created in the user ui in the admin ui


cjellick:
When a user creates a db table and puts data in it in the user UI, we want to be able to see those on the threads page in the admin ui

sangee2004:
We are now able to see the Database created for a thread in thread inspect view.
We are also provided with ability  to query and modify DB entries.

PR #1657, opened 2025-02-05
feat: allow users to update thread tables via llm

ryanhopperlowe:
addresses #1463 

https://www.loom.com/share/66ddd3c21d714efbb2030bd9d786cb4f?sid=dc812ee0-202f-446b-bb90-20f7be08fe63


ivyjeong13:
Silly question, but with the change to make admin-ui more for administrative purposes, would this make more sense happening on the user-ui side?

ryanhopperlowe:
> Silly question, but with the change to make admin-ui more for administrative purposes, would this make more sense happening on the user-ui side? edit: maybe an overall question for threads, should modifying/interacting with a thread happen on the user-ui side and admin-ui be read-only?

That's definitely a good question. I do think it's important for an admin to be able to interract with most aspects of the users' threads

Issue #1480, opened 2025-01-28
Scheduled triggers are  getting invoked using UTC timezone.


sangee2004:
I am testing this on `test` env which is a hosted env.

**Steps to reproduce the problem:**
1. Create a workflow with schedule trigger for `Daily at Midnight`

<img width="612" alt="Image" src="https://github.com/user-attachments/assets/82083b16-d32f-4f48-9c9d-f6b04964a4de" />

This workflow got triggered at 4 pm PST (which is midnight for UTC)

**Expected Behavior:**

Ideally user should be able to pick a timezone in which they would want the trigger time to be associated with.

StrongMonkey:
@cjellick I think for now, we can add API support for cron trigger to specify timezone and the client will be responsible to supply it(default to user's own timezone but can change to a different one just like the O365 UX).

In the future we can have a user setting for that and client can use it by default.

sangee2004:
@StrongMonkey I am not able to see the scheduled tasks get executed in user's timezone.

I created a task to be run at 3:00 PM and i dont see it getting executed at 3:00 PM PST (which is my timezone).

Its 3:35 pm (PST) and task is still not executed.

<img width="1323" alt="Image" src="https://github.com/user-attachments/assets/6a4db0cc-357f-4acd-a9da-92c7bf69a77e" />

StrongMonkey:
yeah, I should test it. It doesn't work in prod due to a bug.

StrongMonkey:
@sangee2004 fix is in, you can retest this now.

sangee2004:
This issue is addressed when testing with latest builds.

Scheduled tasks are getting invoked as expected in user's timezone.

PR #1664, opened 2025-02-06
Chore: support setting timezone for cronjob

StrongMonkey:
https://github.com/obot-platform/obot/issues/1480

Adding support to run cronjob in a specific timezone. It will default to server's own timezone if it is not set in API. The browser will now set the user's own timezone as default but user will be able to change it.

https://www.loom.com/share/052bd151e46b4c1cb37e28b281687a50

### DIFF
--- a/apiclient/types/cronjob.go
+++ b/apiclient/types/cronjob.go
@@ -3,9 +3,10 @@ package types
 type CronJob struct {
 	Metadata
 	CronJobManifest
-	LastRunStartedAt           *Time `json:"lastRunStartedAt,omitempty"`
-	LastSuccessfulRunCompleted *Time `json:"lastSuccessfulRunCompleted,omitempty"`
-	NextRunAt                  *Time `json:"nextRunAt,omitempty"`
+	LastRunStartedAt           *Time  `json:"lastRunStartedAt,omitempty"`
+	LastSuccessfulRunCompleted *Time  `json:"lastSuccessfulRunCompleted,omitempty"`
+	NextRunAt                  *Time  `json:"nextRunAt,omitempty"`
+	Timezone                   string `json:"timezone,omitempty"`
 }
 
 type CronJobManifest struct {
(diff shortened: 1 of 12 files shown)


ui/admin/app/components/workflow-triggers/ScheduleForm.tsx
```diff
@@ -116,6 +120,17 @@ export function ScheduleForm({ cronjob }: { cronjob?: CronJob }) {
...
+						label="Timezone"
+						onChange={(timezone) =>
+							form.setValue("timezone", timezone, {
+								shouldValidate: true,
+								shouldDirty: true,
+							})
+						}
+						value={form.watch("timezone")}
+					/>
+
```
ryanhopperlowe: can you use `ControlledCustomInput` here? it provides some useful accessibility features

ryanhopperlowe: It also provides a standard method for providing a label, so you can omit that from `TimezoneSelection.tsx`

StrongMonkey: @ryanhopperlowe ok, I did some effort and have no idea how to use it outside a form... 🤷 

maybe you can take a look after it. It is used in a different place where form is not provided.

StrongMonkey:
@ryanhopperlowe changes pushed

g-linville (approved):

ryanhopperlowe (approved):
thank you!!

StrongMonkey:
Per Darren's UX suggestion:

We are going to change the UX so that user doesn't need to pick timezone in current UX. Browser will by default send its current timezone and backend will save it and use for cronjob. It will only display the timzeone in UI when browser's current timezone is different than the one set in database.

StrongMonkey:
I changed the UX to match what we have discussed for disabling selection of a timezone now.

https://www.loom.com/share/c12007f78e7f4219970e89056257505f

ryanhopperlowe:
The layout for the schedulers seems to be shifted depending on the existence of a timezone. Looks a bit wonky

StrongMonkey:
@ryanhopperlowe yeah I agree... don't have a a clear solution for that. It is a bit weird if you have one entry for the same timezone but one different one in a different timezone.

StrongMonkey:
Going to open a new PR given a lot of place have changed.

PR #1685, opened 2025-02-07
feat: encryption with Google Cloud KMS

g-linville:
for https://github.com/obot-platform/obot/issues/1686


drpebcak (approved):

tools/package-tools.sh
```diff
@@ -124,6 +124,15 @@ if [ ! -e aws-encryption-provider ]; then
...
 cd aws-encryption-provider
 go build -o "${BIN_DIR}/aws-encryption-provider" cmd/server/main.go
+cd ..
+
+if [ ! -e k8s-cloudkms-plugin ]; then
+	git clone --depth=1 https://github.com/GoogleCloudPlatform/k8s-cloudkms-plugin
+fi
+cd k8s-cloudkms-plugin
+go build -ldflags "-s -w -extldflags 'static'" -installsuffix cgo -tags netgo -o "${BIN_DIR}/gcp-encryption-provider" cmd/k8s-cloudkms-plugin/main.go
+cd ../..
```
thedadams: Previously, for the AWS encryption provider, we weren't changing directories at the end. Now we are. Just wanted to double-check that this is correct because it seemed suspicious to me.

g-linville: Thank you for catching this!!

thedadams (approved):

PR #1695, opened 2025-02-10
enhance: workflows -> user tasks, task runs & chat threads

ivyjeong13:
* workflows -> tasks (we're technically still using workflow from the backend, a task is a workflow like assistant is an agent)
* separate threads -> tasks runs * chat threads
* task/workflow changes


ui/admin/app/components/composed/DataTable.tsx
```diff
@@ -113,3 +122,53 @@ export function DataTable<TData, TValue>({
...
+		keyof {
+			[K in keyof TData as TData[K] extends string | number
+				? K
+				: never]: TData[K];
+		}
+) => {
+	const navigate = useNavigate();
+
+	const handleAction = (row: TData, ctrl: boolean) => {
+		const path = $path(url, { id: String(row[property]) });
```
ryanhopperlowe: it looks like this will cause non-obvious breakages if we provide a route that doesn't take an ID

ryanhopperlowe: this would likely get you the same result, with a bit less TS mess:

```ts
const useRowNavigate = <TData extends object>(
	getPath: (row: TData) => string
) => {
	const navigate = useNavigate();

	const handleAction = (row: TData, ctrl: boolean) => {
		const path = getPath(row);
		if (ctrl) {
			window.open(`/admin${path}`, "_blank");
		} else {
			navigate(path);
		}
	};

	return {
		internal: (row: TData) => handleAction(row, false),
		external: (row: TData) => handleAction(row, true),
	};
};

const navigate = useRowNavigate((agent: Agent) => $path("/agents/:id", { id: agent.id }))

... 

return <Table
    onRowClick={navigate.internal}
    onRowCtrlClick={navigate.external}
/>
```

this will also be a lot more flexible if we need to use multiple parameters

ui/admin/app/routes/_auth.chat-threads._index.tsx
```diff
@@ -0,0 +1,260 @@
...
+	function getColumns(): ColumnDef<(typeof data)[0], string>[] {
+		return [
+			columnHelper.accessor((thread) => thread.parentName, {
+				id: "Agent",
+				header: ({ column }) => (
+					<DataTableFilter
+						key={column.id}
+						field="Agent"
+						values={
+							getAgents.data?.map((agent) => ({
```
ryanhopperlowe, line 156: Nit: this logic could get messy once we implement server side pagination, but I also think there's a good chance that addresing that will be more work than it's worth. Just making a note here

ui/admin/app/components/composed/SearchInput.tsx
```diff
@@ -0,0 +1,31 @@
...
+				type="text"
+				placeholder={placeholder}
+				value={searchQuery}
+				onChange={(e) => {
+					setSearchQuery(e.target.value);
+					debounceOnChange(e.target.value);
+				}}
+				className="w-64 pl-10"
+			/>
+		</div>
```
ryanhopperlowe: nit: I added a `startContent` prop to `Input` to make this easier :)

ui/admin/app/components/ui/button.tsx
```diff
@@ -24,6 +24,7 @@ const buttonVariants = cva(
 				link: "text-primary underline-offset-4 shadow-none hover:text-primary/70 hover:underline hover:shadow-none",
 				warning:
 					"bg-warning text-primary-foreground shadow-sm hover:bg-warning/80",
+				text: "p-0 text-muted-foreground hover:text-muted-foreground/70",
```
ryanhopperlowe: can you add `hover:shadow-none` here too. There's a little silhouette whenever you hover over the table headers

ryanhopperlowe (commented):
couple of things, but other than that looks great

ui/admin/app/routes/_auth.chat-threads._index.tsx
ivyjeong13: Yea, similarly I think the filter in the table headers will also need a revamp with it as well. Once that happens, happy to make the changes too.

ui/admin/app/components/composed/DataTable.tsx
ivyjeong13: Yes! That definitely opens up and makes less of a TS mess, thanks!

ryanhopperlowe (approved):
Thanks!

PR #1699, opened 2025-02-10
Revert "fix: setup tooltip takes user to auth config directly"

cjellick:
Reverts obot-platform/obot#1655


thedadams:
Can be closed with https://github.com/obot-platform/obot/pull/1701#pullrequestreview-2606929439 being the fix for the issue.

PR #1709, opened 2025-02-11
fix: update microsoft 365 oauth instruction text

tybalex:
quick one line fix, the last section is for multi-tenant only.


thedadams (approved):

ui/admin/app/lib/model/oauthApps/providers/microsoft365.ts
```diff
@@ -133,7 +133,7 @@ If you already set your web redirect URI in step 1, you can skip to Step 3.\n
 	{
 		type: "markdown",
 		text: `
-#### For Single-Tenant Applications Only!\n\n
+#### For Multi-Tenant Applications Only!\n\n
```
drpebcak, line 136: Pretty sure this text was correct, and the next line is the one that needs to be fixed. You only need the directory id when you have a single-tenant app.

Issue #1663, opened 2025-02-05
Files and tables show incorrect pagination when there are no entries


sangee2004:
**Steps to reproduce the problem:**
1. Create an agent and chat with it to create a thread.
2. Look at the threads inspect view and expand on Files and tables list.

Notice that there is incorrect pagination of `< 1/0 >`

<img width="320" alt="Image" src="https://github.com/user-attachments/assets/94af2b9b-a68f-4dd1-9807-7bff4c690733" />

sangee2004:
This issue is not seen anymore when testing with latest build.

<img width="334" alt="Image" src="https://github.com/user-attachments/assets/0daa7726-7451-4aff-a262-8b2b3f482900" />

PR #1714, opened 2025-02-11
fix: showing 1/0 for pagination with no pages

ryanhopperlowe:
addresses #1663

image

image

### DIFF
--- a/ui/admin/app/components/composed/PaginationActions.tsx
+++ b/ui/admin/app/components/composed/PaginationActions.tsx
@@ -20,6 +20,8 @@ export function PaginationActions({
 	const hasNextPage = nextPage != null;
 	const hasPreviousPage = previousPage != null;
 
+	const currentPage = totalPages ? page + 1 : 0;
+
 	return (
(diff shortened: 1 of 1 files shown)


ivyjeong13 (approved):

thedadams:
@ryanhopperlowe The second screenshot shows `1/1` which confused me. As I was typing this out, I realized that it is coutning pages and not files there. Can you think of a way to make that more obvious? Or am I just being thick-headed?

ryanhopperlowe:
> @ryanhopperlowe The second screenshot shows `1/1` which confused me. As I was typing this out, I realized that it is counting pages and not files there. Can you think of a way to make that more obvious? Or am I just being thick-headed?

I can't really think about how to indicate better that these are pages without taking up a bunch of horizontal real estate with `next page` and `previous page` buttons

Another option would be to replace `1/1` with `0-2 of 2`. This is fine until numbers get bigger... looks a bit out of place when it says `50-74 of 100`

I think that keeping it simple for now makes the most sense, but I'm happy to bring it to Craig as well.

Issue #1719, opened 2025-02-11
Support enabling multiple auth providers at once


g-linville:

g-linville:
Multiple auth providers can now be enabled at the same time. You can have all four enabled if you want. Just configure multiple and play around with them during testing. You should be able to seamlessly log in with one, then log back out, and log in with another. Identities are mapped to accounts on the basis of email address, so if you have the same email address configured for your accounts upstream (Google, GitHub, Entra, Okta), then you should be logged in as the same user in Obot (though perhaps with a different profile photo) and have access to the same agents, threads, etc.

sangee2004:
Allowed to have multiple auth providers configured at the same time.
Able to work with users logged in from different providers.
When email address for accounts logged in using different providers are the same , they will all be mapped to the same otto user.
There is 1 related bug when testing this use case - https://github.com/obot-platform/obot/issues/1775

PR #1720, opened 2025-02-11
enhance: support multiple auth providers

g-linville:
for https://github.com/obot-platform/obot/issues/1719

This makes it so that a new identity will be mapped to an existing user, if the existing user has the same email address as the new identity. Previously we were doing this based on username, but while usernames in GitHub can be set by the user, the username in Google is just a random number, so this doesn't work.

I also removed the restrictions (in both the frontend and the backend) that prevent you from configuring more than one auth provider at once.

Google and GitHub both require you to verify that you have access to the email for your account, so this implementation is secure. It is important that future auth providers that we add also adhere to this constraint, and do not allow you to have an email address that you don't really own.

### DIFF
--- a/pkg/gateway/client/identity.go
+++ b/pkg/gateway/client/identity.go
@@ -63,7 +63,7 @@ func ensureIdentity(tx *gorm.DB, id *types.Identity, timezone string, role types
 	if user.ID != 0 {
 		userQuery = userQuery.Where("id = ?", user.ID)
 	} else {
-		userQuery = userQuery.Where("username = ?", user.Username)
+		userQuery = userQuery.Where("email = ?", user.Email)
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/gateway/client/identity.go
```diff
@@ -63,7 +63,7 @@ func ensureIdentity(tx *gorm.DB, id *types.Identity, timezone string, role types
 	if user.ID != 0 {
 		userQuery = userQuery.Where("id = ?", user.ID)
 	} else {
-		userQuery = userQuery.Where("username = ?", user.Username)
+		userQuery = userQuery.Where("email = ?", user.Email)
```
g-linville, line 66: Surprisingly, this is the only backend change that was necessary to reconcile identities from different auth providers down to the same user.

njhale (approved):

ryanhopperlowe (approved):
+7 -62 ❤️

Issue #1723, opened 2025-02-12
Cleanup threads that are "ephemeral" on some interval


thedadams:
There are threads, like the ones used to authenticate agents and workflows, that are ephemeral and don't need to be displayed to the user. We should clean up these threads at some interval. Twelve hours seems like the most reasonable choice.

thedadams:
As of the time of this comment, ephemeral threads are those created for authentication and model provider validation.

sejubar:
A note for testing ( from Donnie Adams )

1. Create an agent and add a tool that needs authentication.
2. List all the threads by going to /api/threads (you can do this in a browser)
3. Authenticate the tool for the agent by clicking the globe icon next to the tool.
4. List threads again and find the one that was created for this authentication. You should save the id of the thread.
5. Wait 12 hours
6. Verify that the thread no longer exists by going to /api/threads/THREAD_ID where the THREAD_ID is the id you saved above.

sejubar:
created a slack authentication with thread informations below;

```
{
            "id": "t1szrjr",
            "created": "2025-02-28T15:59:47Z",
            "revision": "59645",
            "type": "thread",
            "name": "",
            "icons": null,
            "prompt": "",
            "knowledgeDescription": "",
            "introductionMessage": "",
            "starterMessages": null,
            "state": "continue",
            "lastRunID": "r126ss6",
            "abort": true,
            "ephemeral": true
        }
```
I will validate in 12 hours to see if the thread is cleaned up.

sejubar:
tested on https://test.acornlabs.com

https://test.acornlabs.com/api/threads/t1szrjr returns 404 after  12 hours + 

**Enviroment**:
tested on https://test.acornlabs.com

**Components**:

```
{
  "authEnabled": true,
  "chrome": "Chromium 133.0.6943.126",
  "dockerSupported": false,
  "emailDomain": "",
  "github.com/gptscript-ai/datasets": "31ce6c3",
  "github.com/gptscript-ai/workspace-provider": "b24dbdd",
  "github.com/kubernetes-sigs/aws-encryption-provider": "91f20e6",
  "github.com/obot-platform/enterprise-tools": "a36994f",
  "github.com/obot-platform/k8s-cloudkms-plugin": "62f15f1",
  "github.com/obot-platform/tools": "fb06251",
  "gptscript": "v0.9.5-86e85f0",
  "obot": "v0.0.0-dev+d96c83e8"
}
```

PR #1724, opened 2025-02-12
enhance: cleanup ephemeral threads after 12 hours

thedadams:
This change introduces the concept of an ephemeral thread. These are threads
that are typically used for system-based or temporary tasks. Such threads will
be cleaned up 12 hours after their creation.

Issue: https://github.com/obot-platform/obot/issues/1723

### DIFF
--- a/pkg/controller/handlers/threads/threads.go
+++ b/pkg/controller/handlers/threads/threads.go
@@ -2,6 +2,7 @@ package threads
 
 import (
 	"context"
+	"time"
 
 	"github.com/gptscript-ai/go-gptscript"
 	"github.com/obot-platform/nah/pkg/name"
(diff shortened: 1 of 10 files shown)


pkg/controller/handlers/threads/threads.go
```diff
@@ -140,3 +141,29 @@ func (t *Handler) CreateKnowledgeSet(req router.Request, _ router.Response) erro
...
+			thread.Spec.KnowledgeSetName == "" &&
+			thread.Spec.KnowledgeSourceName == ""
+		if thread.Spec.Ephemeral {
+			return req.Client.Update(req.Ctx, thread)
+		}
+	}
+	if !thread.Spec.Ephemeral ||
+		thread.CreationTimestamp.After(time.Now().Add(-12*time.Hour)) {
+		return nil
+	}
```
StrongMonkey, line 166: If it is Ephemeral, should it be retry/enqueued after 12 hours? I would feel like this would never get cleaned after 12 hours unless it is restarted.

thedadams: Everything gets re-enqueued every 10 hours.

StrongMonkey (approved):

iwilltry42 (approved):

PR #1731, opened 2025-02-12
Fix: Fix tool reference to properly index gpt files

StrongMonkey:
The previous way of indexing a tool based on localTool doesn't seem to be correct. the local tools parsed from gptscript load doesn't seem to have tools that are referenced from other gptscripts.

Instead, we should be able to look up tools directly from tool.export. It also needs to correctly reference tools from other gptscript files like ./company.gpt when indexing subTool.

This will solve the hubspot tool case where sub tools are located into other files other that tool.gpt.

### DIFF
--- a/pkg/tools/resolve.go
+++ b/pkg/tools/resolve.go
@@ -3,6 +3,7 @@ package tools
 import (
 	"context"
 	"fmt"
+	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/tools/resolve.go
```diff
@@ -51,12 +52,29 @@ func ResolveToolReferences(ctx context.Context, gptClient *gptscript.GPTScript,
...
+	}
+
+	for _, peerToolID := range exportToolIDs {
 		if peerToolID == prg.EntryToolID {
 			continue
 		}
 
 		peerTool := prg.ToolSet[peerToolID]
+		ref, _, _ := strings.Cut(peerToolID, ":")
+		toolRef := reference
```
StrongMonkey, line 70: The toolRef has to be properly parsed and joined by its current reference so that it can refer to tools that are located to files other than default tool.gpt. The current implementation assumes everything is in tool.gpt.

g-linville (approved):

PR #1732, opened 2025-02-12
fix: rearrange sidebar nav items & combobox multi-highlight name fix

ivyjeong13:
* rearrange sidebar nav items per Craig's request
* fix combobox highlighting items with same name (but different id)

fixes:
Screenshot 2025-02-12 at 1 49 54 PM

adds agent name if the workflow has a duplicate title with someone else:
Screenshot 2025-02-12 at 3 16 40 PM

### DIFF
--- a/ui/admin/app/components/composed/ComboBox.tsx
+++ b/ui/admin/app/components/composed/ComboBox.tsx
@@ -1,6 +1,8 @@
 import { CheckIcon, ChevronsUpDownIcon } from "lucide-react";
 import { ReactNode, useState } from "react";
 
+import { cn } from "~/lib/utils";
+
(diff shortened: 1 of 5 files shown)


ui/admin/app/components/composed/ComboBox.tsx
```diff
@@ -244,6 +254,12 @@ function ComboBoxList<T extends BaseOption>({
 			>
 				<span>
 					{renderOption ? renderOption(option) : (option.name ?? option.id)}
+					{option.sublabel && (
+						<span className="text-xs text-muted-foreground">
```
ryanhopperlowe: nit: can you use `small` instead of `<span className="text-xs">...`

ryanhopperlowe (approved):

Issue #1397, opened 2025-01-22
When logged in using github auth provider , show github handle instead of email address as user's identity.


sangee2004:
**Steps to reproduce the problem:**
1. Enable github auth provider.
2. Log in to otto server using github auth

<img width="530" alt="Image" src="https://github.com/user-attachments/assets/dbe5735e-8015-46ac-8536-17902b1584cf" />

3. Once logged in successfully, notice that email associated with the github account (used to log in) is presented as the user's identity.

<img width="338" alt="Image" src="https://github.com/user-attachments/assets/39af3295-d9ae-4f2c-bb71-e308b6cc367a" />

In users page
<img width="1112" alt="Image" src="https://github.com/user-attachments/assets/8f49fc8c-0d1d-4020-b5b6-f8b5e00255bd" />

Expected Behavior:
We should be able to show the github handle of the logged in user instead of email.

ryanhopperlowe:
@cjellick @sangee2004 @thedadams - Github and Google have different "primary keys" for identifying users to themselves: Google displays their `email` while github displays the `username`. Both of these fields are populated on the `/me` and `/users/` apis, and I don't know which one to display at a given time. 

My first thought was to always display both, but that will get cluttered in the User dropdown since we already display the email and role

<img width="276" alt="Image" src="https://github.com/user-attachments/assets/234a44ed-5523-4d82-a267-0c7a5eab7c7d" />

In order to get this to work without always displaying both, I need the backend to give me some sort of `displayName` property to tell me what I should display as the identifier for the user.

cjellick:
Yeah, I agree. Arguably, if the auth provider used to login was made available on /api/me, you could map it, but it isn't.

@g-linville - where this gets more interest for me is okta and entra. As you do these ones, id like to know how username and email appear as well as if we have the user's "real" name like John Smith

Does the framework give you indicator of what the providers "preferred" display name is?

g-linville:
@cjellick for Okta, the email address is the username. For Entra, I believe it is the same thing.
I could modify the API between Obot and the auth providers to return a `displayName` field that we can use as part of the user info, if that is something that would be helpful.

sangee2004:
@ivyjeong13 I am testing with latest build - `obot:v0.0.0-dev+f04e0790`

In the `Users` page , i am still seeing email address for github users.

Wanted to confirm if we want to show the github handle in this case?

PR #1740, opened 2025-02-13
fix: show github handle instead of email for github auth provider

ivyjeong13:
* uses username for github, otherwise defaults to email
* Addresses #1397 

github:
Screenshot 2025-02-13 at 1 43 52 PM

google:
Screenshot 2025-02-13 at 1 44 44 PM

TODO:
- [x] Test case

### DIFF
--- a/apiclient/types/user.go
+++ b/apiclient/types/user.go
@@ -22,7 +22,7 @@ type User struct {
 	Email         string `json:"email,omitempty"`
 	IconURL       string `json:"iconURL,omitempty"`
 	Timezone      string `json:"timezone,omitempty"`
-	AuthProvider  string `json:"authProvider,omitempty"`
+	CurrentAuthProvider  string `json:"currentAuthProvider,omitempty"`
 }
 
 type UserList List[User]
--- a/apiclient/types/user.go
+++ b/apiclient/types/user.go
@@ -22,6 +22,7 @@ type User struct {
 	Email         string `json:"email,omitempty"`
 	IconURL       string `json:"iconURL,omitempty"`
 	Timezone      string `json:"timezone,omitempty"`
+	AuthProvider  string `json:"authProvider,omitempty"`
 }
 
 type UserList List[User]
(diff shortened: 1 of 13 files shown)


pkg/gateway/types/users.go
```diff
@@ -37,6 +37,12 @@ func ConvertUser(u *User, roleFixed bool) *types2.User {
 		IconURL:       u.IconURL,
 		Timezone:      u.Timezone,
 	}
+
+	if authProviderName != nil {
```
ivyjeong13: Wanted to check if there's no issue passing the authProvider for current user?

ryanhopperlowe (approved):

cjellick (approved):

pkg/gateway/types/users.go
```diff
@@ -20,12 +20,12 @@ type User struct {
 	Timezone  string      `json:"timezone"`
 }
 
-func ConvertUser(u *User, roleFixed bool) *types2.User {
+func ConvertUser(u *User, roleFixed bool, authProviderName *string) *types2.User {
```
thedadams: nit: let's just make this a string. Then you can always set it in the user struct. There is no difference between setting a field to the empty string versus not setting it in Go.

apiclient/types/user.go
```diff
@@ -22,6 +22,7 @@ type User struct {
 	Email         string `json:"email,omitempty"`
 	IconURL       string `json:"iconURL,omitempty"`
 	Timezone      string `json:"timezone,omitempty"`
+	AuthProvider  string `json:"authProvider,omitempty"`
```
thedadams: This may change from on request to the next. In other words, the same user may come back with different auth providers when more than one auth provider is enabled and two "identities" have been mapped to the same user (I believe we do this by email).

Because of this, how do you feel about renaming this field to `CurrentAuthProvider` (or something else) to make it more apparent this this is not set permanently?

ivyjeong13: Makes sense, I'm down with that change! Will update!

thedadams (approved):

Issue #1734, opened 2025-02-13
Database tool sql string encoding issues


njhale:
The database tool's context is causing LLM's to double-encode the sqlite statement with an extra layer of quotes, causing the tool call to fail. The LLM can recover from this failure, but we can do better.

sejubar:
![Image](https://github.com/user-attachments/assets/2d41865c-420d-48be-a055-7715c736ab85)

**Enviroment:**
http://localhost:8080/

```
{
  "authEnabled": true,
  "chrome": "Chromium 134.0.6998.88",
  "dockerSupported": false,
  "emailDomain": "",
  "github.com/Azure/kubernetes-kms": "465705b",
  "github.com/gptscript-ai/datasets": "31ce6c3",
  "github.com/gptscript-ai/workspace-provider": "23e4267",
  "github.com/kubernetes-sigs/aws-encryption-provider": "91f20e6",
  "github.com/obot-platform/k8s-cloudkms-plugin": "62f15f1",
  "github.com/obot-platform/tools": "2d46211",
  "gptscript": "v0.9.5-cee1a75",
  "obot": "v0.0.0-dev+d7d50f5a"
}
```

PR #1746, opened 2025-02-13
chore: update admin and user ui to accept new database tool output

njhale:
The database tools output the UIs depend on will change slightly with https://github.com/obot-platform/tools/pull/438, adjust the UI's to account for this.

Part of https://github.com/obot-platform/obot/issues/1734#issue-2850199682


njhale:
Going back to the old format so that we can know the column names when the database is empty.

Issue #1060, opened 2024-12-26
The threads page needs filtering


cjellick:
need to be able to filter by type, user, name, create date

ivyjeong13:
Note: Threads is now divided into two pages: Task Runs & Chat Threads

sangee2004:
We now have "Chat threads" for agent threads  and "Task Runs"  for task threads.
Each of these pages have filters for agent/task , user and created fields.

Created field filter is not working properly which will be tracked in a separate issue.

ivyjeong13:
> Discussed with [@ivyjeong13](https://github.com/ivyjeong13) on the following issues seen with "Created" field filter.
> 
> 1. When a Date range is picked for calendar filter , pill showing the the selected range is not presented.
> 2. When selecting/de selecting a date range , it is not obvious that this action is actually made. We can make this more obvious by having an explicit select/clear button ?
> 3. When only 1 date is picked and clicked once , no action is taken since a date range needs to be picked for the filter to be applied. This is not obvious to the user.
> 4. In the Calendar view the left and right navigation buttons overlap
> 
[...]

PR #1748, opened 2025-02-13
fix: created filter and resizable table columns

ivyjeong13:
* adds resizable columns to tables
* add created filter

Addresses #1060 

https://www.loom.com/share/dddc4b18b04e4dd7a90869d11d252dc4


ivyjeong13:
Note: React Tanstack has an example of a performant resizing: https://tanstack.com/table/v8/docs/framework/react/examples/column-resizing-performant 

If it looks like we're hitting performant issues, will make a follow-up PR to follow example above.

ryanhopperlowe:
Maybe some issues we could create for the future:

1. save table/column sizes to localStorage so that users don't have to resize when we have pagination or on page reload
2. Date filters should be date ranges, so that the user doesn't have to know what specific date they created something. Notion has a GREAT implementation for date filters that we can draw inspiration from

Screenshot 2025-02-14 at 9 42 37 AM
Screenshot 2025-02-14 at 9 43 03 AM

ui/admin/app/components/composed/DataTable.tsx
```diff
@@ -47,6 +54,9 @@ export function DataTable<TData, TValue>({
 	onCtrlClick,
 }: DataTableProps<TData, TValue>) {
 	const table = useReactTable({
```
ryanhopperlowe, line 58: Man, I love the flexibility of @tanstack/table

But I wonder if it would be worth pulling in a pre-built table component to save us the hassle of building out features like this. We should probably discuss this soon

ivyjeong13:
> Maybe some issues we could create for the future:
> 
> 1. save table/column sizes to localStorage so that users don't have to resize when we have pagination or on page reload
> 2. Date filters should be date ranges, so that the user doesn't have to know what specific date they created something. Notion has a GREAT implementation for date filters that we can draw inspiration from
> 
> Screenshot 2025-02-14 at 9 42 37 AM Screenshot 2025-02-14 at 9 43 03 AM

Yeah, I was thinking about whether or not to turn it into a date range! And I like the localStorage idea to save the resize changes! I think I can make the date range change pretty quick but I'll prolly create a follow-on issue for the saved resized states!

ui/admin/app/routes/_auth.chat-threads._index.tsx
```diff
@@ -89,8 +90,20 @@ export default function TaskRuns() {
...
 
+		if (created) {
+			const createdDate = new Date(created);
+			filteredThreads = filteredThreads.filter((thread) => {
+				const threadDate = new Date(thread.created);
+				return (
+					threadDate.getFullYear() === createdDate.getFullYear() &&
+					threadDate.getMonth() === createdDate.getMonth() &&
+					threadDate.getDate() === createdDate.getDate()
+				);
```
ryanhopperlowe: nit: wondering if we should bring in `date-fns`

they have a ton of great date-related helpers:

```ts
import { isSameDay } from 'date-fns'

...

return isSameDay(threadDate, createdDate)
```

reference: https://date-fns.org/v4.1.0/docs/isSameDay

ui/admin/app/components/composed/DataTable.tsx
ivyjeong13: I actually enjoy the flexibility but happy to look into pre-built options too!

ui/admin/app/routes/_auth.chat-threads._index.tsx
ivyjeong13: Changed to use date-fns! edit: and date range instead!

ui/admin/app/lib/service/routeService.ts
```diff
@@ -14,19 +14,21 @@ const QuerySchemas = {
 	}),
 	threadsListSchema: z.object({
 		agentId: z.string().nullish(),
-		created: z.string().nullish(),
 		userId: z.string().nullish(),
 		taskId: z.string().nullish(),
 		from: z.enum(["tasks", "agents", "users"]).nullish().catch(null),
+		start: z.string().nullish(),
+		end: z.string().nullish(),
```
ryanhopperlowe: nit: might want to change this to `createdStart` and `createdEnd` can be a follow up though

ryanhopperlowe (approved):

Issue #1751, opened 2025-02-14
Store oauth app configuration in credentials


g-linville:

g-linville:
To test this using local Obot in a Docker container:

1. Start up Obot and configure an OAuth for a tool like GitHub. (You don't need to put in real credentials. It can just be random strings.)
2. Connect to the database by running `docker exec -it <container ID> psql -U obot -d obot`.
3. Run `select * from gptscript_credentials;` to list out the stored credentials. You should see one for the tool you configured with the client secret you entered.
4. Run `select * from oauthapp;`. You should see the configuration, minus the client secret.

sejubar:
**Environment**: 

http://localhost:8080

**Components**:

```
{
  "authEnabled": false,
  "chrome": "Chromium 133.0.6943.126",
  "dockerSupported": false,
  "emailDomain": "",
  "github.com/gptscript-ai/datasets": "31ce6c3",
  "github.com/gptscript-ai/workspace-provider": "b24dbdd",
  "github.com/kubernetes-sigs/aws-encryption-provider": "91f20e6",
  "github.com/obot-platform/k8s-cloudkms-plugin": "62f15f1",
  "github.com/obot-platform/tools": "b8b32b4",
  "gptscript": "v0.9.5-86e85f0",
  "obot": "v0.0.0-dev+f115e4a1"
}
```

PR #1764, opened 2025-02-15
fix: oauth apps: store client secret as a credential

g-linville:
for https://github.com/obot-platform/obot/issues/1751

With this change, newly configured oauth apps will have their client secrets stored as credentials so that they get encrypted.

Existing oauth apps, when upgrading to this code, will continue to have their client secrets in the `oauthapp` table, alongside the rest of their configuration. They will not get the new credential created unless they are reconfigured. Once reconfigured, the old revision will still exist in the database due to how kinm manages object revisions, but they should get pruned eventually.

### DIFF
--- a/pkg/gateway/server/oauth_apps.go
+++ b/pkg/gateway/server/oauth_apps.go
@@ -465,14 +465,9 @@ func (s *Server) callbackOAuthApp(apiContext api.Context) error {
 		return fmt.Errorf("failed to create token request: %w", err)
 	}
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
-<<<<<<< HEAD
 	if app.Spec.Manifest.Type != types2.OAuthAppTypeGoogle &&
 		app.Spec.Manifest.Type != types2.OAuthAppTypePagerDuty {
 		req.SetBasicAuth(url.QueryEscape(app.Spec.Manifest.ClientID), url.QueryEscape(cred.Env["CLIENT_SECRET"]))
-=======
-	if app.Spec.Manifest.Type != types2.OAuthAppTypeGoogle {
-		req.SetBasicAuth(url.QueryEscape(app.Spec.Manifest.ClientID), url.QueryEscape(cred.Env["CLIENT_SECRET"]))
->>>>>>> d61e017b (fix: store oauth app client secrets as credentials)
 	}
 
 	resp, err := http.DefaultClient.Do(req)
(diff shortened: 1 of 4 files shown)


pkg/gateway/server/oauth_apps.go
```diff
@@ -125,6 +126,20 @@ func (s *Server) createOAuthApp(apiContext api.Context) error {
 		return types2.NewErrHTTP(http.StatusConflict, fmt.Sprintf("OAuth app with alias %s already exists", appManifest.Alias))
 	}
 
+	// Store the client secret as a credential.
+	credential := gptscript.Credential{
+		Context:  fmt.Sprintf("%s-%s", appManifest.Alias, appManifest.ClientID),
+		ToolName: appManifest.Alias,
+		Type:     gptscript.CredentialTypeTool,
+		Env:      map[string]string{"client_secret": appManifest.ClientSecret},
```
thedadams: nit: `Env` here means `CLIENT_SECRET` to me.

thedadams: Also, what are your thoughts on the following?
1. Create the OAuth app first.
2. Put the credential in the credential context corresponding to the generated name of the OAuth app.

My thoughts on this are:
1. Allows the same OAuth credentials to be used for multiple OAuth apps (assuming we support that one day).
2. Allows for multiple OAuth apps with the same `alias` (assuming we support that one day).

g-linville: I can update that string. It's not actually an environment variable, since we are using creds outside the context of tools here. So it doesn't really make a difference whether it's caps or not, as long as we use the same string later when we read the credential.

Regarding this:
> Allows the same OAuth credentials to be used for multiple OAuth apps (assuming we support that one day).

I'm not sure I understand how this would allow for that.

thedadams: This implementation doesn't allow for the same Client ID to be used for the same alias: when you try to create a second OAuth app with the same client credentials, you will get an error each time about the credential already existing.

If you put these credentials in a credential context corresponding to the generated name of the OAuth app, then this type of collision won't occur.

g-linville: Oh now I understand. Thanks.

g-linville: Fixed

g-linville:
Sorry for the rebase. Had to fix a conflict with main.

thedadams (approved):

Issue #1582, opened 2025-02-03
Agent level credentials for the tools is not made available to the tasks(Workflows) that gets created from user ui.


sangee2004:
**Steps to reproduce the problem:**
1. Create an agent and add github tool with PAT authentication.
2. Enable pre authentication

<img width="1293" alt="Image" src="https://github.com/user-attachments/assets/80a5b030-b423-408a-ba87-aea0dd1f0fb3" />

3. Access this agent in user ui using publish url
4. Create a task and a step that uses github tool like `most recent issue in <repo>`

Use "Test" to run the step.

Notice that user is promoted for creds.

<img width="1294" alt="Image" src="https://github.com/user-attachments/assets/cc94e811-1787-45a7-83f9-4cb0900567df" />

The workflow that gets created does not have a credential for github tool  associated with it.

sangee2004:
@ryanhopperlowe This issue of prompting for credentials when global creds are present in not addressed.

I do see this issue addressed - https://github.com/obot-platform/obot/issues/1296

cjellick:
This isnt a UI issue I dont think

g-linville:
Just merged the fix.

sangee2004:
This issue is addressed when testing the latest build.
Credentials created for the agent is now available to the tasks that are created for the agent. Able to execute the task without being prompted for the crendentials.

PR #1774, opened 2025-02-18
fix: tasks: use agent's credential context

g-linville:
for https://github.com/obot-platform/obot/issues/1582

This adds a new field to the Workflows (Tasks) so that we can store the agent name and use that as an additional credential context.

### DIFF
--- a/pkg/controller/migrations.go
+++ b/pkg/controller/migrations.go
@@ -56,3 +56,18 @@ func changeWorkflowStepOwnerGVK(req router.Request, _ router.Response) error {
 	}
 	return nil
 }
+
+func setWorkflowAdditionalCredentialContexts(req router.Request, _ router.Response) error {
+	wf := req.Object.(*v1.Workflow)
+	if len(wf.Spec.AdditionalCredentialContexts) != 0 {
+		return nil
+	}
+
+	var thread v1.Thread
+	if err := req.Get(&thread, wf.Namespace, wf.Spec.ThreadName); err != nil {
+		return err
+	}
+
+	wf.Spec.AdditionalCredentialContexts = []string{thread.Spec.AgentName}
+	return req.Client.Update(req.Ctx, wf)
+}
--- a/pkg/controller/migrations.go
+++ b/pkg/controller/migrations.go
@@ -56,18 +56,3 @@ func changeWorkflowStepOwnerGVK(req router.Request, _ router.Response) error {
 	}
 	return nil
 }
-
-func setWorkflowAdditionalCredentialContexts(req router.Request, _ router.Response) error {
-	wf := req.Object.(*v1.Workflow)
-	if len(wf.Spec.AdditionalCredentialContexts) != 0 {
-		return nil
-	}
-
-	var thread v1.Thread
-	if err := req.Get(&thread, wf.Namespace, wf.Spec.ThreadName); err != nil {
-		return err
-	}
-
-	wf.Spec.AdditionalCredentialContexts = []string{thread.Spec.AgentName}
-	return req.Client.Update(req.Ctx, wf)
-}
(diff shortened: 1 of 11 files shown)


pkg/controller/migrations.go
```diff
@@ -56,3 +56,18 @@ func changeWorkflowStepOwnerGVK(req router.Request, _ router.Response) error {
...
 }
+
+func setWorkflowAdditionalCredentialContexts(req router.Request, _ router.Response) error {
+	wf := req.Object.(*v1.Workflow)
+	if len(wf.Spec.AdditionalCredentialContexts) != 0 {
+		return nil
+	}
+
+	var thread v1.Thread
+	if err := req.Get(&thread, wf.Namespace, wf.Spec.ThreadName); err != nil {
```
thedadams: nit: just in case.
```suggestion
	if err := req.Get(&thread, wf.Namespace, wf.Spec.ThreadName); err != nil || thread.Spec.AgentName == "" {
```

pkg/controller/routes.go
```diff
@@ -72,6 +72,7 @@ func (c *Controller) setupRoutes() error {
 	root.Type(&v1.Workflow{}).HandlerFunc(workflow.EnsureIDs)
 	root.Type(&v1.Workflow{}).HandlerFunc(workflow.CreateWorkspaceAndKnowledgeSet)
 	root.Type(&v1.Workflow{}).HandlerFunc(workflow.BackPopulateAuthStatus)
+	root.Type(&v1.Workflow{}).HandlerFunc(setWorkflowAdditionalCredentialContexts)
```
thedadams: I would be OK with this being a startup migration rather than a controller. What are your thoughts?

pkg/invoke/invoker.go
```diff
@@ -334,6 +334,9 @@ func (i *Invoker) Agent(ctx context.Context, c kclient.WithWatch, agent *v1.Agen
 		credContextIDs = append(credContextIDs, agent.Name)
 	}
 	credContextIDs = append(credContextIDs, agent.Namespace)
+	if len(agent.Spec.AdditionalCredentialContexts) != 0 {
+		credContextIDs = append(credContextIDs, agent.Spec.AdditionalCredentialContexts...)
+	}
```
thedadams: nit: this does the right thing, even for `nil` slices.
```suggestion
	credContextIDs = append(credContextIDs, agent.Spec.AdditionalCredentialContexts...)
```

pkg/render/workflow.go
```diff
@@ -81,6 +81,10 @@ func Workflow(ctx context.Context, c kclient.Client, wf *v1.Workflow, opts Workf
 		agent.Spec.CredentialContextID = wf.Spec.CredentialContextID
 	}
 
+	if len(wf.Spec.AdditionalCredentialContexts) != 0 {
+		agent.Spec.AdditionalCredentialContexts = wf.Spec.AdditionalCredentialContexts
+	}
```
thedadams: same nit here.

thedadams (approved):

StrongMonkey (approved):

pkg/controller/routes.go
g-linville: Good call. That way it doesn't run on every change.

g-linville:
@thedadams I made all the changes you suggested

pkg/invoke/invoker.go
```diff
@@ -334,6 +334,7 @@ func (i *Invoker) Agent(ctx context.Context, c kclient.WithWatch, agent *v1.Agen
 		credContextIDs = append(credContextIDs, agent.Name)
 	}
 	credContextIDs = append(credContextIDs, agent.Namespace)
+	credContextIDs = append(credContextIDs, agent.Spec.AdditionalCredentialContexts...)
```
thedadams, line 336: Let's put this above the `agent.Namespace` credential context so that the namespace is always the "latest resort" credential context.

Issue #1759, opened 2025-02-14
WISIWYG Introductions


ivyjeong13:
* would be nice to add a markdown compatible WISIWYG to introductions
* also fix instructions to match introductions -- it expands on initial focus
Reference: https://acorn-io.slack.com/archives/C07P0EZR917/p1739552109398859

cjellick:
Note that we will want this in the user ui once darren exposes obot buliding there

sangee2004:
Testing wth `obot:v0.0.0-dev+130c4fa0`

Both instruction and introduction fields expands when you click on it. But they donot collapse when focus is taken away from them.

https://www.loom.com/share/74fef0a562014f1fb9959a11efd66fe8?sid=80f4ef9d-140b-4238-b541-6961a13f6495

**Issues seen with WISIWYG introductions**
1. Bulleted list does not get rendered properly in "Introductions" field in Agent view. Gets rendered properly in chat view.
2.  Numbered list does not get rendered properly in "Introductions" field in Agent view. Gets rendered properly in chat view.
3. Check list is truncated a bit in in "Introductions" field in Agent view. But it also shows up as a bulleted list

"Introductions" field in Agent view
<img width="811" alt="Image" src="https://github.com/user-attachments/assets/b2f40131-1f6d-41e1-a993-d5529fa8f7c6" />

Chat view
<img width="533" alt="Image" src="https://github.com/user-attachments/assets/37bf95cf-2c49-4e40-bbd2-16ab839bceac" />

4. Quote shows up as regular text in chat view.

"Introductions" field in Agent view
<img width="928" alt="Image" src="https://github.com/user-attachments/assets/b006b9e3-7ec3-4d8f-b6ad-bca6f59dad20" />
Chat view
<img width="1111" alt="Image" src="https://github.com/user-attachments/assets/74736333-4a0e-4cbd-ac4f-b3d5824a2af7" />

ivyjeong13:
> Both instruction and introduction fields expands when you click on it. But they donot collapse when focus is taken away from them.

Sorry for confusion, should just be expand on initial focus. Added reference to Slack thread in description

sangee2004:
@ivyjeong13 Reopening this issue to follow up on https://github.com/obot-platform/obot/issues/1759#issuecomment-2675003943

ivyjeong13:
Adding as checklist for myself to keep track of: 

[...]

PR #1797, opened 2025-02-20
enhance: wysiwyg introductions

ivyjeong13:
Addresses #1759 

* add markdown wysiwyg for introductions 
Screenshot 2025-02-20 at 10 55 19 AM

* fixes textarea so fields start out minHeight and expand on initial focus
Screenshot 2025-02-20 at 10 55 31 AM
Screenshot 2025-02-20 at 10 55 34 AM


ui/admin/app/components/agent/__tests__/Agent.test.tsx
```diff
@@ -90,7 +90,6 @@ describe(Agent, () => {
 			"placeholder",
 		],
 		["prompt", "Instructions", "textbox", 2],
-		["introductionMessage", "Introductions", "textbox"],
```
ivyjeong13, line 93: removed introduction test case because the mdxeditor doesn't operator like a normal textarea input, suggestions I saw was to mock the MDXEditor component which seems not great cause then I'm not actually testing using that component 😅 So opted to remove it.

ui/admin/app/components/ui/markdown.tsx
```diff
@@ -51,3 +80,75 @@ export function Markdown({
...
+	className,
+	markdown,
+	onChange,
+}: {
+	className?: string;
+	markdown: string;
+	onChange: (markdown: string) => void;
+}) {
+	const { theme } = useTheme();
+	const isDarkMode = theme === AppTheme.Dark;
```
ryanhopperlowe: nit: 

```suggestion
	const { isDark } = useTheme();
```

PR #1800, opened 2025-02-20
fix: dockerfile: copy encryption providers to final stage

g-linville:


tools/package-tools.sh
```diff
@@ -137,7 +137,7 @@ fi
 cd k8s-cloudkms-plugin
 go build -ldflags "-s -w -extldflags 'static'" -installsuffix cgo -tags netgo -o "${BIN_DIR}/gcp-encryption-provider" cmd/k8s-cloudkms-plugin/main.go
 OBOT_SERVER_VERSIONS="$(cat <<VERSIONS
-"github.com/kubernetes-sigs/gcp-encryption-provider": "$(git rev-parse --short HEAD)"
+"github.com/GoogleCloudPlatform/k8s-cloudkms-plugin": "$(git rev-parse --short HEAD)"
```
g-linville, line 140: I noticed that the GH reference here was wrong, so I fixed that too

drpebcak (approved):

StrongMonkey (approved):

PR #1806, opened 2025-02-20
fix: tools description help cursor & conditional tooltip based on content length

ivyjeong13:
* if content is short enough, don't render tooltip -- maybe overkill to check the ref? 😅 
* cursor-help instead of cursor-pointer


ryanhopperlowe (approved):

ryanhopperlowe:
> maybe overkill to check the ref? 

Yeah idk how else we would do this. I'm good with it!

Issue #1807, opened 2025-02-20
Distinguish between verified and unverified email addresses


g-linville:

g-linville:
In order to test this one, you will need to set up local Obot with at least two auth providers: a verified one (Google or GitHub) and an unverified one (Okta or Microsoft Entra). Log in through them separately, but with accounts that have the same email address. Then connect to the database and run `select * from users`. You should see two separate users with the same email address, one marked verified and one marked unverified.

sejubar:
- setup github auth https://www.loom.com/share/0681dd75305747f2b901168e6cd09fa4

- authenticate with github

- Setup Okta: https://www.loom.com/share/5c30d09286b54c52b43df7b7d1c8b39d

- authenticate with Okta

- check users in db and you should see two separate users with the same email address, one marked verified (github) and one marked unverified (Okta).

![Image](https://github.com/user-attachments/assets/d6a1c7ea-ab20-4a48-9533-71e7eec870da)

PR #1814, opened 2025-02-21
enhance: distinguish between verified and unverified email addresses

g-linville:
**Reviewers**: please double-check my logic in this, as well was my design choices surrounding migration that I outlined below.

for https://github.com/obot-platform/obot/issues/1807

This PR adds a distinction between users with a verified email address or an unverified email address. Google and GitHub verify email addresses, so we automatically treat those as verified, and everything else as unverified.

When checking to see if a new identity matches an existing user, we only look for an existing user if the new identity has a verified email address, and the existing user does too. This is the only way we can safely continue to match identities to users based on email addresses.

## Handling Migration

There is no automatic migration to do. The new `VerifiedEmail` field on the user starts out as null. We fill it in with either `true` or `false` depending on whether the first auth provider that the user uses to log in to it is verified or not.

All existing associations between an identity and a user are preserved.

If there is both a verified and an unverified identity associated with the user, the user's `VerifiedEmail` field will be marked `true` permanently as soon as they log in with the verified auth provider, but they will also still be able to log in with the unverified one, for the sake of backward compatibility. I don't know if this case actually exists anywhere out there, since the only unverified auth providers are enterprise only, but I wanted to account for this case anyway.

### DIFF
--- a/pkg/api/server/server.go
+++ b/pkg/api/server/server.go
@@ -69,6 +69,17 @@ func (s *Server) wrap(f api.HandlerFunc) http.HandlerFunc {
 		}
 
 		if !s.authorizer.Authorize(req, user) {
+			if _, err := req.Cookie("obot_access_token"); err == nil && req.URL.Path == "/api/me" {
+				// Tell the browser to delete the obot_access_token cookie.
+				// If the user tried to access this path and was unauthorized, then something is wrong with their token.
+				http.SetCookie(rw, &http.Cookie{
+					Name:   "obot_access_token",
+					Value:  "",
+					Path:   "/",
+					MaxAge: -1,
+				})
+			}
+
 			http.Error(rw, "forbidden", http.StatusForbidden)
 			return
 		}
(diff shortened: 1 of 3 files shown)


pkg/api/server/server.go
```diff
@@ -69,6 +69,17 @@ func (s *Server) wrap(f api.HandlerFunc) http.HandlerFunc {
...
+			if _, err := req.Cookie("obot_access_token"); err == nil && req.URL.Path == "/api/me" {
+				// Tell the browser to delete the obot_access_token cookie.
+				// If the user tried to access this path and was unauthorized, then something is wrong with their token.
+				http.SetCookie(rw, &http.Cookie{
+					Name:   "obot_access_token",
+					Value:  "",
+					Path:   "/",
+					MaxAge: -1,
+				})
+			}
```
g-linville, line 81: This is an unrelated change that I made just to help optimize things and avoid trying to verify an invalid token over and over and over again.

thedadams (approved):

drpebcak (approved):

PR #1839, opened 2025-02-24
fix: use username in user-ui if logged in with github

ivyjeong13:
Screenshot 2025-02-24 at 12 35 05 PM

* use username if logged in with github auth


ui/user/src/lib/components/navbar/Profile.svelte
```diff
@@ -15,9 +15,15 @@
 	async function load() {
 		credentials = await ChatService.listCredentials();
 	}
+
+	function getDisplayName(profile?: Profile) {
```
ryanhopperlowe: maybe this should be added as a getter for the profile store?

ryanhopperlowe (approved):

PR #1840, opened 2025-02-24
chore: gcp encryption: make healthz request

g-linville:
The encryption provider was not ready in time for when we needed to use it, leading to crashes. (At least, that is my theory.) This fix adds a retry loop to wait on the health check to pass before proceeding.

### DIFF
--- a/pkg/credstores/credstore.go
+++ b/pkg/credstores/credstore.go
@@ -3,9 +3,12 @@ package credstores
 import (
 	"context"
 	"fmt"
+	"io"
+	"net/http"
 	"os"
 	"os/exec"
 	"strings"
+	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/credstores/credstore.go
```diff
@@ -92,6 +95,7 @@ func setUpGoogleKMS(ctx context.Context, kmsKeyURI, configFile string) error {
 		"gcp-encryption-provider",
 		"--logtostderr",
 		"--path-to-unix-socket=/tmp/gcp-cred-socket.sock",
+		"--healthz-port=22222",
```
g-linville, line 98: 22222 should be out of whatever port range we use for daemon tools. Let me know if I should change this to something else.

drpebcak (approved):

njhale (approved):

Issue #1793, opened 2025-02-20
When changing color for obot icon , default agent icon is presented once before the colored icon is presented.


sangee2004:
**Step to reproduce the problem:**
1. Create an agent.
2. Set any obot icon for agent.
3. Use "Choose color" option to choose a color for obot icon.

Notice that when changing color for obot icon , default agent icon is presented once before the colored icon is presented.

https://www.loom.com/share/d05fbb5870f247f28b57e14634d92fa7?sid=2e0c3e08-9b32-4161-8652-9023e502a105

shweta8996:
I have tested this issue using the following steps, and color of icon is not retained when switching icons. :

- Create an agent.
- Set any obot icon for agent.
- Use "Choose color" option to choose 'yellow' color for obot icon.
- Notice icon will change to yellow.
- Now, change icon to something else than previously selected. 
- New icon will black in color and yellow color preference is not retained.

_Expected behavior:_
- Color preference should be retained when we switch icon.

shweta8996:
I have tested this issue using the following steps, and color of icon is not retained when switching icons. :

Create an agent.
Set any obot icon for agent.
Use "Choose color" option to choose 'yellow' color for obot icon.
Notice icon will change to yellow.
Now, change icon to something else than previously selected.
New icon will black in color and yellow color preference is not retained.
Expected behavior:

Color preference should be retained when we switch icon.

Please refer attachment:

https://github.com/user-attachments/assets/68553a98-871c-4b9d-a193-a10d86e6b64d

ivyjeong13:
@shweta8996 Realizing you're testing on test and I don't believe latest changes are deployed there, I'd use main.acornlabs!

shweta8996:
> [@shweta8996](https://github.com/shweta8996) Realizing you're testing on test and I don't believe latest changes are deployed there, I'd use main.acornlabs!

I tested the Admin UI by changing the icon and then modifying the color. After changing the icon again, the updated icon reflected the previous color change, but the front-end icon did not update correctly.

Please refer attachment

https://github.com/user-attachments/assets/16f03d15-924b-42fb-8643-82185fa6f5a5

ivyjeong13:
I don't think this valid anymore with the changes to user-ui and icon setting being handled on that end. Maybe appropriate action is to axe being able to set an icon on the admin side?

ivyjeong13:
[...]

PR #1842, opened 2025-02-24
fix: no fallback flash on agent icon change

ivyjeong13:
* avoid flashing of fallback agent icon (first char) when switching between icon logos/colors
note: tried `delayMs` but didn't seem to work even when bumping to 5secs

Addresses #1793


ryanhopperlowe (approved):
nit: should we fix this for all Avatar elements?

PR #1853, opened 2025-02-25
fix: oauth apps: fall back to manifest when credential not found

g-linville:


pkg/gateway/server/oauth_apps.go
```diff
@@ -343,15 +343,27 @@ func (s *Server) refreshOAuthApp(apiContext api.Context) error {
 		return apierrors.NewBadRequest("missing refresh_token query parameter")
 	}
 
+	var clientSecret string
+
 	// Reveal the credential to get the client secret.
 	cred, err := s.gptClient.RevealCredential(apiContext.Context(), []string{app.Name}, app.Spec.Manifest.Alias)
 	if err != nil {
-		return err
+		if errors.Is(err, gptscript.ErrNotFound{}) {
```
StrongMonkey, line 351:
```suggestion
                clientSecret := app.Spec.Manifest.ClientSecret
                cred, err := s.gptClient.RevealCredential(apiContext.Context(), []string{app.Name}, app.Spec.Manifest.Alias)
		if err != nil && !errors.Is(err, gptscript.ErrNotFound{}) {
		    return err
		 } else err == nil {
		    clientSecret = cred.Env["CLIENT_SECRET"]
		 }
		 
```

StrongMonkey: reduced some if else friction...

pkg/gateway/server/oauth_apps.go
```diff
@@ -442,16 +454,28 @@ func (s *Server) callbackOAuthApp(apiContext api.Context) error {
 		return apierrors.NewBadRequest("missing state query parameter")
 	}
 
+	var clientSecret string
+
 	// Reveal the credential to get the client secret.
 	cred, err := s.gptClient.RevealCredential(apiContext.Context(), []string{app.Name}, app.Spec.Manifest.Alias)
 	if err != nil {
-		return err
+		if errors.Is(err, gptscript.ErrNotFound{}) {
```
StrongMonkey, line 462: same suggestion above

StrongMonkey (approved):

PR #1891, opened 2025-03-04
feat: mark finished runs as inactive

thedadams:
Inactive runs will not be processed by the controller and will be removed from
its cache. These runs will still be accessible from the API.

This logic is implemented such that other objects can be marked as inactive and
the controller will similarly ignore them.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/obot-platform/obot
 
-go 1.23.4
+go 1.24.0
 
 replace (
 	github.com/obot-platform/obot/apiclient => ./apiclient
--- a/go.mod
+++ b/go.mod
@@ -23,7 +23,7 @@ require (
 	github.com/liggitt/tabwriter v0.0.0-20181228230101-89fcab3d43de
 	github.com/mhale/smtpd v0.8.3
 	github.com/obot-platform/kinm v0.0.0-20250116162656-270198b40c6d
-	github.com/obot-platform/nah v0.0.0-20250227230402-c8ac47e1add7
+	github.com/obot-platform/nah v0.0.0-20250305151239-35a77e89a2eb
 	github.com/obot-platform/namegenerator v0.0.0-20241217121223-fc58bdb7dca2
 	github.com/obot-platform/obot/apiclient v0.0.0-00010101000000-000000000000
 	github.com/obot-platform/obot/logger v0.0.0-20241217130503-4004a5c69f32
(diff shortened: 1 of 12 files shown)


go.mod
```diff
@@ -1,8 +1,9 @@
 module github.com/obot-platform/obot
 
-go 1.23.4
+go 1.24.0
 
 replace (
+	github.com/obot-platform/nah => github.com/thedadams/nah v0.0.0-20250304005118-f2cfb1983bf3
```
g-linville: just a reminder to remove this before merging

pkg/storage/apis/obot.obot.ai/v1/run.go
```diff
@@ -23,6 +24,11 @@ const (
 	AuthProviderSyncAnnotation  = "obot.ai/auth-provider-sync"
 )
 
+var (
+	_ fields.Fields = (*Run)(nil)
+	_ DeleteRefs    = (*Run)(nil)
+)
```
g-linville, line 30: What does this do?

g-linville (approved):

pkg/storage/apis/obot.obot.ai/v1/run.go
thedadams: It creates a compile-time check that `*Run` implements these interfaces.

pkg/controller/handlers/threads/threads.go
```diff
@@ -287,3 +289,23 @@ func (t *Handler) CleanupEphemeralThreads(req router.Request, _ router.Response)
...
+	// This must be uncached since inactive things aren't in the cache.
+	if err := req.List(untriggered.UncachedList(&runs), &kclient.ListOptions{
+		Namespace:     req.Namespace,
+		FieldSelector: fields.SelectorFromSet(map[string]string{"spec.threadName": req.Object.GetName()}),
+	}); err != nil {
+		return fmt.Errorf("failed to list runs for thread %s: %w", req.Object.GetName(), err)
+	}
+
+	for _, run := range runs.Items {
+		v1.SetActive(&run)
```
StrongMonkey: worth to check so that if it is already active, then don't update it?

thedadams: Yes, definitely.

thedadams: Fixed.

StrongMonkey (approved):

Issue #1345, opened 2025-01-20
Provide a way to see the full file name in the File view.


sangee2004:
**Steps to reproduce the proboem:**
1. Create an agent.
2. Access user ui (using publish url) and upload files with long file names to the workspace

Notice that the full file name is not displayed in File list view

<img width="432" alt="Image" src="https://github.com/user-attachments/assets/ba7c9320-f41d-486b-bcb6-730082067a9d" />

Hovering on the filename will also not provide the full file name.

**Expected Behavior:**
When hovering on the filename , user should be presented with the full file name.

shweta8996:
I have verified on main URL : https://main.acornlabs.com/o/p142zf4/t/t1d8xhs and it is working fine.
When uploading a long file name in the file section, the full name is displayed without truncation. Consider implementing text truncation or wrapping to improve UI readability.

Please refer attachment:

https://github.com/user-attachments/assets/4d8125cd-2290-4c61-904b-9dbd5cf36a64

PR #1898, opened 2025-03-04
chore: improve tooltip behavior in user ui

ryanhopperlowe:
Addresses #1345

- creates tooltip-specific abstraction
- creates reusable Truncate component
- adds truncation to necessary fields

Screenshot 2025-03-04 at 2 48 35 PM


ivyjeong13 (approved):
Wasn't able to test locally (I think cause I ran my local with the new huge changes), but looks good to me! :+1:

PR #1901, opened 2025-03-05
fix: broken syntax highlighting in generated code tags

ryanhopperlowe:
https://www.loom.com/share/98abe0bb878c40adb0ad89bc5913f0a1?sid=6a5ad7f4-07e0-464b-85c3-a3a688c48808

### DIFF
--- a/ui/user/src/lib/components/messages/Message.svelte
+++ b/ui/user/src/lib/components/messages/Message.svelte
@@ -73,7 +73,7 @@
 
 	$effect(() => {
 		// this is a hack to ensure the effect is run each time animatedText updates
-		void animatedText;
+		void markdownText;
 
 		const blocks = document.querySelectorAll('.message-content pre > code');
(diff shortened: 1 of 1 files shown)


ui/user/src/lib/components/messages/Message.svelte
```diff
@@ -72,10 +72,8 @@
 	});
 
 	$effect(() => {
-		// this is a hack to make sure this effect is run after the content is updated
-		if (content.length == 0) {
-			return;
-		}
+		// this is a hack to ensure the effect is run each time animatedText updates
+		void animatedText;
```
ibuildthecloud: do you need to watch markdownText instead?

ryanhopperlowe: that would probably be better. updated

ibuildthecloud (approved):

PR #1932, opened 2025-03-06
enhance: generate thread names based on first message

thedadams:
This change will generate a thread name for all user threads that don't
have names. This will be done by running a task that will generate a
short (1 to 5 words), identifying description for this conversation that
reflects the conversation's theme or tone.


pkg/storage/apis/obot.obot.ai/v1/thread.go
```diff
@@ -28,7 +28,7 @@ func (in *Thread) IsUserThread() bool {
 }
 
 func (in *Thread) IsProjectThread() bool {
-	return in.IsProjectBased() && in.Spec.Project
+	return in.Spec.Project
```
thedadams, line 31: `IsProjectBased()` also checks the `Spec.Project` field, so this was redundant.

ibuildthecloud (approved):

iwilltry42 (approved):

PR #1934, opened 2025-03-06
feat: add azure key vault encryption

g-linville:
See my video in Slack for a demo of this.

### DIFF
--- a/pkg/credstores/credstore.go
+++ b/pkg/credstores/credstore.go
@@ -18,6 +18,9 @@ import (
 type Options struct {
 	AWSKMSKeyARN         string
 	GCPKMSKeyURI         string
+	AzureKeyVaultName    string
+	AzureKeyName         string
+	AzureKeyVersion      string
 	EncryptionProvider   string
 	EncryptionConfigFile string
 }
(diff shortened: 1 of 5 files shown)


drpebcak (approved):

pkg/credstores/credstore.go
```diff
@@ -82,6 +94,45 @@ func Init(ctx context.Context, toolRegistries []string, dsn string, opts Options
...
+func setUpAzureKeyVault(ctx context.Context, keyvaultName, keyName, keyVersion, configFile string) error {
+	if keyvaultName == "" || keyName == "" || keyVersion == "" {
+		return fmt.Errorf("missing Azure Key Vault configuration")
+	}
+
+	if err := os.Setenv("GPTSCRIPT_ENCRYPTION_CONFIG_FILE", configFile); err != nil {
+		return fmt.Errorf("failed to set GPTSCRIPT_ENCRYPTION_CONFIG_FILE: %w", err)
+	}
+
+	if err := os.WriteFile("/tmp/azure.json", []byte(`{"useManagedIdentityExtension": true}`), 0600); err != nil {
```
thedadams, line 106: Does it make sense to use `os.CreateTemp` instead of create a named temporary file?

thedadams (approved):

pkg/credstores/credstore.go
g-linville: I think it's fine as it is. This is really only meant to be run in the container inside an Azure VM, so there's not gonna be a file already there.

PR #1965, opened 2025-03-07
chore: change the config setup via environment variables in the helm chart

iwilltry42:
By switching the naming in the values.yaml to the env var names, we can drop quite a bit of templating and have the generated secret and the existingSecret behave the same.

Slack Thread: https://acorn-io.slack.com/archives/C03FU91U4SX/p1740643210636069

issue: https://github.com/obot-platform/obot/issues/1966


chart/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v2
 name: obot
 description: A Helm chart for Obot
-version: 0.1.0
+version: 0.1.1
```
drpebcak: no need to update this, the version is set dynamically when we do a release of obot.

drpebcak (changes requested):

thedadams (approved):
It looks good after Taylor's comment is addressed.

drpebcak:
I'll merge this later after I update the terraform and stuff to account for the new values

Issue #1831, opened 2025-02-21
Allow for uploading  workspace files > 8MB size.


sangee2004:
We now supports for creating posts with audio/video files from workspace files - https://github.com/obot-platform/obot/issues/1682.

It would be good to be able to upload larger audio/video files to workspace files . Currently  we are allowed to upload a maximum of 8MB size files.

thedadams:
The limit was bumped to 100MB.

shweta8996:
I tested the functionality for Gmail drafts and emails using audio and video file attachments with sizes greater than 10MB, and it worked successfully. This confirms that the feature is functioning as expected.

PR #1979, opened 2025-03-07
enhance: allow 100MB workspace file uploads

thedadams:
Issue: https://github.com/obot-platform/obot/issues/1831

### DIFF
--- a/pkg/api/handlers/files.go
+++ b/pkg/api/handlers/files.go
@@ -71,7 +71,10 @@ func (f *FilesHandler) UploadFile(req api.Context) error {
 		return types.NewErrNotFound("no workspace found")
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/api/handlers/files.go
```diff
@@ -282,16 +285,6 @@ func convertFile(file, prefix string) types.File {
 	}
 }
 
-func uploadFile(ctx context.Context, req api.Context, gClient *gptscript.GPTScript, workspaceName string) error {
```
thedadams, line 285: This was only used in one spot, so I removed its contents to where it is being used to avoid confusion.

g-linville (approved):

ibuildthecloud:
Definitely a 10x engineer PR.

PR #1991, opened 2025-03-07
Slack integration with Obot

StrongMonkey:
This is the exact workflow/UX user will have to follow if they want to configure a custom app in their workspace and being able to delegate task to the task from every message that mentions the bot:

1. Use `/api/assistants/{assistant_id}/projects/{project_id}/slack/configure` to configure a custom oauth app for their obot and get a id. They need to create a slack app, configure the callback url to /api/slack/oauth/callback/{id}, set event redirect url to ` /api/slack/event` and all the right scopes.

2. Use `/api/assistants/{assistant_id}/projects/{project_id}/slack/authorize` to authorize. This will bind Obot with the workspace they have signed in. We will save the team id and name into the obot so that UI can create slack trigger based on the information. Slack trigger is a 1 to 1 mapping bewteen (workspace_id and app_id) to task.

3. Use `/api/slack-triggers` to create trigger to trigger task from your slack workspace.

The UX on the end user would be:
1. User develops their own slack app and configure that on the Obot level.
2. User autheticate slack with their app and install to their workspace.
3. User is able to trigger tasks by mention their own app in their workspace.
4. User will be able to do oauth with their own app using our slack/slack-bot tool.

image

https://github.com/obot-platform/obot/issues/2273


pkg/api/handlers/projects.go
```diff
@@ -673,6 +673,22 @@ func (h *ProjectsHandler) authenticate(req api.Context, local bool) (err error)
 		return err
 	}
 
+	var oauthApps v1.OAuthAppList
```
cjellick: what does this block of code do?

ui/user/vite.config.ts
```diff
@@ -8,7 +8,8 @@ export default defineConfig({
 			'/api': 'http://localhost:8080',
 			'/admin': 'http://localhost:8080',
 			'/oauth2': 'http://localhost:8080'
-		}
+		},
+		allowedHosts: ['04ac-70-184-112-91.ngrok-free.app']
```
cjellick: i assume this is a short term thing?

pkg/render/render.go
```diff
@@ -247,7 +247,7 @@ func OAuthAppEnv(ctx context.Context, db kclient.Client, oauthAppNames []string,
 
 	for _, integration := range slices.Sorted(maps.Keys(activeIntegrations)) {
 		app := activeIntegrations[integration]
-		integrationEnv := strings.ReplaceAll(strings.ToUpper(app.Spec.Manifest.Alias), "-", "_")
+		integrationEnv := strings.ReplaceAll(strings.ToUpper(string(app.Spec.Manifest.Type)), "-", "_")
```
cjellick, line 343: why this change?

pkg/api/handlers/slack.go
```diff
@@ -0,0 +1,262 @@
...
+	if err != nil {
+		return err
+	}
+
+	var app v1.OAuthApp
+	if err := req.Get(&app, getOauthAppFromThreadName(thread.Name)); err != nil {
+		return err
+	}
+
+	scopes := []string{
```
cjellick: should we let the obot author define scopes?

pkg/api/handlers/slackevent.go
```diff
@@ -0,0 +1,148 @@
...
+type SlackEventHandler struct {
+	invoker   *invoke.Invoker
+	gptscript *gptscript.GPTScript
+}
+
+func NewSlackEventHandler(invoker *invoke.Invoker, gptscript *gptscript.GPTScript) *SlackEventHandler {
+	return &SlackEventHandler{invoker: invoker, gptscript: gptscript}
+}
+
+type SlackEvent struct {
```
cjellick, line 32: in their docs, they make a big deal about verify thing token

https://api.slack.com/apis/events-api

> The shared-private callback token that authenticates this callback to the application as having come from Slack. Match this against what you were given when the subscription was created. If it does not match, do not process the event and discard it. Example: JhjZd2rVax7ZwH7jRYyWjbDl

Should we be checking that as well?

pkg/api/router/router.go
```diff
@@ -123,6 +124,14 @@ func Router(services *services.Services) (http.Handler, error) {
 	mux.HandleFunc("GET /api/assistants/{assistant_id}/projects/{project_id}/env", assistants.GetEnv)
 	mux.HandleFunc("PUT /api/assistants/{assistant_id}/projects/{project_id}/env", assistants.SetEnv)
 
+	// Project Slack integration
+	mux.HandleFunc("GET /api/assistants/{assistant_id}/projects/{project_id}/slack/authorize", projects.SlackAuthorize)
```
cjellick: where does this get used? do i drop it into slack somewher?

pkg/api/handlers/slack.go
```diff
@@ -0,0 +1,262 @@
...
+		"im:write",
+	}
+
+	// Construct the Slack OAuth URL
+	redirectURL := fmt.Sprintf("%s?client_id=%s&scope=%s&user_scope=%s&redirect_uri=%s",
+		slackOAuthURL,
+		app.Spec.Manifest.ClientID,
+		url.QueryEscape(strings.Join(scopes, ",")),
+		url.QueryEscape(strings.Join(userScopes, ",")),
+		url.QueryEscape(fmt.Sprintf("https://%s/api/slack/oauth/callback/%s", req.Host, app.Name)))
```
cjellick: need to use something other than req.host.

cjellick: theres an env var

pkg/api/handlers/slackevent.go
StrongMonkey: Yeah this is deprecrated in favor of using signing secret to verify the payload

image

StrongMonkey: Yeah this is deprecated. Using the sign secret is in favor of.

> This deprecated Verification Token can still be used to verify that requests come from Slack, but we strongly recommend using the above, more secure, signing secret instead.

pkg/api/handlers/slack.go
StrongMonkey: It should be removed. We don't really need scopes here, just need to have user go to the oauth flow and install the app to their workspace.

pkg/api/handlers/projects.go
```diff
@@ -370,6 +373,17 @@ func (h *ProjectsHandler) DeleteProject(req api.Context) error {
 		return err
 	}
 
+	if project.Spec.SlackConfiguration != nil {
+		var slackApp v1.OAuthApp
+		if err := req.Get(&slackApp, getSlackOauthAppFromThreadName(project.Name)); err != nil && !apierrors.IsNotFound(err) {
```
ibuildthecloud: add a spec.threadname field to oauthapp to create the relationship and then add  to the oauthapps deleterefs an entry so that the oauth app is deleted when the project/thread is deleted

pkg/api/handlers/projects.go
```diff
@@ -717,6 +731,10 @@ func (h *ProjectsHandler) authenticate(req api.Context, local bool) (err error)
 		return err
 	}
 
+	if err = h.OverrideOauthApps(req, &agent, thread); err != nil {
```
ibuildthecloud: I would expect that this was needed and this would just be done in the render.go

pkg/api/handlers/slack.go
```diff
@@ -0,0 +1,154 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	slackOAuthURL = "https://slack.com/oauth/v2/authorize"
+)
+
+func (p *ProjectsHandler) Configure(req api.Context) error {
```
ibuildthecloud: Can this be a whole new struct, SlackHandler

pkg/api/handlers/slack.go
```diff
@@ -0,0 +1,154 @@
...
+	thread, err := getThreadForScope(req)
+	if err != nil {
+		return err
+	}
+
+	var (
+		input struct {
```
ibuildthecloud: Can this be added as a struct in the types. I would call this `SlackReceiver`.  And this same object should be used for create/update.  User writable fields go into the manifest field (just follow the pattern of other types). and status and server controlled fields are outside of the manifest field. This makes it simple to know what can be modified by a user.  In this situation it might be all fields, but still just follow the pattern.

pkg/api/handlers/slack.go
```diff
@@ -0,0 +1,154 @@
...
+			ClientSecret  string `json:"clientSecret"`
+		}
+	)
+
+	if err := req.Read(&input); err != nil {
+		return err
+	}
+
+	var threads v1.ThreadList
+	if err := req.List(&threads, &client.ListOptions{
```
ibuildthecloud: I would actually create a real resource called SlackReciever.  Don't just embed all the info elsewhere. This way you can do this check but also not polute the other types like thread with this information that's not fully needed.  Then also some of this logic for create/update can be put in a controller.  The oauthapp can then be created by the controller. If the oauthapp has a spec.threadname field then you don't need to add the oauthapps field to the thread because you can just list for oauth apps associated to the thread.

pkg/render/render.go
ibuildthecloud: I think I'm on board with this change.

pkg/render/render.go
```diff
@@ -247,7 +247,7 @@ func OAuthAppEnv(ctx context.Context, db kclient.Client, oauthAppNames []string,
 
 	for _, integration := range slices.Sorted(maps.Keys(activeIntegrations)) {
```
ibuildthecloud, line 341: I would expect the render.go to look up all the oauthapp associated with the thread and then merge them with the agent level oauth apps.  This does mean you have to look at the "project thread", not the current thread, so the thread.spec.parentthreadname of the non-project thread.

pkg/api/handlers/slack.go
```diff
@@ -0,0 +1,143 @@
...
+		slackReceiver.Spec.Manifest.ClientSecret = ""
+		slackReceiver.Spec.Manifest.SigningSecret = ""
+
+		if err := req.Update(&slackReceiver); err != nil {
+			return err
+		}
+	}
+
+	oauthAppName := system.OAuthAppPrefix + thread.Name
+	credential := gptscript.Credential{
```
StrongMonkey, line 104: Notice that the cred name match the oauthapp name. This is so that we can reuse the credential for oauth app to store client secret and signing secret.

pkg/invoke/invoker.go
```diff
@@ -184,6 +184,7 @@ type Options struct {
 	CredentialContextIDs  []string
 	UserUID               string
 	GenerateName          string
+	ParentThreadName      string
```
StrongMonkey, line 187: If you create cred directly in ui, the api is called in a different route which don't have parent thread(project thread) available. So this is to pass the parent thread

pkg/api/handlers/slack.go
StrongMonkey: Changed to have a new resource type `SlackReciever`

pkg/api/handlers/slack.go
StrongMonkey: addressed.

pkg/api/handlers/slack.go
StrongMonkey: done

pkg/api/handlers/projects.go
StrongMonkey: this is moved to render to properly look for parent thread

pkg/render/render.go
StrongMonkey: addressed.

pkg/render/render.go
StrongMonkey: Yeah, since we have multiple oauth apps for a given type, we have to use type to deduplicate the oauth apps. Alias is no longer unique.

StrongMonkey:
The PR is largely overhaul to simplify the UX. The workflow will look like something now

1. User creates an app and call obot api to configure slack with necessary information.
2. User create a task and setup slack trigger.
3. Any mention for the app in their workspace will trigger the task.

Issue #1958, opened 2025-03-07
Admin UI - Top Level Obots Page


ryanhopperlowe:
We want to be able to view obots in the admin UI. So, i think we need a new Obots top level nav in the admin ui
You'll consume the /projects api for them. the only hitch is that right now if you hit /projects, you only get your projects. donnie is going to add a ?all=true query param (or something) that you will be able to use to list all of them in the admin ui
the weird/interesting things with these will be properly conveying the relationships between obots (projects) and agents.
You will have relationships like this:
Agent 1 -> Obot 1 -> Obot 2
(Obot 2 inherits from obot 1 and obot 1 inherits rom obot 2)

PR #2003, opened 2025-03-09
enhance: add created on column for obots table

ryanhopperlowe:
addresses #1958

Screenshot 2025-03-09 at 12 48 09 AM


pkg/api/handlers/projectshare.go
```diff
@@ -87,11 +87,15 @@ func (h *ProjectShareHandler) GetShare(req api.Context) error {
...
+		fields          = kclient.MatchingFields{}
+		all             = req.UserIsAdmin() && req.URL.Query().Get("all") == "true"
 	)
 
-	if err := req.List(&threadShareList, kclient.InNamespace(req.Namespace()), kclient.MatchingFields{
-		"spec.featured": "true",
-	}); err != nil {
+	if !all {
+		fields["spec.featured"] = "true"
+	}
```
thedadams, line 96: In my opinion, `ListFeatured` should list only the featured items. Can the UI use the other API to display all available?

thedadams (changes requested):

pkg/api/handlers/projectshare.go
ryanhopperlowe: discussed offline. Opted to rename method to `ListShares` instead of `ListFeatured`

thedadams (approved):
Go changes look good!

g-linville (approved):

ivyjeong13 (approved):

Issue #2007, opened 2025-03-10
Add APIs to support generating Obot icons from their descriptions


njhale:
Add some first-class APIs to support generating and retrieving Obot icons.

- The API should generate images with Google's Imagen3 model via the Gemini Developer SDK.
- Generated icons should be accessible via relative path in the frontend

shweta8996:
Please provide me the additional steps to replicate this issue.

njhale:
@shweta8996 Yeah, try generating some images for Obot icons using the UI. Check out the video below on how to do this based on their description and from a custom prompt. You should be able to see the APIs being called from your browser's developer console.

https://github.com/user-attachments/assets/db9660b2-80fe-4372-b961-19eb9f2a121c

shweta8996:
I have verified the Obot icon generation using the UI as per the provided video. Verified both description-based and custom prompt-generated icons. Also, checked the API calls in the browser's developer console to ensure requests and responses are functioning correctly. Everything is working as expected.
Please refer attachment:

https://github.com/user-attachments/assets/9aaf7bf4-7f6c-4184-abd7-279fa36d72eb

PR #2008, opened 2025-03-10
feat: add generated images api

njhale:
Add a first-class image generation API to enable generating and uploading custom icons for Obots.

This adds three new endpoints:
- `GET /api/image/{id}` returns the content of generated images by ID.
- `POST /api/image/generate` generates an image with Google's Imagen3
  model using the prompt body field, and returns the path that the image
  is served from (e.g. `/api/image/{id}`)
- `POST /api/image/upload` uploads a JPEG,  image with Google's Imagen3
  model using the prompt body field, and returns the path that the image
  is served from (e.g. `/api/image/{id}`)

In the user UI, this change:
- Adds support for generating custom Obot icons with both default and custom prompts
- Adds support for uploading custom light and dark mode icons
  individually
- Removes support for setting custom URLs for light and dark mode icons

Additional Icon Generation Notes:
- Image generation is disabled by default and is enabled by providing a `GEMINI_API_KEY`
  environment variable to Obot at startup
- The implementation uses the Gemini SDK instead of the new `Gemini
  Image Generator` tool for simplicity and speed
- Generated image IDs are UUIDv4
- There is currently no deletion API, but we'll eventually add a cron to
  cleanup orphaned images
- There is no owner attribution or metadata about the generated image
  (e.g. user, prompt, etc)
- All generated and uploaded images are stored and served in WEBP

Addresses https://github.com/obot-platform/obot/issues/2007


go.mod
```diff
@@ -13,9 +13,10 @@ require (
 	github.com/docker/docker v27.4.1+incompatible
 	github.com/dustin/go-humanize v1.0.1
 	github.com/fatih/color v1.18.0
+	github.com/gen2brain/webp v0.5.4
```
g-linville, line 16: It does not look like this library is well known or widely used. If you haven't done so already, I would suggest looking over its code for any obvious issues before we include it in Obot.

pkg/gateway/client/generatedimage.go
```diff
@@ -0,0 +1,41 @@
...
+		Data:     data,
+		MIMEType: mimeType,
+	}
+
+	err := c.db.WithContext(ctx).Create(img).Error
+	if err != nil {
+		return nil, err
+	}
+
+	return img, nil
```
g-linville: nit:

```suggestion
	return img, c.db.WithContext(ctx).Create(img).Error
```

pkg/gateway/client/generatedimage.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// GetGeneratedImage retrieves an image by its ID
+func (c *Client) GetGeneratedImage(ctx context.Context, id string) (*types.GeneratedImage, error) {
+	img := new(types.GeneratedImage)
+	if err := c.db.WithContext(ctx).Where("id = ?", id).First(img).Error; err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return nil, err
```
g-linville: I assume this was supposed to be this:

```suggestion
			return nil, nil
```

pkg/services/config.go
```diff
@@ -406,6 +411,19 @@ func New(ctx context.Context, config Config) (*Services, error) {
 		go smtp.Start(ctx, storageClient, config.EmailServerName)
 	}
 
+	var geminiClient *gemini.Client
+	if config.GeminiAPIKey != "" {
+		// Enable gemini-powered image generation
+		log.Warnf("using gemini API key from config.\nGEMINI_API_KEY=%q\nGEMINI_IMAGE_GENERATION_MODEL=%q", config.GeminiAPIKey, config.GeminiImageGenerationModel)
```
g-linville: Was this log just used for debugging? Doesn't seem like we need it, to me.

go.mod
njhale: yeah, it was difficult to find a non-cgo go package I could use to convert from png to webp. This one uses a WASM compilation of weblibp via wazero if the library isn't installed. AFACT, there's no overtly malicious code in it (as far as I and cursor could see skimming through it).

pkg/services/config.go
njhale: Wow, that DEFINITELY should not have been included. Thank you for catching that!

pkg/api/authz/ui.go
```diff
@@ -13,6 +13,7 @@ var uiResources = []string{
 	"GET /o/",
 	"GET /s/",
 	"GET /user/images/",
+	"GET /api/generated/images/{id}",
```
njhale: Not sure if this is 100% the correct authorizer to put this in.

pkg/api/authz/authz.go
```diff
@@ -60,6 +60,7 @@ var staticRules = map[string][]string{
 		"POST /api/prompt",
 		"GET /api/models",
 		"GET /api/version",
+		"POST /api/generated/images",
```
njhale: Is the `Authenticated` group the right group to put this in?

pkg/api/router/router.go
```diff
@@ -377,6 +377,10 @@ func Router(services *services.Services) (http.Handler, error) {
 	mux.HandleFunc("PUT /api/workflows/{id}", workflows.Update)
 	mux.HandleFunc("DELETE /api/workflows/{id}", workflows.Delete)
 
+	// Generated Images
+	mux.HandleFunc("POST /api/generated/images", generatedImages.GenerateImage)
+	mux.HandleFunc("GET /api/generated/images/{id}", generatedImages.GetGeneratedImage)
```
njhale: I don't particularly like that the DAO lives in gateway, but the routes live here. I did it this way to make the implementation simpler. That being said, having the APIs on gateway doesn't feel quite right either. Any suggestions?

chart/templates/deployment.yaml
```diff
@@ -170,6 +170,13 @@ spec:
                   name: {{ include "obot.config.secretName" . }}
                   key: workspaceProviderS3Bucket
             {{- end }}
+            {{- if .Values.config.geminiApiKey }}
```
drpebcak: Is this configurable inside the obot admin as well? I think we wanted to avoid having to set a ton of configuration as env vars if we could avoid it.

njhale: No it's not. Do we have any other configuration for non-tools-based features like this in the admin UI? I guess OAuth Apps, but it seems like a lot more moving pieces would be required to do this. I'll ask around to see what we want to do here.

ivyjeong13 (approved):

ivyjeong13:
Screenshot 2025-03-12 at 10 45 54 AM 🙌 UI side looks good to me!

chart/templates/deployment.yaml
cjellick: nick brought this up to me (or in a thread i was in). lets do it like this for now. we will eventually have to add an admin config screen. @njhale if we dont have one can you open an issue for this?

g-linville (approved):

chart/templates/deployment.yaml
drpebcak: i dont know the contents of the thread, but from a 'user' perspective we allow all the other llm/model provider config in the ui.

Issue #1977, opened 2025-03-07
Feature Request: Fine-Grained Tool Access in Obot Configuration


StrongMonkey:
This Feature includes changes in both the Admin UI and the User UI

Admin UI:
- admins should be able to select all combinations of tools.
- Selecting a bundle tool no longer removes granular tools, and the inverse is also removed.
  - bundles and granular tools are no longer mutually exclusive
- Because there will be so many more tools associated with a category, we need to group admin tools in the agent form to hide the mess

User UI
- We need to add a tool catalog to the obot edit side panel (under the tools section)
- it should display in a dialog with a similar ux to the _old_ Admin tool catalog
- one difference is that selecting a bundle will deselect all granular tools (unless the bundle is not included as an option)

sejubar:
I am picking this up

sejubar:
**Environment**:

https://main.acornlabs.com/

```
{
  "authEnabled": true,
  "chrome": "Chromium 134.0.6998.88",
  "dockerSupported": false,
  "emailDomain": "",
  "github.com/Azure/kubernetes-kms": "465705b",
  "github.com/gptscript-ai/datasets": "31ce6c3",
  "github.com/gptscript-ai/workspace-provider": "23e4267",
  "github.com/kubernetes-sigs/aws-encryption-provider": "91f20e6",
  "github.com/obot-platform/enterprise-tools": "cf9a61b",
  "github.com/obot-platform/k8s-cloudkms-plugin": "62f15f1",
  "github.com/obot-platform/tools": "fe3d276",
  "gptscript": "v0.9.5-f9c09f1",
  "obot": "v0.0.0-dev+8ed54c77"
}
```

PR #2063, opened 2025-03-12
enhance: update admin tool catalog list + add tool catalog to user ui

ryanhopperlowe:
addresses #1977

- Updates admin UI tool catalog to allow for combinations of both bundle tools with granular tools on an agent
- Adds a tool catalog (similar to admin UI) to the user UI

### DIFF
--- a/ui/user/src/lib/components/edit/ToolCatalog.svelte
+++ b/ui/user/src/lib/components/edit/ToolCatalog.svelte
@@ -75,15 +75,15 @@
 
 <div
 	use:catalog.tooltip
-	class="left-1/2 top-[15%] flex min-h-[500px] w-[500px] -translate-x-1/2 flex-col rounded-lg rounded-b-lg border border-surface3 bg-surface1 shadow-lg"
+	class="default-dialog left-1/2 top-[15%] flex min-h-[500px] w-[500px] -translate-x-1/2 flex-col"
 >
 	<div class="rounded-t-lg p-2">
 		<input
-			bind:this={input}
-			class="w-full rounded-lg p-2"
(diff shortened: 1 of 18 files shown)


ui/user/src/lib/components/edit/ToolCatalog.svelte
```diff
@@ -0,0 +1,195 @@
...
+				acc[tool.id] = { ...tool };
+				return acc;
+			}, {})
+	);
+
+	let canSelectMore = $derived(
+		Object.values(toolSelection).filter((t) => t.enabled).length < MAX_TOOLS
+	);
+
+	let catalog = popover({ fixed: true });
```
ibuildthecloud, line 43: for "fixed" I've just been using <dialog> elements. You get escape behavior for free.  Is using a popover library better somehow?

ryanhopperlowe: that makes sense. will remove

ryanhopperlowe: Primary thing is that it adds the backdrop which catches clicks outside of the popover

ryanhopperlowe: I spent a bunch of time fighting with `dialog` to see if I could get that to work but figured that was better saved for post-release

cjellick (changes requested):
Can you drop the words "Bundle Selected" when I click the top level tool? just have no text in that scenario. I dont really want the users to know the word "Bundle" its kind of an implementation detail

cjellick (approved):

PR #2073, opened 2025-03-12
chore: docs add azure kms

cloudnautique:
Adding docs for the KMS providers based.

### DIFF
--- a/docs/docs/50-configuration/06-encryption-providers.md
+++ b/docs/docs/50-configuration/06-encryption-providers.md
@@ -78,7 +78,7 @@ az role assignment create \
 
 ### 5. Get the Key Version
 
-1. Get the key version (optional, latest version will be used if not specified):
+1. Get the key version:
 
 ```bash
 az keyvault key show \
--- a/docs/docs/50-configuration/06-encryption-providers.md
+++ b/docs/docs/50-configuration/06-encryption-providers.md
@@ -1,0 +1,102 @@
+# Encryption Providers
+
+## Setting up Google Cloud KMS Encryption Provider
+
+## Setting up Azure Key Vault Encryption Provider
+
+This guide explains how to set up an Azure VM with a managed identity that can access an Azure Key Vault for encryption operations.
+
+### Prerequisites
+
+- Azure CLI installed and logged in (`az login`)
+- A subscription with permissions to create resources
+
+### 1. Create Resource Group
+
+```bash
+# Create a resource group
+az group create \
+  --name your-resource-group \
+  --location eastus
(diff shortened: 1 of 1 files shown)


cloudnautique:
@drpebcak do you have docs for what we would need to do to configure GCP? I can take a pass at adding them here.

docs/docs/50-configuration/06-encryption-providers.md
```diff
@@ -0,0 +1,102 @@
...
+# Assign Key Vault Crypto User role
+az role assignment create \
+  --role "Key Vault Crypto User" \
+  --assignee-object-id $IDENTITY_OBJECT_ID \
+  --scope $KEYVAULT_ID
+```
+
+### 5. Get the Key Version
+
+1. Get the key version (optional, latest version will be used if not specified):
```
g-linville: This is my mistake in the docs that I gave you from the LLM. This actually isn't optional.

```suggestion
1. Get the key version:
```

docs/docs/50-configuration/06-encryption-providers.md
```diff
@@ -0,0 +1,102 @@
...
+  --name your-keyvault-name \
+  --resource-group your-resource-group \
+  --location eastus \
+  --enable-rbac-authorization true
+
+# Create an RSA key in the Key Vault
+az keyvault key create \
+  --name your-key-name \
+  --vault-name your-keyvault-name \
+  --size 2048 \
```
g-linville, line 37: Bigger key for better security.

```suggestion
  --size 4096 \
```

drpebcak:
> @drpebcak do you have docs for what we would need to do to configure GCP? I can take a pass at adding them here.

I do not.. I think it should work just fine on GCE, but we never tested it there. I only ever set it up in GKE.

docs/docs/50-configuration/06-encryption-providers.md
```diff
@@ -0,0 +1,102 @@
...
+  --kty RSA
```
drpebcak, line 38: RSA is generally bad.. We should recommend something like `EC`

drpebcak (changes requested):

docs/docs/50-configuration/06-encryption-providers.md
g-linville: (We can change `--kty` to `EC` and remove the `--size` parameter and that will work.

drpebcak: I retract this statement.. looks like EC doesnt work for this.

PR #2111, opened 2025-03-13
enhance: reorganize UI components and improve copy ux

njhale:
- Add BETA badge to navbar logo
- Move dark mode toggle from navbar to profile dropdown menu
- Add external links (Docs, Discord, GitHub) to navbar
- Simplify icon management UI with combined upload/remove buttons
- Improve custom prompt handling in GenerateIcon with reactive updates
- Reorganize home page to show "Recently Used" and "My Obots" sections
- Sort Obots by creation date (newest first)
- Redirect to Obot edit page on copy
- Update AssistantIcon to use placeholder image as fallback
- Conditionally show delete option based on project type


cjellick (approved):
Conditional approval .... the external links need to open on a new page

ui/user/src/lib/components/navbar/Profile.svelte
```diff
@@ -10,11 +12,29 @@
...
+				}}
+				role="menuitem"
+				class="icon-button"
+			>
+				{#if darkMode.isDark}
+					<Sun class="h-5 w-5" />
+				{:else}
+					<Moon class="h-5 w-5" />
+				{/if}
+			</button>
```
ryanhopperlowe, line 28: what was wrong with the original dark mode toggle component?

ui/user/src/routes/+page.svelte
```diff
@@ -91,26 +90,26 @@
...
+						{#if toolData}
+							<ToolPill tool={toolData} />
+						{/if}
 					{/each}
+					{#if projectShare.tools.length > 3}
+						<ToolPill
+							tools={projectShare.tools
+								.slice(3)
+								.map((t) => tools.get(t))
+								.filter((t): t is ToolReference => !!t)}
```
ryanhopperlowe, line 106: maybe reformat to be `.map().filter().slice(3)` so that there will always be 3 if available?

With this there is a chance where a share has 3 valid tools, but only displays 2 because 1 got prematurely chopped off

ryanhopperlowe (approved):

ui/user/src/routes/+page.svelte
njhale: I pulled this directly from the home page snippet. So I think we should fix all at once in a follow up and potentially factor it out into a shared component

ui/user/src/lib/components/navbar/Profile.svelte
njhale: It's now the only place it's used and the original didn't have `role="menuitem"`

PR #2128, opened 2025-03-14
chore: immediately read runstate then start interval

ibuildthecloud:


pkg/events/events.go
```diff
@@ -327,7 +330,12 @@ func (e *Emitter) printRun(ctx context.Context, state *printState, run v1.Run, r
 					}
 				}
 			}
-		case <-ticker.C:
+		case <-tick:
+			if ticker == nil {
+				// now wait every second for new events
+				ticker = time.NewTicker(time.Second)
```
thedadams, line 336: I think we need to stop the ticker to release the resources and not leak.

Issue #1978, opened 2025-03-07
"File wasn't available on site" error when attempting to download files from the Obot.ai file section


shweta8996:
Steps to Reproduce:

- Navigate to the Obot.ai File Section.
- Upload files (e.g., John_Doe_Resume.docx and cover_letter.txt).
- Attempt to download any file from the available list.
- Observe that a red error message appears stating:
- "File wasn't available on site."

Expected Result:
 Files should download successfully when clicking the download button.

Actual Result:
Attempting to download files results in an error: "File wasn't available on site."

Please refer attachment:

![Image](https://github.com/user-attachments/assets/e7eaa364-f6a8-4585-b18f-a2323a0994c1)

thedadams:
It looks like this functionality has changed. I am able to download files from the site. I previously was not able.

shweta8996:
I have verified on obot.ai and it is not working fine. I performed on below the steps:
Navigate to the Obot.ai File Section.
Upload files (e.g., John_Doe_Resume.docx and cover_letter.txt).
Attempt to download any file from the available list.
Observe that a red error message appears stating:
"File wasn't available on site."
Please refer attachment:

![Image](https://github.com/user-attachments/assets/d097c70b-d9f1-4d74-a914-21b84566eae3)

thedadams:
@shweta8996 Hoe are you creating these files? If you are uploading them, can you give the steps you are using to upload them? Also, can you provide one of these files here?

g-linville:
https://www.loom.com/share/8efc890e5c4c4e86b5a703b24dc915fa

I'm moving this to Testing because as far as I can tell, this issue is gone. I can't reproduce it at all. I can download files generated by the LLM as well as files that I upload to the thread.

g-linville:
I figured out what the issue was today, and it was fixed in #2093

shweta8996:
> I figured out what the issue was today, and it was fixed in [#2093](https://github.com/obot-platform/obot/pull/2093)

@g-linville The issue still persists when you try to download the images that are broken in chat but available in files section.

![Image](https://github.com/user-attachments/assets/55a94a38-23f9-4803-90eb-5a205e0666a3)
You can try downloading images in the below obot thread.
https://obot.ai/o/p1zg686?thread=t15vxsx

g-linville:
Sorry about this. I think I found the real issue. It appears to be a bug in how the user UI is generating the download URL.

g-linville:
just merged the fix

[...]

PR #2135, opened 2025-03-14
fix: authz: fix access to threads

g-linville:
I'm not 100% sure if this is the proper fix, but I think it is. I'll make sure Darren sees it.

for https://github.com/obot-platform/obot/issues/1978


g-linville:
I have updated this PR with the changes that Darren asked, and tested to make sure it still works. Ready for review.

thedadams (approved):

drpebcak (approved):

PR #2141, opened 2025-03-14
feat: add streamlined tooltips to user ui

ryanhopperlowe:
https://www.loom.com/share/48c459d6e51649cdb010fe84183b7daf?sid=3a1a28d1-395d-4f3b-8852-d57a946e6887

Additionally fixes issue where long file names completely break width of page when editor is open

### DIFF
--- a/ui/user/src/lib/actions/overflow.ts
+++ b/ui/user/src/lib/actions/overflow.ts
@@ -1,23 +1,27 @@
 import popover from './popover.svelte.js';
+import type { Placement } from '@floating-ui/dom';
 
 function hasOverflow(element: HTMLElement) {
 	return element.scrollHeight > element.clientHeight || element.scrollWidth > element.clientWidth;
 }
 
-export function overflowToolTip(node: HTMLElement) {
+export function overflowToolTip(
+	node: HTMLElement,
+	{ placement = 'top', tooltipClass }: { placement?: Placement; tooltipClass?: string } = {}
+) {
 	const { ref, tooltip } = popover({
-		placement: 'top',
+		placement,
 		offset: 4,
 		hover: true
 	});
 
-	node.classList.add('text-nowrap', 'overflow-hidden', 'text-ellipsis', 'w-full');
+	node.classList.add('line-clamp-1', 'break-all');
 
-	const span = document.createElement('p') as HTMLSpanElement;
-	span.classList.add('tooltip');
-	span.textContent = node.textContent;
+	const p = document.createElement('p');
+	p.classList.add('tooltip-text', 'break-all', ...(tooltipClass?.split(' ') ?? []));
+	p.textContent = node.textContent;
 
-	node.insertAdjacentElement('afterend', span);
+	node.insertAdjacentElement('afterend', p);
 	node.addEventListener('mouseenter', (e) => {
 		if (!hasOverflow(node)) {
 			e.stopImmediatePropagation();
@@ -25,6 +29,6 @@ export function overflowToolTip(node: HTMLElement) {
 	});
 
 	// Register after the above event listener to ensure we can stop propagation
-	tooltip(span);
+	tooltip(p);
 	ref(node);
 }
--- a/ui/user/src/lib/actions/overflow.ts
+++ b/ui/user/src/lib/actions/overflow.ts
@@ -1,5 +1,6 @@
 import popover from './popover.svelte.js';
 import type { Placement } from '@floating-ui/dom';
+import { twMerge } from 'tailwind-merge';
 
 function hasOverflow(element: HTMLElement) {
 	return element.scrollHeight > element.clientHeight || element.scrollWidth > element.clientWidth;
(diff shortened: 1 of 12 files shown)


ui/user/src/lib/actions/overflow.ts
```diff
@@ -1,30 +1,37 @@
...
+	{
+		placement = 'top',
+		tooltipClass,
+		offset = 4
+	}: { placement?: Placement; tooltipClass?: string; offset?: number } = {}
+) {
+	const { ref, tooltip } = popover({ placement, offset, hover: true });
 
-	node.classList.add('text-nowrap', 'overflow-hidden', 'text-ellipsis', 'w-full');
+	node.classList.add('line-clamp-1', 'break-all');
```
ibuildthecloud, line 19: removing `w-full` will probably break existing things.  The reason `w-full` was in there was that I typically hit situations where overflow would not happen because the containing element didn't have a fixed with. Also I hit subtle differences between line-clamp-1 vs text-ellipsis. It think it had to do with cutting words in half. I really don't know if line-clamp-1 is better.  But for sure removing w-full just make sure anything that already used this method doesn't break.

ui/user/src/lib/actions/overflow.ts
```diff
@@ -1,30 +1,37 @@
...
 
-	const span = document.createElement('p') as HTMLSpanElement;
```
ibuildthecloud, line 16: lol, I created a `p` and cast it to `span` :man_facepalming:

ibuildthecloud (commented):
I just modified/add a class `tooltip` in app.css to try to make the tool tips consistent. Can you use/modify that, or let me know if thats a bad idea.  I just noticed that all the tooltips had a different style.

ui/user/src/lib/components/tasks/Step.svelte
```diff
@@ -110,7 +110,7 @@
 			class="flex-1 resize-none border-none bg-gray-50 outline-none dark:bg-gray-950"
 		></textarea>
 		<div class="flex gap-2 p-2">
-			<button class="rounded-lg p-2 hover:bg-gray-50 dark:hover:bg-gray-950" onclick={doRun}>
+			<button class="icon-button-small" onclick={doRun}>
```
cjellick: ivys revamping tasks. maybe better to not touch it so as to not create a conflict for her?

ui/user/src/lib/actions/overflow.ts
ibuildthecloud: oh I think the difference between line-clamp-1 and text-ellipsis I saw was that "foo bar baz" would become "foo bar..." vs "foo bar b..." if there was space for the b. So in my situation I never wanted a line 2 (like never line -clamp-2 or -3) so what was the point of losing the extra characters.

ryanhopperlowe:
> I just modified/add a class `tooltip` in app.css to try to make the tool tips consistent. Can you use/modify that, or let me know if thats a bad idea. I just noticed that all the tooltips had a different style.

I saw that. I think I have a change that does what you're asking in another branch

ui/user/src/lib/actions/overflow.ts
ryanhopperlowe: disucssed offline... we both agree line-clamp sucks, but not using it will cause more issues than we'd like to fix today. Will revisit after the release

ui/user/src/lib/components/Thread.svelte
```diff
@@ -119,10 +123,10 @@
 <div class="relative w-full max-w-[900px] pb-32">
 	<!-- Fade text in/out on scroll -->
 	<div
-		class="absolute inset-x-0 top-0 z-30 h-14 w-full bg-gradient-to-b from-white dark:from-black"
+		class="absolute inset-x-0 top-0 z-10 h-14 w-full bg-gradient-to-b from-white dark:from-black"
```
cjellick, line 126: what does this change do?

ryanhopperlowe: this was causing the fade filter to cover the profile dropdown

Screenshot 2025-03-14 at 9 53 10 PM

ui/user/src/lib/components/Thread.svelte
```diff
@@ -184,30 +188,35 @@
 				<!-- Vertical Spacer -->
 			</div>
 		</div>
-		<div class="absolute inset-x-0 bottom-0 z-30 flex justify-center py-8">
-			<div class="w-full max-w-[1000px]">
```
cjellick, line 188: why drop this div?

ui/user/src/lib/components/edit/Files.svelte
```diff
@@ -1,5 +1,14 @@
 <script lang="ts">
-	import { FileText, Trash, Upload, X } from 'lucide-svelte/icons';
+	import { overflowToolTip } from '$lib/actions/overflow';
```
cjellick, line 2: its hard to tell what of this is reodering and what is meaningful changes. if doing purely cosmetic code changes, its good to break them out into a separate commit (or just not do them at all) to keep the surface area of meaningful changes small

cjellick (approved):

ui/user/src/lib/components/edit/Files.svelte
ryanhopperlowe: Yeah the formatter sorts imports automatically when I hit save. I'll see about making the formatter a bit less invasive

ui/user/src/lib/components/Thread.svelte
ryanhopperlowe: Hmm... I wonder if this specific diff was pulled in from rebasing, because I thought this was something I fixed earlier. Either that or it's something that I broke here and then fixed up in the next PR. in any case it might be worth pushing in the second PR as a single one.

ryanhopperlowe: Sorry the scope of my PRs have been super out of wack the past few days. I'll do better about keeping them simple

ryanhopperlowe:
Changes merged via #2158  closing this PR

PR #2147, opened 2025-03-14
Fix: change task's alias, matching alias from email

StrongMonkey:
I changed the task.alias to be a short version of 12 character, this will be used for webhook and email trigger. 

Make the email address from task prefixed with task name to be more user friendly, and in the backend we will only match the receiver that match the alias in the last part.

demo: https://www.loom.com/share/071ea961f7d74e34bf99de4f6dadf0b9

### DIFF
--- a/pkg/api/handlers/tasks.go
+++ b/pkg/api/handlers/tasks.go
@@ -704,6 +704,10 @@ func (t *TaskHandler) CreateFromScope(req api.Context) error {
 		return fmt.Errorf("failed to generate alias: %w", err)
 	}
 
+	if len(workflowManifest.Alias) > 12 {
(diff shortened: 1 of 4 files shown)


pkg/api/handlers/tasks.go
```diff
@@ -704,6 +704,10 @@ func (t *TaskHandler) CreateFromScope(req api.Context) error {
 		return fmt.Errorf("failed to generate alias: %w", err)
 	}
 
+	if len(workflowManifest.Alias) > 12 {
```
StrongMonkey, line 707: I can only shorten the alias for email receiver, but let me know if this is going to cause problems. Making it consistent for other things.

ibuildthecloud (approved):

thedadams (approved):

Issue #2047, opened 2025-03-11
Task threads should not show the intro message at the top


cjellick:
you can see in this task thread, i see the (very bad) intro message and starter message buttons

<img width="1363" alt="Image" src="https://github.com/user-attachments/assets/299b78dd-e2a5-4865-a818-4d55d4290b1d" />

shweta8996:
I don't see this issue in obot.ai. Can you please provide where can i test this issue?

shweta8996:
> I don't see this issue in obot.ai. Can you please provide where i can test this issue?

This was a misunderstanding, I re-tested this in obot.ai and I no longer view the starter messages or unnecessary texts in Task Threads.
Sorry for the confusion caused.

PR #2150, opened 2025-03-14
fix: don't show obot icons, intros, or starter messages for task runs

njhale:
Addresses https://github.com/obot-platform/obot/issues/2047

### DIFF
--- a/ui/user/src/lib/components/Thread.svelte
+++ b/ui/user/src/lib/components/Thread.svelte
@@ -26,9 +26,10 @@
 		project: Project;
 		tools: AssistantTool[];
 		version: Version;
+		isTaskRun?: boolean;
 	}
 
-	let { id = $bindable(), project, version, tools }: Props = $props();
+	let { id = $bindable(), project, version, tools, isTaskRun }: Props = $props();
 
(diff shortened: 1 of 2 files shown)


ui/user/src/lib/components/Thread.svelte
```diff
@@ -25,9 +25,10 @@
 		project: Project;
 		tools: AssistantTool[];
 		version: Version;
+		isTaskRun?: boolean;
 	}
 
-	let { id = $bindable(), project, version, tools }: Props = $props();
+	let { id = $bindable(), project, version, tools, isTaskRun = $bindable(false) }: Props = $props();
```
ryanhopperlowe: does this work if `isTaskRun` is not bindable? it seems like this doesn't do anything here

ryanhopperlowe (approved):

cjellick (approved):

ui/user/src/lib/components/Thread.svelte
njhale: ah, good catch. I'll remove it and see how it works out.

njhale: removed

PR #2173, opened 2025-03-15
fix: chat messages spilling over

ryanhopperlowe:
addresses #2174

It looks like this change got lost/reverted somewhere along the way.

This prevents the chat contents from spilling over and filling the container when the viewport is larger

Before:
Screenshot 2025-03-15 at 10 24 53 AM

After:
Screenshot 2025-03-15 at 10 25 17 AM


cjellick:
going to close this. ivy is revamping some stuff and this would slightly conflict with this. it should be fixed in her PR

Issue #2118, opened 2025-03-14
add Recently Used and My Obots to drop down naviation


cjellick:
Now that we have these two distinctions on the home page, we should have them in the obot navigation dropdown as well.

They should do the same thing as threads - show 10 and then a "load more" button to load more if there are any

shweta8996:
I have verified that obot.ai is working fine. On the home page, there are two sections, "My obot" and "Recently used," which are also reflected in the obot navigation dropdown. The dropdown functions similarly to threads, displaying 10 items initially, with a "See More" button appearing if additional items are available in the Thread section. This design ensures a seamless navigation experience, making it easier for users to access relevant items without cluttering the UI.
Please refer attachment:

https://github.com/user-attachments/assets/e65987c3-e3ef-411a-b0a9-c982ef4deffd

PR #2180, opened 2025-03-15
enhance: split obot nav into recently used and my obots sections

njhale:
Addresses https://github.com/obot-platform/obot/issues/2118


ivyjeong13 (approved):
:+1: Looks good to me. Only nit is visually I think the headers might be a little difficult to tell at first glance that they're headers

Screenshot 2025-03-15 at 5 48 26 PM

Quickly fiddled and maybe something like this? Or if the font sizes for the header and the project name(s) were different, that'd help!

Screenshot 2025-03-15 at 5 58 13 PM

Issue #2095, opened 2025-03-13
make button styles consistent


cjellick:
here's several different styles, lets pick one

![Image](https://github.com/user-attachments/assets/925621eb-f136-44d0-992a-6e1b76ae139c)

shweta8996:
I have verified on obot.ai and I see two different styles for Upload and Website button. Is this expected from UI perspective? 
I also noticed that Share button does not have share icon in the button whereas upload and website has respective icons. 
Please refer attachment:

![Image](https://github.com/user-attachments/assets/00a1d505-186a-436e-a707-b1ddd11fea96)

![Image](https://github.com/user-attachments/assets/bb212abc-39d7-4e8d-a023-4fe654195a6f)

shweta8996:
> > I have verified on obot.ai and I see two different styles for Upload and Website button. Is this expected from UI perspective? I also noticed that Share button does not have share icon in the button whereas upload and website has respective icons. Please refer attachment:
> > ![Image](https://github.com/user-attachments/assets/00a1d505-186a-436e-a707-b1ddd11fea96)
> > ![Image](https://github.com/user-attachments/assets/bb212abc-39d7-4e8d-a023-4fe654195a6f)
> 
> [@shweta8996](https://github.com/shweta8996) I don't think these changes have been pushed to prod/obot.ai. Have you checked main/test?

Let me check with main/test and get back to you.

shweta8996:
I tested on main at https://main.acornlabs.com/o/p1hm7ch?thread=t1jrwxs. 

Styling are consistent now for various buttons like Upload, Website, etc on left side menu. 

Please refer attached video. 
https://github.com/user-attachments/assets/721d3d18-a5b5-4523-b90c-57fe690df28e

PR #2195, opened 2025-03-17
fix: button consistency & editor stylings

ivyjeong13:
Addresses #2095 

* consistent buttons in editor mode
* some styling updates for editor (font sizes, editing image)

https://www.loom.com/share/97b34e6b90504d2681cd85fe5aeb3da0

### DIFF
--- a/ui/user/src/lib/actions/reactiveLabel.svelte.ts
+++ b/ui/user/src/lib/actions/reactiveLabel.svelte.ts
@@ -9,7 +9,8 @@ export const reactiveLabel: Action<HTMLElement, ReactiveLabelParams> = (node, pa
 	// Add default classes that don't change
 	node.classList.add('origin-top', 'overflow-hidden', 'transition-all', 'duration-200');
 
-	function update({ value, height = 'h-6' }: ReactiveLabelParams) {
+	$effect(() => {
+		const { value, height = 'h-6' } = params;
 		// Remove previous height class if it exists
 		node.classList.remove('h-0', height);
 		// Remove previous transform classes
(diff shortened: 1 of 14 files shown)


ui/user/src/lib/actions/reactiveLabel.svelte.ts
```diff
@@ -0,0 +1,35 @@
...
+		// Remove previous opacity class
+		node.classList.remove('opacity-0');
+
+		// Add appropriate classes based on value
+		if (value) {
+			node.classList.add(height, 'translate-y-0');
+		} else {
+			node.classList.add('h-0', 'translate-y-full', 'opacity-0');
+		}
+	}
```
ryanhopperlowe: I'm starting to think that adding/removing classes via svelte actions might be an anti-pattern (possibly better handled via components or inline logic). Something we should take up with @ibuildthecloud

ryanhopperlowe (approved):

ui/user/src/lib/actions/reactiveLabel.svelte.ts
ryanhopperlowe: definitely not a blocking comment ^^ feel free to ignore haha

ibuildthecloud: Oh, @ivyjeong13, I didn't even think about the fact that there is a update method on actions.  But's it's actually deprecated somewhat now.  More info here: https://svelte.dev/docs/svelte/use, but the short of it is that you can just use `$effect()` and then things are super straight forward.

ibuildthecloud: since you can do effect in an action I don't see much down side to using them and I really like the pattern as it's nicely composable with raw HTML.  And if not obvious $effect will be more performant than update() or any other dom listener.

Issue #2203, opened 2025-03-17
Remove Bundle Tools from Admin UI tool catalog


ryanhopperlowe:
Hide the concept of bundle tools from the admin ui. They should be added and removed behind the scenes

ryanhopperlowe:
After going through a few scenarios, I found some issues with this change as a whole and had a discussion with @will-chan. There's obviously more discussion to be had on this so we're going to wait for @cjellick to get back to propose the idea of removing bundle tools entirely.

PR #2204, opened 2025-03-17
enhance: hide bundle tools from admin users and auto add/remove as necessary

ryanhopperlowe:
Addresses #2203 

Updated this PR to serve as a POC for removing bundle tools entirely from the Admin UI

### DIFF
--- a/ui/admin/app/components/agent/ToolForm.tsx
+++ b/ui/admin/app/components/agent/ToolForm.tsx
@@ -265,6 +265,8 @@ export function ToolForm({
 	);
 
 	function renderToolEntry(tool: ToolReference) {
+		if (tool.bundle) return null;
+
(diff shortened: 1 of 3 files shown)


ivyjeong13 (approved):

ryanhopperlowe:
After going through a few scenarios, I found some issues with this change as a whole and had a discussion with @will-chan. There's obviously more discussion to be had on this so we're going to wait for @cjellick to get back to propose the idea of removing bundle tools entirely.

ryanhopperlowe:
nit: you could prevent the layout shift when the scrollbar opens by setting `scrollbar-gutter: stable` 

https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter

PR #2214, opened 2025-03-18
feat: add drag n drop upload state to chat input

ryanhopperlowe:
https://www.loom.com/share/2a0e29c0345c454f8938193ca17859a5?sid=9ca8c879-cc45-4f2a-bee8-a499a06363bf

### DIFF
--- a/ui/user/src/lib/components/Thread.svelte
+++ b/ui/user/src/lib/components/Thread.svelte
@@ -21,6 +21,9 @@
 	import AssistantIcon from '$lib/icons/AssistantIcon.svelte';
 	import { responsive } from '$lib/stores';
 	import { Bug } from 'lucide-svelte';
+	import { errors } from '$lib/stores';
+	import { twMerge } from 'tailwind-merge';
+	import { dragstate } from '$lib/actions/dragstate.svelte';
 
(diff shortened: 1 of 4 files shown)


ivyjeong13:
I was wondering does it have to be in the message input or could they drop it anywhere in the chat? Think ChatGPT allows soon as it comes into viewport so was curious!

ivyjeong13 (approved):

ryanhopperlowe:
Here's what that would look like
https://www.loom.com/share/28534b94248e4f9d8a83222d6262e80b?sid=efb3cb10-acc8-41a0-9180-21ff598023d8

ivyjeong13:
> Here's what that would look like https://www.loom.com/share/28534b94248e4f9d8a83222d6262e80b?sid=efb3cb10-acc8-41a0-9180-21ff598023d8

Feel like that's much easier to allow a user to drop files in!

ui/user/src/lib/components/Thread.svelte
ryanhopperlowe, line 1: This file really just has additions for dragging and upload logic.

The diff went crazy because I moved the wrapping div from `Obot.svelte` into here

PR #2217, opened 2025-03-18
enhance: input details in thread

ivyjeong13:
https://www.loom.com/share/2cbec5dfa1bf42ed8947b4e5bf438b94

* move Input details into thread
* clean up json preview of Output


ui/user/src/lib/components/messages/Message.svelte
```diff
@@ -265,17 +304,47 @@
 				toolTT.toggle();
 			}}
 		>
-			<X class="h-4 w-4" />
```
ibuildthecloud, line 268: i love that you replace every h-x w-x with size-x :)

PR #2235, opened 2025-03-19
feat: add user activities

thedadams:
Authenticated users are not tracked for last activity, updated at most once in a 24-hour period.


ui/admin/app/lib/model/primitives.ts
```diff
@@ -11,6 +11,7 @@ export type EntityMeta<
 	links?: TLinks;
 	metadata?: TMetadata;
 	type?: string;
+	lastActiveTime?: string; // date
```
ryanhopperlowe: Is this available on every single object type? If not can you just add it to the User type in `~/lib/model/users.ts`

thedadams: Yeah, I didn't pay close enough attention to that. Sorry about that.

ryanhopperlowe: All good. That's what I'm here for :)

ryanhopperlowe (approved):

njhale (approved):

Issue #2100, opened 2025-03-13
Add a couple buttons to obot responses


cjellick:
In chat, add the following pieces of functionality at the end of any message sent by the obot:
- copy - just copy the entire message to your clip board
- edit - open the entire message as a file in our file editor

sejubar:
**Evironment**
https://main.acornlabs.com/

```
{
  "authEnabled": true,
  "chrome": "Chromium 134.0.6998.88",
  "dockerSupported": false,
  "emailDomain": "",
  "github.com/Azure/kubernetes-kms": "465705b",
  "github.com/gptscript-ai/datasets": "31ce6c3",
  "github.com/gptscript-ai/workspace-provider": "23e4267",
  "github.com/kubernetes-sigs/aws-encryption-provider": "91f20e6",
  "github.com/obot-platform/enterprise-tools": "799f996",
  "github.com/obot-platform/k8s-cloudkms-plugin": "62f15f1",
  "github.com/obot-platform/tools": "322e905",
  "gptscript": "v0.9.5-31ce029",
  "obot": "v0.0.0-dev+87806433"
}
```

PR #2241, opened 2025-03-19
feat: add copy and edit buttons to obot responses

njhale:
Add copy and edit buttons to chat messages that allow users to copy Obot responses to the clipboard and edit them as a file.

Addresses https://github.com/obot-platform/obot/issues/2100

https://github.com/user-attachments/assets/96290daf-3b81-4fb1-a3ad-4a0c6dd18a39

### DIFF
--- a/ui/user/src/lib/components/messages/Message.svelte
+++ b/ui/user/src/lib/components/messages/Message.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
 	import MessageIcon from '$lib/components/messages/MessageIcon.svelte';
-	import { FileText, Pencil } from 'lucide-svelte/icons';
+	import { FileText, Pencil, Copy, Edit } from 'lucide-svelte/icons';
 	import { Tween } from 'svelte/motion';
 	import { ChatService, type Message, type Project } from '$lib/services';
 	import highlight from 'highlight.js';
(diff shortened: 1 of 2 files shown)


ui/user/src/lib/components/messages/Message.svelte
```diff
@@ -183,6 +185,35 @@
...
+			if (!fileExists) {
+				const file = new File([content], filename, { type: 'text/plain' });
+				await ChatService.saveFile(project.assistantID, project.id, file, {
+					threadID: currentThreadID
+				});
+			}
+
+			onLoadFile(filename);
+		} catch (err) {
+			console.error('Failed to create or open file:', err);
```
njhale, line 220: should these errors get sent to an error component?

ui/user/src/lib/components/messages/Message.svelte
```diff
@@ -479,8 +510,37 @@
 				{#if msg.sent}
 					{@render time()}
 				{/if}
+
+				{#if !msg.sent && msg.done && !msg.toolCall && msg.time && !animating}
+					{@const copyTT = popover({ hover: true, placement: 'bottom' })}
+					{@const editTT = popover({ hover: true, placement: 'bottom' })}
```
njhale: Is there a better way to add nice tool-tips?

ui/user/src/lib/components/messages/Message.svelte
```diff
@@ -49,6 +52,9 @@
 	let showToolInputDetails = $state(false);
 	let showToolOutputDetails = $state(false);
 
+	let copyTT = popover({ hover: true, placement: 'bottom' });
+	let editTT = popover({ hover: true, placement: 'bottom' });
```
njhale, line 56: Is there a better way to add nice-looking tool-tips for these buttons?

ui/user/src/lib/components/messages/Message.svelte
ivyjeong13: There is an `errors` in store that you can push into thatll show up as snackbar messages. I can't recall if it doesn't already handle those calls in the ChatService though. 

edit: http.ts `handleResponse` does this! It pushes into `errors` and those show up under Notifications.svelte. There is a `dontLogErrors` opts you can use so you can push this error through instead. I think this error is probably more user-friendly for the user.

ivyjeong13 (approved):

cjellick (approved):

PR #2256, opened 2025-03-22
bug: dialogs are not centered by default

ibuildthecloud:
Quick fix to make dialogs centered by default.  Maybe that's bad, but I'm guessing something changed between tailwind 3 and 4 regarding this.

image


ibuildthecloud:
Oh yes, this is the case. And wow I feel so smart because I did the exact right fix :)

https://tailwindcss.com/docs/upgrade-guide#dialog-margins-removed

PR #2280, opened 2025-03-24
chore: use built tools image

drpebcak:
requires https://github.com/obot-platform/tools/pull/527 and https://github.com/obot-platform/enterprise-tools/pull/40

### DIFF
--- a/.github/workflows/build-tools.yaml
+++ b/.github/workflows/build-tools.yaml
@@ -14,6 +14,9 @@ jobs:
   enterprise-build:
     if: github.event.repository.full_name == 'obot-platform/enterprise-tools' || github.event.repository.full_name == 'obot-platform/tools'
     runs-on: depot-ubuntu-22.04
+    concurrency:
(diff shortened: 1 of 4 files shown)


.github/workflows/build-tools.yaml
```diff
@@ -0,0 +1,86 @@
...
+          repository: 'obot-platform/obot'
+
+      - name: Log in to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ secrets.GHCR_USERNAME }}
+          password: ${{ secrets.GHCR_TOKEN }}
+
+      - name: Build and push enterprise-tools Docker image
```
njhale:
```suggestion
      - name: Build and push oss-tools Docker image
```

njhale (approved):
One nit, but looks like this should work.

thedadams (approved):

Issue #2248, opened 2025-03-21
Files aren't saved on edit in the user UI


njhale:
**Describe the bug**
Edits made to files through the user UI sometimes aren't persisted.

**To Reproduce**
1. Go to chat with an Obot
2. Upload a file (plaintext or markdown)
3. Select the file to open the file editor
4. Make a change to the file
5. Close the file editor
6. Open the file editor for the file again
7. Sometimes the change made in step 4 isn't present

**Expected behavior**
Edits to files made through the user UI should always be saved to the file

**Screenshots**

https://github.com/user-attachments/assets/426806a1-e6cf-43d3-8d97-5e142b0b2c45

PR #2283, opened 2025-03-25
fix: file editor not saving when file contents changed

ivyjeong13:
Addresses #2248 
* when onFileChanged is fired, need to call saveContents

### DIFF
--- a/ui/user/src/lib/components/Editors.svelte
+++ b/ui/user/src/lib/components/Editors.svelte
@@ -4,6 +4,7 @@
 	import FileEditors from '$lib/components/editor/FileEditors.svelte';
 	import { getLayout } from '$lib/context/layout.svelte';
 	import { ChatService, EditorService, type InvokeInput, type Project } from '$lib/services';
+	import type { EditorItem } from '$lib/services/editor/index.svelte';
 	import { Download } from 'lucide-svelte';
 	import { X } from 'lucide-svelte/icons';
 	import { twMerge } from 'tailwind-merge';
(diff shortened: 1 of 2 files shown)


ibuildthecloud:
There's a very small nuance that nobody would have known about that needs to be addressed here.  The modified bit and buffer field should only be reset here: https://github.com/obot-platform/obot/blob/c7ac341b83977c96bf652dc4c32af024d590a078/ui/user/src/lib/components/messages/Input.svelte#L59

The basic idea is that the modified flag means that we have modified the file without AI knowing about it.  So on the next chat input we tell the AI that we changed the file.  This isn't perfect, but kind of best effort because various things like refreshing the page will mess it up.  I think the existing `download()` logic was already wrong in that it reset the modified bit.

When tell the LLM that we modified a file we actually send a special json input as the chat input.  You can see the input when you look at the thread in the admin UI.

ivyjeong13:
> There's a very small nuance that nobody would have known about that needs to be addressed here. The modified bit and buffer field should only be reset here:
> 
> https://github.com/obot-platform/obot/blob/c7ac341b83977c96bf652dc4c32af024d590a078/ui/user/src/lib/components/messages/Input.svelte#L59
> 
> The basic idea is that the modified flag means that we have modified the file without AI knowing about it. So on the next chat input we tell the AI that we changed the file. This isn't perfect, but kind of best effort because various things like refreshing the page will mess it up. I think the existing `download()` logic was already wrong in that it reset the modified bit.
> 
> When tell the LLM that we modified a file we actually send a special json input as the chat input. You can see the input when you look at the thread in the admin UI.

Taking a look and will update! Thanks Darren!

ibuildthecloud (approved):

PR #2287, opened 2025-03-25
fix: be able to scroll outside chat well

ivyjeong13:
* some restructuring to get user to be able scroll outside the chat well

https://github.com/user-attachments/assets/30a034a4-1af7-4baa-91cb-8963ff29bc0e

### DIFF
--- a/ui/user/src/lib/components/Obot.svelte
+++ b/ui/user/src/lib/components/Obot.svelte
@@ -78,7 +78,7 @@
 				</Navbar>
 			</div>
 
-			<div class="flex h-[calc(100%-76px)] max-w-full grow">
+			<div class="relative flex h-[calc(100%-76px)] max-w-full grow">
 				{#if !responsive.isMobile || (responsive.isMobile && !layout.fileEditorOpen)}
 					{#if layout.editTaskID && layout.tasks}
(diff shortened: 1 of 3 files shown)


ui/user/src/lib/components/Obot.svelte
```diff
@@ -96,14 +96,12 @@
 							{/if}
 						{/each}
 					{:else}
-						<div id="main-input" class="flex h-full max-w-full flex-1 justify-center">
```
ibuildthecloud, line 99: The main-innput ID was used here. Either that was already broken, it this will break. But the basic 
 idea was to focus the thread input when we switch or create threads from the sidebar. 
https://github.com/obot-platform/obot/blob/1d6764b1d1c6acf067c1d160cd728038574025db/ui/user/src/lib/components/sidebar/Threads.svelte#L81

ivyjeong13: Oh yes! I thought I moved it into Thread since that querySelector should be able to find it anywhere. Thank you, will add that back in!

ivyjeong13:
@ibuildthecloud Addressed! : )

ibuildthecloud (approved):

Issue #2140, opened 2025-03-14
Support the PAT or Oauth cred paradigm in user ui


cjellick:
Some tools give you the choice of authing via oauth or a personal access token. right now, we only let you make that choice in the admin-ui at the base agent level. this choice probably needs pushed down to the obot level

sangee2004:
User is now prompted to pick PAT or Oauth cred when chatting with agent that has tools that support both authentication mode.

User is also prompted to pick  PAT or Oauth cred when setting credentials for the tools (that support both PAT and oauth) at the obot level.

Few issues relating to this feature:

https://github.com/obot-platform/obot/issues/2340
https://github.com/obot-platform/obot/issues/2367
https://github.com/obot-platform/obot/issues/2369

PR #2289, opened 2025-03-26
Enhance: add ability to choose PAT token using user ui

StrongMonkey:
This PR adds support to choose PAT/Oauth in user ui. it change the oauth app config from agent to thread/project. 

It also migrated existing thread/project to inherit existing oauth apps in agent, and will be using thread's oauth app to determine whether it is doing oauth or PAT.

https://github.com/obot-platform/obot/issues/2140

https://www.loom.com/share/0a151cf5b67648449b92b05fd201ec21


pkg/controller/handlers/threads/threads.go
```diff
@@ -601,3 +601,37 @@ func (t *Handler) CopyTasksFromParent(req router.Request, _ router.Response) err
 
 	return nil
 }
+
+func MigrateOauthAppsFromParent(ctx context.Context, c kclient.Client) error {
```
StrongMonkey, line 605: @thedadams I did change so that we use thread's oauthapp by default and in order to not break existing project, we have to migrate the oauthapp from agent to project.

apiclient/types/thread.go
```diff
@@ -52,4 +52,5 @@ type ThreadManifest struct {
 	StarterMessages      []string          `json:"starterMessages"`
 	SharedTasks          []string          `json:"sharedTasks,omitempty"`
 	WebsiteKnowledge     *WebsiteKnowledge `json:"websiteKnowledge,omitempty"`
+	OauthApps            *[]string         `json:"oauthApps,omitempty"`
```
ibuildthecloud, line 55: Why is this nil-able?

StrongMonkey:
doing a different approach now.

Issue #1982, opened 2025-03-07
User UI: thread gets aborted when hitting enter on a prompt


g-linville:
https://www.loom.com/share/03dbbd29c3774ba5b1a24854860546bc

cjellick:
dangit...this was fixed previously. i guess its back

StrongMonkey:
should be fixed here https://github.com/obot-platform/obot/pull/2315

g-linville:
yeah I can confirm that this doesn't happen anymore. just gonna close it

PR #2315, opened 2025-03-27
Fix: fix an issue where aborted message is duplicated

StrongMonkey:
This fixed an issue where aborted message is duplicated if you cancel the run. 

https://github.com/obot-platform/obot/issues/1982


StrongMonkey:
probably fixed by https://github.com/obot-platform/obot/pull/2317

StrongMonkey:
solution is not good.

Issue #2314, opened 2025-03-27
Non-admins can't delete credentials


njhale:
**Describe the bug**
Non admins get a 403 error when attempting to delete a credential.

**To Reproduce**
1. Go to the user UI
2. Add a tool that requires authentication (e.g. Gmail)
3. Ask a question to trigger an OAuth prompt for the tool and authenticate
4. Open the credentials menu and attempt to delete the credential

**Expected behavior**
Non admins should be able to delete their credentials.

**Screenshots**

<img width="1080" alt="Image" src="https://github.com/user-attachments/assets/673819a8-63b8-444f-87d1-35c88773c251" />

njhale:
Note: This issue should have also affected admins, so we need to verify that works as well.

shweta8996:
Link: https://obot.ai/o/p1fjcm7?edit

Error: redirect_uri_mismatch

When I try to add Gmail credentials, i am getting 400 redirect_uri_mismatch from gmail. 
Please refer attached video. 

https://github.com/user-attachments/assets/f8a0390d-39ec-47ee-8405-f6d0e69999da

njhale:
> Link: https://obot.ai/o/p1fjcm7?edit
> 
> Error: redirect_uri_mismatch
> 
> When I try to add Gmail credentials, i am getting 400 redirect_uri_mismatch from gmail. Please refer attached video.
> 
>  Recording.171.mp4

@shweta8996 The error you're encountering seems like a recent (probably unrelated) regression. You should be able to work around it by creating a credential for a non-oauth app and then deleting that.

shweta8996:
I have verified on https://test.acornlabs.com and logged with okta(non-admin user). non admin user is able to delete the Gmail credential.
Please refer attachment:

https://github.com/user-attachments/assets/2e53553f-b75f-408a-8f3d-bdc814d4a289

PR #2316, opened 2025-03-28
fix: allow deauthentication of non-custom tools

njhale:
Update the custom tools authorization logic to pass requests when a tool
resource is not found. This allows authorization for deauthentication requests on
non-custom tools (which don't have an associated Tool resource) to
succeed.

Addresses: https://github.com/obot-platform/obot/issues/2314

Note: I'm not sure if this is the _correct_ fix.


pkg/api/authz/tool.go
```diff
@@ -2,14 +2,17 @@ package authz
...
 	"github.com/obot-platform/nah/pkg/router"
 	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
+	"github.com/obot-platform/obot/pkg/system"
 	"k8s.io/apiserver/pkg/authentication/user"
 )
 
 func (a *Authorizer) checkTools(req *http.Request, resources *Resources, _ user.Info) (bool, error) {
-	if resources.ToolID == "" {
+	// Skip this auth check if the request isn't for a custom tool
+	if resources.ToolID == "" || !strings.HasPrefix(resources.ToolID, system.ToolPrefix) {
```
thedadams, line 15: If this isn't a custom tool, does it make sense to get the `ToolReference` and check to see if that tool is available for the agent or obot?

njhale: I'm not sure. This was the solution Darren suggested here, but you may have a point. IIUC, the current implementation would allow you to call deauthenticate (as well as other endpoints) for tools that may not be available on the agent/obot. AFAICT, this was the original behavior before custom tools went in, so may have been a pre-existing gap.

thedadams: The endpoint you're fixing here also allows a comma-separated list of tools, too. So this is probably the right thing for now.

thedadams (approved):

PR #2321, opened 2025-03-28
fix: abort on chat ui

ivyjeong13:
Screenshot 2025-03-28 at 1 27 16 PM

* fix abort on chat-ui


ui/user/src/lib/constants.ts
```diff
@@ -1,2 +1,5 @@
 export const DEFAULT_PROJECT_NAME = 'My Obot';
 export const DEFAULT_PROJECT_DESCRIPTION = 'Do more with AI';
+
+export const ABORTED_THREAD_MESSAGE = 'thread was aborted, cancelling run';
```
StrongMonkey, line 4: Oh you are using the old message. I will revert back my change :)

StrongMonkey (approved):

Issue #2331, opened 2025-03-31
OAuth for github and slack tool is broken due to incorrect redirect url.


sangee2004:
**Describe the bug**
Not able to authenticate with github and slack tool.

**To Reproduce**
1. Create a new obot using "+ Create New Obot"
2. Add github and slack tool to this obot.
3. Chat with this agent about github issues.
4. When following the "Click here" option for authentication , user is presented with following error page

<img width="1219" alt="Image" src="https://github.com/user-attachments/assets/ae236d02-feaf-44f1-8e79-81412411fa1d" />

Similar error is also seen when authenticating with slack tool.

<img width="1114" alt="Image" src="https://github.com/user-attachments/assets/3bff2fd2-bc92-4413-9a23-035d7e97c303" />

**Expected behavior**
Should be  able to authenticate with github and slack tool successfully.

sangee2004:
This issue is not seen anymore when testing with latest commit on https://obot.ai/

PR #2334, opened 2025-03-31
Fix: Look up oauth app in a right way

StrongMonkey:
Fix the oauth app to look it up in a correct way.

https://github.com/obot-platform/obot/issues/2331

### DIFF
--- a/pkg/gateway/server/oauth_apps.go
+++ b/pkg/gateway/server/oauth_apps.go
@@ -746,5 +746,21 @@ func convertOAuthAppRegistrationToOAuthApp(app v1.OAuthApp, baseURL string) type
 
 func getOAuthAppFromName(apiContext api.Context) (*v1.OAuthApp, error) {
 	var oauthApp v1.OAuthApp
-	return &oauthApp, apiContext.Get(&oauthApp, apiContext.PathValue("id"))
+	id := apiContext.PathValue("id")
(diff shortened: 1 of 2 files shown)


pkg/gateway/server/oauth_apps.go
```diff
@@ -746,5 +746,21 @@ func convertOAuthAppRegistrationToOAuthApp(app v1.OAuthApp, baseURL string) type
...
+			Namespace: apiContext.Namespace(),
+		}); err != nil {
+			return nil, err
+		}
+		if len(oauthApps.Items) == 0 {
+			return nil, fmt.Errorf("oauth app %s not found", id)
+		}
+		return &oauthApps.Items[0], nil
+	}
+	return &oauthApp, apiContext.Get(&oauthApp, id)
```
thedadams, line 765: Can we use this: https://github.com/thedadams/obot/blob/otel/pkg/alias/get.go#L19?

StrongMonkey: darren removed the alias in oauthapp so we can't really use that.

StrongMonkey: I think we can remove the alias later. This is for backwards compabities so we don't have to update existing oauth apps to use id instead of alias

thedadams (approved):

PR #2335, opened 2025-03-31
feat: add support for audit logging

thedadams:


pkg/api/server/audit/store/s3.go
```diff
@@ -0,0 +1,77 @@
...
+	if options.AuditLogsStoreS3Bucket == "" {
+		return nil, errors.New("Audit log store S3 bucket is required")
+	}
+
+	cfg, err := config.LoadDefaultConfig(context.Background())
+	if err != nil {
+		return nil, err
+	}
+
+	client := s3.NewFromConfig(cfg, func(o *s3.Options) {
```
StrongMonkey, line 38: I assume the requirement is to have aws env variable setup?

StrongMonkey (approved):

pkg/api/server/audit/store/s3.go
thedadams: Yeah, the standard `AWS_` environment variables are required.

StrongMonkey (approved):

pkg/api/server/audit/store/s3.go
```diff
@@ -0,0 +1,77 @@
...
+
+func NewS3Store(host string, compress bool, options S3StoreOptions) (Store, error) {
+	if options.AuditLogsStoreS3Bucket == "" {
+		return nil, errors.New("Audit log store S3 bucket is required")
+	}
+
+	cfg, err := config.LoadDefaultConfig(context.Background())
+	if err != nil {
+		return nil, err
+	}
```
g-linville, line 36: Will this cause problems in situations where we're using an alternative S3 provider? Like if I had my own MinIO instance or something, would this still look for the AWS_* variables, and return an error if there are none?

g-linville (approved):

pkg/api/server/audit/store/s3.go
thedadams: That's the way the workspace-provider works now, I believe, and this uses the same ideas.

PR #2341, opened 2025-03-31
Encrypt-run-states

thedadams:
RunStates can have some personally identifiable information in them depending on
tool calls. This change encrypts the sensitive parts of the run state object.

Signed-off-by: Donnie Adams <donnie@acorn.io>

### DIFF
--- a/aws-encryption.yaml
+++ b/aws-encryption.yaml
@@ -3,9 +3,12 @@ kind: EncryptionConfiguration
 resources:
   - resources:
       - credentials
+      - runstates.obot.obot.ai
     providers:
       - kms:
           apiVersion: v2
           name: aws-kms
           endpoint: unix:///tmp/aws-cred-socket.sock
           timeout: 3s
+      - identity: {} # this fallback allows reading unencrypted secrets;
+        # for example, during initial migration
--- a/pkg/gateway/client/runstate.go
+++ b/pkg/gateway/client/runstate.go
@@ -70,11 +70,7 @@ func (c *Client) DeleteRunState(ctx context.Context, namespace, name string) err
 }
 
 func (c *Client) encryptRunState(ctx context.Context, runState *types.RunState) error {
-	if c.encryptionConfig == nil {
-		return nil
-	}
-	transformer := c.encryptionConfig.Transformers[gr]
-	if transformer == nil {
+	if c.transformer == nil {
 		return nil
 	}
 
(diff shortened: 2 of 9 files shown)


StrongMonkey (approved):

pkg/gateway/client/runstate.go
```diff
@@ -51,3 +73,60 @@ func (c *Client) DeleteRunState(ctx context.Context, namespace, name string) err
...
+			errs = append(errs, err)
+		} else {
+			runState.Output = append(b, encryptionSuffix...)
+		}
+		if b, err = transformer.TransformToStorage(ctx, runState.CallFrame, nil); err != nil {
+			errs = append(errs, err)
+		} else {
+			runState.CallFrame = append(b, encryptionSuffix...)
+		}
+		if b, err = transformer.TransformToStorage(ctx, runState.ChatState, nil); err != nil {
```
g-linville: I don't think we can pass `nil` as the third argument to `TransformToStorage` here, at least, not safely. There is this `Context` interface that needs to be fulfilled. See https://github.com/obot-platform/tools/blob/782f6c8145b292c90fe37b5edcea7ddf2199334d/credential-stores/pkg/common/db.go#L16-L25 and the code link in the comment there for an example. I don't know all the technical details behind this, but I want to follow k8s' example as much as we can.

g-linville (changes requested):

pkg/gateway/client/runstate.go
thedadams: From the documentation:
>// AuthenticatedData should return an array of bytes that describes the current value. If the value changes,
	// the transformer may report the value as unreadable or tampered. This may be nil if no such description exists
	// or is needed. For additional verification, set this to data that strongly identifies the value, such as
	// the key and creation version of the stored data.

It's fine to be `nil` for implementation purposes. Not sure if there is some other reason why we should set it to something else, though. And not sure what that something else should be.

g-linville (approved):

thedadams:
I made a change to the way unencrypted data is handled, so rerequesting reviews.

g-linville (approved):

aws-encryption.yaml
```diff
@@ -3,9 +3,12 @@ kind: EncryptionConfiguration
...
       - credentials
+      - runstates.obot.obot.ai
     providers:
       - kms:
           apiVersion: v2
           name: aws-kms
           endpoint: unix:///tmp/aws-cred-socket.sock
           timeout: 3s
+      - identity: {} # this fallback allows reading unencrypted secrets;
+        # for example, during initial migration
```
thedadams, line 14: @g-linville I would appreciate your input on this. What do you think?

g-linville: fine with me!

g-linville (approved):

pkg/gateway/client/runstate.go
```diff
@@ -51,3 +68,57 @@ func (c *Client) DeleteRunState(ctx context.Context, namespace, name string) err
...
+	if err != nil {
+		errs = append(errs, err)
+	}
+	runState.ChatState, _, err = transformer.TransformFromStorage(ctx, runState.ChatState, nil)
+	if err != nil {
+		errs = append(errs, err)
+	}
+
+	return errors.Join(errs...)
+}
```
njhale: nit: would be interesting to know how much this impacts performance at some point.

njhale (approved):
One more LGTM for good measure

PR #2343, opened 2025-03-31
chore: remove the concept of "inactive runs"

thedadams:
We realized that this can't work the way we need it to work, so are removing it.


StrongMonkey (approved):
LGTM, what is the exact reason that we have to revert it back?

njhale (approved):

chart/values.yaml
```diff
@@ -73,9 +73,6 @@ config:
   # config.KINM_DB_CONNECTIONS -- Advanced - the number of connections in the database pool for kinm
   KINM_DB_CONNECTIONS: "5"
 
-  # config.OBOT_SERVER_IGNORE_INACTIVE -- whether the controller processes objects that have been marked as inactive
-  OBOT_SERVER_IGNORE_INACTIVE: false
-
```
thedadams, line 78: @drpebcak Will removing this have adverse effects on the terraform?

thedadams:
https://github.com/obot-platform/terraform/pull/2 is required before this can be merged.

chart/values.yaml
drpebcak: should be fine, but we can remove it from there as well

PR #2346, opened 2025-04-01
chore: share postgres DSN with auth providers for session storage

g-linville:


pkg/api/handlers/authprovider.go
```diff
@@ -221,6 +230,42 @@ func (ap *AuthProviderHandler) Deconfigure(req api.Context) error {
...
+		if tablePrefix != "" {
+			if err := db.Exec("DROP TABLE IF EXISTS " + tablePrefix + "sessions;").Error; err != nil {
+				return fmt.Errorf("failed to drop sessions table: %w", err)
+			}
+			if err := db.Exec("DROP TABLE IF EXISTS " + tablePrefix + "session_locks;").Error; err != nil {
+				return fmt.Errorf("failed to drop session_locks table: %w", err)
+			}
+		}
+	}
+
```
njhale: Feels like this is low-level implementation details bleeding up. Is it possible to move this into the auth providers themselves? maybe have them expose a `destroy` tool that does this instead?

If not, we should at least make the table prefix an field on the provider meta so we don't need to maintain this list here.

njhale (changes requested):

pkg/api/handlers/authprovider.go
g-linville: Yeah I'll just add it to the metadata. This needs to be a function outside of the auth providers themselves, in order to safely ensure that the tables get dropped.

g-linville: Fixed

njhale (approved):

pkg/api/handlers/authprovider.go
```diff
@@ -143,6 +149,10 @@ func (ap *AuthProviderHandler) Configure(req api.Context) error {
 	}
 	envVars[providers.CookieSecretEnvVar] = cookieSecret
 
+	if ap.postgresDSN != "" {
+		envVars[providers.PostgresConnectionEnvVar] = ap.postgresDSN
+	}
```
thedadams, line 154: nit: Is there any harm in always including this?

thedadams (approved):

pkg/api/handlers/authprovider.go
g-linville: I don't think so? I guess I could remove the if statement. Doesn't really make a difference.

PR #2361, opened 2025-04-02
enhance: add rate limiting

njhale:
## Description

Implement middleware that applies rate limits to all HTTP endpoints exposed by Obot.

## Request Flow

1. A request is sent to any HTTP endpoint
2. The request immediately encounters the rate limiting middleware
3. The middleware inspects the request for any headers or cookies that appear
   in its recently authenticated users cache (LRU with entry TTL and automatic eviction)
4. The request is checked against the appropriate rate limiter based on
   the result of the cache lookup
   - If it contained a matching credential for an admin, no rate limits
     are applied to the request
   - If a non-admin entry was found, authenticated rate limits (200
     req/sec) are applied to the user ID associated with the cached
     credential
   - If the credential wasn't found in the cache, unauthenticated rate
     limits are applied to the client's IP address (100 req/sec)
   - If a limit is exceeded a 429 response code is returned with headers
     indicating when the client can make their next request
5. The request is sent through the authentication chain
   - If the request is successfully authenticated, the respective authenticator
     adds the hash of the credential and the info necessary to extract it from
     the request to the user info embedded in its response
6. After authentication is complete, the credential hash and lookup info
   are added to the rate limiting middleware's user cache
   - No entries are added to the cache if the request doesn't pass
     authentication
   - This step enables the middleware to roughly (and inexpensively)
     determine what limits should be applied to future requests with the
     same credential (steps 3 and 4)

## Justifying the Design

This approach should help reduce the amount of high-frequency traffic that can reach the more resource intensive operations performed by the authenticator chain (e.g. auth provider proxy and database token lookups) while still enabling higher rate limits for authenticated users and
admins.

Addresses: https://github.com/obot-platform/sensitive-issues/issues/18

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,6 +22,7 @@ require (
 	github.com/gptscript-ai/cmd v0.0.0-20250324222528-f16f18548238
 	github.com/gptscript-ai/go-gptscript v0.9.6-0.20250331192455-415de950d72d
 	github.com/gptscript-ai/gptscript v0.9.6-0.20250328194144-ce3b7262ed0e
+	github.com/hashicorp/golang-lru/v2 v2.0.7
 	github.com/liggitt/tabwriter v0.0.0-20181228230101-89fcab3d43de
 	github.com/mhale/smtpd v0.8.3
 	github.com/obot-platform/kinm v0.0.0-20250328185534-d9c9de49cc20
@@ -34,6 +35,7 @@ require (
 	github.com/pterm/pterm v0.12.80
 	github.com/rs/cors v1.11.1
 	github.com/sendgrid/sendgrid-go v3.16.0+incompatible
+	github.com/sethvargo/go-limiter v1.0.0
 	github.com/spf13/cobra v1.8.1
 	github.com/tidwall/gjson v1.18.0
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.60.0
@@ -163,7 +165,6 @@ require (
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.26.1 // indirect
 	github.com/hashicorp/errwrap v1.1.0 // indirect
 	github.com/hashicorp/go-multierror v1.1.1 // indirect
-	github.com/hashicorp/golang-lru/v2 v2.0.7 // indirect
 	github.com/imdario/mergo v0.3.16 // indirect
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/jackc/pgpassfile v1.0.0 // indirect
--- a/pkg/api/ratelimit/authgroup.go
+++ b/pkg/api/ratelimit/authgroup.go
@@ -1,0 +1,214 @@
+// Package ratelimit implements HTTP middleware for rate limiting API requests
+package ratelimit
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"slices"
+	"sync"
+	"time"
+
+	"github.com/hashicorp/golang-lru/v2/expirable"
+	"github.com/obot-platform/obot/pkg/api/authz"
+	"github.com/sethvargo/go-limiter/httplimit"
+	"github.com/sethvargo/go-limiter/memorystore"
+	"k8s.io/apiserver/pkg/authentication/user"
+)
+
+// Config contains all configuration options for the rate limiter middleware.
+// These values can be set via environment variables or command-line flags.
+type Config struct {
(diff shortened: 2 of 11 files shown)


go.mod
```diff
@@ -34,6 +35,7 @@ require (
 	github.com/pterm/pterm v0.12.80
 	github.com/rs/cors v1.11.1
 	github.com/sendgrid/sendgrid-go v3.16.0+incompatible
+	github.com/sethvargo/go-limiter v1.0.0
```
g-linville, line 38: did you vet this one very much? Does not look like it is widely used.

pkg/api/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,121 @@
...
+// It contains only the information needed for rate limiting decisions.
+type User struct {
+	// ID is the unique identifier for the user, used as the rate limit key
+	ID string
+
+	// IsAdmin determines if the user bypasses rate limits completely
+	IsAdmin bool
+}
+
+// EnableGroupRateLimit adds credential information to a user.DefaultInfo for later retrieval.
```
g-linville, line 80: nit

```suggestion
// EnableAuthGroupRateLimit adds credential information to a user.DefaultInfo for later retrieval.
```

pkg/api/ratelimit/authgroup.go
```diff
@@ -0,0 +1,214 @@
...
+)
+
+// Config contains all configuration options for the rate limiter middleware.
+// These values can be set via environment variables or command-line flags.
+type Config struct {
+	// UnauthenticatedRateLimit is the number of requests allowed for unauthenticated users
+	UnauthenticatedRateLimit int `usage:"Rate limit for unauthenticated requests (req/sec)" default:"200"`
+
+	// AuthenticatedRateLimit is the number of requests allowed for authenticated non-admin users
+	AuthenticatedRateLimit int `usage:"Rate limit for authenticated non-admin requests (req/sec)" default:"400"`
```
g-linville, line 26:
```suggestion
	UnauthenticatedRateLimit int `usage:"Rate limit for unauthenticated requests (req/sec)" default:"100"`

	// AuthenticatedRateLimit is the number of requests allowed for authenticated non-admin users
	AuthenticatedRateLimit int `usage:"Rate limit for authenticated non-admin requests (req/sec)" default:"200"`
```

pkg/proxy/proxy.go
```diff
@@ -112,6 +122,12 @@ func (pm *Manager) AuthenticateRequest(req *http.Request) (*authenticator.Respon
 						provider:  system.DefaultNamespace + "/" + configuredProvider,
 						createdAt: time.Now(),
 					}
+					ratelimit.EnableAuthGroupRateLimit(
+						ratelimit.CredSourceTypeCookie,
+						tokenCookie.Name,
+						tokenCookie.Value,
+						resp.User.(*user.DefaultInfo),
+					)
 					pm.lock.Unlock()
```
g-linville, line 131:
```suggestion
					pm.lock.Unlock()
					ratelimit.EnableAuthGroupRateLimit(
						ratelimit.CredSourceTypeCookie,
						tokenCookie.Name,
						tokenCookie.Value,
						resp.User.(*user.DefaultInfo),
					)
```

pkg/services/config.go
```diff
@@ -415,6 +421,12 @@ func New(ctx context.Context, config Config) (*Services, error) {
 		}); err != nil {
 			return nil, fmt.Errorf(`failed to remove "nobody" user and identity from database: %w`, err)
 		}
+
+		// Enable rate limiting for all requests.
+		rateLimiter, err = ratelimit.NewAuthGroupMiddleware(ratelimit.Config(config.RateLimitConfig))
+		if err != nil {
+			return nil, fmt.Errorf(`failed to  create rate limiter middleware: %w`, err)
```
g-linville, line 428: nano nit

```suggestion
			return nil, fmt.Errorf(`failed to create rate limiter middleware: %w`, err)
```

g-linville (changes requested):

go.mod
cjellick: oh, @njhale - lol - this _is_ the one i used in acorn-dns

https://github.com/acorn-io/acorn-dns/pull/40/files

@g-linville it's used in hashicorp projects/products and the author has a good reputation (at least in my experience)

g-linville: Sounds good to me

njhale:
Going to open a new PR that trades some robustness for a simpler design.

(Thanks for the feedback anyway @g-linville)

PR #2370, opened 2025-04-02
chore: return 401 when the anonymous user is unauthorized

thedadams:
If the anonymous user is unauthorized, then whoever made the request should authenticate.


g-linville (approved):

ibuildthecloud:
This looks terrible

PR #2373, opened 2025-04-03
Feat: Add slack integration ui

StrongMonkey:
1. Add slack integration UI.
2. Add slack triggering for task.
3. Some other tweaks to make UI nicely.

### DIFF
--- a/pkg/api/handlers/tasks.go
+++ b/pkg/api/handlers/tasks.go
@@ -20,6 +20,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/util/retry"
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -520,6 +521,10 @@ func (t *TaskHandler) updateTrigger(req api.Context, workflow *v1.Workflow, task
 		return nil, err
 	}
 
+	if err := t.updateSlack(req, workflow, task); err != nil {
+		return nil, err
+	}
+
 	return &trigger, nil
 }
 
@@ -642,6 +647,21 @@ func (t *TaskHandler) updateCron(req api.Context, workflow *v1.Workflow, task ty
 	}
 
 	return nil
+}
+
+func (t *TaskHandler) updateSlack(req api.Context, workflow *v1.Workflow, task types.TaskManifest) error {
+	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		if err := req.Storage.Get(req.Context(), kclient.ObjectKeyFromObject(workflow), workflow); err != nil {
+			return err
+		}
+		if task.OnSlackMessage == nil {
+			workflow.Spec.Manifest.OnSlackMessage = nil
+		} else {
+			workflow.Spec.Manifest.OnSlackMessage = &types.TaskOnSlackMessage{}
+		}
+		return req.Update(workflow)
+	})
+	return err
 }
 
 func (t *TaskHandler) getThreadAndManifestFromRequest(req api.Context) (*v1.Thread, types.WorkflowManifest, types.TaskManifest, error) {
--- a/pkg/api/handlers/tasks.go
+++ b/pkg/api/handlers/tasks.go
@@ -650,7 +650,7 @@ func (t *TaskHandler) updateCron(req api.Context, workflow *v1.Workflow, task ty
 }
 
 func (t *TaskHandler) updateSlack(req api.Context, workflow *v1.Workflow, task types.TaskManifest) error {
-	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
 		if err := req.Storage.Get(req.Context(), kclient.ObjectKeyFromObject(workflow), workflow); err != nil {
 			return err
 		}
@@ -661,7 +661,6 @@ func (t *TaskHandler) updateSlack(req api.Context, workflow *v1.Workflow, task t
 		}
 		return req.Update(workflow)
 	})
-	return err
 }
 
 func (t *TaskHandler) getThreadAndManifestFromRequest(req api.Context) (*v1.Thread, types.WorkflowManifest, types.TaskManifest, error) {
--- a/pkg/gateway/server/oauth_apps.go
+++ b/pkg/gateway/server/oauth_apps.go
@@ -654,6 +654,15 @@ func (s *Server) storeSlackTrigger(apiContext api.Context, tokenResp types.OAuth
 		return nil
 	}
 
+	slackReceiverName := system.SlackReceiverPrefix + thread.Spec.ParentThreadName
+
+	var slackReceiver v1.SlackReceiver
+	if err := apiContext.Get(&slackReceiver, slackReceiverName); apierrors.IsNotFound(err) {
+		return nil
+	} else if err != nil {
+		return err
+	}
+
 	id := name.SafeHashConcatName(slackAppID, slackTeamID, thread.Spec.ParentThreadName)
 	name := system.SlackTriggerPrefix + hash2.String(id)[:12]
 	err := apiContext.Create(&v1.SlackTrigger{
@@ -662,9 +671,10 @@ func (s *Server) storeSlackTrigger(apiContext api.Context, tokenResp types.OAuth
 			Namespace: apiContext.Namespace(),
 		},
 		Spec: v1.SlackTriggerSpec{
-			AppID:      slackAppID,
-			TeamID:     slackTeamID,
-			ThreadName: thread.Spec.ParentThreadName,
+			AppID:             slackAppID,
+			TeamID:            slackTeamID,
+			ThreadName:        thread.Spec.ParentThreadName,
+			SlackReceiverName: slackReceiverName,
 		},
 		Status: v1.SlackTriggerStatus{},
 	})
--- a/ui/user/src/lib/components/EditMode.svelte
+++ b/ui/user/src/lib/components/EditMode.svelte
@@ -29,6 +29,8 @@
 	import Sites from '$lib/components/edit/Sites.svelte';
 	import { errors, responsive, tools } from '$lib/stores';
 	import { twMerge } from 'tailwind-merge';
+	import Slack from '$lib/components/slack/Slack.svelte';
+
 	interface Props {
 		project: Project;
 		currentThreadID?: string;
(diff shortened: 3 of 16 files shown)


pkg/gateway/server/oauth_apps.go
```diff
@@ -662,9 +671,10 @@ func (s *Server) storeSlackTrigger(apiContext api.Context, tokenResp types.OAuth
 			Namespace: apiContext.Namespace(),
 		},
 		Spec: v1.SlackTriggerSpec{
```
StrongMonkey, line 673: @ibuildthecloud this is so that when slack receiver is deleted, the trigger will be deleted too

pkg/api/handlers/tasks.go
```diff
@@ -644,6 +649,21 @@ func (t *TaskHandler) updateCron(req api.Context, workflow *v1.Workflow, task ty
 	return nil
 }
 
+func (t *TaskHandler) updateSlack(req api.Context, workflow *v1.Workflow, task types.TaskManifest) error {
+	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
thedadams: nit:
```suggestion
	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```

thedadams (approved):
Backend looks good.

ui/user/src/lib/components/slack/Slack.svelte
```diff
@@ -0,0 +1,380 @@
...
+			project = await ChatService.updateProject(project);
+		}
+		addSlackBotTool.close();
+	}
+</script>
+
+<CollapsePane header="Slack Integration">
+	<div class="flex flex-col gap-2 p-4">
+		<p class="text-gray text-sm">
+			Enable this to trigger tasks from slack messages that mention the slack bot you configured
```
g-linville:
```suggestion
			Enable this to trigger tasks from Slack messages that mention the Slack bot you configured
```

ui/user/src/lib/components/slack/Slack.svelte
```diff
@@ -0,0 +1,380 @@
...
+			with Obot.
+		</p>
+		<div class="flex items-center justify-end">
+			<button class="button" onclick={() => dialog.showModal()}>
+				<div class="flex items-center gap-2">
+					<Settings size={16} />
+					Configure
+				</div></button
+			>
```
g-linville: nit

```suggestion
				</div>
			</button>
```

ui/user/src/lib/components/EditMode.svelte
```diff
@@ -164,6 +166,7 @@
 							{/if}
 							<Files {project} />
 							<Tasks {project} bind:currentThreadID />
+							<Slack {project} />
```
ivyjeong13, line 169: Super minor nit: 
Screenshot 2025-04-07 at 12 52 54 PM

I think padding on this is off from the other sections (ex. end of Configure button isn't aligned with + New Task, "Enable this..." helper text isn't aligned with "The following tasks..."

ui/user/src/lib/components/slack/Slack.svelte
```diff
@@ -0,0 +1,380 @@
...
+				<div>
+					<h4 class="font-medium">Step 2: Add the Redirect URL</h4>
+					<p class="text-sm text-gray-600">
+						From the Slack API Dashboard, click on your app and select "OAuth & Permissions"
+					</p>
+					<p class="text-sm text-gray-600">
+						In the "Redirect URLs" section, click "Add New Redirect URL"
+					</p>
+					<div class="mt-2 flex max-w-fit items-center gap-2 rounded bg-gray-100 p-2">
+						<CopyButton text={redirectUrl} />
```
ivyjeong13, line 186: Screenshot 2025-04-07 at 12 53 01 PM

dark mode: think a dark mode bg color change is needed! Text is hard to read on the copy / pills

ui/user/src/lib/components/slack/Slack.svelte
```diff
@@ -0,0 +1,380 @@
...
+				<div class="flex items-center gap-2">
+					<Settings size={16} />
+					Configure
+				</div>
+			</button>
+		</div>
+	</div>
+</CollapsePane>
+
+<dialog bind:this={dialog} class="default-dialog w-1/2">
```
ivyjeong13: Screenshot 2025-04-07 at 1 01 03 PM

May need mobile conditional dialog classes like `mobile-screen-dialog`

ui/user/src/lib/components/slack/Slack.svelte
```diff
@@ -0,0 +1,380 @@
...
+
+	let toolSelection = $state<Record<string, AssistantTool>>({});
+
+	toolSelection = getSelectionMap();
+
+	let { project = $bindable() }: Props = $props();
+	let dialog: HTMLDialogElement;
+	let confirmDisable: HTMLDialogElement;
+	let addSlackBotTool: HTMLDialogElement;
+	let redirectUrl = $state(
```
ivyjeong13: nit: $state isn't needed for `redirectUrl` and `eventUrl` if they stay the same and are never modified!

ivyjeong13: Oh actually redirectUrl could use `derived`, just realized the `project` prop!

ui/user/src/lib/components/slack/Slack.svelte
```diff
@@ -0,0 +1,380 @@
...
+	let config = $state({
+		appId: '',
+		clientId: '',
+		clientSecret: '',
+		signingSecret: ''
+	});
+
+	let defaultMask = '***********';
+
+	if (project.capabilities?.onSlackMessage) {
```
ivyjeong13: Looks like this reactive to `project` so maybe needs to be in an `$effect`?

ui/user/src/lib/components/slack/Slack.svelte
```diff
@@ -0,0 +1,380 @@
...
+			project = await ChatService.updateProject(project);
+		}
+		addSlackBotTool.close();
+	}
+</script>
+
+<CollapsePane header="Slack Integration">
+	<div class="flex flex-col gap-2 p-4">
+		<p class="text-gray text-sm">
+			Enable this to trigger tasks from Slack messages that mention the slack bot you configured
```
ivyjeong13, line 132: Screenshot 2025-04-07 at 1 26 03 PM

Kinda confusing this area doesn't change after you've configured the Slack Integration. Maybe either button should say 'Enabled' or this area displays the four masked values to show the user has configured them?

Also not sure if this allowed but you can click Configure without setting any of the values:

https://github.com/user-attachments/assets/6a67ffce-c209-486b-b838-aa01cdbb96bb

ui/user/src/lib/components/slack/Slack.svelte
ivyjeong13: Maybe could save a few lines with setting of `config` with something like: 
```
	let config = $state<SlackConfig>({
		appId: '',
		clientId: '',
		clientSecret: '',
		signingSecret: ''
	});

	function setConfig(val: string) {
		Object.keys(config).forEach((key) => {
			config[key as keyof SlackConfig] = val;
		});
	}

	$effect(() => { // maybe?
		if (project.capabilities?.onSlackMessage) {
			setConfig(defaultMask);
		}
	});
        
	async function handleSubmit() {
                // ...
			addSlackBotTool.showModal();
			setConfig('');

	async function disableSlack() {
		// ...
		confirmDisable.close();
		setConfig('');
	}
```

edit: updated with use of type `SlackConfig`

ui/user/src/lib/components/slack/Slack.svelte
```diff
@@ -0,0 +1,380 @@
...
+		config.clientSecret = defaultMask;
+		config.signingSecret = defaultMask;
+	}
+	async function handleSubmit() {
+		try {
+			if (config.appId === defaultMask) {
+				dialog.close();
+				return;
+			}
+			if (project.capabilities?.onSlackMessage) {
```
ivyjeong13: nit: Since this is called no matter what, you could probably reduce this to just:
```
await configureProjectSlack(project.assistantID, project.id, config, project.capabilities?.onSlackMessage ? 'PUT' : 'POST');
```
without the if/else

ui/user/src/lib/components/slack/Slack.svelte
ivyjeong13: Maybe a silly/rare case:

After configuring the Slack Integration, I went back and change the appId (like maybe I realized I added the wrong appId) and clicked Configure.

Screenshot 2025-04-07 at 2 01 33 PM

It sent the defaultMask as part of the payload and wondering if the backend handles/ignores that?

ui/user/src/lib/components/EditMode.svelte
StrongMonkey: image
fixed!

ui/user/src/lib/components/slack/Slack.svelte
StrongMonkey: image
yeah fixed

ui/user/src/lib/components/slack/Slack.svelte
StrongMonkey: done 👍

ui/user/src/lib/components/slack/Slack.svelte
StrongMonkey: yeah, we don't expose those values in api(due to those being secrets), so I compensate those values in frontend to show masked value indicating those are set...but I realize it is a bad idea because user will realize that it is set..\ let me fix this

ui/user/src/lib/components/slack/Slack.svelte
StrongMonkey: image
image

added an enable button and error message to not allowing empty value

ui/user/src/lib/components/slack/Slack.svelte
StrongMonkey: Should be fixed

StrongMonkey:
@ivyjeong13 Pushed a commit that addresses your comment. Thanks for looking into this!

ui/user/src/lib/components/slack/Slack.svelte
ivyjeong13: I think the `w-1/2` is overriding one of its properties. This should do it!:
```
<dialog
	bind:this={dialog}
	class="default-dialog md:w-1/2"
	class:mobile-screen-dialog={responsive.isMobile}
>
```

StrongMonkey: done!

ui/user/src/lib/components/slack/Slack.svelte
ivyjeong13: :+1: Just have a nit that if the user does receive an error that a user-friendly error should display in the dialog instead of an errors notification behind the dialog 

Screenshot 2025-04-07 at 4 54 51 PM

ivyjeong13 (approved):
🙌

ui/user/src/lib/components/slack/Slack.svelte
StrongMonkey: image
done 👍

PR #2374, opened 2025-04-03
enhance/post authn rate limits

njhale:
Rate limit all HTTP requests after authentication completes. Authenticated users get a higher quota than unauthenticated users by default and, as a special case, admins are exempt from rate limits entirely.

Addresses https://github.com/obot-platform/sensitive-issues/issues/18

### DIFF
--- a/pkg/api/server/ratelimiter/ratelimiter.go
+++ b/pkg/api/server/ratelimiter/ratelimiter.go
@@ -1,0 +1,121 @@
+package ratelimiter
+
+import (
+	"errors"
+	"fmt"
+	"net"
+	"net/http"
+	"slices"
+	"strconv"
+	"time"
+
+	"github.com/obot-platform/obot/pkg/api/authz"
+	"github.com/obot-platform/obot/pkg/api/server/requestinfo"
+	"github.com/sethvargo/go-limiter"
+	"github.com/sethvargo/go-limiter/memorystore"
+	"k8s.io/apiserver/pkg/authentication/user"
+)
+
+const (
+	// HeaderRateLimitLimit, HeaderRateLimitRemaining, and HeaderRateLimitReset
+	// are the recommended return header values from IETF on rate limiting. Reset
+	// is in UTC time.
+	headerRateLimitLimit     = "X-RateLimit-Limit"
+	headerRateLimitRemaining = "X-RateLimit-Remaining"
+	headerRateLimitReset     = "X-RateLimit-Reset"
+
+	// HeaderRetryAfter is the header used to indicate when a client should retry
+	// requests (when the rate limit expires), in UTC time.
+	headerRetryAfter = "Retry-After"
+)
+
+var ErrRateLimitExceeded = errors.New("rate limit exceeded, please try again later")
+
+type Options struct {
+	UnauthenticatedRateLimit int `usage:"Rate limit for unauthenticated requests (req/sec)" default:"100"`
+	AuthenticatedRateLimit   int `usage:"Rate limit for authenticated non-admin requests (req/sec)" default:"200"`
+}
+
+// RateLimiter limits the number of HTTP requests per second a user can make.
+// It tracks limits for unauthenticated and authenticated users separately:
+// - Authenticated requests are tracked by user ID or name.
+// - Unauthenticated requests are tracked by IP address.
+// - Admins are exempt from rate limiting.
+type RateLimiter struct {
+	unauthenticatedStore limiter.Store
+	authenticatedStore   limiter.Store
+}
+
+func New(opts Options) (*RateLimiter, error) {
+	unauthenticatedStore, err := memorystore.New(&memorystore.Config{
+		Tokens:   uint64(opts.UnauthenticatedRateLimit),
+		Interval: time.Second,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create unauthenticated store: %w", err)
+	}
+
+	authenticatedStore, err := memorystore.New(&memorystore.Config{
+		Tokens:   uint64(opts.AuthenticatedRateLimit),
+		Interval: time.Second,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create authenticated store: %w", err)
+	}
+
+	return &RateLimiter{
+		unauthenticatedStore: unauthenticatedStore,
+		authenticatedStore:   authenticatedStore,
+	}, nil
+}
+
+// ApplyLimit applies the user's rate limit to the request, sets the rate limit headers, and returns a ErrRateLimitExceeded error if the limit has been exceeded.
+// It returns nil if the user is exempt from rate limiting or if the user has not exceeded their limit.
+func (l *RateLimiter) ApplyLimit(u user.Info, rw http.ResponseWriter, req *http.Request) error {
+	groups := u.GetGroups()
+
+	if slices.Contains(groups, authz.AdminGroup) {
+		// Admins are exempt from rate limiting
+		return nil
+	}
+
+	var store limiter.Store
+	key := u.GetUID()
+	if key == "" {
+		key = u.GetName()
+	}
+
+	if slices.Contains(groups, authz.AuthenticatedGroup) && key != "" {
+		store = l.authenticatedStore
+	} else {
+		// Get the source IP address from the request.
+		key = requestinfo.GetSourceIP(req)
+
+		// Strip the port from the IP address if present.
+		if ip, _, err := net.SplitHostPort(key); err == nil {
+			key = ip
+		}
+
+		store = l.unauthenticatedStore
+	}
+
+	limit, remaining, reset, ok, err := store.Take(req.Context(), key)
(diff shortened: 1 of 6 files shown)


pkg/api/server/audit/logger.go
```diff
@@ -168,3 +169,26 @@ func (l *noOpLogger) LogEntry(_ LogEntry) error {
 func (l *noOpLogger) Close() error {
 	return nil
 }
+
+// GetSourceIP extracts the real client IP address from the request.
+// It checks X-Forwarded-For and X-Real-IP headers before falling back to RemoteAddr.
+func GetSourceIP(req *http.Request) string {
```
njhale: There's probably a better place to put this. I'm open to suggestions.

pkg/api/server/server.go
```diff
@@ -82,78 +85,81 @@ func (s *Server) wrap(f api.HandlerFunc) http.HandlerFunc {
...
-					Method:    req.Method,
-					Path:      req.URL.Path,
-					UserAgent: req.UserAgent(),
-					SourceIP:  getSourceIP(req),
-					Host:      req.Host,
-				},
-				auditLogger: s.auditLogger,
-			}
+		// Apply per-user rate limits to all endpoints
+		s.rateLimiter.Limit(user, rw, req, func(rw http.ResponseWriter, req *http.Request) {
```
njhale: I'm not thrilled with how this method signature turned out, especially with the nested anonymous functions.

Besides that, this is actually creating a new anonymous function for _every_ request made to the server. I'm not sure how badly that will affect performance.

Anyone have an alternative approach in mind for this?

njhale: Came up with something slightly less bad.

pkg/api/server/audit/logger.go
```diff
@@ -168,3 +169,26 @@ func (l *noOpLogger) LogEntry(_ LogEntry) error {
...
+	// Check X-Forwarded-For header first
+	if xff := req.Header.Get("X-Forwarded-For"); xff != "" {
+		// X-Forwarded-For can contain multiple IPs (client, proxy1, proxy2, ...).
+		// A typical deployment of Obot will have a proxy in front of it that appends the request client's IP address (req.RemoteAddr) to the X-Forwarded-For header before forwarding the request to Obot.
+		// With that in mind, we choose the rightmost IP in the X-Forwarded-For header since it's the only IP address that is not spoofable by the client.
+		ips := strings.Split(xff, ",")
+		if len(ips) > 0 {
+			return strings.TrimSpace(ips[len(ips)-1])
```
njhale: Fixed the issue we were discussing yesterday @thedadams

pkg/api/server/audit/logger.go
thedadams: I'd probably put it in its own package, but I think this is also fine.

pkg/api/server/audit/logger.go
```diff
@@ -168,3 +169,26 @@ func (l *noOpLogger) LogEntry(_ LogEntry) error {
...
+// GetSourceIP extracts the real client IP address from the request.
+// It checks X-Forwarded-For and X-Real-IP headers before falling back to RemoteAddr.
+func GetSourceIP(req *http.Request) string {
+	// Check X-Forwarded-For header first
+	if xff := req.Header.Get("X-Forwarded-For"); xff != "" {
+		// X-Forwarded-For can contain multiple IPs (client, proxy1, proxy2, ...).
+		// A typical deployment of Obot will have a proxy in front of it that appends the request client's IP address (req.RemoteAddr) to the X-Forwarded-For header before forwarding the request to Obot.
+		// With that in mind, we choose the rightmost IP in the X-Forwarded-For header since it's the only IP address that is not spoofable by the client.
+		ips := strings.Split(xff, ",")
+		if len(ips) > 0 {
```
thedadams: I know AI wrote this for me previously, but this check is superfluous.

pkg/api/server/ratelimiter/ratelimiter.go
```diff
@@ -0,0 +1,124 @@
...
+	headerRateLimitReset     = "X-RateLimit-Reset"
+
+	// HeaderRetryAfter is the header used to indicate when a client should retry
+	// requests (when the rate limit expires), in UTC time.
+	headerRetryAfter = "Retry-After"
+)
+
+type Options struct {
+	// UnauthenticatedRateLimit is the number of requests per second an unauthenticated user can make before being rate limited.
+	UnauthenticatedRateLimit int `usage:"Rate limit for unauthenticated requests (req/sec)" default:"100"`
```
thedadams: I understand the desire to document your code, but I would argue that this is in detriment of the DRY principle. You now have two documentation strings to deal with.

pkg/api/server/ratelimiter/ratelimiter.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// RateLimiter limits the number of HTTP requests per second a user can make.
+// It tracks limits for unauthenticated and authenticated users separately:
+// - Authenticated requests are tracked by user ID or name.
+// - Unauthenticated requests are tracked by IP address.
+// - Admins are exempt from rate limiting.
+//
+// Rate limits are tracked in-memory using a token bucket algorithm.
+// The bucket is refilled at a rate of 1 token per second.
```
thedadams: Seems like the rate isn't 1 token per second, but, rather, is determined by the configuration. Is that right?

pkg/api/server/ratelimiter/ratelimiter.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	return &RateLimiter{
+		unauthenticatedStore: unauthenticatedStore,
+		authenticatedStore:   authenticatedStore,
+	}, nil
+}
+
+// ApplyLimit applies the user's rate limit to the request and returns true if the user has exceeded their limit.
+// When the user has exceeded their limit, the response headers are set and the caller should return a 429 error.
```
thedadams: Is there a reason we shouldn't set the response header here, too? Since we are already setting the other headers.

pkg/api/server/ratelimiter/ratelimiter.go
```diff
@@ -0,0 +1,124 @@
...
+		return false, nil
+	}
+
+	var store limiter.Store
+	key := u.GetUID()
+	if key == "" {
+		key = u.GetName()
+	}
+
+	if slices.Contains(groups, authz.AuthenticatedGroup) && key != "" && key != "anonymous" {
```
thedadams: nit: The anonymous user should never be in the Authenticated group.

pkg/api/server/ratelimiter/ratelimiter.go
njhale: Sorry, yes I let cursor do its thing there and didn't prune it back. Will fix

pkg/api/server/ratelimiter/ratelimiter.go
njhale: The idea was to make it less fragile in case that changes unintentionally somehow, but yeah I'll drop.

pkg/api/server/ratelimiter/ratelimiter.go
njhale: I was on the fence about whether it's easier to read in the wrap function this way.

pkg/api/server/ratelimiter/ratelimiter.go
njhale: This comment was cursor being overzealous when I told it to apply comments everywhere -- I missed this.

The number of available tokens per key is reset at the `Interval` configured on the store (which is hard-coded to one second), so this comment doesn't really make sense since the "refill rate" depends on how many tokens a user consumed within a given interval.

pkg/api/server/ratelimiter/ratelimiter.go
njhale: OK, I found some time today to look at this again and I still feel the current approach is more clear because the caller is always responsible for sending the request, and I can read through the `wrap` function and understand what error code is being returned right away. I think it aligns better with the way the code is currently set up.

Note: I've update this a bit so that it returns a specific error when rate limits are exceeded.

thedadams: I don't think errors are necessary for this.

pkg/api/server/audit/logger.go
njhale: Moved it out

pkg/api/server/audit/logger.go
njhale: Fixed

pkg/api/server/ratelimiter/ratelimiter.go
njhale: Removed

pkg/api/server/ratelimiter/ratelimiter.go
njhale: Removed

g-linville (approved):

PR #2392, opened 2025-04-04
feat: add ability to log out of all other sessions

g-linville:


pkg/gateway/client/identity.go
```diff
@@ -17,6 +17,14 @@ var verifiedAuthProviders = []string{
 	"default/github-auth-provider",
 }
 
+func (c *Client) FindIdentitiesForUser(ctx context.Context, userID uint) ([]*types.Identity, error) {
```
thedadams: nit: the slice of pointers seems kind of strange. Slices are already passed as pointers, so not sure if this gives us any advantage.

pkg/gateway/server/router.go
```diff
@@ -61,4 +61,6 @@ func (s *Server) AddRoutes(mux *server.Server) {
 
 	// LLM proxy
 	mux.HandleFunc("POST /api/llm-proxy/{path...}", s.llmProxy)
+
+	mux.HandleFunc("POST /api/logout-all", wrap(s.logoutAll))
```
thedadams: nit: how would you feel about putting this up with the `/api/me` endpoint?

pkg/gateway/server/logout_all.go
```diff
@@ -0,0 +1,114 @@
...
+import (
+	"crypto/sha256"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/obot-platform/obot/pkg/api"
+	kcontext "github.com/obot-platform/obot/pkg/gateway/context"
```
thedadams: nit:
```suggestion
	gcontext "github.com/obot-platform/obot/pkg/gateway/context"
```

pkg/gateway/server/logout_all.go
```diff
@@ -0,0 +1,114 @@
...
+		userHash := sha256.Sum256([]byte(user))
+
+		logger.Debug("deleting sessions for provider", "provider", identity.AuthProviderName)
+
+		if meta, ok := ref.Status.Tool.Metadata["providerMeta"]; ok {
+			tablePrefix := gjson.Get(meta, "postgresTablePrefix").String()
+			if tablePrefix != "" {
+				var err error
+				if sessionIDFound {
+					err = s.db.WithContext(apiContext.Context()).Exec("DELETE FROM "+tablePrefix+"sessions WHERE key NOT LIKE ? AND \"user\" = decode(?, 'hex') AND \"email\" = decode(?, 'hex')", sessionID+"%", fmt.Sprintf("%x", userHash), fmt.Sprintf("%x", emailHash)).Error
```
thedadams: Add a method to the client for these. We want to have a better separation of concerns.

pkg/gateway/server/logout_all.go
```diff
@@ -0,0 +1,114 @@
...
+			// The GitHub auth provider stores the username as the user ID in the sessions table.
+			// This is because of an annoying quirk of the oauth2-proxy code for GitHub,
+			// where we do not know the real user ID until after the user has logged in and the session is created,
+			// and we have to manually fetch it from the GitHub API.
+			// The oauth2-proxy is only aware of the username, which is why that's in the sessions table.
+			user = identity.ProviderUsername
+		}
+
+		emailHash := sha256.Sum256([]byte(identity.Email))
+		userHash := sha256.Sum256([]byte(user))
```
thedadams: We do have `hash.String` that will take care of this for you (and convert to `hex`).

pkg/gateway/server/logout_all.go
StrongMonkey: Any reason we are not directly comparing the value for user and email in query, instead of doing hash and decode with hash too?

g-linville: yeah, PII can't be stored in plaintext, and email and username are considered PII

g-linville: the decode call in there is just to convert from a hex string representation of a hash, to the postgres `bytea` type

pkg/gateway/client/identity.go
g-linville: fixed

pkg/gateway/server/router.go
g-linville: fixed

pkg/gateway/server/logout_all.go
g-linville: fixed

pkg/gateway/server/logout_all.go
g-linville: fixed

pkg/gateway/server/logout_all.go
g-linville: fixed

thedadams (approved):

ivyjeong13:
Just one small thing I noticed: If logout-all isn't supported, should we not display the logout all option?

Screenshot 2025-04-07 at 2 31 56 PM

Issue #2394, opened 2025-04-04
Opening any featured obot  and not taking any other action ( no chatting/editing) results in a new obot instance being created.


sangee2004:
When we open on any featured obot , make no chat conversations and quit, we already see this featured obot being listed in "My obots" . In this case,  we have actually created an instance of the obot for this user. 
Should we  prevent creating an obot instance in this case ?

ivyjeong13:
This is related to the change where we create a new thread if a thread hasn't been created for them but we are reverting that change. Gonna hijack this ticket and do the revert with it as it will address this.

sangee2004:
Discussed with @ivyjeong13 .This issue is not yet addressed yet.

PR #2399, opened 2025-04-04
fix: revert set initial thread change

ivyjeong13:
* revert the setting of first thread or creating a new thread change
* do similar thing with files, create thread if user clicks thread tool catalog
* remove `navigating` check, swore there was an error before but now think I'm going crazy 🤷‍♀️


ivyjeong13:
Addresses #2394

PR #2410, opened 2025-04-07
feat: implement retention policy for obots and threads

g-linville:


thedadams:
How would you feel about running this as a controller on the thread objects and using `resp.RetryAfter()`. You would only have to do the `RetryAfter` if the time to retry was within 10 hours (because that is the automatic requeue duration).

In my mind, there are a few advantages to using a controller:
1. We don't have to list all projects and threads uncached to process a few of them.
2. We would have better control of when the threads are processed. For example, the implementation here would very likely not cleanup threads during the week because we merge and deploy more frequently than when the cleanup would run.

What are your thoughts here? Are there benefits to this goroutine approach that I am not seeing?

g-linville:
@thedadams that is a good point about deployments resetting the clock for how often we check - perhaps I should just do cron instead of a ticker in order to get around that problem.

I think the main advantage of doing it the way that I implemented it is that it allows me to process the threads in the order that I want to. If I implement it in a controller, they will be processed in a random order. Though I suppose I could implement it such that, each time we delete a chat thread, we check its project thread and list out the threads for that, and if there are none remaining, we then delete the project thread. Though this would likely result in several unnecessary List calls, if, for example, for a single project, there are 30 threads, and we end up deleting 10 of them. But maybe that doesn't matter, since we will have the caching. Let me know what you think.

g-linville:
@thedadams oh wow, I had not considered that either. Yeah I think that is the right approach. I'm going to move this back into draft and unassign reviewers.

pkg/controller/handlers/retention/lastused.go
```diff
@@ -0,0 +1,27 @@
...
+		if err := c.Get(ctx, router.Key(thread.Namespace, thread.Spec.ParentThreadName), &parentThread); err != nil {
+			return err
+		}
+		parentThread.Status.LastUsedTime = thread.Status.LastUsedTime
+		if err := c.Status().Update(ctx, &parentThread); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
g-linville: I wanted to capture this logic in a function since we use it three times. It feels kinda weird and arbitrary, since the caller has to call `.Status().Update()` for the thread after calling this, but I didn't know of a better way to do it.

pkg/invoke/invoker.go
```diff
@@ -479,6 +480,9 @@ func (i *Invoker) createRun(ctx context.Context, c kclient.WithWatch, thread *v1
 				return err
 			}
 			thread.Status.CurrentRunName = run.Name
+			if err := retention.SetLastUsedTime(ctx, c, thread); err != nil {
```
g-linville, line 486: I implemented this so that, every time we change the `CurrentRunName` on a thread, we also update its last used time. I figured this was the easiest way to determine thread usage.

pkg/controller/handlers/retention/retention.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	if !thread.Status.LastUsedTime.IsZero() && time.Since(thread.Status.LastUsedTime.Time) > m.policy {
+		log.Infof("retention: deleting thread %s/%s", thread.Namespace, thread.Name)
+		return m.c.Delete(req.Ctx, thread)
+	}
+
+	if m.frequency < 10*time.Hour {
+		resp.RetryAfter(m.frequency)
+	}
```
thedadams: I would imagine something like this. Then you don't need the frequency parameter.
```suggestion
	if since := time.Since(thread.Status.LastUsedTime.Time); m.policy - since < 10*time.Hour {
		resp.RetryAfter(until)
	}
```

pkg/controller/handlers/retention/lastused.go
```diff
@@ -0,0 +1,27 @@
...
+	thread.Status.LastUsedTime = metav1.Now()
+	if thread.Spec.ParentThreadName != "" {
+		var parentThread v1.Thread
+		if err := c.Get(ctx, router.Key(thread.Namespace, thread.Spec.ParentThreadName), &parentThread); err != nil {
+			return err
+		}
+		parentThread.Status.LastUsedTime = thread.Status.LastUsedTime
+		if err := c.Status().Update(ctx, &parentThread); err != nil {
+			return err
+		}
```
thedadams: We probably need some sort of retry on conflict here. You can imagine two tasks running simultaneously for a single obot conflicting here.

My suggestion would be to have a `now := metav1.Now()` at the top of this function, so that after we only update the parent thread if the `LastUsedTime` on it is after `now` (to avoid multiple unnecessary updates.

pkg/invoke/workflow.go
```diff
@@ -159,6 +160,10 @@ func (i *Invoker) rerunThread(ctx context.Context, c kclient.WithWatch, wf *v1.W
 
 	if thread.Status.CurrentRunName != "" || thread.Status.LastRunName != "" {
 		thread.Status.CurrentRunName = ""
+		if err := retention.SetLastUsedTime(ctx, c, &thread); err != nil {
+			return nil, nil, err
+		}
+
```
thedadams: nit: breaking up the assignments to empty strings here seems rubs me the wrong way.

pkg/controller/handlers/retention/retention.go
g-linville: Fixed. Good call

pkg/controller/handlers/retention/lastused.go
g-linville: Fixed

pkg/invoke/workflow.go
g-linville: Fixed

g-linville:
Sorry for the rebase. I needed to resolve conflicts with main.

pkg/controller/handlers/retention/retention.go
```diff
@@ -0,0 +1,72 @@
...
+
+func NewRetentionManager(c kclient.Client, policy time.Duration) *Manager {
+	if policy == 0 {
+		log.Infof("retention policy: disabled")
+	} else {
+		log.Infof("retention policy: %s", policy)
+	}
+
+	return &Manager{
+		c:      c,
```
thedadams: As I understand it, the only reason to pass this client is so that you can make the list request below with multiple field selectors. Since that would be an uncached list, I don't think it's worth doing this because this handler will be called a lot.

pkg/controller/handlers/retention/retention.go
```diff
@@ -0,0 +1,72 @@
...
+		return nil
+	}
+
+	if thread.Spec.Project {
+		// If this thread is a project, there is a chance it is a featured Obot.
+		// We do not want to clean up featured Obots. Check the thread shares to see if it is one.
+		shares := &v1.ThreadShareList{}
+		if err := m.c.List(req.Ctx, shares, kclient.InNamespace(thread.Namespace), &kclient.ListOptions{
+			FieldSelector: fields.SelectorFromSet(map[string]string{
+				"spec.featured":          "true",
```
thedadams: Remove this line, and check each item to see if it is featured.

thedadams (changes requested):

pkg/controller/handlers/retention/retention.go
g-linville: Curious what the reason for this is? Is it more performant to just cycle through them in a loop instead of using the field selector?

g-linville: (fixed btw)

pkg/controller/handlers/retention/retention.go
g-linville: fixed

pkg/controller/handlers/retention/retention.go
thedadams: The cached client doesn't allow you to list with multiple field selectors. So, you have list with one and adjust.

thedadams (approved):

njhale (approved):

Issue #2386, opened 2025-04-03
cards arent consistent height on safari


cjellick:
![Image](https://github.com/user-attachments/assets/e04d21fd-b2ca-4c21-90a4-ea8675fc3e06)

shweta8996:
I have verified on obot.ai and it is working fine as expected. now cards consistent height on safari.
Please refer attachment:

<img width="1596" alt="Image" src="https://github.com/user-attachments/assets/ccd6e41e-d74d-4e84-8f9f-590f332b3b76" />

PR #2411, opened 2025-04-07
Fix: Cards consistent height issue on safari

tony-dotnet-fullstack:


ivyjeong13 (approved):

ivyjeong13:
Sometimes it slip my mind as well but also good to reference the issue the PR is tied to! Ex. Addresses #2386

Issue #2412, opened 2025-04-07
Allow users to delete their account


thedadams:
It should be possible for a user to delete their account. Doing so, should also delete their obots, threads, credentials and any thing else owned by the user.

thedadams:
The backend portion of this is done in https://github.com/obot-platform/obot/pull/2416

thedadams:
THe backend has been merged. Moving to Triage for frontend work.

cjellick:
@rob-fullstack-star - when you implement this, here is a high level of what id like:

In the user's drop down menu in the top right (pictured below) I would like a "My Account" link. I'd like that to take the user to an account page that has basic information from `/api/me` on it. Id like this page to also have a "Delete my account" button that hooks into the api donnie implemented here. AND I'd like the "Log out all other sessions" action removed from the drop down and moved to this page. It's text can be made a little friendlier too, like "Log me everywhere else" or "Log me out on all my other devices" or something like that.

![Image](https://github.com/user-attachments/assets/4bc591b7-ca72-471b-bff2-d7dafaa7087f)

thedadams:
I forgot to add that a `DELETE` call to `/api/me` will delete the logged-in user's account.

sangee2004:
Users are now given the ability to delete their account  in `My Accounts` page

<img width="700" alt="Image" src="https://github.com/user-attachments/assets/dda04402-d386-452f-906f-68ece2672ac3" />

Issues found when testing this feature is tracked in separate issues.

Issue #2252, opened 2025-03-21
Expose user delete in admin UI


cjellick:
We have an api endpoint for deleting users. We should expose that as an option in the admin ui.

We should also verify that it fully deletes all obots, threads, credentials, and other resources owned by the user.

thedadams:
I am fairly sure that deleting an account does not cleanup the owned objects. I will ensure this happens with this issue: https://github.com/obot-platform/obot/issues/2412

thedadams:
The backend portion for this is done in https://github.com/obot-platform/obot/pull/2416

thedadams:
The backend code has been merged. Moving to Triage for frontend work.

cjellick:
@rob-fullstack-star - adding this to your stack. it is coupled with https://github.com/obot-platform/obot/issues/2412

thedadams:
To delete a specific user's account, make a `DELETE` call to `/api/users/{username}`

sangee2004:
Admin users are now given the ability to delete users in `User` list view.

Issues found when testing this feature is tracked in separate issues.

PR #2416, opened 2025-04-08
feat: add the ability for a user to remove themselves

thedadams:
This change adds the backend work to allow a user to delete themselves and have all of their Obots, threads, runs, credentials, etc cleaned up. This change also ensures that these items are cleaned up when an admin deletes another user.

The cleanup is accomplished asynchronously, in the controller. This seemed like a prudent way to do the cleanup for two reasons:
1. To make the API as speedy as possible; and
2. To centralize the cleanup so that other objects could be cleaned up easily.

There is a slight issue with this: the order in which the items are cleaned up and the user is removed from the database. There could be situations where the starting of the cleanup is successful, but the removal of the user fails. Or visa-versa. This change will start the cleanup of the objects before trying to remove the user from the database. This seemed like the most reasonable thing to do because the deletion of the user could be retried in this scenario. In the reverse case, if the removal of user object from the database was successful, but the starting of the cleanup failed, then there is really no way to start the cleanup again because the user is gone.

Issues: https://github.com/obot-platform/obot/issues/2252, https://github.com/obot-platform/obot/issues/2412


pkg/controller/handlers/cleanup/user.go
```diff
@@ -0,0 +1,57 @@
...
+
+func NewUserCleanup(gatewayClient *gclient.Client) *UserCleanup {
+	return &UserCleanup{
+		gatewayClient: gatewayClient,
+	}
+}
+
+func (u *UserCleanup) Cleanup(req router.Request, _ router.Response) error {
+	userDelete := req.Object.(*v1.UserDelete)
+	var projects v1.ThreadList
```
StrongMonkey: Nit: `Threads` might be more appropriate for variable name

pkg/gateway/server/user.go
```diff
@@ -129,11 +132,32 @@ func (s *Server) updateUser(apiContext api.Context) error {
...
+	if err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return types2.NewErrNotFound("user %s not found", username)
+		}
+		return fmt.Errorf("failed to get user: %v", err)
+	}
+
+	if err = apiContext.Create(&v1.UserDelete{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      system.UserDeletePrefix,
```
StrongMonkey: This should be generated name?

pkg/controller/handlers/cleanup/user.go
thedadams: Changed.

pkg/gateway/server/user.go
thedadams: Yes, good catch. Updated.

StrongMonkey (approved):

g-linville (approved):

Issue #2404, opened 2025-04-07
Custom Tools TLC


ivyjeong13:
Custom Tools (not from catalog) TLC update

shweta8996:
@ivyjeong13 I can't see the custom tool in my obot , can you help me with it?

![Image](https://github.com/user-attachments/assets/281679ce-389e-49a2-a7cd-254d3b0ecd1f)

shweta8996:
I have verified on obot. ai and it is working fine as expected. Everything is working as expected 
Please refer attachment:

https://github.com/user-attachments/assets/89667faa-c8f7-4e7d-94a2-291f5abbaaed

PR #2420, opened 2025-04-08
enhance: custom tools TLC

ivyjeong13:
* custom tools TLC 
* some UI consistency in arguments/running with arguments in tasks & custom tools
* tweak to popover action to disable portaling if needed (keeps tooltips working in dialogs moved to the #topLayer)
* also tacked on: fix to make sure the fade-in/fade-out gradients do not overlap over the scrollbars

Addresses #2404 

https://www.loom.com/share/12c1adf656a341659df968bc49aa76d5

### DIFF
--- a/ui/user/src/lib/components/Thread.svelte
+++ b/ui/user/src/lib/components/Thread.svelte
@@ -18,6 +18,7 @@
 	import EditIcon from './edit/EditIcon.svelte';
 	import { DEFAULT_PROJECT_DESCRIPTION, DEFAULT_PROJECT_NAME } from '$lib/constants';
 	import { tooltip } from '$lib/actions/tooltip.svelte';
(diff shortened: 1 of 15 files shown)


ui/user/src/lib/components/Thread.svelte
```diff
@@ -181,7 +181,7 @@
 
 <div
 	id="main-input"
-	class="thread-container default-scrollbar-thin flex w-full grow justify-center overflow-y-auto"
```
ivyjeong13, line 184: TODO: seeing case of the fade-out/fade-in crossing over scrollbar when file editor is open, adding self task to make sure that case is also taken care of

ivyjeong13: Fixed

cjellick (approved):

Issue #2400, opened 2025-04-05
Download button in file editor overlaps with file content.


sangee2004:
**Describe the bug**
Download button in file editor overlaps with file content

**To Reproduce**
1. Chat with any obot and ask to generate a code snippet
2. Use "Open message in Editor" option to open the response in file editor.

<img width="1150" alt="Image" src="https://github.com/user-attachments/assets/2272c01d-b963-49b7-a1ad-74e13adebfe2" />

With download option highlighted

<img width="729" alt="Image" src="https://github.com/user-attachments/assets/98899bdd-4587-4556-a9d6-8ca7443b13ab" />

Notice that the download option overlaps with the file content .
**Expected behavior**
Download button  should not overlaps with file content.

cjellick:
While the current behavior definitely looks bad/off, I don't think simply push the text down (such that button gets it's own "line") will look right either. Not sure of the right solution, but I just wanted to preemptively give this opinion.

shweta8996:
I have verified on obot.ai and it is working fine as expected. Download button is not overlaps with file content.
Please refer attachment:

https://github.com/user-attachments/assets/1b15ad98-b676-4e24-b861-a3de738b95c8

PR #2421, opened 2025-04-08
Fix: Download button overlaps issue in file editor

tony-dotnet-fullstack:


ui/user/src/app.css
```diff
@@ -196,6 +196,12 @@
 		&:focus {
 			outline: none;
 		}
+		&.bg-gray-100 {
```
ivyjeong13: I believe `bg-gray-100` is an already supported with the bg utility class in tailwind so it may be confusing to conflict with this one and if someone wants to use the utility class

If all you need is the hover, you can do `class="bg-gray-100 hover:bg-surface3"` on the button!

ivyjeong13:
I think we'll still have the issue of the text being overlapped by this button, wondering if on desktop, the button only shows on hover/focus of the file content? Or some padding on the right to keep text from running into it?

ui/user/src/lib/components/Editors.svelte
```diff
@@ -118,7 +118,7 @@
 
 		{#if downloadable}
 			<button
-				class="icon-button absolute top-5 right-5"
+				class="icon-button hover:bg-surface3 absolute top-5 right-5 bg-gray-100 shadow-sm"
```
ivyjeong13: https://github.com/user-attachments/assets/2cfbd43a-17b8-4f5a-b6a3-b016d4451402

Looks like download button still is overlapping the text. Added possible suggestions in comments!

And may need to adjust bg colors for dark mode, the unhovered bg color doesn't match the hovered bg-surface3. Instead of bg-gray-100, another surface color may work better?

tony-dotnet-fullstack: It would be better to add some padding on the right to prevent the text from running into the editor.

tony-dotnet-fullstack: Here’s the new design I implemented.
image
image

ui/user/src/lib/components/editor/Codemirror.svelte
```diff
@@ -291,7 +291,14 @@
 	use:editor
 	onfocusin={() => (focused = true)}
 	onfocusout={() => (focused = false)}
-	class={twMerge('mx-2 mt-4 h-full border-l-2 border-gray-100 dark:border-gray-900', klass)}
+	class={twMerge(
+		'mx-2 mt-4 h-full border-l-2 border-gray-100 pr-[56px] dark:border-gray-900 dark:bg-[#0d1117]',
```
ivyjeong13: I think these stylings may need to be in Editors rather than Codemirror specifically
* Codemirror gets used in CustomTools too and it doesn't have a download option so we'd want to avoid this showing up there
* There are other file preview renders that will have the download button (think it's everything besides PDF! Should be based on `downloadable` in Editors!) Might be good to make sure content doesn't get overlapped by the button there too -- image is probably okay though?
```
{#if file.name.toLowerCase().endsWith('.md')}
			<Milkdown {file} {onFileChanged} {onInvoke} {items} class="p-5" />
		{:else if file.name.toLowerCase().endsWith('.pdf')}
			<Pdf {file} {height} />
		{:else if file.table?.name}
			<Table tableName={file.table?.name} {project} {currentThreadID} {items} />
		{:else if isImage(file.name)}
			<Image {file} />
		{:else if [...(file?.file?.contents ?? '')].some((char) => char.charCodeAt(0) === 0)}
			{@render unsupportedFile()}
		{:else}
			<Codemirror
				{file}
				{onFileChanged}
				{onInvoke}
				{items}
				class="m-0 overflow-hidden rounded-b-2xl"
			/>
		{/if}
```

tony-dotnet-fullstack: refactored code

ivyjeong13: Screenshot 2025-04-09 at 4 27 42 PM

Think just md files are left! Still seeing text overlap in there (the `<Milkdown />` component)

ui/user/src/lib/components/editor/FileEditors.svelte
```diff
@@ -21,31 +21,42 @@
...
-				{items}
-				class="m-0 overflow-hidden rounded-b-2xl"
-			/>
-		{/if}
-	</div>
+		</div>
+	{:else}
+		<div
+			class:hidden={!file.selected}
+			class="default-scrollbar-thin h-full flex-1 pr-[56px] dark:bg-[#0d1117]"
```
ivyjeong13: Screenshot 2025-04-10 at 10 13 00 AM

Only thing: background color does not match surface color here to me. 😅 

I'd also generally avoid using hardcoded hex colors if you can assign them a var in the app.css for theme. I'm guessing the base stylings for Codemirror is conflicting with Milkdown which is set to our theming. I'd lean towards making Codemirror use our colors but not in the scope of this issue so no worries!

ui/user/src/lib/components/editor/FileEditors.svelte
```diff
@@ -21,31 +21,42 @@
...
+			{:else}
+				<Codemirror
+					{file}
+					{onFileChanged}
+					{onInvoke}
+					{items}
+					class="m-0 overflow-hidden rounded-b-2xl"
+				/>
+			{/if}
+			<div
```
ivyjeong13, line 55: Nit: Ah, just realized! I think with the absolute positioning, the scrollbar has been lost here and you can't scroll when your mouse is on the bar:

Before:
https://github.com/user-attachments/assets/80a28dc9-e948-4455-85e0-d1f52cb9cf3a

After:
https://github.com/user-attachments/assets/456e2569-bfb0-44bd-8702-4dc523bc05ef

ivyjeong13 (approved):
Other than background color mismatch, approving!

ivyjeong13:
Addresses #2400

PR #2424, opened 2025-04-08
fix: detect task calling cycles

thedadams:
It is possible for users to construct tasks that call other tasks. A problem arises when a task call cycle is accidentally constructed because the task will never complete. This change tracks task breadcrumbs so the task invoke tool can detect cycles.


StrongMonkey (approved):
LGTM, assuming the actual task tool will handle the BreadCrumb and detect the cycle?

PR #2428, opened 2025-04-09
fix: report errors for workflow executions

thedadams:
If a workflow execution failed for some reason, this was not being reported correctly in the external call information. It would look like the task would run forever.

This change will detect errors and report them on the external run output.

### DIFF
--- a/pkg/controller/handlers/workflowexecution/workflowexecution.go
+++ b/pkg/controller/handlers/workflowexecution/workflowexecution.go
@@ -27,10 +27,8 @@ func New(invoker *invoke.Invoker) *Handler {
 }
 
 func (h *Handler) UpdateRun(req router.Request, _ router.Response) error {
-	var (
-		we = req.Object.(*v1.WorkflowExecution)
-	)
-	if !(we.Status.State == types.WorkflowStateComplete && we.Spec.RunName != "") {
+	we := req.Object.(*v1.WorkflowExecution)
+	if we.Status.State != types.WorkflowStateComplete && we.Status.State != types.WorkflowStateError || we.Spec.RunName == "" {
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/handlers/workflowexecution/workflowexecution.go
```diff
@@ -27,10 +27,8 @@ func New(invoker *invoke.Invoker) *Handler {
 }
 
 func (h *Handler) UpdateRun(req router.Request, _ router.Response) error {
-	var (
-		we = req.Object.(*v1.WorkflowExecution)
-	)
-	if !(we.Status.State == types.WorkflowStateComplete && we.Spec.RunName != "") {
+	we := req.Object.(*v1.WorkflowExecution)
+	if we.Status.State != types.WorkflowStateComplete && we.Status.State != types.WorkflowStateError || we.Spec.RunName == "" {
```
thedadams, line 31: Also process this workflow execution if it is in an error state.

pkg/controller/handlers/workflowexecution/workflowexecution.go
```diff
@@ -49,9 +47,13 @@ func (h *Handler) UpdateRun(req router.Request, _ router.Response) error {
 	}
 
 	if run.Status.ExternalCall != nil && run.Status.ExternalCall.ID == we.Name {
+		data := we.Status.Output
+		if we.Status.Error != "" {
+			data = we.Status.Error
```
thedadams, line 52: If there is no output, then use the error value.

pkg/controller/handlers/workflowstep/invoke.go
```diff
@@ -54,15 +54,20 @@ func (h *Handler) RunInvoke(req router.Request, _ router.Response) error {
...
 
-func (h *Handler) setStepStateFromRun(step *v1.WorkflowStep, run *v1.Run) error {
+func (h *Handler) setStepStateFromRun(step *v1.WorkflowStep, run *v1.Run) {
 	switch run.Status.State {
 	case v1.Finished:
 		step.Status.State = types.WorkflowStateError
 		step.Status.LastRunName = step.Status.RunNames[0]
 		step.Status.Error = "Aborted"
+		if run.Status.Output != "" {
+			step.Status.Error += ": " + run.Status.Output
```
thedadams, line 69: If the run was aborted and provided some output, provide that as context for why the run was aborted.

PR #2450, opened 2025-04-09
fix: sanitize markdown html before rendering it

njhale:
Sanitize the HTML produced by user/llm sourced message markdown before
rendering it in the user UI. This should help prevent XSS attacks.

Addresses https://github.com/obot-platform/sensitive-issues/issues/8

Note: It looks like the admin ui is already using `rehype-sanitize` when displaying/editing markdown


tybalex (approved):
lint fails though

ivyjeong13 (approved):

PR #2453, opened 2025-04-10
Fix: don't assign slack receiver to triggers

StrongMonkey:
It turns out that assigning slack receiver name to trigger doesn't work well in copied obot, it is hard to trace the original slack receiver. Just relying on parent thread to delete should be fine for now.


thedadams:
Looks like `go generate` needs to be run on this.

StrongMonkey:
Oh, my bad. Stump other changes on it.

cjellick (approved):
Assuming donnie approves, im good with it. thanks for answering my questions about it

njhale (approved):

Issue #2369, opened 2025-04-02
"Submit" option in the credential authentication mode picker screen is not needed


sangee2004:
**Describe the bug**
"Submit" option in the credential authentication mode picker screen is not needed

**To Reproduce**
1. Create an obot with github tool
2. Use "Open Sidebar" option and click on "Credentials"
3. Add credential for GitHub tool 
4. In the following  authentication mode picker screen there is no need for "Submit" button.

<img width="304" alt="Image" src="https://github.com/user-attachments/assets/ba52ec41-d757-4ed4-b6f1-cf93d68991e7" />

Clicking on "PAT" or "OAUTH" option puts user into the workflow flow for authentication.

Also if submit button is clicked without doing any action, it puts the user into OAUTH workflow which is not intuitive.

**Expected behavior**
"Submit" option in the credential authentication mode picker screen can be removed.

cjellick:
Yeah, submit is not needed, but I guess it makes sense to keep cancel. While we're in here, can we do some restyling? In particular:
- I dont like that the buttons go however wide the modal is. Can we have a max width on them? 
- If we're dropping "Submit", the cancel button might be out of place. Maybe it's moved next to the other buttons and maybe the text could change "Nevermind, don't authenticate"
- Drop the "Authentication Method" header above the buttons as well as "The authentication method to use for this tool"
- "*The submitted contents are not visible to AI." doesnt need to be on this screen (but it should be on the next modal where the user actually prompted for creds
- "This tool has personal access token (PAT) and OAuth support. Select the authentication method you would like to use for this tool." ` change the wording to "This tool supports two ways to authentication: personal access tokens and OAuth. Select the method you'd like to use."

<img width="994" alt="Image" src="https://github.com/user-attachments/assets/c42d6ab0-a1a8-4db2-a7a1-2048c554c290" />


This is going to be a mixture of front and backend code. Note that if we make these changes, it's important that other forms of prompting/gathering credentials do not loose the functionality (they should still have field headers, submit, etc).

sangee2004:
Current authentication picker screen:
<img width="893" alt="Image" src="https://github.com/user-attachments/assets/b5bc36f8-2910-457f-bae4-3b7c7dba5fda" />

It addresses the following issues
1. Width of the buttons are fixed now and much smaller 
2. Submit option is not present and cancel option is changed to `Nevermind, don't authenticate"`
3. "Authentication Method" header is dropped
[...]

PR #2454, opened 2025-04-10
enhance: refactor authn method selection prompts

njhale:
Change how authentication method selection prompts are displayed in the
user UI so that:
- The submit button isn't displayed
- The field name and "not visible to AI" disclaimer are
  hidden
- The text of the cancel button is changed to "Nevermind, don't
  authenticate"
- Buttons are all the same width and are centered below the description

### e.g. In chat

https://github.com/user-attachments/assets/bfbeb3ad-1315-4361-8e8e-7e07c80b650f

### e.g. In drawer creds

https://github.com/user-attachments/assets/afdd52eb-a22c-4c1b-bfda-573e2555ce55

### e.g. In Obot "advanced configuration" menu

https://github.com/user-attachments/assets/797fb60e-7738-4ef4-9d2e-6430b4affab3

Note: The logic to use this formatting for the prompt form is predicated
on having a single field named "Authentication Method" that has defined
`options`. This is probably not the best way to discriminate here.
Another option is to add metadata to the prompt and plumb that through.
Would like to see what folks think of this approach first before I go
down that route.

Part of https://github.com/obot-platform/obot/issues/2369

### DIFF
--- a/ui/user/src/lib/components/messages/Message.svelte
+++ b/ui/user/src/lib/components/messages/Message.svelte
@@ -417,6 +417,10 @@
 
 {#snippet promptAuth()}
 	{#if msg.fields && !credentialsSubmitted}
+		{@const isAuthMethodSelectionPrompt =
+			msg.fields.length === 1 &&
+			msg.fields[0].name === 'Authentication Method' &&
+			msg.fields[0].options}
 		<form
 			class="flex flex-col gap-2"
 			onsubmit={(e) => {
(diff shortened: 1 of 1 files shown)


ui/user/src/lib/components/messages/Message.svelte
```diff
@@ -429,45 +433,72 @@
...
 									{option}
 								</button>
 							{/each}
+							{#if onSendCredentialsCancel}
+								<button
+									class="button-secondary whitespace-nowrap"
+									type="button"
+									onclick={() => onSendCredentialsCancel(msg.promptId ?? '')}
+								>
+									Nevermind, don't authenticate
```
g-linville:
```suggestion
									Never mind, don't authenticate
```

(Nirvana was wrong)

ui/user/src/lib/components/messages/Message.svelte
```diff
@@ -429,45 +433,72 @@
...
-					</button>
-				{/if}
-				<button class="button-primary" type="submit">Submit</button>
-			</div>
-			<span class="text-gray mt-1 flex grow items-end self-end text-sm"
-				>*The submitted contents are not visible to AI.</span
-			>
+				<span class="text-gray mt-1 flex grow items-end self-end text-sm">
+					*The submitted contents are not visible to AI.
+				</span>
```
g-linville, line 500: Tbh I'm still not a huge fan of having this. I'm not sure if we currently have any tools that do this, but there may be times when we want to hand the result of a prompt directly to the LLM (like in the survey tool that we had at the end of the hack-a-thon).

g-linville (changes requested):

g-linville:
> Note: The logic to use this formatting for the prompt form is predicated
on having a single field named "Authentication Method" that has defined
options. This is probably not the best way to discriminate here.
Another option is to add metadata to the prompt and plumb that through.
Would like to see what folks think of this approach first before I go
down that route.

I'm fine with this.

ui/user/src/lib/components/messages/Message.svelte
njhale: yeah, I'm wondering if this should be an option controlled by the tool generating the prompt.

g-linville (approved):

ui/user/src/lib/components/messages/Message.svelte
```diff
@@ -417,6 +417,10 @@
 
 {#snippet promptAuth()}
 	{#if msg.fields && !credentialsSubmitted}
+		{@const isAuthMethodSelectionPrompt =
```
ivyjeong13, line 420: Just a super small thing but wondering if the background color should not be there when it's an auth prompt in chat?

When Obot replies, it usually doesn't have a bubble, but when I type something in it does, so maybe that's what is tripping me up. 


Screenshot 2025-04-10 at 10 53 48 AM

ui/user/src/lib/components/messages/Message.svelte
```diff
@@ -429,45 +433,72 @@
...
 			{@html toHTMLFromMarkdown(msg.message.join('\n'))}
 
-			{#each msg.fields as field}
-				<div class="flex flex-col gap-1">
-					<label for={field.name} class="mt-1 text-sm font-medium">{field.name}</label>
-					{#if field.options}
-						<div class="flex flex-col gap-2">
-							{#each field.options as option}
-								<button class="button" onclick={() => (promptCredentials[field.name] = option)}>
+			{#if isAuthMethodSelectionPrompt && msg.fields && msg.fields[0].options}
```
ivyjeong13, line 436: Screenshot 2025-04-10 at 11 03 06 AM

Nit (or out of scope): Realizing in the advanced configuration and drawer creds, you can briefly lose the context of what tool you're setting credentials for. From the screenshot I don't know which tool I'm authing unless I click PAT (OAuth is also ambiguous until I click the link) Would be nice to always have that context in case the user walks away or memory lapse lol

ivyjeong13 (approved):

Issue #2269, opened 2025-03-24
Add thread id column to admin chat threads ui


cjellick:
Do this for all our tables in the admin UI

cjellick:
Here is a link to the slack thread where this was discussed: https://acorn-io.slack.com/archives/C07P0EZR917/p1744212533105879

As mentioned in the loom there, there are a few other things I want as well:
1. the column should be filter-able
2. we should add IDs to our other tables too (this is also mentioned in the issue description)
3. Add thread name as a column as well

sangee2004:
All tables now have ID column - `Base Agent`,  `Obots`,  `Chat Threads`,  `Tasks`, and `Task Runs`
`Chat Threads` page has ID and Thread Name field. Both these fields are filterable and user can search using id and thread name.

Task and Task Runs are not filterable/searchable by id which is tracked in https://github.com/obot-platform/obot/issues/2498

PR #2458, opened 2025-04-10
feat: add thread id column to admin chat threads ui

tony-dotnet-fullstack:


ui/admin/app/components/composed/Filters.tsx
```diff
@@ -60,6 +64,18 @@ export function Filters({
 		};
 
 		return [
+			"threadId" in filters &&
+				filters.threadId &&
+				threadMap && {
```
ivyjeong13, line 69: nit: Can lessen the threadMap.get calls with something like:
```
		const filteredThread = filters.threadId
			? threadMap?.get(filters.threadId)
			: null;

		return [
			filteredThread && {
				key: "threadId",
				label: "Thread",
				value: filteredThread?.id
					? [filteredThread.id, filteredThread.name].filter(Boolean).join(" - ")
					: filters.threadId,
				onRemove: () => deleteFilters("threadId"),
			},
```

ivyjeong13:
Addresses #2269

Issue #2254, opened 2025-03-21
add warning about using shared obots


cjellick:
If an obot is shared, but not featured (featured indicates we "own" it) and a user uses it(and they are not the owner/author, then we should expose a warning to the user as a modal on first launch explaining that it was created by a third party

sangee2004:
Following Warning page is presented to the user when using a shared link to launch an obot

<img width="572" alt="Image" src="https://github.com/user-attachments/assets/163b06e6-d190-4454-932a-ee34978b6207" />

If "I understand" option is picked , obot gets created.
If "Go Hone" option is picked , obot does not get created and user is taken to the home page (`My Obots` page).


Even after the user has accessed the share link and created the obot, using the shared link again results in the warning page shown.

@ivyjeong13 Reopening this issue to not have the warning page presented to the user if shared link is re visited when user already has an obot instance created for the shared link.

ivyjeong13:
```
Reopening this issue to not have the warning page presented to the user if shared link is re visited when user already has an obot instance created for the shared link.
```

This should be addressed (forgot which PR it was though, will attach once I find it!)

sangee2004:
This issue is addressed when testing with latest builds.

Warning page is not presented to the user if shared link is re visited when user already has an obot instance created for the shared link.

PR #2471, opened 2025-04-10
enhance: shared obot warning

ivyjeong13:
Related Loom: https://www.loom.com/share/beed6d3f3f254215983bd950104b59ab
Addresses #2254 

Screenshot 2025-04-10 at 4 11 45 PM
Screenshot 2025-04-10 at 4 11 37 PM

### DIFF
--- a/pkg/api/authz/authz.go
+++ b/pkg/api/authz/authz.go
@@ -72,6 +72,7 @@ var staticRules = map[string][]string{
 		"POST /api/image/generate",
 		"POST /api/image/upload",
 		"POST /api/logout-all",
+		"GET /api/shares/{share_public_id}",
 	},
 	MetricsGroup: {
 		"/debug/metrics",
--- a/pkg/api/authz/authz.go
+++ b/pkg/api/authz/authz.go
@@ -72,7 +72,6 @@ var staticRules = map[string][]string{
 		"POST /api/image/generate",
 		"POST /api/image/upload",
 		"POST /api/logout-all",
-		"GET /api/shares/{share_public_id}",
 	},
 	MetricsGroup: {
 		"/debug/metrics",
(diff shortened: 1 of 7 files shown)


pkg/api/handlers/projectshare.go
```diff
@@ -257,3 +256,23 @@ func (h *ProjectShareHandler) CreateProjectFromShare(req api.Context) error {
 
 	return req.Write(convertProject(newProject))
 }
+
+func (h *ProjectShareHandler) GetShareFromShareID(req api.Context) error {
```
ivyjeong13, line 260: Added `GET /api/shares/{share_public_id}` and wanted to make sure that was the right thing to do. Wanted to be able to get the Share based on the publicID without creating one and be able to display the shared obot warning if it isn't featured.

pkg/api/authz/authz.go
```diff
@@ -72,6 +72,7 @@ var staticRules = map[string][]string{
 		"POST /api/image/generate",
 		"POST /api/image/upload",
 		"POST /api/logout-all",
+		"GET /api/shares/{share_public_id}",
```
thedadams: I believe this exposes too much information to non-admin users. I believe this would make it possible for a non-admin user to see thread shares that don't give access to anything.

For example, if Ivy shares a thread with me, then non-admin user Daishan would be able to see and inspect the thread share. Daishan should not be able to see the thread share.

You can remove this path here and add it here instead: https://github.com/obot-platform/obot/blob/a285e06e5bdb4f34bad1c7acf4c38a141d68e456/pkg/api/authz/resources.go#L110 I believe the framework will take care of this case for us. If that is not the case, please let me know and I will take a closer look.

thedadams (changes requested):

pkg/api/authz/authz.go
ivyjeong13: Thanks Donnie! Made the change and took a closer look at resources.go to get familiar, I see the checkThreadShare in there! Appreciated!

thedadams (approved):

PR #2483, opened 2025-04-11
feat: allow users to delete their account

tony-dotnet-fullstack:
Allow users to delete their account
Addresses [#2412](https://github.com/obot-platform/obot/issues/2412)

### DIFF
--- a/ui/user/src/routes/profile/+page.svelte
+++ b/ui/user/src/routes/profile/+page.svelte
@@ -102,34 +102,27 @@
 						alt=""
 						class="mx-auto mb-3 h-28 w-28 rounded-full object-cover"
 					/>
-					<div class="flex flex-row py-4">
+					<div class="flex flex-row py-3">
 						<div class="w-1/2 max-w-[150px]">Display Name:</div>
 						<div class="w-1/2 break-words">{profile.current.getDisplayName?.()}</div>
 					</div>
 					<hr />
-					<div class="flex flex-row py-2">
+					<div class="flex flex-row py-3">
 						<div class="w-1/2 max-w-[150px]">Email:</div>
 						<div class="w-1/2 break-words">{profile.current.email}</div>
 					</div>
 					<hr />
-					<div class="flex flex-row py-2">
+					<div class="flex flex-row py-3">
 						<div class="w-1/2 max-w-[150px]">Username:</div>
 						<div class="w-1/2 break-words">{profile.current.username}</div>
 					</div>
 					<hr />
-					<div class="flex flex-row py-2">
+					<div class="flex flex-row py-3">
 						<div class="w-1/2 max-w-[150px]">Role:</div>
 						<div class="w-1/2 break-words">{profile.current.role === 1 ? 'Admin' : 'User'}</div>
 					</div>
 					<hr />
-					<div class="flex flex-row py-2">
-						<div class="w-1/2 max-w-[150px]">AuthProvider:</div>
-						<div class="w-1/2 break-words">
-							{profile.current.currentAuthProvider?.split('-')[0]}
-						</div>
-					</div>
-					<hr />
-					<div class="mt-2 flex flex-col gap-4 py-2">
+					<div class="mt-2 flex flex-col gap-4 py-3">
 						{#if version.current.sessionStore === 'db'}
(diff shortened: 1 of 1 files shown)


ui/user/src/routes/profile/+page.svelte
```diff
@@ -0,0 +1,132 @@
...
+		class="colors-background relative flex w-full max-w-(--breakpoint-2xl) flex-col justify-center md:pb-12"
+	>
+		<div class="mt-8 flex w-full flex-col gap-8">
+			<div class="flex w-full flex-col gap-4">
+				<div
+					class="sticky top-0 z-30 flex items-center gap-4 bg-white px-4 py-4 md:px-12 dark:bg-black"
+				>
+					<h3 class="flex flex-shrink-0 text-2xl font-semibold">My Account</h3>
+				</div>
+				<div class="mx-auto max-w-sm rounded-xl bg-white p-6 shadow-md">
```
ivyjeong13: dark mode: 

Screenshot 2025-04-14 at 10 22 46 AM

Maybe a bg-surface1 instead to have the card pop out from the background that is bg-black/bg-white?

ui/user/src/routes/profile/+page.svelte
```diff
@@ -0,0 +1,132 @@
...
+						<div class="w-1/2">{profile.current.getDisplayName?.()}</div>
+					</div>
+					<hr />
+					<div class="flex flex-row py-2">
+						<div class="w-1/2">Email:</div>
+						<div class="w-1/2 break-words">{profile.current.email}</div>
+					</div>
+					<hr />
+					<div class="flex flex-row py-2">
+						<div class="w-1/2">Username:</div>
```
ivyjeong13: Screenshot 2025-04-14 at 10 25 05 AM

Looks like a long username breaks out of this card, maybe a truncate?

ui/user/src/routes/profile/+page.svelte
```diff
@@ -0,0 +1,132 @@
...
+					</div>
+				</div>
+			</div>
+		</div>
+	</main>
+
+	<Notifications />
+</div>
+
+<Confirm
```
ivyjeong13, line 151: Screenshot 2025-04-14 at 10 26 39 AM

Probably Will or Craig input, but maybe a blurb here to nail down the deletion is permanent and cannot be restored? Maybe could be a checkbox or a type your username to confirm delete too? Kind of a way to avoid accidental decision and letting the user know what will happen.

tony-dotnet-fullstack: image
image
image
image
I've attached some screenshots of the updated UI.

tony-dotnet-fullstack: Here is the updated UI.
image
image

ui/user/src/routes/profile/+page.svelte
```diff
@@ -0,0 +1,176 @@
...
+						<div class="w-1/2 break-words">{profile.current.username}</div>
+					</div>
+					<hr />
+					<div class="flex flex-row py-2">
+						<div class="w-1/2 max-w-[150px]">Role:</div>
+						<div class="w-1/2 break-words">{profile.current.role === 1 ? 'Admin' : 'User'}</div>
+					</div>
+					<hr />
+					<div class="flex flex-row py-2">
+						<div class="w-1/2 max-w-[150px]">AuthProvider:</div>
```
ivyjeong13: nit: I agree with @g-linville that the Auth Provider isn't needed here. It shows the current auth provider they logged in with but I can't remember if this is case, couldn't they technically have both a Github and Google Sign-On associated with the same email/same account?

g-linville: Yes, Google and GitHub logins get tied to the same underlying user in Obot, as long as they both use the same email address.

ivyjeong13: Thanks Grant! Yeah, then let's axe this row!

ivyjeong13 (approved):
Pulled down and tested. Would just double check with @cjellick that you have the green light to merge this in. Don't think the narrow well vs. full-width is a blocker but doesn't hurt to make sure.

Issue #2439, opened 2025-04-09
Delete workfile confirmation popup window gets hidden behind the file list


sangee2004:
**Describe the bug**

**To Reproduce**
1. Create an obot
2. Upload few files (5 files in my case) using the "File" option
3. Try to delete one of the uploaded files.

Notice that the delete workfile confirmation popup window gets hidden behind the file list
<img width="1373" alt="Image" src="https://github.com/user-attachments/assets/b79f78c5-c4e4-403a-9aa9-44241e344e04" />

**Expected behavior**

Delete workfile confirmation popup window should be shown on top of the file flisting.

sangee2004:
Issue mentioned in https://github.com/obot-platform/obot/issues/2439#issuecomment-2797047603 is addressed now.

@rob-fullstack-star When deleting a file ,  file list is not shown anymore and delete workfile conformation popup window is shown . The `file` helper text continues to be present even after the file list view is dismissed.

<img width="1214" alt="Image" src="https://github.com/user-attachments/assets/d72ee331-211d-40b8-b99a-a03b72106399" />

Reopening this issue to address this.

cjellick:
I agree - I'd like the helper text to not be displayed when the delete modal opens

sangee2004:
When testing with latest builds , `File` helper text is not presented when delete workfile conformation popup window is shown.

PR #2487, opened 2025-04-11
Fix: Delete workfile confirmation popup style issue

tony-dotnet-fullstack:
Delete workfile confirmation popup window gets hidden behind the file list
Addresses [#2439](https://github.com/obot-platform/obot/issues/2439)

### DIFF
--- a/ui/user/src/lib/components/edit/Files.svelte
+++ b/ui/user/src/lib/components/edit/Files.svelte
@@ -173,10 +173,7 @@
 							class="icon-button-small invisible ms-2 group-hover:visible"
 							onclick={() => {
 								fileToDelete = file.name;
-								const element = document.querySelector('#click-catch');
-								if (element instanceof HTMLElement) {
-									element.click();
-								}
+								menu?.toggle(false);
 							}}
 						>
 							<Trash class="text-gray h-5 w-5" />


ui/user/src/lib/components/edit/Files.svelte
```diff
@@ -253,3 +257,9 @@
 	onsuccess={deleteFile}
 	oncancel={() => (fileToDelete = undefined)}
 />
+
+<style>
+	.group:hover .group-hover\:visible {
```
ivyjeong13, line 259: Should this be here? These are tailwind utilities classes

tony-dotnet-fullstack: Tailwind utility classes(.group-hover:visible) aren't working in the Ubuntu browser. 
I've attached a screen recording to the issue ticket: [#2439](https://github.com/obot-platform/obot/issues/2439).

ivyjeong13: Ah okay, would opacity work instead then? Instead of using visible/invisible opacity-100/opacity-0?

ui/user/src/lib/components/edit/Files.svelte
```diff
@@ -173,6 +173,10 @@
 							class="icon-button-small invisible ms-2 group-hover:visible"
 							onclick={() => {
 								fileToDelete = file.name;
+								const element = document.querySelector('#click-catch');
```
ivyjeong13: Just wondering, does `menu?.toggle(false)` not work here?

ui/user/src/lib/components/edit/Files.svelte
tony-dotnet-fullstack: I've already tried using opacity-100 and opacity-0, but that didn't work either. 
I also need to define a custom style.

ivyjeong13: I see! Thanks for checking and yeah, just wondering cause there are other places that utilize this (either visibility or opacity) so worried if they're affected too and if this is a broader issue. 

I did see someone mention their laptop is touchscreen and tailwind v4 "thinks" they don't have hover support

The Tailwind v4 change:
```
[Hover styles on mobile](https://tailwindcss.com/docs/upgrade-guide#hover-styles-on-mobile)
In v4 we've updated the hover variant to only apply when the primary input device supports hover:

CSS
@media (hover: hover) {
  .hover\:underline:hover {
    text-decoration: underline;
  }
}
This can create problems if you've built your site in a way that depends on touch devices triggering hover on tap. If this is an issue for you, you can override the hover variant with your own variant that uses the old implementation:

CSS
@custom-variant hover (&:hover);
Generally though we recommend treating hover functionality as an enhancement, and not depending on it for your site to work since touch devices don't truly have the ability to hover.
```

Will see if I can find anybody reporting issues with Tailwind & Ubuntu, hopefully they have a solution.

ivyjeong13 (approved):

PR #2489, opened 2025-04-11
fix: allow user to delete their account from the UI

thedadams:
There were a couple issues keeping a user from being able to do this. First, when encryption is enabled, the user was not able to remove their account because it could not be found. This change switches the delete user method to use the hashed_username field.

Secondly, when the user successfully removed their account, the cookie they used to authenticate would still be valid. The UI would immediately called /api/me and the user would be authenticated and a new user created. This change tells the browser to remove the cookie so this does not happen.

### DIFF
--- a/pkg/gateway/client/user.go
+++ b/pkg/gateway/client/user.go
@@ -82,7 +82,7 @@ func (c *Client) UserByID(ctx context.Context, id string) (*types.User, error) {
 func (c *Client) DeleteUser(ctx context.Context, username string) (*types.User, error) {
 	existingUser := new(types.User)
 	if err := c.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
-		if err := tx.Where("username = ?", username).First(existingUser).Error; err != nil {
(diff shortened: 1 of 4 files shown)


pkg/gateway/client/user.go
```diff
@@ -82,7 +82,7 @@ func (c *Client) UserByID(ctx context.Context, id string) (*types.User, error) {
 func (c *Client) DeleteUser(ctx context.Context, username string) (*types.User, error) {
 	existingUser := new(types.User)
 	if err := c.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
-		if err := tx.Where("username = ?", username).First(existingUser).Error; err != nil {
+		if err := tx.Where("hashed_username = ?", hash.String(username)).First(existingUser).Error; err != nil {
```
StrongMonkey, line 85: Does this work when encryption is off?

njhale (approved):

StrongMonkey (approved):

pkg/gateway/client/user.go
thedadams: Yes, it does. We always have the hashed_username.

Issue #2384, opened 2025-04-03
No tools shown for obots created from featured obots in `My obots` view.


sangee2004:
**Describe the bug**
No tools shown for obots created from featured obots in `My obots` view.

**To Reproduce**
1. Go to "View Obot Catalog"
2. Click on any featured obot - "Slack2Gitbub" in my case
3. Go to home page and look for the `Slack2Gitbub` obot that gets created in `My Obots` list

Notice that no tools are shown

`My obots` list
<img width="479" alt="Image" src="https://github.com/user-attachments/assets/e5bec85b-215a-4fd8-b549-20b36b7c56e4" />

Obot Catalog `Featured` list that shows the built in tools 
<img width="659" alt="Image" src="https://github.com/user-attachments/assets/87f142a2-df27-4740-a153-ec94cf7c250c" />

**Expected behavior**
All tools should be shown for this obot.

sangee2004:
This issue is fixed when testing with latest builds

Tools for obots created from featured obots are presented in the `My Obots` view.
<img width="431" alt="Image" src="https://github.com/user-attachments/assets/59e1f7ef-ba8b-4dcf-8a70-26cdf32993c5" />

PR #2497, opened 2025-04-11
fix: include built-in tools for homepage obot cards

ivyjeong13:
Addresses #2384 

## Before:

https://github.com/user-attachments/assets/b0ab98ae-9f6b-4506-9128-8374b587aac4

## After:

https://github.com/user-attachments/assets/ea4b47a4-b502-493e-a550-65c2996ed717

### DIFF
--- a/pkg/api/handlers/projects.go
+++ b/pkg/api/handlers/projects.go
@@ -509,9 +509,7 @@ func convertProject(thread *v1.Thread, req api.Context) types.Project {
 	if thread.Spec.ParentThreadName != "" {
 		var parentThread v1.Thread
 		if err := req.Get(&parentThread, thread.Spec.ParentThreadName); err == nil {
-			for _, tool := range parentThread.Spec.Manifest.Tools {
-				p.Tools = append(p.Tools, tool)
-			}
+			p.Tools = append(p.Tools, parentThread.Spec.Manifest.Tools...)
 		}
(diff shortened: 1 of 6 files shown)


pkg/api/handlers/projects.go
```diff
@@ -504,6 +504,15 @@ func convertProject(thread *v1.Thread) types.Project {
 		UserID:          thread.Spec.UserID,
 		Capabilities:    types.ProjectCapabilities(thread.Spec.Capabilities),
 	}
+
+	// Include tools from parent project
+	if thread.Spec.ParentThreadName != "" {
```
ivyjeong13: Wanting to check with backend experts, this is right/okay thing to do? 😄

thedadams: I think this is good. The only thing is that we don't want to use a client in these `convert` functions. It would be better if this function was called with a `thread` and a `parentThread` and we check here if `parentThread != nil`.

thedadams (changes requested):

pkg/api/handlers/projects.go
ivyjeong13: Thanks Donnie! Updated it to `func convertProject(thread *v1.Thread, parentThread *v1.Thread)`

thedadams (approved):

Issue #2381, opened 2025-04-03
PDF Content Not Returned to Context


StrongMonkey:
**Issue Summary**

It appears that the functionality to return the contents of a PDF to the context is no longer working as expected. When a PDF is uploaded, its content should be read into the knowledge base and returned to the context. However, this process seems to be failing, as evidenced by the recent discussions in the Slack thread.

**Steps to Reproduce**
1. Upload a PDF file via the normal UI-based upload.
2. Observe whether the content of the PDF is returned to the context.

**Expected Behavior**
The content of the PDF should be read and returned to the context after being uploaded.

**Actual Behavior**
The content of the PDF is not being returned to the context.

**Additional Context**
- The issue was discussed in a Slack thread, which can be found [here](https://acorn-io.slack.com/archives/C03FU91U4SX/p1743632140588549?thread_ts=1743632140.588549&cid=C03FU91U4SX).
- Users have previously used pdf2image and image analyzer with success, but the current method seems to be failing.

**Request**
Please investigate the issue and restore the functionality to ensure that PDF contents are correctly returned to the context.

shweta8996:
I have verified on obot.ai and it is working fine. 
Steps:
-  I uploaded couple of PDFs and Obot is able to provide correct context after each upload. 
- Asked question related to PDF like 'provide summary for the file uploaded'  , and Obot was able to provide the data. 

Please refer the attachment:

PR #2531, opened 2025-04-14
fix: generate knowledge summaries from more than the first page

njhale:
Knowledge summaries were being generated from only the first page of
content. This resulted in poor summaries for documents with little or
no information on the first page (e.g. slide deck PDF starting with a title
slide). To fix this, gather content from each page until the 3000
character knowledge summary limit is reached.

Addresses https://github.com/obot-platform/obot/issues/2381


StrongMonkey (approved):

tybalex (approved):
why is the limit=3000 though

njhale:
> why is the limit=3000 though

@tybalex I'm assuming this was an arbitrary limit chosen to keep summaries to a reasonable length.

Issue #2507, opened 2025-04-14
Instructions: Clicking into large textarea scrolls to top


ivyjeong13:
**Describe the bug**
When trying to click at the end of text content in Instructions, and it already has a lot of text, it focuses to the top of the input instead of where I had clicked. This is annoying when trying to modify Instructions, having to scroll down constantly when I click into it to modify the Instructions at the bottom. 

**To Reproduce**
1. Go to a Obot with already a large text in 'Instructions'
2. Click the area of the very last text
3. See it scroll to top of input and have to scroll down to see where your typing indicator is actually set (at the bottom)

**Expected behavior**
Textarea should stay in focus of where I set my typing indicator

**Screenshots**
https://www.loom.com/share/2d7712b52dd043b6b26e393e9858f6f6 from Slack thread: https://acorn-io.slack.com/archives/C07P0EZR917/p1744391152350449

**Additional context**

ivyjeong13:
@StrongMonkey is working on this atm, assigning to you Daishan!

PR #2536, opened 2025-04-15
Fix: fix unexpected scroll behavior in textarea

StrongMonkey:
https://github.com/obot-platform/obot/issues/2507

### DIFF
--- a/ui/user/src/lib/actions/textarea.ts
+++ b/ui/user/src/lib/actions/textarea.ts
@@ -1,38 +1,63 @@
-export function resize(node: HTMLTextAreaElement) {
-	node.style.height = 'auto';
-	node.style.height = node.scrollHeight + 'px';
-}
-
 export function autoHeight(node: HTMLTextAreaElement) {
+	const scrollContainer = findScrollContainer(node);
+	let lastValue = node.value;
+
+	const resize = () => {
+		if (node.value === lastValue) return;
+
+		lastValue = node.value;
+		const prevScrollTop = scrollContainer?.scrollTop ?? 0;
+		const { selectionStart, selectionEnd } = node;
+
+		node.style.height = 'auto';
+		node.style.height = node.scrollHeight + 'px';
+
+		requestAnimationFrame(() => {
+			if (selectionStart != null && selectionEnd != null) {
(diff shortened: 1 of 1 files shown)


ui/user/src/lib/actions/textarea.ts
```diff
@@ -1,38 +1,63 @@
...
+
 	return {
 		destroy() {
 			resizeObserver.disconnect();
+			mutationObserver.disconnect();
 		}
 	};
 }
+
+function findScrollContainer(el: HTMLElement): HTMLElement | null {
```
ivyjeong13, line 53: Wondering if this is unnecessary? I think we should be using this action on a `textarea` and setting that with a scroll overflowY if we're using autoHeight

ui/user/src/lib/actions/textarea.ts
```diff
@@ -1,38 +1,63 @@
-export function resize(node: HTMLTextAreaElement) {
-	node.style.height = 'auto';
-	node.style.height = node.scrollHeight + 'px';
-}
-
 export function autoHeight(node: HTMLTextAreaElement) {
+	const scrollContainer = findScrollContainer(node);
```
ivyjeong13, line 2: https://github.com/user-attachments/assets/f9558dd3-cea2-4a73-bd16-3c27de9584fa

Looks like it turns the text area into a crazy long height for me?

ui/user/src/lib/actions/textarea.ts
StrongMonkey: I think the problem here is that the parent div is scrollable, not the textarea. I tried to set on textarea and it didn't work.

ui/user/src/lib/actions/textarea.ts
StrongMonkey: Hmm, can't see your video. but the height looks fine to me?

image

ivyjeong13: > Hmm, can't see your video. but the height looks fine to me?
> 
> image

Oh oops, sorry! Didn't realize the video isn't working, here's a Loom!

https://www.loom.com/share/f94268d7ccf24c6a941366b928a02296

Issue #2539, opened 2025-04-15
Modifications to Shared Tasks Revert to Original State


mighty-pup-acorn:
### Issue Summary

When a task is shared by someone, any modifications made to the task seem to revert back to the original task’s step. This behavior is unexpected and may hinder collaboration.

### Steps to Reproduce
1. Share a task with another user.
2. Attempt to modify the task.
3. Observe that the modifications revert back to the original task’s step.

### Expected Behavior
Modifications to a shared task should be retained and not revert to the original state.

### Actual Behavior
Modifications revert back to the original task’s step.

### Additional Information
This issue was reported by a user in a Slack thread. Further investigation is needed to determine the root cause and potential solutions.

sangee2004:
This issue is not seen when testing with latest builds:

1. As user1 , Create an obot. 
2. Create a task and make it sharable
3. Share the obot
4. As a different user user2 , access the share link from step 3.
5. User will get the task created in step 2.
6. As user1 , update the task -> change the task name , description and step 
This change made to task gets saved and persists as expected.
This change does not get copied over to user2's shared obot as expected 
User2 has task as created from step1

7. As user2 , update the task -> change the task name , description and step 
This change made to task gets saved and persists as expected.
This change does not get copied over to user1's  obot as expected 
User1 has task as in step 6 where he made updates.

PR #2540, opened 2025-04-15
Fix: only copy task from shared parent once

StrongMonkey:
This PR fixed an issue so that if a task is shared from a shared project, modification to the task will not be reverted.

https://github.com/obot-platform/obot/issues/2539


g-linville (approved):

cjellick:
i dont know what darren's intended behavoir is. trying to get at that

cjellick (approved):

PR #2542, opened 2025-04-15
enhance: add clickOutside to dialogs

ivyjeong13:


g-linville (approved):
Does the clickOutside thing also make it close when you hit ESC? I find that useful.

ivyjeong13:
> Does the clickOutside thing also make it close when you hit ESC? I find that useful.

Luckily ESC should already be supported for `dialog` elements! It's the ones that don't use `dialog` (using the use:popover action instead) that don't have it. But vice versa, they already support click outside to close :sweat_smile: 

But I am looking to consolidate and get everything to use the same thing so we're not fighting two different ways to get dialog/popovers : )

PR #2544, opened 2025-04-15
chore: add retention to chart and docs; add retention migration

g-linville:
Forgot to take care of these earlier.

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -88,7 +88,7 @@ config:
   # config.OBOT_SERVER_HOSTNAME -- The hostname of your Obot instance, including protocol
   OBOT_SERVER_HOSTNAME: ""
   # config.OBOT_SERVER_RETENTION_POLICY_HOURS -- The retention policy for the system. Set to 0 to disable retention. Default is 2160 (90 days) if left blank. This field should just be a number in a string, no `h` suffix.
-  OBOT_SERVER_RETENTION_POLICY_HOURS: "2160"
+  OBOT_SERVER_RETENTION_POLICY_HOURS: ""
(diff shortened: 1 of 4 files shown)


chart/values.yaml
```diff
@@ -87,6 +87,8 @@ config:
   OBOT_SERVER_DSN: ""
   # config.OBOT_SERVER_HOSTNAME -- The hostname of your Obot instance, including protocol
   OBOT_SERVER_HOSTNAME: ""
+  # config.OBOT_SERVER_RETENTION_POLICY_HOURS -- The retention policy for the system. Set to 0 to disable retention. Default is 2160 (90 days) if left blank. This field should just be a number in a string, no `h` suffix.
+  OBOT_SERVER_RETENTION_POLICY_HOURS: "2160"
```
thedadams: nit: if we are going to use the default value, then I say we leave this blank to avoid confusion.

pkg/controller/handlers/retention/retention.go
```diff
@@ -61,3 +62,19 @@ func RunRetention(policy time.Duration) func(req router.Request, resp router.Res
...
+// Migrate sets the last used time of the thread to the current time if it is not set.
+func Migrate(req router.Request, _ router.Response) error {
+	thread := req.Object.(*v1.Thread)
+
+	if thread.Spec.SystemTask {
+		return nil
+	}
+
+	if thread.Status.LastUsedTime.IsZero() {
+		thread.Status.LastUsedTime = metav1.NewTime(time.Now())
```
thedadams: nit:
```suggestion
		thread.Status.LastUsedTime = metav1.Now()
```

g-linville: fixed

chart/values.yaml
g-linville: fixed

StrongMonkey (approved):

PR #2545, opened 2025-04-15
fix: authz: enable admins to list all threads on projects

g-linville:


StrongMonkey (approved):

pkg/api/handlers/projects.go
```diff
@@ -623,10 +623,18 @@ func (h *ProjectsHandler) ListProjectThreads(req api.Context) error {
...
+			"spec.project":          "false",
+			"spec.parentThreadName": projectThread.Name,
+		}
+	} else {
+		selector = kclient.MatchingFields{
+			"spec.project":          "false",
+			"spec.parentThreadName": projectThread.Name,
+			"spec.userUID":          req.User.GetUID(),
+		}
 	}
```
njhale: nit: could you do something like this instead?

```suggestion
	selector := kclient.MatchingFields{
	    "spec.project":          "false",
	    "spec.parentThreadName": projectThread.Name,
	}
	if !req.UserIsAdmin() {
		selector["spec.userUID"] = req.User.GetUID()
	}
```

njhale (approved):

pkg/api/handlers/projects.go
g-linville: Good call, done

Issue #2465, opened 2025-04-10
Change task runs table columns


cjellick:
In the admin UI, task runs table columns should be changed/updated:
1. As mentioned elsewhere, add an ID column
2. Drop the user column
3. Created column is fine to keep, but I don't like the "X minutes/days ago" paradigm. Can I get the actually date/time (adjusted for the viewers local timezone)
4. Add filters to all the columns

shweta8996:
I have verified on main URL and it is working fine as expected.
As mentioned elsewhere, add an ID column > **Done**
Drop the user column > **Done**
But, I noticed some points are not working fine.
Created column is fine to keep, but I don't like the "X minutes/days ago" paradigm. Can I get the actually date/time (adjusted for the viewers local timezone) >The date is not initially displayed in the 'Created' column, but when I apply a filter on the time column, the full date becomes visible.
Add filters to all the columns > Not exist data is displayed in filter list.

Please refer attachment:

https://github.com/user-attachments/assets/5fca9ee1-aa66-4bfc-bed2-8f18133638bc

shweta8996:
I have verified on main URL and it is working fine as expected.
Created column is fine to keep, but I don't like the "X minutes/days ago" paradigm. Can I get the actually date/time (adjusted for the viewers local timezone) > Done
Add filters to all the columns > Done
Please refer attachment:

https://github.com/user-attachments/assets/eae67582-a104-4ac1-af9c-f0d2e3c53f6f

PR #2558, opened 2025-04-16
Fix: Validate the filter option

tony-dotnet-fullstack:
Change task runs table columns
Addresses [#2465](https://github.com/obot-platform/obot/issues/2465)

### DIFF
--- a/ui/admin/app/routes/_auth.task-runs._index.tsx
+++ b/ui/admin/app/routes/_auth.task-runs._index.tsx
@@ -9,6 +9,7 @@ import {
 import { $path } from "safe-routes";
 import useSWR, { preload } from "swr";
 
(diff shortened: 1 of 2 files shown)


ui/admin/app/routes/_auth.task-runs._index.tsx
```diff
@@ -222,7 +235,7 @@ export default function TaskRuns() {
 						key={column.id}
 						field="Task"
 						values={
-							getTasks.data?.map((task) => ({
+							taskList?.map((task) => ({
```
ivyjeong13, line 238: small nit: The `?` and ` ?? []` isn't needed since looks like `taskList` always return an array

ivyjeong13 (approved):

PR #2562, opened 2025-04-16
Chore: Add necessary library and permission to upload files in client

StrongMonkey:
This PR allows the obot token passed to run to upload file/knowledge files that are used in https://github.com/obot-platform/tools/pull/587

https://github.com/obot-platform/obot/issues/2563

### DIFF
--- a/apiclient/thread.go
+++ b/apiclient/thread.go
@@ -1,6 +1,7 @@
 package apiclient
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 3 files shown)


apiclient/thread.go
```diff
@@ -82,3 +83,21 @@ func (c *Client) ListThreads(ctx context.Context, opts ListThreadsOptions) (resu
...
 	return
 }
+
+func (c *Client) UploadThreadFile(ctx context.Context, projectID string, threadID string, assistantID string, filename string, fileContent []byte) error {
+	_, resp, err := c.doRequest(ctx, http.MethodPost, fmt.Sprintf("/assistants/%s/projects/%s/threads/%s/files/%s", assistantID, projectID, threadID, filename), bytes.NewReader(fileContent))
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	return nil
```
g-linville, line 93: Does this also work?

```suggestion
	defer resp.Body.Close()
	return err
```

Just a bit more concise if it does. This could also be changed in the other function in this PR.

g-linville (approved):

apiclient/thread.go
StrongMonkey: yeah I think it works, resp.body should always be non-nil

thedadams: I don't think this works because `resp` is probably `nil` when a non-nil error is returned.

thedadams (approved):

apiclient/thread.go
StrongMonkey: Oh. donnie is correct

PR #2570, opened 2025-04-16
Fix: hide helper text when opening the delete modal

rob-fullstack-star:
Delete workfile confirmation popup window gets hidden behind the file list
Addresses [#2439](https://github.com/obot-platform/obot/issues/2439)


ivyjeong13:
@rob-fullstack-star https://github.com/obot-platform/obot/pull/2573 Can you check if this change fixes it? Some of the changes here didn't seem necessary. I just axed this from the structure: `<div role="dialog" class="relative max-h-full w-full max-w-md">` and looks like it addresses it for me. Can you verify? Thanks!

PR #2579, opened 2025-04-16
fix: check table existence before deleting sessions

g-linville:
Turns out that it was possible to get into a state where the user would get a 500 when trying to log out of all other sessions. This would happen if they had an identity tied to a previously configured but now deconfigured auth provider, so that sessions table didn't exist. This fixes that by checking for the table before we try to delete sessions from it.

### DIFF
--- a/pkg/gateway/client/session.go
+++ b/pkg/gateway/client/session.go
@@ -81,22 +81,15 @@ func (c *Client) deleteSessionsForUser(ctx context.Context, db *gorm.DB, storage
 	return errors.Join(errs...)
 }
 
-func (c *Client) tableExists(db *gorm.DB, tableName string) (bool, error) {
-	var exists bool
-	if err := db.Raw("SELECT EXISTS (SELECT FROM information_schema.tables WHERE table_name = ?)", tableName).Scan(&exists).Error; err != nil {
-		return false, err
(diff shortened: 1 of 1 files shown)


StrongMonkey (approved):

pkg/gateway/client/session.go
```diff
@@ -81,7 +81,22 @@ func (c *Client) deleteSessionsForUser(ctx context.Context, db *gorm.DB, storage
 	return errors.Join(errs...)
 }
 
+func (c *Client) tableExists(db *gorm.DB, tableName string) (bool, error) {
+	var exists bool
+	if err := db.Raw("SELECT EXISTS (SELECT FROM information_schema.tables WHERE table_name = ?)", tableName).Scan(&exists).Error; err != nil {
+		return false, err
+	}
+	return exists, nil
```
thedadams: nit: Does something like this work?
```suggestion
	return db.Migrator().TableExists(tableName)
```

thedadams (approved):

pkg/gateway/client/session.go
g-linville: Yep, this is much better, thanks

thedadams (approved):

Issue #2267, opened 2025-03-24
Implement for each loop over datasets


cjellick:

g-linville:
this one is mine now

g-linville:
this comment is from a demo workflow

sangee2004:
When testing with latest build, we have the option "Convert to loop step" that allows users to add multiple steps under the selected step that will be iterated as a loop for the data set received in the current step.

This is the task I created for getting the latest price range for each items I have in a table and update the table with the optimal price.

<img width="754" alt="Image" src="https://github.com/user-attachments/assets/5bbdfd45-e2ce-451a-b23b-5056fe297c32" />

When this task was executed , i had 20 iterations( as many entries in db) for each  steps in the loop.

https://www.loom.com/share/ebb0614c4c284ce3b8c2883db37dea65?sid=eef28b37-8a25-4030-a091-ab7793b62d73

sangee2004:
Using "Convert to regular Step" option results in deleting all the loop step without asking user for any confirmation . Logged https://github.com/obot-platform/obot/issues/2888 for tracking.

PR #2585, opened 2025-04-17
feat: loop over datasets in workflows

g-linville:
This adds a new feature to allow you to add a group of prompts to one step in a workflow. The workflow will first look for the dataset described by the user, and then run through the loop once for each element in the dataset. UI will come in a future PR.

for https://github.com/obot-platform/obot/issues/2267


pkg/controller/handlers/workflowstep/loop.go
```diff
@@ -0,0 +1,211 @@
...
+	for elementIndex, element := range dataset.GetAllElements() {
+		steps, err := defineLoop(elementIndex, element.Contents, lastStepName, rootStep)
+		if err != nil {
+			return err
+		}
+
+		objects = append(objects, steps...)
+
+		if len(steps) > 0 {
+			lastStepName = steps[len(steps)-1].(*v1.WorkflowStep).Name
```
thedadams: nit:
```suggestion
			lastStepName = steps[len(steps)-1].GetName()
```

pkg/controller/handlers/workflowstep/loop.go
```diff
@@ -0,0 +1,211 @@
...
+	for i, s := range rootStep.Spec.Step.Loop {
+		afterStepName := dataStepName
+		if i > 0 {
+			afterStepName = lastStepName
+		}
+
+		if i == 0 {
+			// For the very first step, we need to add the element to the prompt.
+			s = elementPrompt(element, s)
+		}
```
thedadams: nit:
```suggestion
		if i > 0 {
			afterStepName = lastStepName
		} else {
			// For the very first step, we need to add the element to the prompt.
			s = elementPrompt(element, s)
		}
```

pkg/controller/handlers/workflowstep/loop.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	completeResponse = true
+	rootStep.Status.State = types.WorkflowStateComplete
+	rootStep.Status.LastRunName = runName
+	return nil
+}
+
+func defineLoop(elementIndex int, element string, dataStepName string, rootStep *v1.WorkflowStep) (result []kclient.Object, _ error) {
+	var lastStepName string
```
thedadams: nit: I got confused on this.
```suggestion
	var previousStepName string
```

pkg/controller/handlers/workflowstep/loop.go
```diff
@@ -0,0 +1,211 @@
...
+		parentStep.Status.Error = "no dataset ID found in output"
+		parentStep.Status.State = types.WorkflowStateError
+		return "", "", true, nil
+	}
+
+	return run.Name, datasetID, false, nil
+}
+
+func getDatasetID(output string) string {
+	return regexp.MustCompile(`gds://[a-z0-9]+`).FindString(output)
```
thedadams: `MustCompile` can panic, so it should be a package variable so that it panics on startup, and we would only have to compile it once.

thedadams (changes requested):

pkg/controller/handlers/workflowstep/loop.go
g-linville: Good call

g-linville: Fixed

thedadams (approved):

Issue #2276, opened 2025-03-24
Memory feature


cjellick:
Implement a memory feature that meets the following requirements:
- The user can directly view and delete the individual items in memory. This should be on their profile page.
- The user can disable the feature entirely
- All memory items are available to all chats. Lets just keep it simple and inject all of the memory into the system prompt
- Saving an item to memory should be a tool call. For now, let’s just let it show up like a normal tool call in the UI. We may later choose to suppress it.
- To guide the llm on when/how it should save to memory, we should have some context similar to this: https://github.com/modelcontextprotocol/servers/tree/main/src/memory#system-prompt
- The sample i just linked to has a pretty complicated solution creating a knowledge graph. I dont think we need that. I think we just need a simple list memories
- Tools we should have:
    - the context tool that injects the memories into the chat
    - a tool for adding new memories
    - a tool for deleting a memory ... this could get tricky bc the llm will have to know an “id” by which to delete the memory, so we may need a corresponding list memory.

njhale:
Spoke with @cjellick and the requirements shifted a bit; memory should be Obot-scoped, not account wide.

cjellick:
So from a UX perspective, we should:

Memory lives at the same level as threads and tasks. We should have a menu item over there that just says "Memory". When you click it, you get a modal displaying the list of items in memory. You can trash can any individual item.

tybalex:
Should we consider setting a limit or cap on the number of memories as well?

njhale:
> Should we consider setting a limit or cap on the number of memories as well?

@tybalex Darren, Craig, and I spoke about this briefly and came to the conclusion that it wasn't necessary for now.

sangee2004:
Able to add `memory` tool to my obot and work with it successfully.

Scenarios that were tested

- Able to add memory entries and share them across threads.

  1. Create an obot.
  2. Add `memory` tool to obot.
  3. Chat with obot and entries to memory like `Cricket is my favorite sport. Add it to memory` . I have to ask to add entry to memory once in the thread. I then see subsequent messages about me getting added to the memory automatically (Sometimes i will be prompted to add to memory)
  4. Start another thread and ask about information stored in the memory. Able to get information stored in the memory successfully. 
  
- Able to list, view and delete memory entries at obot level using "Memories" option 
- Sharing/copying obot with memory entries will result in the shared/copied instance to have the `memory` tool but not the memory entries

sangee2004:
[...]

PR #2586, opened 2025-04-17
feat: memory

njhale:
Add APIs and user UI integration for managing Obot memories at the project scope.

Addresses part of https://github.com/obot-platform/obot/issues/2276


pkg/api/handlers/memorysets.go
```diff
@@ -0,0 +1,172 @@
...
+
+		// Update existing entries or add new ones
+		for id, memory := range newMemories {
+			if _, exists := existingMap[id]; exists {
+				continue
+			}
+
+			// Add new memory
+			memorySet.Spec.Manifest.Memories = append(memorySet.Spec.Manifest.Memories, memory)
+		}
```
g-linville: I don't think this loop can update existing entries?

pkg/api/handlers/memorysets.go
```diff
@@ -0,0 +1,172 @@
...
+
+		if err := req.Update(&memorySet); err != nil {
+			return err
+		}
+	}
+
+	// Convert the map back to a list for the response
```
g-linville:
```suggestion
```

pkg/api/handlers/memorysets.go
njhale: Ah, that's an old comment. Thanks for catching.

ivyjeong13 (approved):
Approving for UI changes 🙌

pkg/api/handlers/memorysets.go
```diff
@@ -0,0 +1,168 @@
...
+
+	if err := req.Read(&memories); err != nil {
+		return err
+	}
+
+	// Current time for new memories
+	currentTime := types.NewTime(time.Now())
+
+	// Assign IDs based on content hash and deduplicate
+	newMemories := make(map[string]types.Memory)
```
thedadams: nit:
```suggestion
	newMemories := make(map[string]types.Memory, len(memories))
```

pkg/api/handlers/memorysets.go
```diff
@@ -0,0 +1,168 @@
...
+		for _, memory := range newMemories {
+			memorySet.Spec.Manifest.Memories = append(memorySet.Spec.Manifest.Memories, memory)
+		}
+
+		if err := req.Create(&memorySet); err != nil {
+			return err
+		}
+	} else {
+		// Add memories to the existing MemorySet
+		existingMap := make(map[string]struct{})
```
thedadams: nit
```suggestion
		existingMap := make(map[string]struct{}, len(memorySet.Spec.Manifest.Memories))
```

pkg/api/handlers/memorysets.go
```diff
@@ -0,0 +1,168 @@
...
+		for _, memory := range memorySet.Spec.Manifest.Memories {
+			existingMap[memory.ID] = struct{}{}
+		}
+
+		for id, memory := range newMemories {
+			if _, exists := existingMap[id]; exists {
+				continue
```
thedadams, line 91: Do we want to update existing memories? Like, make sure the content is updated?

pkg/api/handlers/memorysets.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+
+		if err := req.Update(&memorySet); err != nil {
+			return err
+		}
+	}
+
+	return req.Write(&types.MemorySet{
+		Metadata: types.Metadata{
+			ID: memorySet.Name,
```
thedadams: Should we include the entire `Metadata` field? Maybe use `MetadataFrom`?

pkg/api/handlers/memorysets.go
```diff
@@ -0,0 +1,168 @@
...
+	if err != nil {
+		return err
+	}
+
+	if err := req.Get(&memorySet, thread.Name); err != nil {
+		return err
+	}
+
+	return req.Write(&types.MemorySet{
+		Metadata: types.Metadata{
```
thedadams: Same question here about Metadata.

pkg/api/handlers/memorysets.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	// Create a new slice without the memory to delete
+	var updatedMemories []types.Memory
+	for _, memory := range memorySet.Spec.Manifest.Memories {
+		if memory.ID == memoryID {
+			continue
+		}
+		updatedMemories = append(updatedMemories, memory)
+	}
```
thedadams: Do we need to create an entire new slice? Seems like something like `slices.DeleteFunc` would work here.

If we do need to create a new slice, then:
```suggestion
	// Create a new slice without the memory to delete
	updatedMemories := make([]types.Memory, 0, len(memorySet.Spec.Manifest.Memories))
	for _, memory := range memorySet.Spec.Manifest.Memories {
		if memory.ID == memoryID {
			continue
		}
		updatedMemories = append(updatedMemories, memory)
	}
```

thedadams (changes requested):

pkg/api/handlers/memorysets.go
njhale: the id's are a hash of the content, so there shouldn't be a case where the content needs to be updated, but I can add that logic for robustness if you'd like

thedadams: I think the point is that this is an "add memory" endpoint and not an "update memories" endpoint, so I think this is good.

pkg/api/handlers/memorysets.go
```diff
@@ -0,0 +1,166 @@
...
+
+		if err := req.Update(&memorySet); err != nil {
+			return err
+		}
+	}
+
+	response := types.MemorySet{
+		MemorySetManifest: memorySet.Spec.Manifest,
+	}
+	response.Metadata = MetadataFrom(&memorySet)
```
thedadams:
```suggestion
	response := types.MemorySet{
		Metadata:          MetadataFrom(&memorySet)
		MemorySetManifest: memorySet.Spec.Manifest,
	}
```

pkg/api/handlers/memorysets.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	if err := req.Get(&memorySet, thread.Name); err != nil {
+		return err
+	}
+
+	response := types.MemorySet{
+		MemorySetManifest: memorySet.Spec.Manifest,
+	}
+	response.Metadata = MetadataFrom(&memorySet)
```
thedadams:
```suggestion
	response := types.MemorySet{
		Metadata:          MetadataFrom(&memorySet)
		MemorySetManifest: memorySet.Spec.Manifest,
	}
```

thedadams (approved):
Minor changes, so approving for expediency.

tybalex (approved):

Issue #2528, opened 2025-04-14
When running a task within Chat Threads, an error message briefly appears (for about 2 seconds), even though the task ultimately completes successfully.


shweta8996:
**Steps to Reproduce**

1. Create a new obot.
2. Create a new task and attempt to run it.
3. Navigate to Chat Threads and trigger the task.
4. Observe the interface—an error message briefly flashes before the task finishes.
5. The message shown is: “gzip: invalid header”

**Expected Behavior**

 The task should complete cleanly without displaying a transient error message. If there is an issue, the error should be clearly shown and persist until resolved, with an explanation of what went wrong.

**Please refer attachment:**

https://github.com/user-attachments/assets/f62fc9c6-fb67-48a0-8b0e-29f2a7c9113c

mighty-pup-acorn:
Debug Output:

```
{"runID":"r1chatznvdf","time":"2025-04-16T23:44:13.403177375Z","content":"\n","input":"run the greeter task"}
{"runID":"r1chatznvdf","time":"2025-04-16T23:44:13.403177375Z","content":"","contentID":"call_bk5WGxXxJU6jfNO8y29OTvQl","toolCall":{"name":"Task greeter","description":"Invokes task named greeter","taskID":"w1bw5b2","taskRunID":"we1gz2dl","input":"{}","output":"Goodbye! If you need anything else, feel free to reach out. Have a great day!\n"}}
{"runID":"r1chatznvdf","time":"2025-04-16T23:44:13.403177375Z","content":"The greeter task has been executed, and it says: \"Goodbye! If you need anything else, feel free to reach out. Have a great day!\"","contentID":"r1chatznvdf-1-0"}
{"runID":"r1chatznvdf","time":"2025-04-16T23:50:40.520970723Z","content":"","runComplete":true}
{"threadID":"t1ktvt6","content":"","replayComplete":true}
{"threadID":"t1ktvt6","content":"","replayComplete":true}
{"runID":"r1chatbbwj6","time":"2025-04-16T23:50:40.508130087Z","content":"\n","input":"run the greeter task"}
{"runID":"r1chatbbwj6","time":"2025-04-16T23:50:40.508130087Z","content":"","contentID":"r1chatbbwj6-0-0","waitingOnModel":true}
[...]

PR #2587, opened 2025-04-17
Fix: assign chatState before encryption

StrongMonkey:
When decompress chatState, it accidentally use the encrypted chatState value after creating runState. It should use the decrypted one when running decompress.

[https://github.com/obot-platform/obot/issues/2528](https://github.com/obot-platform/obot/issues/2528#event-17292949980)

### DIFF
--- a/pkg/gateway/client/runstate.go
+++ b/pkg/gateway/client/runstate.go
@@ -34,9 +34,9 @@ func (c *Client) RunState(ctx context.Context, namespace, name string) (*types.R
 
 func (c *Client) CreateRunState(ctx context.Context, runState *types.RunState) error {
 	// Copy the run state to avoid modifying the original
-	r := runState
+	r := *runState
 
-	if err := c.encryptRunState(ctx, r); err != nil {
+	if err := c.encryptRunState(ctx, &r); err != nil {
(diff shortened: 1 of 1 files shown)


thedadams:
Thanks for finding this issue. The intention was that the encrypted values would never leak out of the implementation. Since the values are still encrypted here, then the fix should be in the client.

I confirmed that the copying I was doing was wrong. The fix should be to change [this line](https://github.com/obot-platform/obot/blob/2a17ea1178b965c98e27e1e49ad8c2df0dae3207/pkg/gateway/client/runstate.go#L37) to `r := *runState`. Then [this line](https://github.com/obot-platform/obot/blob/2a17ea1178b965c98e27e1e49ad8c2df0dae3207/pkg/gateway/client/runstate.go#L39) and [this line](https://github.com/obot-platform/obot/blob/2a17ea1178b965c98e27e1e49ad8c2df0dae3207/pkg/gateway/client/runstate.go#L49) should have `&r`.

A similar change should also happen in `UpdateRunState`.

StrongMonkey:
@thedadams changed to use pointer.

thedadams (approved):

PR #2611, opened 2025-04-18
add storageClass name to pvc

amsuggs37:
This PR simply adds the use of the persistence.storageClass value to set the storageClassName on the pvc if it is provided.

### DIFF
--- a/chart/templates/pvc.yaml
+++ b/chart/templates/pvc.yaml
@@ -4,6 +4,9 @@ apiVersion: v1
 metadata:
   name: {{ .Release.Name }}-pvc
 spec:
+  {{- if .Values.persistence.storageClass }}
+  storageClassName: {{ .Values.persistence.storageClass }}
+  {{- end }}
   accessModes:
   {{- range .Values.persistence.accessModes }}
     - {{ . | quote }}


thedadams:
Seems good to me, but would like @drpebcak's input.

chart/templates/pvc.yaml
```diff
@@ -4,6 +4,9 @@ apiVersion: v1
 metadata:
   name: {{ .Release.Name }}-pvc
 spec:
+  {{- if .Values.persistence.storageClass }}
+  storageClassName: {{ .Values.persistence.storageClass }}
```
drpebcak, line 8: Can you add this to the values.yaml with a comment so that it shows up in the generated documentation?

thedadams: It's already there.

amsuggs37: Thanks @thedadams. Yes that was what led me to create this PR. The documentation has information there about storageClassName but when I went to use it, the PVC did not have the storage class set.

drpebcak: lol oh, my bad

drpebcak (approved):

Issue #2576, opened 2025-04-16
No way to get  to home page from "My Account" page.


sangee2004:
**Describe the bug**
No way to get  to home page from "My Account" page.

**To Reproduce**
1. Click on profile pic and use "My Account" option to launch  "My Account" page.

Notice that there is no way to get to home page from "My Account" page.

<img width="1091" alt="Image" src="https://github.com/user-attachments/assets/3345fee3-9714-45a7-a07c-f0e433440904" />

**Expected behavior**
Provide a way to get to home page

cjellick:
Let's just make the obot icon in the top left go to the https://obot.ai

One other thing i noticed while on this page (this should be a separate issue, but im lazy and am just going to post it here) - the My Account header is kind of just floating out in space, especially on a wide monitor. Can we do something about that? Maybe even just drop it.

![Image](https://github.com/user-attachments/assets/cb239550-2795-4f95-bd4a-b97f77b21c43)

PR #2616, opened 2025-04-18
Fix: No way to get to home from My Account page

tony-dotnet-fullstack:
No way to get to home page from "My Account" page.
Addresses [#2576](https://github.com/obot-platform/obot/issues/2576)


ivyjeong13:
Probably can close this one out! With the navigation updates, instead of home, we go to /catalog instead.

PR #2627, opened 2025-04-18
Fix: allow uploading markdown files

StrongMonkey:
We should allow uploading markdown file in knowledge.

### DIFF
--- a/ui/user/src/lib/components/navbar/KnowledgeUpload.svelte
+++ b/ui/user/src/lib/components/navbar/KnowledgeUpload.svelte
@@ -71,7 +71,12 @@
 		{#if !uploadInProgress}
 			<Upload class="size-4" />
 		{/if}
-		Upload (.pdf, .pptx, .txt)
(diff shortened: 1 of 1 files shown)


thedadams (approved):

cjellick (approved):

cjellick (changes requested):
oh actually, since we support more than what we say in the parenthesis, can we drop that part?

ui/user/src/lib/components/navbar/KnowledgeUpload.svelte
```diff
@@ -72,6 +72,6 @@
 			<Upload class="size-4" />
 		{/if}
 		Upload (.pdf, .pptx, .txt)
```
cjellick:
```suggestion
		Upload
```

StrongMonkey: Changed, all added the rest of the types that knowledge supports.

cjellick (approved):

Issue #2598, opened 2025-04-17
Drop introduction and starter messages from base agent and clarify instructions field in Admin UI


mighty-pup-acorn:
## Summary of Slack Thread

- There is confusion among prospects regarding the purpose and function of the "instructions," "introduction," and "starter messages" fields in the Admin Agent UI.
- The current UI does not clearly explain how these fields relate to the obot or the overall product experience.
- There is consensus that the "instructions" field is still valuable as the base system prompt for all obots inheriting from the base agent, but the "introduction" and "starter messages" should be dropped.
- The need for better alignment and clarity in the Admin UI was highlighted, especially around the instructions field.

## Feature Request

1. **Remove Introduction and Starter Messages:**
   - Drop the "introduction" and "starter messages" fields from the base agent configuration in the Admin UI, as they add confusion and are not essential to the core agent setup.

2. **Improve Instructions Field Explanation:**
   - Add clear, user-friendly documentation or inline help in the Admin UI to explain the purpose of the "instructions" field.
   - Make it explicit that the instructions serve as the base system prompt for all obots inheriting from the base agent, and clarify how this impacts agent behavior.

3. **UI/UX Alignment:**
   - Review the Admin UI for consistency and alignment with the overall product experience, ensuring that all configuration fields are intuitive and well-documented.

### Additional Considerations
- Consider whether other fields (e.g., knowledge, workspace files) also need review or clearer documentation.

---

This issue is based on feedback from both internal team members and prospects, as discussed in the Slack thread on 4/17/2025.

shweta8996:
I have verified below points on main url and it is working fine as expected.
Remove Introduction and Starter Messages > **Done**.
Improve Instructions Field Explanation > **Done**
UI/UX Alignment > **Done**

Please refer attachment:

https://github.com/user-attachments/assets/86beddec-296d-48c8-8528-d521d58dc731

PR #2656, opened 2025-04-21
Fix: Drop introduction and starter messages from base agent

tony-dotnet-fullstack:
Drop introduction and starter messages from base agent and clarify instructions field in Admin UI
Addresses [#2598](https://github.com/obot-platform/obot/issues/2598)


ivyjeong13 (approved):
If the AgentIntro is no longer used, you can prolly delete that component too

PR #2658, opened 2025-04-21
Enhanced analytics

thedadams:
This change does three things and I plan to keep them as three separate commits:
1. Add the concept of internal users, and ignores these users when calculating daily active users
2. Adds the ability to track tokens for LLM requests. APIs are added for user and system reporting
3. Changes the user authorization so that we don't have to check usernames in the API layer. The impetus for this is so that a user can look at their own usages.

Related issue: https://github.com/obot-platform/sensitive-issues/issues/24


pkg/gateway/server/user.go
```diff
@@ -83,26 +83,16 @@ func (s *Server) encryptAllUsersAndIdentities(apiContext api.Context) error {
...
-	var (
-		getByID      = apiContext.URL.Query().Get("by-id") == "true"
-		usernameOrID = apiContext.PathValue("username_or_id")
-		user         *types.User
-	)
+	userID := apiContext.PathValue("user_id")
 
-	if usernameOrID == "" {
+	if userID == "" {
 		return types2.NewErrHTTP(http.StatusBadRequest, "username path parameter is required")
```
g-linville:
```suggestion
		return types2.NewErrHTTP(http.StatusBadRequest, "user_id path parameter is required")
```

pkg/gateway/server/user.go
```diff
@@ -111,18 +101,11 @@ func (s *Server) getUser(apiContext api.Context) error {
...
 
 func (s *Server) updateUser(apiContext api.Context) error {
-	requestingUsername := apiContext.User.GetName()
-	actingUserIsAdmin := apiContext.UserIsAdmin()
-
-	username := apiContext.PathValue("username")
-	if username == "" {
+	userID := apiContext.PathValue("user_id")
+	if userID == "" {
 		return types2.NewErrHTTP(http.StatusBadRequest, "username path parameter is required")
```
g-linville:
```suggestion
		return types2.NewErrHTTP(http.StatusBadRequest, "user_id path parameter is required")
```

g-linville (approved):

pkg/gateway/server/user.go
thedadams: Fixed.

pkg/gateway/server/user.go
thedadams: Fixed.

pkg/api/authz/user.go
```diff
@@ -0,0 +1,13 @@
+package authz
+
+import (
+	"slices"
+
+	"k8s.io/apiserver/pkg/authentication/user"
+)
+
+func (a *Authorizer) checkUser(user user.Info, userID string) bool {
+	return userID == "" ||
```
njhale, line 10: My only concern here would be that any user can pass this check with `/api/users//activities` somehow, but -- aside from failing further downstream -- I think the path wouldn't match, so probably not a valid concern. Just thought I'd bring it up, just in case.

thedadams: This isn't a concern because Go cleans the path and will redirect the user to `/api/users/activities` in this case.

PR #2665, opened 2025-04-22
enhance: allow looping over JSON string lists in addition to datasets 

g-linville:
depends on https://github.com/obot-platform/tools/pull/599

### DIFF
--- a/pkg/controller/handlers/workflowstep/loop.go
+++ b/pkg/controller/handlers/workflowstep/loop.go
@@ -150,8 +150,8 @@ func dataPrompt(description, fileName string) string {
 	%q
 
 	If the data is not already available in the chat history, call any tools you need in order to find it.
-	If you find a dataset ID (begins with gds://), call the loop-data tool with it as the dataset_id argument.
-	If you do not find a dataset ID, create a JSON list of strings and call the loop-data tool with it as the data_list argument.
+	If the data includes a dataset ID (begins with gds://), call the loop-data tool with it as the dataset_id argument.
+	If you do not find a dataset ID, create a JSON list of strings with the relevant data and call the loop-data tool with it as the data_list argument.
 
 	When you call the loop-data tool, include %s as the file_name argument.
 	`, description, fileName)
--- a/pkg/controller/handlers/workflowstep/loop.go
+++ b/pkg/controller/handlers/workflowstep/loop.go
@@ -99,6 +99,8 @@ func (h *Handler) RunLoop(req router.Request, _ router.Response) (err error) {
 	rootStep.Status.State = types.WorkflowStateComplete
 	rootStep.Status.LastRunName = runName
 
+	// We ignore the error here because it does not really matter if we fail to delete the file.
+	// We're just making a best effort to clean up after ourselves.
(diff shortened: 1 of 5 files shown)


cjellick (approved):
Seems good to me, but would like @thedadams to review too

pkg/controller/handlers/workflowstep/loop.go
```diff
@@ -111,6 +98,10 @@ func (h *Handler) RunLoop(req router.Request, _ router.Response) (err error) {
 	completeResponse = true
 	rootStep.Status.State = types.WorkflowStateComplete
 	rootStep.Status.LastRunName = runName
+
+	_ = h.gptscriptClient.DeleteFileInWorkspace(req.Ctx, fileName, gptscript.DeleteFileInWorkspaceOptions{
```
thedadams, line 104: Either handle the error or add a comment explaining why it is being ignored.

pkg/controller/handlers/workflowstep/loop.go
```diff
@@ -146,66 +137,80 @@ func elementPrompt(element, prompt string) string {
 	`, element, prompt)
 }
 
-func defineDataStep(rootStep *v1.WorkflowStep) *v1.WorkflowStep {
+func defineDataStep(rootStep *v1.WorkflowStep, fileName string) *v1.WorkflowStep {
 	return NewStep(rootStep.Namespace, rootStep.Spec.WorkflowExecutionName, rootStep.Spec.AfterWorkflowStepName, rootStep.Spec.WorkflowGeneration, types.Step{
-		ID:   rootStep.Spec.Step.ID + "-loopdata",
-		Step: dataPrompt(rootStep.Spec.Step.Step),
+		ID:   rootStep.Spec.Step.ID + "{loopdata}",
```
thedadams, line 144: This is a fine hack, though I worry what trouble this might produce.

pkg/controller/handlers/workflowstep/loop.go
```diff
@@ -146,66 +137,80 @@ func elementPrompt(element, prompt string) string {
...
+			return "", nil, false, err
+		}
 	}
 
-	return thread.Status.WorkspaceID, nil
+	return thread.Status.WorkspaceID, data, false, nil
+}
+
+func isDatasetID(output string) bool {
+	return datasetIDRegex.MatchString(output)
```
thedadams: nit: if `datasetIDRegex.Match` is used here, then you can pass the `[]byte` instead of casting to `string`.

thedadams (commented):
Just one issue with ignoring the error.

pkg/controller/handlers/workflowstep/loop.go
g-linville: So for this I am following the older pattern of how we did things for while loops and if statements. For "substeps" generated by the top-level if- and while-steps, there would be additional information surrounded by braces. There is code in the UI that expects this as well. So I'm just trying to follow that convention here.

pkg/controller/handlers/workflowstep/loop.go
g-linville: left a comment

pkg/controller/handlers/workflowstep/loop.go
g-linville: fixed

thedadams (approved):

thedadams (approved):

Issue #2612, opened 2025-04-18
Improve user experience around edit/deletes of memory items attempted when chatting.


sangee2004:
**Describe the bug**
Improve user experience around edit/deletes of memory items attempted when chatting. 

**To Reproduce**
1. Create an obot.
2. Add memory tool to obot.
3. Chat with obot and add entries to memory like `Cricket is my favorite sport. Add it to memory `. 
4. In another thread , ask to `list my favorite sport` and then ask to `delete this info`

<img width="1050" alt="Image" src="https://github.com/user-attachments/assets/9ea88636-1fa8-49eb-97ca-fcffc050fa3c" />

At this time , entry continues to be in the memory

<img width="959" alt="Image" src="https://github.com/user-attachments/assets/3273e962-3cd5-4057-beb3-9099f82c15c4" />

5. In another thread, ask to `list my favorite sport` and we see the info from memory still being listed

<img width="1059" alt="Image" src="https://github.com/user-attachments/assets/0160840b-0c3f-4e83-9764-f576f9cef0a9" />

**Expected Behavior:**
We should be able to inform user that deletion from chat is not possible and that he needs to use the delete option in ui.

**Scenario 2 (update)**

1. Create an obot.
2. Add memory tool to obot.
3. Chat with obot and add entries to memory like `Cricket is my favorite sport. Add it to memory `. 
4. In another thread , ask to `list my favorite sport` and then ask to `Update my favorite sport to volleyball`

<img width="979" alt="Image" src="https://github.com/user-attachments/assets/e053ba73-6250-4f4d-bbee-661b79778163" />

After this we see 2 entries in the memory
<img width="940" alt="Image" src="https://github.com/user-attachments/assets/1721f731-ddef-4567-87c6-f9ff36802203" />

When I ask for my favorite sport , both entries get returned

**Expected Behavior:**
We should be able to inform user that updating entries is not possible and  he could to use the edit memory option (which will be made available soon) in ui.

sangee2004:
Tested with latest build. 

When User asks to update/delete memory entries when chatting, there is an explicit tool call made to `Update Memory`\`Delete memory` tool calls. This results in  entry from memory being updated or deleted as expected.

Scenarios tested to make sure that the change in memory entry is reflected in other threads

**Deleting entries**
1. Thread 1 - No queries made to memory
2. Thread 2 - `Remember my pets name is Bonda ==> Memory entry gets added as expected
[...]

PR #2666, opened 2025-04-22
Add ability to update memories from the user UI and memory tools.

njhale:
Add the ability to update memories from the user UI and API/client.

This change also refactors the API to hide `MemorySets` from clients. They are an implementation detail of how memories are stored in the backend and don't need to be exposed to clients.

e.g. Editing a memory via the UI

https://github.com/user-attachments/assets/f53c5691-3f62-4d81-8905-46924c2cc692

e.g. Editing a memory via chat

https://github.com/user-attachments/assets/64d18171-7ac6-4742-8a80-1fa73ccc5943

e.g. Deleting a memory via chat

https://github.com/user-attachments/assets/f527ac28-72ee-4d42-8424-62b8e2f84ef7

Addresses part of https://github.com/obot-platform/obot/issues/2612


cjellick (approved):

pkg/api/handlers/memories.go
```diff
@@ -0,0 +1,213 @@
...
+	if memory.Content == "" {
+		return apierrors.NewBadRequest("content cannot be empty")
+	}
+
+	// Return error if ID or timestamp is provided
+	if memory.ID != "" {
+		return apierrors.NewBadRequest("id should not be provided")
+	}
+
+	if memory.CreatedAt != (types.Time{}) {
```
thedadams, line 48:
```suggestion
	if memory.CreatedAt != types.Time{} {
```

pkg/api/handlers/memories.go
```diff
@@ -0,0 +1,213 @@
...
+
+	// Get the memory set
+	var memorySet v1.MemorySet
+	if err := req.Get(&memorySet, thread.Name); err != nil {
+		return err
+	}
+
+	// Find and update the memory by ID
+	var updatedMemory types.Memory
+	memoryFound := false
```
thedadams: nit:
```suggestion
	var (
		updatedMemory types.Memory
		memoryFound   bool
	)
```

thedadams (approved):

pkg/api/handlers/memories.go
njhale: This change doesn't compile

PR #2668, opened 2025-04-22
Workflow loops UI

rob-fullstack-star:
Implement for each loop over datasets
Addresses [#2267](https://github.com/obot-platform/obot/issues/2267)

### DIFF
--- a/ui/user/src/lib/components/messages/Message.svelte
+++ b/ui/user/src/lib/components/messages/Message.svelte
@@ -47,7 +47,6 @@
 			: ''
 	);
 	let fullWidth = !msg.sent && !msg.oauthURL && !msg.tool;
-	let showBubble = msg.sent;
 	let isPrompt = msg.fields && msg.promptId;
 	let renderMarkdown = !msg.sent && !msg.oauthURL && !msg.tool;
 	let shell = $state({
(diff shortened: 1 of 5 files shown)


cjellick (changes requested):
This doesnt do what i want it to. I need indented substeps for the loop
https://www.loom.com/share/777ac42e0d3d463685f049ba12df805e

cjellick:
@rob-fullstack-star - actually your PR works with substeps but i happened to have rebased your PR locally on top of main and that is what caused it to break. can you rebase and retest it?

ui/user/src/lib/components/messages/Message.svelte
```diff
@@ -289,10 +288,8 @@
 
 {#snippet messageBody()}
 	<div
-		class:flex={showBubble}
-		class:contents={!showBubble}
 		class:message-content={renderMarkdown}
-		class="bg-gray-70 flex w-full flex-col rounded-2xl px-6 py-3 text-black dark:bg-gray-950 dark:text-white"
+		class="bg-gray-70 flex max-h-[50vh] w-full flex-col overflow-auto rounded-2xl px-6 py-3 text-black dark:bg-gray-950 dark:text-white"
```
ivyjeong13: This breaks regular chat when the agent writes a long message. Also not sure, are we changing the agent to have a background color for responding? 

Screenshot 2025-04-23 at 10 30 44 AM

PR #2680, opened 2025-04-23
Token limits

thedadams:
The defaults here are 1,000,000 tokens for prompt and 100,000 for completion. The discrepancy is caused by our input tokens (tools, system prompt, contexts, etc) are long and seem to average 7,500 tokens per run. Back of the envelope calculations indicate that users should be able to send 120 messages/simple tasks per day.

This change also moves away from tracking token usages using the run states because a user can delete threads, which deletes the run states. We need this information to stick around.


pkg/gateway/server/llmproxy.go
```diff
@@ -17,6 +18,15 @@ func (s *Server) llmProxy(req api.Context) error {
 		return types2.NewErrHTTP(http.StatusUnauthorized, fmt.Sprintf("invalid token: %v", err))
 	}
 
+	if token.UserID != "" {
+		promptTokensRemaining, completionTokensRemaining, err := s.client.RemainingTokenUsageForUser(req.Context(), token.UserID, 24*time.Hour, s.dailyUserTokenPromptTokenLimit, s.dailyUserTokenCompletionTokenLimit)
```
njhale: nit: maybe the period should be be a constant (though I'm sure the compiler optimizes this).

njhale (approved):
Just a small nit, otherwise LGTM.

pkg/gateway/server/llmproxy.go
thedadams: Done.

PR #2683, opened 2025-04-23
fix: project user groups onto JWT tokens

njhale:
This fixes an issue where admins get a 403 from the memory tool when attempting to chat in a new thread created in another user's project.

Addresses https://github.com/obot-platform/sensitive-issues/issues/33

### DIFF
--- a/pkg/jwt/jwt.go
+++ b/pkg/jwt/jwt.go
@@ -2,6 +2,7 @@ package jwt
 
 import (
 	"net/http"
+	"slices"
 	"strings"
 
 	"github.com/golang-jwt/jwt/v5"
@@ -32,6 +33,7 @@ type TokenContext struct {
 	UserID         string
 	UserName       string
 	UserEmail      string
+	UserGroups     []string
 }
 
 type TokenService struct{}
@@ -42,13 +44,13 @@ func (t *TokenService) AuthenticateRequest(req *http.Request) (*authenticator.Re
 	if err != nil {
 		return nil, false, nil
 	}
+
+	groups := append([]string{authz.AuthenticatedGroup}, tokenContext.UserGroups...)
 	return &authenticator.Response{
 		User: &user.DefaultInfo{
-			UID:  tokenContext.UserID,
-			Name: tokenContext.Scope,
-			Groups: []string{
-				authz.AuthenticatedGroup,
-			},
+			UID:    tokenContext.UserID,
+			Name:   tokenContext.Scope,
+			Groups: groups,
 			Extra: map[string][]string{
 				"obot:runID":     {tokenContext.RunID},
 				"obot:threadID":  {tokenContext.ThreadID},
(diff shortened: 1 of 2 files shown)


thedadams:
How do you feel about just adding the groups [here](https://github.com/obot-platform/obot/blob/8aae9e2b0b58b74a47adda0671dc03a29f964c69/pkg/invoke/invoker.go#L608-L619) instead of hitting the database in the TokenService method?

njhale:
> How do you feel about just adding the groups [here](https://github.com/obot-platform/obot/blob/8aae9e2b0b58b74a47adda0671dc03a29f964c69/pkg/invoke/invoker.go#L608-L619) instead of hitting the database in the TokenService method?

I like that better.

>  My concern is that this change causes all tools run by an admin to run with admin privileges to Obot.

@thedadams do you share this concern at all?

thedadams:
I don't. And the fix for that is pretty straightforward with no migrations. So I think this is fine for now.

pkg/jwt/jwt.go
```diff
@@ -42,20 +43,25 @@ func (t *TokenService) AuthenticateRequest(req *http.Request) (*authenticator.Re
 	if err != nil {
 		return nil, false, nil
 	}
+
+	groups := []string{authz.AuthenticatedGroup}
+	if len(tokenContext.UserGroups) > 0 {
+		groups = append(groups, tokenContext.UserGroups...)
+	}
```
thedadams:
```suggestion
	groups = append(groups, tokenContext.UserGroups...)
```

pkg/jwt/jwt.go
```diff
@@ -83,11 +90,25 @@ func (t *TokenService) DecodeToken(token string) (*TokenContext, error) {
 		UserID:         claims["UserID"].(string),
 		UserName:       claims["UserName"].(string),
 		UserEmail:      claims["UserEmail"].(string),
-	}, nil
+	}
+
+	if userGroups, ok := claims["UserGroups"]; ok && userGroups != nil {
```
thedadams:
```suggestion
	if userGroups := claims["UserGroups"]; userGroups != nil {
```

pkg/jwt/jwt.go
```diff
@@ -83,11 +90,25 @@ func (t *TokenService) DecodeToken(token string) (*TokenContext, error) {
...
+		if userGroupsSlice, ok := userGroups.([]any); ok {
+			groups := make([]string, 0, len(userGroupsSlice))
+			for _, group := range userGroupsSlice {
+				if groupStr, ok := group.(string); ok {
+					groups = append(groups, groupStr)
+				}
+			}
+			context.UserGroups = groups
```
thedadams: All of this makes me want to make the `UserGroups` claim a comma-delimited string to avoid all the casting. What are your thoughts?

njhale: Had the same thought. I'm going to go ahead and do that instead.

njhale: done

g-linville (approved):

pkg/jwt/jwt.go
```diff
@@ -83,11 +85,14 @@ func (t *TokenService) DecodeToken(token string) (*TokenContext, error) {
 		UserID:         claims["UserID"].(string),
 		UserName:       claims["UserName"].(string),
 		UserEmail:      claims["UserEmail"].(string),
-	}, nil
+		UserGroups:     strings.Split(claims["UserGroups"].(string), ","),
```
thedadams: I believe that `UserGroups` will be `[]string{""}` here for non-admin users. Can we clean that up?

njhale: you're right. cleaned it up by calling slices.DeleteFunc.

thedadams (approved):

Issue #2626, opened 2025-04-18
`403 error`  presented when launching any featured obot for the first time


sangee2004:
**Describe the bug**

**To Reproduce**
1. Go to Catalog page
2. Click on any featured obot

Notice that a 403 error message is presented

<img width="1236" alt="Image" src="https://github.com/user-attachments/assets/d587a8a9-0dd6-43dc-9b6a-9732169967a5" />

<img width="982" alt="Image" src="https://github.com/user-attachments/assets/385d9686-2ecc-463b-87dd-007aa803b7cb" />

**Expected Behavior:**
There should be no error messages presented .

sangee2004:
This issue is not seen anymore when testing with latest builds.

No 403 errors seen when launching any featured obot for the first time.

PR #2689, opened 2025-04-23
fix: 403 on non-admin user visiting shared featured agent

ivyjeong13:
Addresses #2626 

* do not do GetProject call to find out if they're an editor, instead include in the ProjectShare information

### DIFF
--- a/pkg/api/handlers/projectshare.go
+++ b/pkg/api/handlers/projectshare.go
@@ -205,6 +205,7 @@ func convertProjectShare(threadShare v1.ThreadShare) types.ProjectShare {
 		Description:          threadShare.Status.Description,
 		Icons:                threadShare.Status.Icons,
 		Tools:                threadShare.Status.Tools,
+		Editor:               threadShare.Spec.Editor,
 	}
 }
 func (h *ProjectShareHandler) CreateProjectFromShare(req api.Context) error {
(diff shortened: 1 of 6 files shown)


cjellick (approved):

pkg/api/handlers/projectshare.go
```diff
@@ -277,11 +278,19 @@ func (h *ProjectShareHandler) GetShareFromShareID(req api.Context) error {
...
+	// Get the original project to check ownership
+	if err := req.Get(&baseProject, threadShare.Spec.ProjectThreadName); err != nil {
+		return err
+	}
+
 	// Checking if user has a project created from this share
 	if err := req.Get(&baseProject, id); err == nil {
 		// User does have a project instance, include its ID in the response
 		threadShare.Spec.ProjectThreadName = id
 	}
```
g-linville, line 290: We're writing into the `baseProject` variable here twice without checking the first value.

g-linville (changes requested):

pkg/api/handlers/projectshare.go
```diff
@@ -277,11 +278,19 @@ func (h *ProjectShareHandler) GetShareFromShareID(req api.Context) error {
 
 	threadShare := threadShareList.Items[0]
 
+	// Get the original project to check ownership
+	if err := req.Get(&baseProject, threadShare.Spec.ProjectThreadName); err != nil {
+		return err
+	}
+
```
g-linville:
```suggestion
```

g-linville (approved):

pkg/api/handlers/projectshare.go
g-linville: (we already get the same project right below this)

Issue #2365, opened 2025-04-02
Provide ability to see more than 10 most recent conversations in a thread.


sangee2004:
Currently we have the ability to show only  10 most recent conversations in a thread.

It would be good to provide the user the ability to see more (and all) the past conversations in a thread.

cjellick:
id like this to be implemented as ininite scrollback where we keep loading more of the chat history as the user scrolls back

cjellick:
Thread on impl details theres'a parentRunID in the events. and then a /api/runs/{id}/events to get the events for it.  we have to create an API thats' exposed to the user. but if they can get that to work with admin user then exposing the API is simple.

cjellick:
We need backend api changes before UI can be changed (see the slack convo)

sangee2004:
We are now presented with `Load older messages` option in the chat view that allows users to load 10 more older messages.

When there are more than 10 messages in a thread, we are now presented with `Load older messages` option when we scroll up  to the first message in the chat view. When this option is used , 10 messages from the thread history gets loaded. We are able to use this option multiple times to get to load all the threads in the chat view.

@njhale One related issue found when visiting other threads after scrolling in a thread - https://github.com/obot-platform/obot/issues/2785 (This issue is not restricted to threads having `Load older messages` option)

PR #2713, opened 2025-04-25
feat: add the ability to load previous messages in chat

njhale:
In the user UI, add a `Load older messages` button to the top of the chat that allows users to view older messages when available.

Addresses https://github.com/obot-platform/obot/issues/2365

https://github.com/user-attachments/assets/a585ca42-180d-461f-a020-9633e14a1e86

### DIFF
--- a/pkg/api/handlers/assistants.go
+++ b/pkg/api/handlers/assistants.go
@@ -242,17 +242,23 @@ func (a *AssistantHandler) ListCredentials(req api.Context) error {
 
 func (a *AssistantHandler) Events(req api.Context) error {
 	var (
-		runID  = req.Request.Header.Get("Last-Event-ID")
-		thread v1.Thread
+		follow  = req.URL.Query().Get("follow") == "true"
+		history = req.URL.Query().Get("history") == "true"
+		runID   = req.URL.Query().Get("runID")
+		thread  v1.Thread
 	)
+
+	if runID == "" {
+		runID = req.Request.Header.Get("Last-Event-ID")
+	}
 
 	if err := req.Get(&thread, req.PathValue("thread_id")); err != nil {
 		return err
 	}
 
 	_, events, err := a.events.Watch(req.Context(), req.Namespace(), events.WatchOptions{
-		Follow:        true,
-		History:       runID == "",
+		Follow:        follow,
+		History:       history,
 		LastRunName:   strings.TrimSuffix(runID, ":after"),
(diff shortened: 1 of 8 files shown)


cjellick (approved):

pkg/api/handlers/assistants.go
```diff
@@ -242,17 +242,22 @@ func (a *AssistantHandler) ListCredentials(req api.Context) error {
...
+
 	if err := req.Get(&thread, req.PathValue("thread_id")); err != nil {
 		return err
 	}
 
 	_, events, err := a.events.Watch(req.Context(), req.Namespace(), events.WatchOptions{
-		Follow:        true,
-		History:       runID == "",
+		Follow:        follow,
+		History:       true,
```
thedadams: I have definitely used passing a `runID` to get the events for only that run. Is there a reason this is changing?

njhale: I believe that's using the `/api/runs/{id}/events` endpoint, which I think is only for admins. Also the response I got back from that endpoint wasn't in the format I needed to display events. I can double check though.

thedadams: Are you saying that leaving this as-is and passing a `runID` (so not getting any history) breaks something?

njhale: Spoke with @thedadams in a different channel and we're going to add another query parameter for setting the history field here.

njhale: I added the `history` query parameter.

thedadams (approved):
Backend looks good.

cjellick (approved):

ui/user/src/lib/components/Thread.svelte
```diff
@@ -74,6 +76,23 @@
 		}
 	});
 
+	$effect(() => {
+		// Only update if messages change
+		const messages_copy = messages; // Create a local reference
```
ivyjeong13, line 81: nit: I think you can just use `messages` here, nothing seems to be done with messages_copy?

ivyjeong13 (approved):
Just one nit, otherwise looks good to me! Pulled down and tested as well

ui/user/src/lib/components/Thread.svelte
njhale: Claude thought this made the effect more reactive. That was definitely a hallucination.

PR #2731, opened 2025-04-28
Fix: fix timezone for triggering tasks

StrongMonkey:
This fixed the timezone bug where it is not respecting the correct timezone when running for the second time. It also adds some tests for it.

---

**Tracking Issue:** #2732

See the tracking issue for details and context from the Slack thread. This PR aims to resolve the bug where only the first scheduled run respects the timezone, but subsequent runs do not.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,27 @@
+name: Run Tests on Pull Request
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.24"
+
+      - name: Install Dependencies
+        run: |
+          go mod download
+
+      - name: Run Tests
+        run: |
+          make test
--- a/pkg/controller/handlers/cronjob/cronjob.go
+++ b/pkg/controller/handlers/cronjob/cronjob.go
@@ -48,26 +48,9 @@ func GetScheduleAndTimezone(cronJob v1.CronJob) (string, string) {
 
 func (h *Handler) Run(req router.Request, resp router.Response) error {
 	cj := req.Object.(*v1.CronJob)
-	lastRun := cj.Status.LastRunStartedAt
-	schedule, timezone := GetScheduleAndTimezone(*cj)
-	var location *time.Location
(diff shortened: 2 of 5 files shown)


.github/workflows/test.yaml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.22" # Adjust to your project's Go version
```
thedadams:
```suggestion
          go-version: "1.24"
```

StrongMonkey: thanks! addressed

thedadams (approved):

pkg/controller/handlers/cronjob/cronjob_test.go
```diff
@@ -0,0 +1,72 @@
+package cronjob
+
+import (
+	"testing"
+	"time"
```
tybalex, line 5: shall we `import  _ time/tzdata ` , so `time.LoadLocation()`  wouldn't rely on the timezone info from OS system? I encountered a similar issue here https://github.com/obot-platform/tools/pull/626/files

pkg/controller/handlers/cronjob/cronjob.go
```diff
@@ -103,6 +86,32 @@ func (h *Handler) Run(req router.Request, resp router.Response) error {
...
+	lastRun := cronJob.Status.LastRunStartedAt
+	if lastRun.IsZero() {
+		lastRun = &metav1.Time{Time: cronJob.CreationTimestamp.Time}
+	}
+
+	schedule, timezone := GetScheduleAndTimezone(cronJob)
+	var location *time.Location
+	if timezone != "" {
+		loc, err := time.LoadLocation(timezone)
+		if err == nil {
```
tybalex, line 99: if timezone load failed , should there be any logic for that?

pkg/controller/handlers/cronjob/cronjob_test.go
StrongMonkey: yeah it is included here https://github.com/obot-platform/obot/blob/main/main.go#L5

pkg/controller/handlers/cronjob/cronjob.go
StrongMonkey: It is probably fine that if it fails, the code just defaults to use system's timezone which is UTC.

PR #2761, opened 2025-04-30
enhance: add ability to specify a model provider at the Obot level

thedadams:
Specifying a model provider at the Obot level requires a user to specify their own API keys. APIs for configuring, deconfiguring, validating, and list models at this level are added.

Additionally, quota is tracked for such LLM requests, but these are not counted towards the user's limits.

Finally, the default model for runs is changed such that tools honoring OBOT_DEFAULT_LLM_MODEL environment variable will use the user's personal API token when making LLM requests.


thedadams:
Requires: https://github.com/obot-platform/tools/pull/636

pkg/api/handlers/availablemodels.go
```diff
@@ -87,8 +88,23 @@ func (a *AvailableModelsHandler) List(req api.Context) error {
 }
 
 func (a *AvailableModelsHandler) ListForModelProvider(req api.Context) error {
+	assistantID := req.PathValue("assistant_id")
+	projectID := req.PathValue("project_id")
+	modelProviderID := req.PathValue("id")
```
njhale: I would think changing the name of the path param from `model_provider` to `id` would break the pre-existing route:

```go
mux.HandleFunc("GET /api/available-models/{model_provider}", availableModels.ListForModelProvider)
```

cjellick:
@thedadams in this paradigm, any model provider we configure in the admin (like if we added keys for anthropic) will automatically become one of our free models?

pkg/api/handlers/modelprovider.go
```diff
@@ -74,6 +75,35 @@ func (mp *ModelProviderHandler) ByID(req api.Context) error {
 }
 
 func (mp *ModelProviderHandler) List(req api.Context) error {
+	assistantID := req.PathValue("assistant_id")
+	projectID := req.PathValue("project_id")
+
+	var allowedModelProviders []string
+	if assistantID != "" {
+		// Ensure that this agent allows this model provider.
```
cjellick: is this comment accurate? feels out of place

cjellick (approved):

thedadams:
> @thedadams in this paradigm, any model provider we configure in the admin (like if we added keys for anthropic) will automatically become one of our free models?

It becomes a possible free model, but it would have to be configured to a Base Agent. If the Base Agent isn't configured to use it, then a user won't be able to use it unless they are able to figure out what the names of the models are in our system.

njhale (approved):

pkg/api/handlers/availablemodels.go
thedadams: I changed them all to `model_provider_id`.

Issue #2577, opened 2025-04-16
"Log out of all other sessions" option does not work


sangee2004:
**Describe the bug**
"Log out of all other sessions" option does not work

**To Reproduce**
1. Log in to obot.ai as regular user
2. Log in to obot.ai in a Incognito Browser session as the same user from step 1.
3. From session open in step 1, click on profile pic and use "My Account" option to launch "My Account" page .
4. Use ""Log out of all other sessions"  option and use "Yes, I'm Sure" when asked for confimation.

User is presented with `Successfully logged out of all other sessions`  message but is not logged out of other sessions.

obot.ai in a Incognito Browser session is still active and i am able to continue to work with it.

**Expected behavior**
User should be presented with "Session Expired" message and user should be logged out.

**Note** - 
This feature works as expected in other obot setups for regular users.
This option does not work for admin users in other setups as well.

g-linville:
There was a bug preventing me from testing this that I just merged a fix for, but I will be unable to troubleshoot this until we re-enable deployments to obot.ai.

cjellick:
re-enabled

g-linville:
This should be fixed now. Some of the identities in the database had their `provider_user_id` field double-encrypted, which prevented us from properly finding and deleting the sessions for those users.

sangee2004:
Tested on `main`

I am able to see the `Log out of all other sessions` work as expected for both my accounts I have.

I see it also work with my admin account (which was not working when I reported the issue).

PR #2767, opened 2025-04-30
fix: ensure identities aren't double-encrypted when encrypting all users

thedadams:
This change also includes a migration to ensure that identities that do have their provider user ID double-encrypted are fixed.


thedadams:
Issue: https://github.com/obot-platform/obot/issues/2577

g-linville (approved):

PR #2777, opened 2025-05-01
feat: add project membership

g-linville:
This replaces the existing ProjectAuthorizations and ThreadAuthorizations with our new project membership and invitations.

I dropped ProjectAuthorizations altogether, since they were an unnecessary wrapper around ThreadAuthorizations. ThreadAuthorizations still exist and serve as a mapping between a project and its members (other than the project owner, whose ID exists on the spec of the project itself). However, I remove the `accepted` field from ThreadAuthorizations, because we handle user consent through the new ProjectInvitations.

ProjectInvitations are another kinm-level object that I added. They contain a uuid that we use as a code. Any user who visits `<proto>://<host>[:<port>]/i/<code>` will be able to accept or reject the invitation. If they accept it, a new ThreadAuthorization gets created to map their user to the project.

Only project owners (the user who originally created the project) can create and manage ProjectInvitations for the project. Only project owners can remove other members from the project as well.


cjellick (approved):
approving from a ux standpoint, not backend code

pkg/api/handlers/projectinvitation.go
```diff
@@ -0,0 +1,332 @@
...
+	thread, err := getProjectThread(req)
+	if err != nil {
+		return err
+	}
+
+	if !thread.Spec.Project {
+		return types.NewErrBadRequest("only projects can have invitations")
+	}
+
+	if thread.Spec.UserID != req.User.GetUID() {
```
thedadams: Should it be possible for an admin to invite someone to anyone else's project?

pkg/api/handlers/projectinvitation.go
```diff
@@ -0,0 +1,332 @@
...
+	// Generate a random code for the invitation
+	code := strings.ReplaceAll(uuid.New().String(), "-", "")
+
+	invitation := v1.ProjectInvitation{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      code,
+			Namespace: req.Namespace(),
+		},
+		Spec: v1.ProjectInvitationSpec{
+			Status:   string(types.ProjectInvitationStatusPending),
```
thedadams: Is there a reason why `Status` shouldn't be of type `ProjectInvitationStatus`?

pkg/api/handlers/projectinvitation.go
```diff
@@ -0,0 +1,332 @@
...
+	thread, err := getProjectThread(req)
+	if err != nil {
+		return err
+	}
+
+	if !thread.Spec.Project {
+		return types.NewErrBadRequest("only projects can have invitations")
+	}
+
+	if thread.Spec.UserID != req.User.GetUID() {
```
thedadams: Same question here about admins doing this.

pkg/api/handlers/projectinvitation.go
```diff
@@ -0,0 +1,332 @@
...
+	// Verify the invitation belongs to this project
+	if invitation.Spec.ThreadID != thread.Name {
+		return types.NewErrBadRequest("invitation does not belong to this project")
+	}
+
+	if err := req.Delete(&invitation); err != nil {
+		return err
+	}
+
+	return req.Write(nil)
```
thedadams: This is equivalent:
```suggestion
	return req.Delete(&invitation)
```

pkg/api/handlers/projectinvitation.go
```diff
@@ -0,0 +1,332 @@
...
+func (h *ProjectInvitationHandler) GetInvitation(req api.Context) error {
+	var (
+		code       = req.PathValue("code")
+		invitation v1.ProjectInvitation
+	)
+
+	if err := req.Get(&invitation, code); err != nil {
+		if apierrors.IsNotFound(err) {
+			return types.NewErrNotFound("invitation not found")
+		}
```
thedadams: If you just return the error, then the framework will detect the `NotFoundError` and do the same thing here.

There are a few places where you do this.

pkg/api/handlers/projectinvitation.go
```diff
@@ -0,0 +1,332 @@
...
+		return types.NewErrBadRequest("invitation is no longer valid")
+	}
+
+	var thread v1.Thread
+	if err := req.Get(&thread, invitation.Spec.ThreadID); err != nil {
+		return err
+	}
+
+	// Update invitation status to rejected
+	invitation.Spec.Status = string(types.ProjectInvitationStatusRejected)
```
thedadams: Is there a reason do you don't do the userID check in this function? I think it's the only one you don't do it in.

pkg/controller/handlers/projectinvitation/projectinvitation.go
```diff
@@ -0,0 +1,56 @@
...
+	if invitation.Spec.Status != string(types.ProjectInvitationStatusPending) {
+		return nil
+	}
+
+	if time.Since(invitation.CreationTimestamp.Time) > 7*24*time.Hour {
+		invitation.Spec.Status = string(types.ProjectInvitationStatusExpired)
+		invitation.Status.RespondedTime = &metav1.Time{Time: time.Now()}
+		return req.Client.Status().Update(req.Ctx, invitation)
+	}
+	resp.RetryAfter(7*24*time.Hour - time.Since(invitation.CreationTimestamp.Time))
```
thedadams: Only retry if the next time is within 10 hours, otherwise the auto re-enqueue the world will handle it.

pkg/controller/handlers/projectinvitation/projectinvitation.go
```diff
@@ -0,0 +1,56 @@
...
+
+// Cleanup deletes the project invitation if it was accepted, rejected, or marked as expired more than 7 days ago.
+func Cleanup(req router.Request, resp router.Response) error {
+	invitation := req.Object.(*v1.ProjectInvitation)
+
+	if !invitation.Status.RespondedTime.IsZero() {
+		if time.Since(invitation.Status.RespondedTime.Time) > 7*24*time.Hour {
+			return req.Client.Delete(req.Ctx, invitation)
+		}
+		resp.RetryAfter(7*24*time.Hour - time.Since(invitation.Status.RespondedTime.Time))
```
thedadams: Same comment about 10 hours.

pkg/api/handlers/projects.go
```diff
@@ -18,214 +19,108 @@ import (
...
+	if !thread.Spec.Project {
+		return types.NewErrBadRequest("only projects can have members")
 	}
 
-	auth := auths.Items[0]
-	if !auth.Spec.Accepted {
-		auth.Spec.Accepted = true
-		return req.Update(&auth)
+	// Only the project creator can remove members
+	if thread.Spec.UserID != req.User.GetUID() {
```
thedadams: Should admins be able to do this?

pkg/api/handlers/projects.go
```diff
@@ -18,214 +19,108 @@ import (
...
-
-	for _, threadAuth := range threadAuths.Items {
-		result.Items = append(result.Items, types.ProjectAuthorization{
-			Target:   threadAuth.Spec.UserID,
-			Accepted: threadAuth.Spec.Accepted,
-		})
-	}
-
-	return req.Write(result)
+	return req.Write(nil)
```
thedadams:
```suggestion
	return nil
```

pkg/api/handlers/projectinvitation.go
g-linville: Nope, fixed

pkg/api/handlers/projectinvitation.go
g-linville: I don't do it in GetInvitation, AcceptInvitation, and RejectInvitation. We allow all authenticated users to do this. The security comes from the code being unguessable. This is just like using an obot share link.

pkg/api/handlers/projects.go
g-linville: fixed

pkg/controller/handlers/projectinvitation/projectinvitation.go
g-linville: fixed

pkg/controller/handlers/projectinvitation/projectinvitation.go
g-linville: fixed

pkg/api/handlers/projectinvitation.go
g-linville: fixed

pkg/api/handlers/projectinvitation.go
g-linville: fixed

pkg/api/handlers/projectinvitation.go
g-linville: Yes, fixed

pkg/api/handlers/projectinvitation.go
g-linville: Yes, fixed

thedadams (approved):

PR #2780, opened 2025-05-02
fix: public chatbot user ui

njhale:
Ensure the editor field is set on project shares for owners of a chatbot so they can access the chatbot.
This also ensures owners never encounter the warning page and fixes a problem where the warning page doesn't close after clicking "accept" (note this applied to both users and
owners).


njhale:
Superseded by https://github.com/obot-platform/obot/pull/2783

PR #2788, opened 2025-05-03
feat: implement loop feature

### LINKED ISSUE #2267: Implement for each loop over datasets

ryu-man:
Addresses  #2267

- Implement task loop messages
- Implement loop iteration navigation


ui/user/src/lib/actions/size.svelte.ts
```diff
@@ -0,0 +1,48 @@
+import gsap from "gsap";
+
+export type AnimateHeightActionProps = {
+    isActive?: () => boolean,
+    vars?: Pick<gsap.TweenVars, 'duration' | 'delay' | 'ease'>
+}
+
+export function animateHeight(node: HTMLElement, { isActive, vars }: AnimateHeightActionProps) {
```
ivyjeong13: Haven't pulled this down, but is this to get an element to expand when it comes into the DOM? Does svelte's slide not work for this? Generally, we're trying to avoid adding libraries if we don't need to.

ryu-man: Not exactly, but this does animate the height of a message container during the task run to make changes look smooth; unlike Svelte's slide transition, this works only when the DOM element is live on the tree, which is something that cannot be done using Svelte's built-in api.

ui/user/src/lib/components/tasks/Step.svelte
```diff
@@ -107,62 +188,208 @@
...
+
+		await run?.($state.snapshot(step));
+	}
+
+	function getTaskRunProgress(stepId: string, messageIds: string[]) {
+		// Define a regex pattern to extract iterations data
+		const pattern = new RegExp(`^${stepId}{element=(\\d+)}{step=(\\d+)}`);
+
+		return messageIds
+			.filter((key) => pattern.test(key))
```
ivyjeong13: I believe these four array iterations (filter, map, map, reduce) can be compacted into one reduce!

ryu-man: - `filter` has to stay to strip out non matched elements and reduce only the needed ones;
- `map` can be compacted into `reduce`, but I used it to make the code more readable as it separates the data preparation step from the reduction step however, the two `map` fns can be compacted

ivyjeong13: Can't the filter happen in the reduce, and if it fails the check, it doesn't affect the result of the `reduce`? 

Like: 
```
(acc, key) => {
			const match = key.match(pattern);
			if (!match) return acc;
```

And this is mostly a nit atm, unless we're worried about messageIds becoming a potentially large array

ui/user/src/lib/components/tasks/LoopStep.svelte
```diff
@@ -0,0 +1,89 @@
+<script lang="ts">
+	import type { Messages, Project } from '$lib/services';
```
ivyjeong13, line 2: Looks like I'm getting some mirroring issues:
https://www.loom.com/share/174d3a7d18bd46bc9111f6e00bf3bfd4

* step changes is mirrored across all steps & substeps
* when trying to delete, it deletes the one step but keeps the dialog up and continues to delete the next one if I click confirm

ui/user/src/lib/components/tasks/Step.svelte
ryu-man: We can do that. I just don't find it be cleaner and if we have to go with that, I think switching to a `for of` loop will be better
```
for (key of messageIds){
   const match = key.match(pattern);
   if (!match) continue;
   ...rest of the code
}
```

ui/user/src/lib/components/tasks/LoopStep.svelte
ryu-man: Oh, I forget to update the array index; it was set to 0

ui/user/src/lib/actions/size.svelte.ts
ivyjeong13: Ah I see, while it's running, you want the height to expand smoothly!

Just small things:
* Looks like some buttons need to account for light mode
* nit: large blank spaces in step when thread content hasn't yet been supplied, would be nice if it started at one line height's height instead 
* nit: Looks like while the height is tweening, the text runs into the outlined focus area?
https://www.loom.com/share/6ac4af245ba047f2bffa96ee86ddb6a1

Screenshot 2025-05-06 at 9 47 17 AM

ivyjeong13: I did try the smooth height with just css (ease-in-out transform, max height). Caveat is a max height does have to be set, but probably can be dynamically changed with a lil bit of js:
https://www.loom.com/share/d736133bc7ad4fd0a0b3374134d68bb7

ui/user/src/lib/components/tasks/LoopStep.svelte
ivyjeong13: :+1: See this fixed!

ui/user/src/lib/actions/size.svelte.ts
ivyjeong13: Screenshot 2025-05-06 at 10 24 43 AM

Sorry, one last thing/question! I see the results of the parent step, but should I also see the contents of loop steps in a previous task run? I had 18 iterations but they're all empty when I click through them.

ryu-man: > nit: Looks like while the height is tweening, the text runs into the outlined focus area?

That was intended to keep tracking only the current running step or the current running loop step; it is not related to height tweening

ryu-man: > Sorry, one last thing/question! I see the results of the parent step, but should I also see the contents of loop steps in a previous task run? I had 18 iterations but they're all empty when I click through them.

Yes, you should be able to navigate forward and backward, but this does seems to be an issue, could send your example to reproduce that in my side

ryu-man: >nit: Looks like while the height is tweening, the text runs into the outlined focus area?

This should be fixed

ui/user/src/app.css
```diff
@@ -191,6 +191,19 @@
 		font-family: inherit;
 		line-height: inherit;
 	}
+
+	.scrollbar-gutter-stable {
+		scrollbar-gutter: stable;
+	}
+
+	.transition-height {
```
cjellick, line 199: what does this do?

ui/user/src/lib/actions/size.svelte.ts
```diff
@@ -0,0 +1,32 @@
+export function transitionParentHeight(node: HTMLElement, fn: () => unknown) {
```
cjellick, line 1: can you explain how i can see this in action? In the past, darren tends to be particuarly senstive to changes in how we output chat messages. so i want to understand the impact

ui/user/src/lib/components/tasks/Step.svelte
```diff
@@ -44,20 +45,110 @@
...
+				// set task as running
+				isRunning = true;
+
+				inProgressFalseCount = 0;
+			}
+		});
+	});
+
+	// Check whether the current step has looping steps (sub steps)
+	// It should have the
```
cjellick: an incomplete comment

cjellick (approved):

ui/user/src/lib/components/tasks/Step.svelte
ryu-man: Not sure what I want to say; the comment above it have enough explanation atm; I am going to remove it

ui/user/src/app.css
ryu-man: the `scrollbar-gutter: stable;` is used to force the browser to reserve a space for the scrollbar, so when a scrollbar shown, it will not cause layout shifts

ui/user/src/lib/components/messages/Message.svelte
```diff
@@ -397,7 +400,10 @@
 				return null;
 			}
 		})()}
-		<div transition:slide={{ duration: 300 }} class="mb-4 flex w-full flex-col justify-start gap-4">
+		<div
+			transition:slide={{ duration: 300, easing: linear }}
```
ivyjeong13, line 404: https://www.loom.com/share/815801caafc34235b8a3f577c22a41cf Loom to show but I'm having issues with loading my agents and tracked it down to the `easing: linear`

Just wanted to make sure this wasn't causing issues with you?

edit: Just got message from Craig about deleting node_modules and will try that!

ui/user/src/lib/actions/size.svelte.ts
ryu-man: This should be explained in a loom video, I think

ui/user/src/lib/components/messages/Message.svelte
ryu-man: I haven't faced something similar before! is there any errors on the console?

ivyjeong13: Got something like this in the console 
`GET http://localhost:8080/node_modules/.vite/deps/svelte_easing.js?t=1746656362547&v=0780ebfd net::ERR_ABORTED 504 (Gateway Timeout)` 

But deleting node_modules and reinstalling seems to have addressed it!

ivyjeong13 (approved):

PR #2804, opened 2025-05-06
enhance: add the ability to deploy MCP servers in a deployment

thedadams:
The necessary secret, configmap, and server are also created.

### DIFF
--- a/chart/templates/mcp.yaml
+++ b/chart/templates/mcp.yaml
@@ -1,0 +1,33 @@
+{{- if and (ne .Values.config.OBOT_SERVER_MCPBASE_IMAGE "") (ne .Values.config.OBOT_SERVER_MCPNAMESPACE "") }}
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: {{ .Values.config.OBOT_SERVER_MCPNAMESPACE }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: {{ .Values.config.OBOT_SERVER_MCPNAMESPACE }}
+  name: obot-mcp-role
+rules:
+  - apiGroups: [""]
+    resources: ["secrets", "services"]
+    verbs: ["create", "get", "list", "watch", "update", "patch", "delete"]
+  - apiGroups: ["apps"]
+    resources: ["deployments"]
+    verbs: ["create", "get", "list", "watch", "update", "patch", "delete"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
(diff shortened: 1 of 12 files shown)


pkg/api/handlers/mcp.go
```diff
@@ -332,11 +329,57 @@ func (m *MCPHandler) UpdateServer(req api.Context) error {
 		return err
 	}
 
-	cred, err := m.gptscript.RevealCredential(req.Context(), []string{fmt.Sprintf("%s-%s", existing.Spec.ThreadName, existing.Name)}, existing.Name)
+	// Shutdown the MCP server using any shared credentials.
```
g-linville: We shut down the server here because it is going to get re-deployed into k8s by the controller, right?

g-linville (approved):

pkg/api/handlers/mcp.go
thedadams: This is correct except for the controller part. It will get redeployed by the loader, typically when getting the tools.

Slightly longer explanation: the loader will start the server if it doesn't have a session for it when it is doing tool discovery. This is a GPTScript thing when it dynamically generates the GPTScript tools needed for the LLM call. We are shutting down the existing one with the expectation that a new server with the new configuration will be deployed when it's needed.

thedadams:
Re-requesting reviews. Mainly just the last commit: cleaning up all MCP servers when they are updated or their credentials change.

g-linville (approved):

chart/templates/mcp.yaml
```diff
@@ -0,0 +1,33 @@
...
+    verbs: ["create", "get", "list", "watch", "update", "patch", "delete"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  namespace: {{ .Values.config.OBOT_SERVER_MCPNAMESPACE }}
+  name: obot-mcp-rolebinding
+subjects:
+  - kind: ServiceAccount
+    name: obot
```
drpebcak: instead of hardcoding the name, use `{{ include "obot.serviceAccountName" . }}`

drpebcak (changes requested):

chart/templates/mcp.yaml
thedadams: Crap, thanks. I meant to do that.

drpebcak (approved):

cjellick:
For testing, the takeaway here is that all MCPs now get deployed in kubernetes deployment. Legacy tool bundles DO NOT. thye are still deployed as-is.

@sangee2004 i dont think you need to test this explicitly, i just want you to be aware that this is how its functioning now. So you can just ACK this issue and move it to done

PR #2823, opened 2025-05-07
Model thread selection

g-linville:


apiclient/types/thread.go
```diff
@@ -37,6 +37,8 @@ type Thread struct {
 	Project         bool     `json:"project,omitempty"`
 	Env             []string `json:"env,omitempty"`
 	Ready           bool     `json:"ready,omitempty"`
+	ModelProvider   string   `json:"modelProvider,omitempty"`
+	Model           string   `json:"model,omitempty"`
```
g-linville: TODO: why did I add this both here and on the manifest?

pkg/api/handlers/threads.go
```diff
@@ -66,6 +66,8 @@ func convertThread(thread v1.Thread) types.Thread {
 		Project:         thread.Spec.Project,
 		Env:             env,
 		Ready:           thread.Status.Created,
+		Model:           thread.Spec.Manifest.Model,
+		ModelProvider:   thread.Spec.Manifest.ModelProvider,
```
g-linville: TODO: probably unneeded since it is already in the manifest

pkg/controller/handlers/threads/threads.go
```diff
@@ -579,3 +579,28 @@ func (t *Handler) RemoveOldFinalizers(req router.Request, _ router.Response) err
 	}
 	return nil
 }
+
+func (t *Handler) MigrateProjectModelSettings(req router.Request, _ router.Response) error {
```
g-linville: TODO: add a comment to this

pkg/api/handlers/threads.go
thedadams: Yeah, I say remove these fields.

pkg/controller/handlers/threads/threads.go
thedadams: We don't need to do this. It's only on main and doesn't have a UI (until your changes go in), so no one is actually using the old fields. Just drop them.

pkg/invoke/invoker.go
```diff
@@ -400,7 +417,7 @@ func (i *Invoker) Agent(ctx context.Context, c kclient.WithWatch, agent *v1.Agen
 		PreviousRunName:       opt.PreviousRunName,
 		ForceNoResume:         opt.ForceNoResume,
 		GenerateName:          opt.GenerateName,
-		DefaultModel:          project.Spec.Model,
+		DefaultModel:          model,
```
thedadams: I say we simplify this and remove this `DefaultModel` field and the one on the Run's Spec. That way, we can calculate what the default model should be in one place (in the Resume method). The base case should be `types.DefaultModelAliasTypeLLM`.

pkg/storage/apis/obot.obot.ai/v1/thread.go
```diff
@@ -99,13 +99,21 @@ type ThreadSpec struct {
 	SystemTools []string `json:"systemTools,omitempty"`
 	// Capabilities are the capabilities of this thread
 	Capabilities ThreadCapabilities `json:"capabilities,omitempty"`
-	// ModelProvider is the model provider the user has chosen for this thread.
-	// Currently, only supported for projects.
+	// DEPRECATED: Use DefaultModelProvider instead.
```
thedadams: Yeah, just remove them. No one is using them.

pkg/controller/handlers/threads/threads.go
g-linville: this is fantastic news

pkg/render/render.go
```diff
@@ -64,7 +59,6 @@ func Agent(ctx context.Context, db kclient.Client, gptClient *gptscript.GPTScrip
 		Temperature:  agent.Spec.Manifest.Temperature,
 		Cache:        agent.Spec.Manifest.Cache,
 		Type:         "agent",
-		ModelName:    model,
```
g-linville, line 67: There was no need to specify a model here on the main tool, since we specify it for all of GPTScript in the invoker.

pkg/thread/threadmodel.go
```diff
@@ -0,0 +1,65 @@
...
+	"slices"
+
+	"github.com/obot-platform/nah/pkg/router"
+	"github.com/obot-platform/obot/apiclient/types"
+	"github.com/obot-platform/obot/pkg/projects"
+	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
+	kclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func GetModelAndModelProviderForThread(ctx context.Context, c kclient.Client, thread *v1.Thread) (string, string, error) {
```
g-linville, line 15: This is the one place with the logic to determine the default model for a thread.

apiclient/types/assitant.go
```diff
@@ -2,16 +2,17 @@ package types
...
+	Default               bool              `json:"default"`
+	Description           string            `json:"description"`
+	Icons                 AgentIcons        `json:"icons"`
+	Alias                 string            `json:"alias,omitempty"`
+	IntroductionMessage   string            `json:"introductionMessage"`
+	StarterMessages       []string          `json:"starterMessages"`
+	EntityID              string            `json:"entityID"`
+	MaxTools              int               `json:"maxTools,omitempty"`
+	WebsiteKnowledge      *WebsiteKnowledge `json:"websiteKnowledge,omitempty"`
+	AllowedModelProviders []string          `json:"allowedModelProviders,omitempty"`
```
thedadams, line 15: Oh, I missed this. Thanks!

thedadams (approved):

PR #2828, opened 2025-05-07
Add support for creating, managing, and consuming Agent Templates

njhale:
Support creating, managing, and consuming Agent Templates

### Backend Changes

- Introduces a new public-facing `ProjectTemplate` resource and APIs to interact with it
- `ProjectTemplates` are implemented as an extension of `Thread` and `ThreadShare`, reusing much of the existing "copy Obot" logic

### Frontend Changes

- Adds support for adding, managing, publishing, and consuming Agent Templates in the user UI
- Overhauls the Agent Catalog to be template-backed
- Implements the same search-based UX pattern used by the MCP Server Catalog


ui/user/src/lib/components/Obot.svelte
```diff
@@ -230,7 +230,7 @@
 							/>
 						{/key}
 					{:else if layout.sidebarConfig}
-						<SidebarConfig bind:project bind:currentThreadID {assistant} />
+						<SidebarConfig bind:project {assistant} />
```
njhale: I think this was dropped by accident. I'll double-check.

apiclient/types/template.go
```diff
@@ -2,12 +2,19 @@ package types
 
 type ProjectTemplate struct {
```
cjellick, line 3: so projectTemplate existed as a thing before? what did it do?

njhale: Nothing. It existed, but nothing depended on it and it wasn't exposed by any API.

ui/user/src/lib/components/Obot.svelte
njhale: Didn't affect anything AFAIK. I'm going to drop this change to avoid breaking anything unintentionally.

cjellick (approved):

Issue #2842, opened 2025-05-08
Implement task run auto-follow


ryu-man:
- Add the ability for user to keep following the task run process (scroll to the bottom and keep scrolled as new messages came in)
- Make it possible to scroll-up to to check on results

shweta8996:
I am not able to verify the task due to this issue [(https://github.com/obot-platform/obot/issues/2978)]

shweta8996:
I have verified this on Obot Main, and it is working as expected.

The task run view automatically scrolls to the bottom as new messages come in, and users can still scroll up to view previous results without any issues. Both behaviors are functioning correctly.

https://github.com/user-attachments/assets/7a21a861-cb2c-41ff-965e-3700b08f1a41

PR #2850, opened 2025-05-08
feat: add task run autoscroll & scroll navigation

ryu-man:
Addresses #2842 

- add auto scroll feature during task run
- add scroll navigation 
- enhance steps animations and UI

### DIFF
--- a/ui/user/src/lib/components/tasks/Step.svelte
+++ b/ui/user/src/lib/components/tasks/Step.svelte
@@ -16,6 +16,7 @@
 	import { tooltip } from '$lib/actions/tooltip.svelte';
 	import LoopStep from './LoopStep.svelte';
 	import { transitionParentHeight } from '$lib/actions/size.svelte';
+	import { linear } from 'svelte/easing';
 
 	interface Props {
 		parentStale?: boolean;
@@ -258,10 +259,10 @@
 				{#if loopDataMessages.length > 0 && showOutput}
 					<!-- Show step message -->
 					<div
-						class="transition-height relative my-3 -ml-4 box-content flex min-h-11 flex-col gap-4 overflow-hidden rounded-lg bg-white p-5 dark:bg-black"
+						class="transition-height relative my-3 -ml-4 box-content flex min-h-6 flex-col gap-4 overflow-hidden rounded-lg bg-white p-5 dark:bg-black"
 						class:outline-2={isRunning}
 						class:outline-blue={isRunning}
-						transition:slide
+						transition:slide={{ duration: !readOnly ? 200 : 0, easing: linear }}
 					>
 						<div
 							class="message-container flex w-full flex-col gap-4"
@@ -295,7 +296,7 @@
 									</div>
 								</div>
 
-								{#each step.loop! as _, j}
+								{#each step.loop! as _, j (_)}
 									<!-- Get the current step messages array -->
 									{@const stepMessages = messages[j] ?? []}
 
@@ -314,6 +315,7 @@
 										{stale}
 										onKeydown={onkeydown}
 										onDelete={() => step.loop!.splice(j, 1)}
+										index={j}
 									/>
 								{/each}
 							</div>
@@ -354,8 +356,8 @@
 			{/if}
 		</div>
 
-		<div class="flex items-start">
-			{#if !readOnly}
+		{#if !readOnly}
+			<div class="flex items-start">
 				<button
 					class="icon-button"
 					class:text-blue={isLoopStep}
@@ -417,18 +419,18 @@
 						{/if}
 					</div>
 				</div>
-			{/if}
-		</div>
+			</div>
+		{/if}
 	</div>
 </li>
 
 <!-- This code section is responsible for showing messages in a !loop task -->
 {#if !isLoopStep && messages.length > 0 && showOutput}
 	<div
-		class="transition-height relative my-3 box-content flex min-h-11 flex-col gap-4 overflow-hidden rounded-lg bg-white p-5 dark:bg-black"
+		class="transition-height relative my-3 box-content flex min-h-6 flex-col gap-4 overflow-hidden rounded-lg bg-white p-5 dark:bg-black"
 		class:outline-2={isRunning}
 		class:outline-blue={isRunning}
-		transition:slide
+		transition:slide={{ duration: !readOnly ? 200 : 0, easing: linear }}
 	>
 		<div
 			class="messages-container flex w-full flex-col gap-4"
--- a/ui/user/src/lib/components/tasks/Step.svelte
+++ b/ui/user/src/lib/components/tasks/Step.svelte
@@ -296,7 +296,7 @@
 									</div>
 								</div>
 
-								{#each step.loop! as _, j (_)}
+								{#each step.loop! as _, j (j)}
 									<!-- Get the current step messages array -->
 									{@const stepMessages = messages[j] ?? []}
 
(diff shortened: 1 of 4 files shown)


ui/user/src/lib/components/tasks/Step.svelte
```diff
@@ -295,7 +296,7 @@
 									</div>
 								</div>
 
-								{#each step.loop! as _, j}
+								{#each step.loop! as _, j (_)}
```
ivyjeong13: nit: Is the `(_)` needed? Is that you're not setting a key? Then can it be dropped?

ui/user/src/lib/components/tasks/Task.svelte
```diff
@@ -51,6 +51,55 @@
...
 
+	/**************************************************************************************************/
+	// HACK: fix glitch that happens when in messages.inProgress when loop steps executed
+	// Make sure that the .inProgress stays true until the end of the current run task
+
+	let timeoutId: number | undefined = undefined;
+	// save how many step.inProgress === false we got
+	let inProgressFalseCount = $state(0);
+
+	// svelte-ignore state_referenced_locally
```
ivyjeong13: don't use svelte-ignore

ui/user/src/lib/components/tasks/Task.svelte
```diff
@@ -51,6 +51,55 @@
...
+	let isRunning = $state(allMessages.inProgress);
+
+	// Indicate whther the user is activating the task run following or not
+	let shouldFollowTaskRun = $state(true);
+
+	$effect(() => {
+		(() => allMessages?.inProgress)();
```
ivyjeong13, line 68: Wondering what's this for?

ui/user/src/lib/components/tasks/Steps.svelte
```diff
@@ -53,29 +234,63 @@
...
+				onclick={onNavigationClick}
+				in:fade={{ duration: 100, delay: 0, easing: linear }}
+				out:fade={{ duration: 50, delay: 0, easing: linear }}
+			>
+				<div
+					class="h-4 w-4 duration-200"
+					class:rotate={!isFollowModeActive && scrollDirection === 'up'}
+				>
+					{#if isFollowModeActive}
+						<UsersRound class="h-full w-full" />
```
ivyjeong13, line 276: nit: Not sure what's the best icon to use here 🤔 But I would have assumed it's something to do with multiple users instead
Screenshot 2025-05-09 at 5 49 26 PM

ui/user/src/lib/components/tasks/Task.svelte
ryu-man: svelte-check complains about this statement and suggest to use a $derived instead of $state; which does not works for me, so this was the only way to make it pass

ui/user/src/lib/components/tasks/Steps.svelte
```diff
@@ -29,13 +34,189 @@
...
+			scrollableElement!.scrollHeight - scrollableElement!.clientHeight
+		);
+
+		scrollableElement!.scrollTo({
+			top,
+			behavior: running && shouldFollowTaskRun ? 'instant' : 'smooth'
+		});
+	}
+
+	function scrollUp() {
```
ivyjeong13, line 207: Not PR blocking but lil odd thing I noticed where I get bounced back up: 
https://www.loom.com/share/263577b1f2bf4b1786e19cff14cedb5a

Thinking it's something to do with when the sticky header comes into play?

ui/user/src/lib/components/tasks/Task.svelte
ryu-man: - This code section is hack to fix a glitch issue happens when loop task is running; the `.inProgress` property keeps switch on & off on each iteration.
- The use of statement `(() => allMessages?.inProgress)();` is to make `svelte-check` pass (this `allMessages?.inProgress` throws an error); it servers as a trigger for the effect, because I am setting $state variable inside an `untrack` function  to avoid re-trigerring and I need it to run whenever the in progress status of messages changed.

ui/user/src/lib/components/tasks/Step.svelte
ryu-man: only need as the key for the loop item, I would prefer not dropping it

ui/user/src/lib/components/tasks/Steps.svelte
ryu-man: I don't think that, maybe it has something to do with task header transition!

ui/user/src/lib/components/tasks/Task.svelte
ivyjeong13: Looks like it's because using $state will make it so it just captures the initial value. So instead of using allMessages.inProgress, couldn't you just start with $state(false) since that's the initial value of inProgress?

ui/user/src/lib/components/tasks/Steps.svelte
ryu-man: The thinking behind using this icon is because it shows a person behind another person which could be used to indicate "following" status. I couldn't find an ideal icon in this context, I would like if you can throw some suggestions.
I thought about the eye but we are already using it in anther context

ui/user/src/lib/components/tasks/Task.svelte
ryu-man: make sense

ui/user/src/lib/components/tasks/Task.svelte
ivyjeong13: I'm curious what error you're getting for allMessages?.inProgress! I just tried and didn't have any issues. 

https://www.loom.com/share/c03100a3d1ac46a09069034b78600c83

ui/user/src/lib/components/tasks/Steps.svelte
ivyjeong13: Yup, that's the sticky header I'm talking about

ui/user/src/lib/components/tasks/Step.svelte
ivyjeong13: Does `(_)` set a key? Underscore looks like specifying unused syntax to me 😅 I thought you'd need to do something like `{#each step.loop! as _, j (j)} // <--- setting what the key should be in the parentheses`

ui/user/src/lib/components/tasks/Task.svelte
ryu-man: https://www.loom.com/share/76e5c1df050f477c86856973fa34d599?sid=b72af6f6-ea0d-46ee-a613-7cb0611cbc71

ryu-man: using this `if (allMessages.inProgress || !allMessages.inProgress)` would work but it is also another hack!

ryu-man: This comes with a tradeoff
https://www.loom.com/share/76e5c1df050f477c86856973fa34d599?sid=b72af6f6-ea0d-46ee-a613-7cb0611cbc71

ryu-man: using this `if (allMessages.inProgress || !allMessages.inProgress)` would work, but it is also another hack!

ui/user/src/lib/components/tasks/Steps.svelte
ryu-man: Oh, I've got you now

ui/user/src/lib/components/tasks/Step.svelte
ryu-man: Oh, I got you. I don't feel comfortable using array index as a key; that's why I am using the item. but let change it for this one

cjellick (approved):

PR #2867, opened 2025-05-09
enhance: enable shared chatbot config

njhale:
- Prevent chatbot config creds from being shared automatically with chatbot instances
- Add APIs to share creds with chatbot instances


pkg/api/authz/mcpserver.go
```diff
@@ -33,10 +33,11 @@ func (a *Authorizer) checkMCPServer(req *http.Request, resources *Resources, _ u
...
 	}
 
-	if resources.Authorizated.Project.Name != mcpServer.Spec.ThreadName {
-		return false, nil
+	if resources.Authorizated.Project.Name == mcpServer.Spec.ThreadName ||
+		resources.Authorizated.Project.Spec.ParentThreadName == mcpServer.Spec.ThreadName {
+		return true, nil
 	}
 
 	resources.Authorizated.MCPServer = &mcpServer
```
g-linville: This definitely needs to be moved into the if statement, since we flipped the return values.

pkg/api/handlers/mcp.go
```diff
@@ -262,6 +307,15 @@ func (m *MCPHandler) UpdateServer(req api.Context) error {
...
 	)
 
+	project, err := getProjectThread(req)
+	if err != nil {
+		return err
+	}
+
+	if project.Name != existing.Spec.ThreadName {
+		return types.NewErrForbidden("cannot edit MCP server from this project")
+	}
```
g-linville, line 324: This needs to be moved down after we call `req.Get` to fetch the existing MCP server.

pkg/render/mcp.go
```diff
@@ -22,10 +22,21 @@ func (e *UnconfiguredMCPError) Error() string {
...
+		credCtxs := []string{
+			fmt.Sprintf("%s-%s", thread.Name, mcpServer.Name),
+		}
+		if parent := thread.Spec.ParentThreadName; parent != "" {
+			credCtxs = append(credCtxs, fmt.Sprintf("%s-%s", parent, mcpServer.Name))
+
+			if parent != mcpServer.Spec.ThreadName {
+				credCtxs = append(credCtxs, fmt.Sprintf("%s-%s-shared", mcpServer.Spec.ThreadName, mcpServer.Name))
+			}
+		}
```
g-linville: This seems bad? If I understand right, it should be this:

```suggestion
		if parent := thread.Spec.ParentThreadName; parent != "" {
			if parent != mcpServer.Spec.ThreadName {
				credCtxs = append(credCtxs, fmt.Sprintf("%s-%s-shared", mcpServer.Spec.ThreadName, mcpServer.Name))
			} else {
				credCtxs = append(credCtxs, fmt.Sprintf("%s-%s", parent, mcpServer.Name))
			}
		}
```

g-linville (changes requested):

pkg/api/handlers/mcp.go
```diff
@@ -147,26 +167,33 @@ func (m *MCPHandler) getServer(req api.Context, withTools bool) error {
 		return err
 	}
 
-	cred, err := m.gptscript.RevealCredential(req.Context(), []string{fmt.Sprintf("%s-%s", server.Spec.ThreadName, server.Name)}, server.Name)
+	thread, err := getThreadForScope(req)
+	if err != nil {
+		return err
+	}
```
thedadams: It looks like this didn't have to move.

pkg/api/handlers/mcp.go
```diff
@@ -185,6 +212,15 @@ func (m *MCPHandler) DeleteServer(req api.Context) error {
 		id     = req.PathValue("mcp_server_id")
 	)
 
+	project, err := getProjectThread(req)
+	if err != nil {
+		return err
+	}
+
+	if project.Name != server.Spec.ThreadName {
```
thedadams, line 228: Can we add a comment here? It look a little while for me to figure out why this was necessary.

pkg/api/handlers/mcp.go
```diff
@@ -208,6 +244,15 @@ func (m *MCPHandler) CreateServer(req api.Context) error {
 		return err
 	}
 
+	project, err := getProjectThread(req)
+	if err != nil {
+		return err
+	}
+
+	if project.Name != t.Name {
```
thedadams: Same here.

pkg/api/handlers/mcp.go
thedadams: Yes, and add a comment please.

pkg/render/mcp.go
thedadams: I agree with this.

thedadams:
Realized this after my review: you also need to add these shared credentials to the MPC server credential cleanup: https://github.com/obot-platform/obot/blob/main/pkg/controller/handlers/cleanup/credentials.go#L76

pkg/api/authz/mcpserver.go
njhale: yup, thanks for catching that!

pkg/render/mcp.go
njhale: IIUC, way it's currently implemented:
1. the chatbot's threads get _both_ their local credential context _and_ the shared context from from the agent
2. the agent's threads get their local context _but_ they don't get the shared context (we don't want their threads defaulting to what they've configured for chatbots)

It seems like the change you're suggesting would prevent chatbot threads from seeing their locally configured MCP credentials.

g-linville: Ohh okay I think that makes sense. In this case, the `thread` variable is the actual thread, not a project, so `thread.Spec.ParentThreadName` is the project for the thread, which, in the case of the chatbot, is not the "parent project" whose unshared credentials we need to avoid taking. That makes sense.

njhale: yup. This definitely warrants a comment to clarify the intent/reasoning.

pkg/render/mcp.go
```diff
@@ -22,10 +22,21 @@ func (e *UnconfiguredMCPError) Error() string {
...
 
-func mcpServerTool(ctx context.Context, gptClient *gptscript.GPTScript, mcpServer v1.MCPServer, allowTools []string) (gptscript.ToolDef, error) {
+func mcpServerTool(ctx context.Context, thread *v1.Thread, gptClient *gptscript.GPTScript, mcpServer v1.MCPServer, allowTools []string) (gptscript.ToolDef, error) {
 	var credEnv map[string]string
 	if len(mcpServer.Spec.Manifest.Env) != 0 || len(mcpServer.Spec.Manifest.Headers) != 0 {
-		cred, err := gptClient.RevealCredential(ctx, []string{fmt.Sprintf("%s-%s", mcpServer.Spec.ThreadName, mcpServer.Name)}, mcpServer.Name)
+		credCtxs := []string{
+			fmt.Sprintf("%s-%s", thread.Name, mcpServer.Name),
+		}
+		if parent := thread.Spec.ParentThreadName; parent != "" {
```
thedadams, line 32: Can we add a nil-check on the thread here? I am not convinced we always set `opt.Thread` where this is called.

thedadams (approved):

pkg/render/mcp.go
njhale: sure thing

g-linville (approved):

pkg/render/mcp.go
```diff
@@ -22,10 +22,26 @@ func (e *UnconfiguredMCPError) Error() string {
...
+
+		if parent := thread.Spec.ParentThreadName; parent != "" {
+			// Add any local MCP credentials from the parent project scope.
+			// For non-project child threads of an agent project, this will include local credentials from the agent project.
+			// For non-project child threads of a chatbot project, this will include local credentials from the chatbot project.
+			credCtxs = append(credCtxs, fmt.Sprintf("%s-%s", parent, mcpServer.Name))
+
+			if parent != mcpServer.Spec.ThreadName {
+				// Add shared MCP credentials from the agent project to chatbot threads.
+				credCtxs = append(credCtxs, fmt.Sprintf("%s-%s-shared", mcpServer.Spec.ThreadName, mcpServer.Name))
```
thedadams, line 40: I think we always want this context, even if the thread is `nil`, right?

thedadams, line 40: I think we always want this context, even if the thread is nil, right?

thedadams (changes requested):

pkg/render/mcp.go
njhale: No, we don't want the context in the case where the thread belongs to the agent project. I can explicitly add it when the thread is nil, but we really don't have enough info to make that decision at that point.

PR #2895, opened 2025-05-12
Feat: task refactoring

StrongMonkey:


cjellick (approved):
UX is good (enough)

pkg/api/handlers/projects.go
```diff
@@ -163,10 +168,12 @@ func (h *ProjectsHandler) UpdateProject(req api.Context) error {
 			}
 		}
 
+		thread.Spec.Capabilities = v1.ThreadCapabilities(*project.Capabilities)
```
thedadams: Why do we need `Capabilities` on both the manifest and the spec? Can we get away with just one?

If we only need one, then it should be on the manifest since it can be set via the API.

Also, it seems like it is possible for `project.Capabilities` to be `nil` here. We don't want to panic.

pkg/api/handlers/projects.go
```diff
@@ -474,6 +513,21 @@ func convertProject(thread *v1.Thread, parentThread *v1.Thread) types.Project {
...
+	result := types.ProjectCapabilities{
+		OnSlackMessage:   capabilities.OnSlackMessage,
+		OnDiscordMessage: capabilities.OnDiscordMessage,
+		OnEmail:          capabilities.OnEmail,
+	}
+	if capabilities.OnWebhook != nil {
+		result.OnWebhook = &types.OnWebhook{}
+		result.OnWebhook.ValidationHeader = capabilities.OnWebhook.ValidationHeader
+		result.OnWebhook.Headers = capabilities.OnWebhook.Headers
+		result.OnWebhook.Secret = "********"
```
thedadams, line 527: Also, the secret is hashed, so I think it's fine to show it.
```suggestion
		result.OnWebhook = &types.OnWebhook{
			ValidationHeader: capabilities.OnWebhook.ValidationHeader
			Headers: capabilities.OnWebhook.Headers
			Secret: "********"
		}
```

EDIT: Nope, I am wrong about the secret being hashed. Ignore that part.

pkg/controller/handlers/task/task.go
```diff
@@ -0,0 +1,373 @@
...
+}
+
+func (t *Handler) HandleTaskCreationForCapabilities(req router.Request, _ router.Response) error {
+	thread := req.Object.(*v1.Thread)
+
+	if !thread.Spec.Project {
+		return nil
+	}
+
+	if err := Validate(thread, false); err != nil {
```
thedadams, line 32: Why should `forSlack` be false here? What if this task is for slack?

pkg/controller/handlers/task/task.go
```diff
@@ -0,0 +1,373 @@
...
+			if err := req.Client.Create(req.Ctx, &workflow); err != nil {
+				return err
+			}
+			thread.Status.WorkflowNameFromIntegration = workflow.Name
+		}
+	} else {
+		if err := req.Get(&workflow, workflow.Namespace, workflow.Name); kclient.IgnoreNotFound(err) != nil {
+			return err
+		} else if err == nil {
+			if err := req.Client.Delete(req.Ctx, &workflow); err != nil && !apierrors.IsNotFound(err) {
```
thedadams: Just to be consistent, here and similar places.
```suggestion
			if err := req.Client.Delete(req.Ctx, &workflow); kclient.IgnoreNotFound(err) != nil {
```

pkg/controller/handlers/task/task.go
```diff
@@ -0,0 +1,373 @@
...
+				return err
+			}
+			if thread.Status.WorkflowNameFromIntegration == workflow.Name {
+				thread.Status.WorkflowNameFromIntegration = ""
+			}
```
thedadams: No harm in just setting it, here and similar places.
```suggestion
			thread.Status.WorkflowNameFromIntegration = ""
```

pkg/api/handlers/projects.go
StrongMonkey: My bad, this is extra field and not needed. Removed

pkg/controller/handlers/task/task.go
StrongMonkey: The slack one is different. We don't directly set project.capabilities.onSlack, and it is set through controller when slackReceiver is set. So for slack we just have to validate if other configs has been set. So this validate method will onlt be called when interfaces other than slack are set.

pkg/controller/handlers/task/task.go
StrongMonkey: addressed

ivyjeong13 (approved):
Approving for UI changes

thedadams (approved):

PR #2906, opened 2025-05-12
enhance: tools: mark tool bundles as official

g-linville:
This adds a `categories: "Official"` metadata field to all of our tool bundles. The UI can use this to make certain ones as official.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,6 @@ obot*.db*
 
 # Ignore http dir for http client files
 http/
+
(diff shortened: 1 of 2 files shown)


g-linville:
Converted to draft because I forgot to add a migration.

g-linville:
Done, ready for review again.

g-linville:
Re-requesting review because I made a change based on Craig's feedback. We're just using the existing `categories` field rather than adding a new one.

cjellick (approved):

PR #2910, opened 2025-05-12
feat: built-in capabilities section

ivyjeong13:
* `Built-In Capabilities` section for sidebar

### DIFF
--- a/ui/user/src/lib/components/Navbar.svelte
+++ b/ui/user/src/lib/components/Navbar.svelte
@@ -18,7 +18,7 @@
 		{#if leftContent}
 			{@render leftContent()}
 		{:else}
-			<div class="relative flex items-end">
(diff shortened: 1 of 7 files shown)


cjellick:
@ivyjeong13 - my mistake - website knowledge SHOULD NOT be in this list at all...it is special and weird, unforunately. you previously pointed out a field you key off of to show or hide the website knowlege bit. just got back to keying off that

ivyjeong13:
> @ivyjeong13 - my mistake - website knowledge SHOULD NOT be in this list at all...it is special and weird, unforunately. you previously pointed out a field you key off of to show or hide the website knowlege bit. just got back to keying off that

:+1: Got it, will drop from the section and move back to keying off that value

cjellick (approved):

cjellick (approved):

PR #2923, opened 2025-05-13
chore: loop steps UI refactor

ryu-man:


ui/user/src/lib/components/tasks/LoopStep.svelte
```diff
@@ -43,23 +46,36 @@
...
-				placeholder="Instructions..."
-				class="ghost-input border-surface2 h-auto grow resize-none"
-				disabled={isReadOnly}
-				onkeydown={onKeydown}
-			></textarea>
-		</div>
+	<div
+		class={twMerge(
+			'flex items-center gap-2 overflow-hidden px-4',
+			(!isReadOnly || isStepRunned || isStepRunning) && ''
```
ivyjeong13: nit: can just make this class="flex items-center gap-2 overflow-hidden px-4"

ui/user/src/lib/components/tasks/LoopStep.svelte
```diff
@@ -74,9 +90,9 @@
 
 	{#if (isStepRunning || isStepRunned) && shouldShowOutput}
 		<div
-			class="loop-step-messages transition-loop-step-message relative box-content flex min-h-11 flex-col gap-4 overflow-hidden rounded-none p-5 duration-200"
-			class:outline-2={isStepRunning && isLoopStepRunning}
-			class:outline-blue={isStepRunning && isLoopStepRunning}
+			class={twMerge(
```
ivyjeong13: also twMerge can be dropped here

ui/user/src/lib/components/tasks/Task.svelte
```diff
@@ -212,10 +215,13 @@
 
 	async function click() {
 		error = '';
-		showAllOutput = true;
-
 		shouldFollowTaskRun = true;
 
+		// Avoid glitch in iterations UI
+		delay(1000).then(() => {
```
ivyjeong13: What is the glitch occurring here? Wondering if it can be resolved another way (this is giving me flaky E2E test vibes here 😅)

ryu-man: I think this was happening on the rerun; let me try get you a video

ryu-man: Interesting, I think that behavior is gone! going to remove the code

ryu-man: Sounds like an old bug, I removed the code

ui/user/package.json
```diff
@@ -73,5 +73,8 @@
 		"typescript-eslint": "^8.25.0",
 		"vite": "^5.4.14"
 	},
-	"type": "module"
+	"type": "module",
+	"dependencies": {
+		"es-toolkit": "^1.37.2"
```
ivyjeong13: Nit: With removal of delay, this isn't needed

ivyjeong13 (approved):
Still wanna encourage I don't think we should need to import library for simple util functions but not PR blocking!

ui/user/package.json
ryu-man: removed

Issue #2926, opened 2025-05-13
Enhance model providers UX


ryu-man:

sangee2004:
All feedback from https://github.com/obot-platform/obot/pull/2927#issuecomment-2884506466 are addressed.
This one line item is not done as mentioned in the feedback
```
In chat, change this text: "No model is available. Configure you provider first!" to this:
You can add additional models under the Model Providers tab in the Configuration section.
Also, pad that out a bit so it is nicely presented. it felt really cramped.
```

cjellick:
yeah, i oked this. its fine

PR #2927, opened 2025-05-13
feat: enhance model providers ux

ryu-man:
Addresses #2926


ui/user/package.json
```diff
@@ -73,5 +73,8 @@
 		"typescript-eslint": "^8.25.0",
 		"vite": "^5.4.14"
 	},
-	"type": "module"
+	"type": "module",
+	"dependencies": {
+		"es-toolkit": "^1.37.2"
```
ivyjeong13, line 78: Wondering if es-toolkit is needed or if the imported util functions are simple enough for us to make (just again making sure, we avoid unnecessary dependencies unless the use case is complex that it is needed!)

ui/user/src/lib/components/ModelProviderCard.svelte
```diff
@@ -0,0 +1,629 @@
...
+			{/if}
+		</div>
+
+		{#if provider.description}
+			<div class="markdown-content text-gray text-sm">
+				{@html toHTMLFromMarkdown(provider.description)}
+			</div>
+		{:else}
+			<div class="bg-surface1 text-gray flex h-14 w-full items-center rounded-lg px-4 py-3 text-sm">
+				<p>Description is not available</p>
```
ivyjeong13: Screenshot 2025-05-13 at 1 36 56 PM

Feel free to resolve if this already went through with Craig and Will! Wondering if showing that description is not available or the provider is not configured text is needed? Seems like the sentence above is enough for me to know I need to configure at least one provider. If we can axe those empty spaces, would allow the user to see more model providers in their view.

ui/user/src/lib/components/ModelProviderCard.svelte
```diff
@@ -0,0 +1,629 @@
...
+								indeterminate={selectedModels.length > 0 && models.length > selectedModels.length}
+							/>
+							<label class="inline" for={`model-${provider.id}-toggle-all`}>Available Models</label>
+
+							{#if models.length}
+								<span class="opacity-50">({models.length})</span>
+							{/if}
+						</h5>
+
+						{#if selectedModels.length}
```
ivyjeong13, line 462: Screenshot 2025-05-13 at 1 47 28 PM

Would be nice to bubble the model provider(s) I did configure to the top, maybe in two sections: Configured Model Providers & Available Model Providers, if we want consistent height because more vertical space is needed for when it's configured?

ui/user/src/lib/components/ModelProviderCard.svelte
```diff
@@ -0,0 +1,629 @@
...
+
+	let defaultModelProvider = $derived(project.defaultModelProvider || '');
+	let defaultModel = $derived(project.defaultModel || '');
+
+	let isConfigured = $derived(provider?.configured ?? false);
+	let providerId = $derived(provider.id);
+	// let selectedModels = $derived<Record<string, string[]>>(project.models || {});
+
+	let configuration: Record<string, string> = $state({});
+	// svelte-ignore state_referenced_locally
```
ivyjeong13: remove svelte-ignore, do not hide warnings/errors

ui/user/src/lib/components/sidebar/ModelProviders.svelte
```diff
@@ -80,32 +65,57 @@
...
+				<p class="text-xs italic">Loading model providers...</p>
+			{:else if project?.models && Object.keys(project.models).length > 0}
+				<div class="pb-2">
+					<p class="mb-1 text-sm font-medium">Configured providers:</p>
+					<ul class="flex flex-col text-xs">
+						{#each projectProviderModelIds as providerId}
+							{@const provider = providersWithMoreData.get(providerId)}
+
+							<li class="model-provider flex items-center gap-1 py-2">
+								<div class="size-4">
```
ivyjeong13: Screenshot 2025-05-13 at 1 49 28 PM

If we're keeping the way it is, it would nice if I could click on the model provider and it takes me to it. A small QOL instead of having to click Manage Model Providers then scroll down to find it.

ui/user/src/lib/components/ModelProviders.svelte
```diff
@@ -282,194 +163,64 @@
...
-						value={defaultModelProvider && defaultModel
-							? `${defaultModel}|||${defaultModelProvider}`
-							: ''}
-						onchange={(e) => updateDefaultModel(e.currentTarget.value)}
+		<div class={twMerge('mb-6 flex flex-col gap-2', !hasOneModelSelected && 'opacity-50')}>
+			<h4 class="text-md font-medium">Default Model</h4>
+
+			<div>
+				<select
+					class="border-surface2 dark:bg-surface1 w-full rounded-md border px-4 py-4 text-sm"
```
ivyjeong13: Screenshot 2025-05-13 at 1 55 56 PM

nit: chevron icon is not evenly spaced away, right up against the border 😅

ui/user/package.json
ryu-man: At the moment it is not used for complex stuff. I am using the delay and throttle functions. I thought maybe using something already made will save us effort and time rather than re-implementing that. the library is tree-shakeable, so unused imports will not be included in the final bundle

ui/user/src/lib/components/ModelProviderCard.svelte
ryu-man: Is this about resorting the configured model providers or creating another section for configured providers?

ui/user/src/lib/components/ModelProviderCard.svelte
ryu-man: I made the provider's card to have a fixed height to prevent UI jumping in & out when the content grows and shrinks. Those placeholders are used to fill up the empty space and help the user expecting what to see behind them. I will see what @cjellick thinks about it

ui/user/src/lib/components/sidebar/ModelProviders.svelte
ryu-man: I will add that

ui/user/src/lib/components/ModelProviders.svelte
ryu-man: Nice catch. That's a select element; its chevron does not respond to padding. We probably need a UI library or to create our own at some point. I have to create a custom chevron

cjellick:
Feedback:

- In the left nav, I don’t like the notification background color for "No model providers configured". It is the only thing we have like that in the left nave and it sticks out too much. lets drop the background color
- Put this description under the Model Providers title at the top:
    - Configure model providers and select models to make them available to all threads and tasks. The default model will be used in place of Obot’s built-in model.
- Drop the bit of text under the default model drop down
- Drop the description sections in each model provider. not needed.
- The model selector needs a box around it. i don't know where it ends and it makes me over scroll
- The unconfigure button opens an old school modal. do not want those in the app at all
- Change the text inside the configure square of each model provider to this:
    - Click on the "Configure" button below to set up this provider.  We’ll then validate your configuration and display available models.
- The API key's help text cuts off in the openai provider
- The model search bar does not work how i want it. i want to be able to type gpt- and it limits to just models that start exactly that way. it should match form the start of the word
- In the openai provider, bubble models that start with gpt to the top of the list
- In chat, the model selector is too skinny. It should be as wide as the names in the selector
- In chat, change this text: "No model is available. Configure you provider first!" to this:
    - You can add additional models under the Model Providers tab in the Configuration section.
    - Also, pad that out a bit so it is nicely presented. it felt really cramped.

You can do these in a follow up pr, they are lower priority:
- if i configure with a bad key i get a 500 error, id like that displayed in that model provider’s section, guiding them to update their configuration
- When i create a new thread, i would like to be able to pick the model before every sending a message. right now the drop down doesnt show up until after my first message is sent.

Here is a loom of me going over this feedback:
https://www.loom.com/share/dbc68472302f4098923573f8c2e6d854

ryu-man:
@cjellick I am not sure I understand what you mean by adding a box around the selector!

cjellick:
just want a subtle boundary around the selector box of models
Screenshot 2025-05-16 at 4 59 00 PM

ryu-man:
@cjellick errors and models selector in chat are not addressed in these changes

ui/user/src/lib/components/ModelProviders.svelte
```diff
@@ -282,189 +170,59 @@
...
+				<select
+					class="border-surface2 dark:bg-surface1 w-full appearance-none rounded-md border px-4 py-4 text-sm"
+					value={defaultModelProvider && defaultModel
+						? `${defaultModel}|||${defaultModelProvider}`
+						: ''}
+					onchange={(e) => updateDefaultModel(e.currentTarget.value)}
+					onclick={() => (isDefaultModelSelectOpen = !isDefaultModelSelectOpen)}
+				>
+					<option class="dark:bg-surface3" value="" disabled
+						>Select the default model for {project.name} project</option
```
cjellick: we do not use the term project. can you change this text to Select the default model for this agent

cjellick (approved):

PR #2941, opened 2025-05-14
enhance: enable configuring shared credentials for chatbots

njhale:
Enhance the sharing UI for chatbots to allow authors to configure shared credentials for MCP servers and tools.


ui/user/src/lib/components/edit/ChatbotConfig.svelte
```diff
@@ -0,0 +1,653 @@
...
+
+			// If we get a non-empty result, there's a shared configuration
+			const hasSharedConfig = result && Object.keys(result).length > 0;
+			sharedConfigs = { ...sharedConfigs, [mcpServerId]: hasSharedConfig };
+
+			// If there's a shared config, set the dropdown selection to "Configure it for everyone"
+			if (hasSharedConfig) {
+				selectedMcpConfig = {
+					...selectedMcpConfig,
+					[mcpServerId]: 'Configure it for everyone'
```
ivyjeong13: nit: `Configure it for everyone` could be a `const`! In case you needed to change the verbiage, then you'd just need to change it in one place

ui/user/src/lib/components/edit/ChatbotConfig.svelte
```diff
@@ -0,0 +1,653 @@
...
+			sharedConfigs = { ...sharedConfigs, [mcpServerId]: false };
+			return null;
+		}
+	}
+
+	async function checkToolCredentials(server: ProjectMCP) {
+		if (!server.catalogID) return;
+
+		try {
+			console.log(`Checking credentials for tool: ${server.name} (${server.catalogID})`);
```
ivyjeong13: Can the console.logs be dropped?

njhale: sure thing! sorry about that, thought I already dropped them

ui/user/src/lib/components/edit/ChatbotConfig.svelte
```diff
@@ -0,0 +1,653 @@
...
+		currentMcpServer = server;
+
+		const bundleId =
+			server.catalogID && toolBundleMap.get(server.catalogID) ? server.catalogID : null;
+		const isAuthTool = bundleId && isToolAuthRequired(bundleId);
+
+		if (isAuthTool) {
+			handleToolAuth(server, bundleId);
+		} else {
+			mcpConfig = {
```
ivyjeong13: Could probably call mcpConfig = initConfigfromManifest(server), it looks like this is doing the same thing here

ui/user/src/lib/components/edit/ChatbotConfig.svelte
```diff
@@ -0,0 +1,653 @@
...
+				icon: server.icon,
+				exists: false
+			};
+		}
+
+		currentCredential = credential;
+		showAuthDialog = true;
+	}
+
+	function isToolAuthRequired(bundleId: string) {
```
ivyjeong13: McpInfoConfig also has a similar check, might be nice to move this as a util function they both can import to avoid having to update multiple places

njhale: yeah, as I was writing the consumption side of this I noticed that too. I'll do that.

cjellick (approved):

Issue #2917, opened 2025-05-12
Tasks - Step details and execution results  for the older runs are shown incorrectly when changes are made to the steps


sangee2004:
**Describe the bug**
Tasks - Step details and execution results  for the older runs are not preserved when changes are made to the steps

**To Reproduce**
Scenario 1:
1. Create an agent.
2. Add a task with step say `Gardening ideas`
3. Execute the task 
4. Edit the existing task step to `Chess rules`
5. Execute the task
6. Go to the task run from step3.

Notice that the step name will still be - `Chess rules` with response from `Gardening ideas` presented.
Once you click in the task run again , you will see the correct step name `Gardening ideas` presented.

https://www.loom.com/share/0017c6066e4d4862b2a03c3a631314d4?sid=10d117b8-1f7a-4db0-9eb9-cc8ef13c24f7

Scenario 2:

1. Create an agent.
2. Add a task with  2 steps  say `chess rules` and `top 10 chess players`
3. Execute the task 
4. Edit the existing task step and delete `top 10 chess players` step.
6. Go to the task run from step3.
Notice that the step will not show execution details of `top 10 chess players` even though this step was executed.

Once you click in the task run again , you will see the `top 10 chess players` step show up

https://www.loom.com/share/f2111a638cad4993a5ea81e8fccf4393?sid=8e1db9bb-b995-480e-87db-cea83b282ad1

**Expected Behavior:**
Step details and execution results  for the older runs should be shown correctly when accessing the task runs.

cjellick:
@ryu-man - i think somewhere in the run data you should have the steps as they were at the time of execution. if you need help @njhale or @g-linville might be able to help you find the right data

ryu-man:
Yeah, this will be solved by relying on run data to keep history. Currently steps are rendered based on the task loop data, which is mutable. I will see if can find something helpful

sangee2004:
This issue is addressed when testing with latest builds.

Step details and execution results for the older runs are shown correctly even after updates are made to the task steps.

PR #2954, opened 2025-05-14
fix: show steps and loop steps correctly in a task run

ryu-man:
Addresses #2917

- load task run data in read only mode and recalculate the steps and loop steps

### DIFF
--- a/ui/user/src/lib/components/tasks/Task.svelte
+++ b/ui/user/src/lib/components/tasks/Task.svelte
@@ -56,7 +56,7 @@
 	let taskRun: Task | undefined = $state(undefined);
 
 	$effect(() => {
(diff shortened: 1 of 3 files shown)


ui/user/src/lib/components/tasks/Task.svelte
```diff
@@ -52,6 +53,17 @@
 	let isTaskInfoVisible = $state(true);
 	let observer: IntersectionObserver;
 
+	let taskRun: Task | undefined = $state(undefined);
+
+	$effect(() => {
+		// Load task run data only in readonly mode; so we can get the correct steps & loop steps related to the current task
```
cjellick: forcing a dumb change to re-trigger build
```suggestion
		// Load task run data only in readonly mode; so that we can get the correct steps & loop steps related to the current task
```

PR #2991, opened 2025-05-15
chore: filter out SSE-based MCP servers with localhost URLs

thedadams:


pkg/controller/handlers/toolreference/toolreference.go
```diff
@@ -354,6 +355,9 @@ func (h *Handler) readMCPCatalog(catalog string) ([]client.Object, error) {
 					},
 				}
 			} else if c.URL != "" {
+				if u, err := url.Parse(c.URL); err != nil || u.Hostname() == "localhost" || u.Hostname() == "127.0.0.1" {
```
g-linville, line 358: I think we need to be more thorough than this. Block `::1` as well, at a minimum. ChatGPT recommended to me that we do something like this:

```go
ips, err := net.LookupIP(host)
// handle err
for _, ip := range ips {
    if ip.IsLoopback() {
        return errors.New("request to loopback address is blocked")
    }
}
```

It also mentioned trying this, to catch things like `127.1`:
```go
if ip := net.ParseIP(host); ip != nil {
    if ip.IsLoopback() {
        return errors.New("loopback IP blocked")
    }
}
```

I think it's fine to merge this for now, because it's better than nothing, but we should come back and do more here.

g-linville (approved):

pkg/controller/handlers/toolreference/toolreference.go
thedadams: Talked "offline" about this concern. It is valid, but not what this change is trying to accomplish. It should be addressed in a different PR.

Issue #2571, opened 2025-04-16
Ability to re-arrange steps in a workflow via drag and drop


cjellick:
As the title says, user should be able to rearrange the steps in a workflow via dragging them around

ryu-man:
@cjellick I guess this applies to loop steps also!

sangee2004:
@ryu-man I dont see the ability to rearrange steps within a loop

<img width="851" alt="Image" src="https://github.com/user-attachments/assets/acf198aa-c053-4abd-9994-c7e532848771" />

sangee2004:
Able to rearrange steps within as task as expected when testing with latest build.

We do not have the ability to rearrange steps within a loop which is tracked in https://github.com/obot-platform/obot/issues/3097

PR #3003, opened 2025-05-16
feat: implement steps re-ordering using drag & drop

ryu-man:
Addresses #2571


ui/user/src/lib/components/primitives/draggable/DraggableItem.svelte
```diff
@@ -0,0 +1,182 @@
...
+	import { getDraggableContext } from './contextRoot';
+	import { setDraggableItemContext } from './contextItem';
+
+	const draggableContext = getDraggableContext();
+
+	if (!draggableContext) {
+		throw new Error('Draggable context was not found');
+	}
+
+	let {
```
ivyjeong13, line 24: nit: Use `type Props` for consistency in defining component props. Drop defaults that aren't needed

ui/user/src/lib/components/primitives/draggable/DraggableList.svelte
```diff
@@ -0,0 +1,191 @@
...
+		as: string;
+		disabled?: boolean;
+		onChange?: (items: unknown[]) => void;
+		children?: Snippet<[]>;
+	};
+	let {
+		class: klass = '',
+		as = 'div',
+		order = [],
+		disabled = false,
```
ivyjeong13: Drop defaults that aren't needed (ex. onChange and children will be undefined already if not supplied)

ui/user/src/lib/components/primitives/draggable/DraggableItem.svelte
```diff
@@ -0,0 +1,182 @@
...
+		ev.preventDefault();
+
+		if (isActive) return;
+
+		isPointerEntered = false;
+	}
+</script>
+
+<svelte:element
+	this={as ?? 'div'}
```
ivyjeong13, line 149: nit: since `as` is defaulted to `div`, is this needed? Or drop the default in the props and use this instead?

ui/user/src/lib/components/tasks/Step.svelte
```diff
@@ -33,22 +34,27 @@
...
 		runID,
 		pending,
 		stepMessages,
 		project,
 		showOutput: shouldShowOutputGlobal,
 		readOnly,
 		lastStepId,
-		isTaskRunning = false
+		isTaskRunning = false,
+		onChange = undefined,
```
ivyjeong13: Drop unneeded undefined defaults

ui/user/src/lib/components/primitives/draggable/DraggableItem.svelte
ryu-man: to make sure we handle a scenario when user passed an undefined value, default prop droped

ui/user/src/lib/components/primitives/draggable/DraggableItem.svelte
ryu-man: done

ui/user/src/lib/components/primitives/draggable/DraggableList.svelte
ryu-man: done

ui/user/src/lib/components/tasks/Step.svelte
ryu-man: done

ivyjeong13 (approved):

Issue #2993, opened 2025-05-15
Slack Interface - Link to `Slack App page` in configuration text should open in a new browser tab.


sangee2004:
**Describe the bug**
Slack Interface - Link to `Slack App page` in configuration text should open in a new browser tab.

**To Reproduce**
1. Create an agent.
2. Configure Slack Interface 

Notice that when clicking the  link to `Slack App page` in  step 1 , it opens the slack app page in the current browser session loosing the current obot view.

https://www.loom.com/share/b5ad480186b34c89ae7b20f40f0d4117?sid=c2631406-5d9e-4905-a87b-b578902413b4

**Expected behavior**
 Link to `Slack App page` in configuration text should open in a new browser tab.

sangee2004:
This issue is fixed when testing with latest builds.

 Link to Slack App page in configuration text now opens up in a new browser tab.

PR #3017, opened 2025-05-16
Fix: Open slack app page in new tab

cjellick:


cjellick:
addresses https://github.com/obot-platform/obot/issues/2993

PR #3019, opened 2025-05-17
chore: add a second catalog file for Zapier and GitMCP

g-linville:


pkg/controller/handlers/toolreference/toolreference.go
```diff
@@ -467,6 +470,7 @@ type mcpServerConfig struct {
 	Args           []string          `json:"args,omitempty"`
 	HTTPHeaders    []types.MCPHeader `json:"httpHeaders,omitempty"`
 	URL            string            `json:"url,omitempty"`
+	Remote         bool              `json:"remote,omitempty"`
```
g-linville, line 473: This field was added to support remote MCP servers with no default URL (which we wanted for Zapier).

thedadams (approved):

Issue #2702, opened 2025-04-24
Allow users to select an exact hour for recurring Obot tasks


mighty-pup-acorn:
**Summary:**
Users are unable to select an exact hour for scheduling recurring Obot tasks. The current scheduling options only allow for broad time windows (e.g., 6am or 9am), which is frustrating for users who need more precise control (e.g., scheduling at exactly 7am).

**User Feedback:**
A user expressed annoyance at not being able to pick an exact hour for their recurring Obot task, specifically needing 7am but only being able to choose between 6am and 9am.

**Proposed Solution:**
Enhance the scheduling interface for recurring Obot tasks to allow users to select any hour of the day, rather than being limited to preset time windows. This would provide greater flexibility and improve user satisfaction.

**Benefits:**
- Increased user control and satisfaction
- More precise task scheduling
- Better alignment with user workflows

**Context:**
This request was raised in a Slack thread by a user who needs to run a recurring task at 7am, but is currently forced to choose between 6am and 9am.

---

*Thread summary:*
A user is frustrated by the lack of precise scheduling options for recurring Obot tasks, specifically wanting to schedule a task at 7am but only being able to choose between 6am and 9am. They requested that this limitation be addressed.

ryu-man:
@cjellick Is the run on schedule option removed from the UI?

ryu-man:
@cjellick I think this change should be applied to all intervals except 'weekly' interval; the best solution is to use a combobox so user can select pre-defined values from the menu and enter custom values as well

PR #3035, opened 2025-05-19
chore: allow users to type custom schedule values

ryu-man:
Addresses #2702

- create combobox component that allows users to both select pre-defined and type custom values

### DIFF
--- a/ui/user/src/lib/components/tasks/Schedule.svelte
+++ b/ui/user/src/lib/components/tasks/Schedule.svelte
@@ -4,6 +4,7 @@
 	import type { Schedule } from '$lib/services';
 	import Dropdown from '$lib/components/tasks/Dropdown.svelte';
 	import Combobox from './Combobox.svelte';
+	import AmPmSwith from './AmPmSwith.svelte';
 
 	interface Props {
 		schedule?: Schedule;
@@ -35,14 +36,52 @@
 	};
 
 	let dailyComboboxFocused = $state(false);
+	let dailyComboboxAmPm: 'am' | 'pm' = $state('am');
+	let dailyComboboxFocusTimeout: number | undefined = undefined;
+
+	function to12H(value: number) {
+		if (value === 0 || value > 23) {
+			return 0;
+		}
+
+		if (value === 12) {
+			return 12;
+		}
+
+		// value in 24H format
+		if (value > 12) {
+			return Math.min(11, Math.max(0, Math.trunc(value - 12)));
+		}
+
+		// value in 12H AM, just make sure it is within the boudaries
+		return Math.min(11, Math.max(0, value));
+	}
+
+	function from12H(value: number, ampm: 'am' | 'pm' = 'am') {
+		let v = value;
+		let factor = 0;
+
+		// if value is greateer than 11; subtract 12 hours to make it 12H format
+		if (value > 11) {
+			v -= 12;
+		}
+
+		// if pm is selected then add 12 hours to make it 24H format
+		if (ampm === 'pm') {
+			factor = 12;
+		}
+
+		// sum and check for boundaries
+		return Math.trunc(Math.min(23, Math.max(0, v + factor)));
+	}
 </script>
 
 <div class="flex h-12 items-center self-start">
 	<h4 class="text-base font-medium">Schedule</h4>
 </div>
-<div class="flex flex-col gap-4 md:flex-row">
+<div class="md:min-w-auto flex min-w-[220px] flex-col gap-4 lg:flex-row">
 	<Dropdown
-		class="schedule-dropdown w-[144px] md:w-[172px]"
+		class="schedule-dropdown w-full md:w-[172px]"
 		values={{
 			hourly: 'hourly',
 			daily: 'daily',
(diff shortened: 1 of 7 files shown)


ivyjeong13:
When I'm not a logged in user, the login dialog is broken for me. Clicking it or any of the cards on the unauthorized home page does nothing.

ryu-man:
I made some changes regarding `clickout`, that could be the issue; let me check

ivyjeong13:
> I made some changes regarding `clickout`, that could be the issue; let me check

No worries, I suspected that may be the case when I saw the clickOutside action for `<dialog>` was modified. I recommend quickly doing an eyeball check on related components when you modify something like this just in case!

ryu-man:
Yeah, I agree

ryu-man:
@ivyjeong13 issue fixed

ui/user/src/lib/components/tasks/Schedule.svelte
```diff
@@ -33,8 +36,9 @@
 	/>
 
 	{#if schedule?.interval === 'hourly'}
-		<Dropdown
-			class="schedule-dropdown"
+		<Combobox
```
ivyjeong13, line 100: Screenshot 2025-05-21 at 11 01 56 AM

Can it show the appropriate label (ex. Midnight or On the Hour when value = 0) on the Combobox if the user has selected one of our options?

Also for 'Daily' maybe some kind of helpful hint that the value they have to input is 24H integer? The dropdown shows am/pm so my first instinct would have been to write like 1AM if I hadn't read the code. 

nit: For the input, the value is squished right up against its container
Screenshot 2025-05-21 at 11 10 00 AM

ivyjeong13: My other thought for Daily is a time/clock picker so they can just pick the time they want. edit: Maybe a silly question, but what if they want to pick like Daily at 7:15AM? Prolly a Craig question

ui/user/src/lib/components/tasks/Schedule.svelte
```diff
@@ -109,22 +114,36 @@
...
-			selected={schedule?.day.toString()}
+			selected={(day >= 0 ? day + 1 : day).toString()}
 			onSelected={(value) => {
 				if (schedule) {
-					schedule.day = parseInt(value);
+					const valueAsNumber = parseInt(value);
+
+					if (valueAsNumber >= 0) {
+						// reset counting to 0-based
+						schedule.day = Math.min(31, Math.max(0, valueAsNumber - 1));
```
ivyjeong13, line 264: Wondering for months that do not have 31 days or February with 28 days, how does this work with scheduling? Will it run on the month's last day or get skipped? Should we default to, if they pick 29-31, it gets set as last day (-1)?

ui/user/src/lib/components/tasks/Schedule.svelte
ryu-man: > For the input, the value is squished right up against its container

Fixed

ryu-man: >The dropdown shows am/pm so my first instinct would have been to write like 1AM if I hadn't read the code.

maybe not let the user fully type '1 AM,' but add an AM/PM switch

ui/user/src/lib/components/tasks/Schedule.svelte
ryu-man: >Will it run on the month's last day or get skipped

they should run on the last day of the month, but it needs to be handled in the back side as well

>if they pick 29-31, it gets set as last day (-1)?

in my opinion I don't think the user will be comfortable with changing an entered value of (28, 29, 30, 31) automatically to -1, that would seems like unexpected behavior

ui/user/src/lib/components/tasks/Schedule.svelte
ivyjeong13: That works too, would make it more consistent with the existing options

ui/user/src/lib/components/tasks/Schedule.svelte
ryu-man: Probably the best approach is to keep the entered value by the user and show some hint that the schedule will always run on the last day of the month if this can be handled on the backside otherwise, we have to auto set it to -1

ui/user/src/lib/components/tasks/Schedule.svelte
ryu-man: I believe Yes. The user will see values from 1...12 along the switch and under the hood we convert the value back and forth between 24H and 12H to keep compatibility with the server

ryu-man: > Can it show the appropriate label (ex. Midnight or On the Hour when value = 0) on the Combobox if the user has selected one of our options?

Can be, by using an absolute element above input that show/hide on input blur/focus

ryu-man: @ivyjeong13 I have added an AM-PM switch for hourly interval values as well as a custom label that show simple view

ivyjeong13:
Gonna go ahead and close this out especially with the upcoming chat changes!

Issue #2915, opened 2025-05-12
Tasks - Deletion of loop steps results in step getting deleted without asking user for any confirmation.


sangee2004:
**Describe the bug**
Tasks - Deletion of loop steps results in step getting deleted without asking user for any confirmation.

**To Reproduce**
1. Create an agent and add a task.
2. Have some loop step with few steps created for the loop

<img width="912" alt="Image" src="https://github.com/user-attachments/assets/ff890379-0a80-4aa1-a01c-a630dbcebd1f" />

3. Attempt to delete one of the steps under the loop step ( like `Insert optimal price into cost column`)

Notice that loop step gets deleted with asking user for any confirmation.

**Expected behavior**
User should be asked for confirmation before deleting the loop step (similar to regular steps)

sangee2004:
Following confirmation is presented to the user when deleting a loop step:

<img width="912" alt="Image" src="https://github.com/user-attachments/assets/bb3f0340-1a5d-4c9f-8547-80989a903ce0" />

PR #3036, opened 2025-05-19
fix: confirm action before delete a loop step

ryu-man:
Address #2915 

- Show a confirmation dialog before deleting a loop step.

### DIFF
--- a/ui/user/src/lib/components/tasks/Step.svelte
+++ b/ui/user/src/lib/components/tasks/Step.svelte
@@ -56,6 +56,9 @@
 	let toDelete = $state<boolean>();
 	let isConvertToRegularDialogShown = $state(false);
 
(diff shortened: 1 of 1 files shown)


ui/user/src/lib/components/tasks/Step.svelte
```diff
@@ -56,6 +56,9 @@
 	let toDelete = $state<boolean>();
 	let isConvertToRegularDialogShown = $state(false);
 
+	let loopStepToDeleteIndex: number | undefined = $state(undefined);
```
ivyjeong13, line 59: You can just do `let loopStepToDeleteIndex: number = $state();` and typescript should infer it as number | undefined. If you use null though, you'll need to do this.

ivyjeong13 (approved):

Issue #3060, opened 2025-05-20
Replace "Verified" category with "Community"


cjellick:
We have trimmed down our MCP catalog to *only* include the ones we've marked verified, so contextually "Verified" doesnt make sense, but "Community" would.

sangee2004:
We are now presented with `Community` catalog which includes `Verified` community MCP catalog entries .
<img width="1498" alt="Image" src="https://github.com/user-attachments/assets/70329abc-afb6-400f-bd69-0d74d18169a4" />

Issue #3025, opened 2025-05-18
Should be able to add the same server twice from the catalog


cjellick:
best example is zapier - i should be able to add it more than once (bc i can supply different url each time).

The catalog should allow for this, but give a warning because lit isnt necessarily the normal flow and if the config was exactly the same, things could get weird in the llm

ivyjeong13:
Checks for the `allows-multiple` metadata property. Should just be Zapier atm

shweta8996:
I have verified on main and it is working fine as expected.
Same server is not able to add as twice from catalog.
Please refer attachment:

https://github.com/user-attachments/assets/a8c4bb07-9c7e-4d79-9f34-6ae52f579828

PR #3062, opened 2025-05-20
fix: can setup mcp server more than once & community tag

ivyjeong13:
* mcp server from catalog can be set up more than once
* change displaying 'Verified' as 'Community' instead
Addresses #3060 
Addresses #3025

### DIFF
--- a/ui/user/src/lib/components/agents/AgentWizard.svelte
+++ b/ui/user/src/lib/components/agents/AgentWizard.svelte
@@ -326,7 +326,7 @@
 							class="h-inherit default-scrollbar-thin flex w-full grow flex-col overflow-y-auto pr-4"
 						>
 							<!-- todo: update when agent wizard gets used-->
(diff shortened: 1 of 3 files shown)


cjellick:
for adding twice, i gave the following feedback in slack. copying here

i tested out the "allow multiple" bit and i think its going to be weird for users. the vast majority of the time, it's a bad idea bc of tool conflicts. in fact, right now i think the only time it makes sense is for Zapier.
Can we key off of a metadata field for allowing multiple?
```
metadata: {
   "categories": ...
   "allow-multiple": "true"
}
```
if an mcp server DOESNT have that, then you shoiuld grey it out like before



also in this pr
- Change "Available Parameters" to just "Parameters"
- Hide Parameters by default
- When Show All Descriptions is on, the up and down arrow doesnt do anything. maybe we can hide it or grey it out? ...or make it work, but i guess that would get funky with the top radio button

cjellick (approved):

Issue #2919, opened 2025-05-13
Agent level Model provider - Chatbots encounter "500 Internal Server Error" relating to credentials when chatting with agents that have  model providers configured.


sangee2004:
**Describe the bug**
Agent level Model provider - chatbots encounter "500 Internal Server Error" relating to credentials when chatting with agent level model provider is set.

**To Reproduce**
1. Create an agent
2. Configure agent level Model provider

<img width="373" alt="Image" src="https://github.com/user-attachments/assets/845a4806-19f6-470d-ad6e-bbc46e5743ba" />

and have default model provider set to `claude-3-7-sonnet` model

<img width="761" alt="Image" src="https://github.com/user-attachments/assets/e7bce8f0-1708-4b3b-b359-1b92ab4d2157" />

4. Enable chatbot.
5. As another regular user, use the chatbot link from step4 to launch chatbot.
6. Notice that the default model provider for the chat gets set to  `claude-3-7-sonnet` model as expected.
7. Try to chat with this chatbot

Notice that chat fails with following error

<img width="969" alt="Image" src="https://github.com/user-attachments/assets/f4ca96b2-780a-4dca-b670-ba563dbc8507" />

**Expected behavior**
Should be able to chat successfully with the default agent level Model provider.

njhale:
Adding @thedadams because he already had an (untested) fix ready for this.

PR #3064, opened 2025-05-20
fix: use top-level project when getting credentials for model provider

thedadams:
Addresses https://github.com/obot-platform/obot/issues/2919


njhale:
@thedadams tested this out with both the top-level agent and chatbots from a separate account and it seems to work fine. The chosen agent-level default is set on the chatbot and it doesn't error out when chatting with it.

njhale (approved):

Issue #2973, opened 2025-05-15
long delay before anything happens when i click Run on a task


cjellick:
When you click run on a task, theres a 2 or 3 second delay before the task kicks off. I'd like some kind of visual indication that something is happening.

shweta8996:
I have verified main and it is working fine as expected.
When you click run on a task, theres a 2 or 3 second delay before the task kicks off. I'd like some kind of visual indication that something is happening. >Done
Please refer attachment:

https://github.com/user-attachments/assets/991eec9d-3b6d-4c1f-a033-3a831d1da0cb

PR #3066, opened 2025-05-20
fix: show running indicatorsd as soon as user clicks run

ryu-man:
Addresses #2973

### DIFF
--- a/ui/user/src/lib/components/tasks/Task.svelte
+++ b/ui/user/src/lib/components/tasks/Task.svelte
@@ -241,10 +241,17 @@
 			return;
 		}
 
-		await run();
+		try {
+			// make sure this runs after run() function and before promise is resolved
+			setTimeout(() => {
+				isRunning = true;
+				shouldFollowTaskRun = true;
+			}, 0);
 
-		isRunning = true;
-		shouldFollowTaskRun = true;
+			await run();
+		} catch (_) {
+			isRunning = false;
+		}
 	}
 
 	async function run(step?: TaskStep) {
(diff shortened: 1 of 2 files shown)


ui/user/src/lib/components/tasks/Task.svelte
```diff
@@ -233,10 +233,17 @@
...
 		}
 
-		await run();
-
-		isRunning = true;
-		shouldFollowTaskRun = true;
+		try {
+			// make sure this runs after run() function and before promise is resolved
+			setTimeout(() => {
+				isRunning = true;
```
ivyjeong13, line 247: Curious if you ran into issues just having the checks before `run()` like:
```
try {
    isRunning = true;
    shouldFollowTaskRun = true;
    await run();
// ...
```

ryu-man: yeah, so inside `run()` it will check if isRunning === true, then break

ivyjeong13: And you can't set isRunning to true in `run` after making the `isRunning` check?

ryu-man: I tried, but inside `run` there are multiple conditions that checks for `isRunning`, so I have to do it multiple times, but with timeout I can do it only once; It is not the best approach but it helps avoid refactoring

ivyjeong13: Hmm I see

PR #3098, opened 2025-05-21
enhance: mcp servers tool bad request or long load exit option

ivyjeong13:
Related Loom: https://www.loom.com/share/aa2a78ddd07e4d44bf98018df6d79ee7

* shows bad request error when trying to view tools and necessary configurations haven't been made
* adds an exit option for long hanging tool calls when setting an mcp server up


ui/user/src/lib/components/PageLoading.svelte
```diff
@@ -1,13 +1,41 @@
...
 		show: boolean;
 		text?: string;
+		longLoadContent?: Snippet;
+		longLoadDuration?: number;
 	}
 
-	let { show, text }: Props = $props();
+	let { show, text, longLoadContent, longLoadDuration = 30000 }: Props = $props();
+	let isLongLoad = $state(false);
+	let timeout = $state<ReturnType<typeof setTimeout>>();
```
ryu-man, line 16: just for simplicity, I think `number` can be used here over `ReturnType<typeof setTimeout>`

ryu-man (approved):

thedadams (approved):

PR #3099, opened 2025-05-21
chore: store the verified MCP servers in their own catalog folder

g-linville:
I refactored the way that we parse catalog files in Obot so that we can read them from a directory, and we no longer require the manifest and metadata to be nested JSON strings.


pkg/controller/handlers/toolreference/toolreference.go
```diff
@@ -447,27 +479,20 @@ func isCommandPreferred(existing, newer string) bool {
...
-	License         string `json:"license"`
-	Icon            string `json:"icon"`
-	Manifest        string `json:"manifest"`
-	ToolDefinitions string `json:"toolDefinitions"`
-}
-
-type mcpServerManifest struct {
-	Name        string            `json:"name"`
-	Description string            `json:"description"`
-	Category    string            `json:"category"`
```
g-linville, line 469: The `Name`, `Description`, and `Category` fields were completely unused, so I removed this type and we just use `mcpServerConfig` directly as the manifest type.

remote_catalog.json
g-linville, line 1: I moved the two manifests in here to their own files in the `catalog` directory.

Issue #3096, opened 2025-05-21
Tasks - When trying to type some text in the middle of an existing text , user is taken back to the end of the text.


sangee2004:
**Describe the bug**
Tasks - When trying to type some text in the middle of an existing text , user is taken back to the end of the text.

**To Reproduce**
1. Create an agent
2. Create task with some steps

Try to type some text in the middle of an existing text .
Notice that user is taken back to the end of the text after the character is entered.

https://www.loom.com/share/fb147713e3314f85a0f54954be49443c?sid=b6991ac0-0391-49a0-80e3-12f3fb9aa1b7

Same behavior is also seen when user tries to delete something from the middle of a step string.
**Expected behavior**
User should continue to present where he started to type.

sangee2004:
This issue is addressed when testing with latest builds.

When trying to type some text in the middle of an existing text , control stays were it is suppose to be. Able to type texts in the middle of existing step strings as expected.

PR #3103, opened 2025-05-21
fix: use input event over value binding

ryu-man:
Addresses #3096


thedadams:
@ryu-man It seems that the issue you linked is not addressed by this change. Did you link the wrong issue?

ryu-man:
@thedadams you are right, wrong issue. I updated the description

thedadams:
This seems to address the loop field (the bottom field in the screenshot), but the top field in the screenshot still has the issue.

Screenshot 2025-05-21 at 19 43 24

ryu-man:
@thedadams should be good now

thedadams (approved):

Issue #3058, opened 2025-05-20
For legacy tool mcp servers, filter them if the default base agent does not include them


ivyjeong13:
For MCP Catalog, if the default agent doesn't contain the tool with the matching legacy mcp server, filter out the mcp server from showing.

ivyjeong13:
> I have verified on Obot main it is not working fine as expected, The MCP servers that should be filtered out (because their mapped tools are not enabled for the default agent) are still showing up in the MCP Catalog.
> 
>  Recording.122.mp4

Ah, MCP Catalog is only for bundle tools, not sub tools! Nothing should change in the Manage Tools section and I'm not sure if messing with the sub tools actually should do anything now. 

Tested on main:
https://www.loom.com/share/8f1f6e5ccb764cdd974a1a622d5d4564

shweta8996:
I have verified on Obot main it is not working fine as expected, The MCP servers that should be filtered out are still showing up in the MCP Catalog.

https://github.com/user-attachments/assets/60b05c12-84b6-434d-8eed-68df94497bf7

PR #3104, opened 2025-05-21
enhance: get tools on assistant & appropriately filter out legacy servers from mcp catalog

ivyjeong13:
Addresses #3058

### DIFF
--- a/apiclient/types/assitant.go
+++ b/apiclient/types/assitant.go
@@ -13,6 +13,9 @@ type Assistant struct {
 	MaxTools              int               `json:"maxTools,omitempty"`
 	WebsiteKnowledge      *WebsiteKnowledge `json:"websiteKnowledge,omitempty"`
 	AllowedModelProviders []string          `json:"allowedModelProviders,omitempty"`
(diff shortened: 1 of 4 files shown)


apiclient/types/assitant.go
```diff
@@ -13,6 +13,9 @@ type Assistant struct {
 	MaxTools              int               `json:"maxTools,omitempty"`
 	WebsiteKnowledge      *WebsiteKnowledge `json:"websiteKnowledge,omitempty"`
 	AllowedModelProviders []string          `json:"allowedModelProviders,omitempty"`
+	AvailableThreadTools  []string          `json:"availableThreadTools,omitempty"`
```
ivyjeong13, line 16: @thedadams Wanted to check if it's alright to add these properties to be viewable when getting them as Assistants!

thedadams: Yep. I think it's fine.

While you're here, do you mind changing the name of this file? It took me a while to find it because it is misspelled.

ivyjeong13: File name has been updated! : )

thedadams (approved):

njhale (approved):

Issue #3107, opened 2025-05-21
External Interface - Email triggers are not working as expected.


sangee2004:
**Describe the bug**
External Interface - Email triggers are not working as expected.

**To Reproduce**
1. Create an agent
2. Configure Email External interface (With default value "*" for `Allowed Sender`)
3. This results in `Email Integration Task` being created as expected which has an email address .
4. Send email address to the email address from step 3.

This does not result in the task getting executed.

@thedadams mentioned that there could be a bug around this:

```
level=info msg="Skipping mail for email-integration-task-6zln45hc6xzc@triggers.obot.ai: no receiver found" logger=/app/pkg/emailtrigger/emailtrigger.go
```
**Expected behavior**
Sending email to the email address specified in the task should result in `Email Integration Task` to be triggered.

sangee2004:
This issue is addressed when testing with latest build.

Emails sent to the email address assigned to `Email Integration Task` results in task being triggered as expected.

PR #3109, opened 2025-05-22
fix: set an alias on the email receiver when creating

thedadams:
The alias is in the email address and used to identify which receiver to use to execute the workflow.

Issue: https://github.com/obot-platform/obot/issues/3107


pkg/controller/handlers/task/task.go
```diff
@@ -215,6 +203,18 @@ func (t *Handler) UpdateTaskForEmail(thread *v1.Thread, req router.Request) erro
 			}
 			thread.Status.WorkflowNamesFromIntegration.EmailWorkflowName = workflow.Name
 		}
+
+		emailReceiver.Spec.Alias = workflow.Spec.Manifest.Alias
```
thedadams, line 207: This is the real fix: ensure that that workflow was created and set the alias here instead of above (where the alias on the workflow is actually blank).

Issue #3050, opened 2025-05-19
Modify Thread Tools - Provide a way to close the page without having to use "Update" option.


sangee2004:
**Describe the bug**
Modify Thread Tools - Provide a way to close the page without having to use "Update" option.

**To Reproduce**
1. Create an agent.
2. Add a MCP server
3. On the chat view , click on `Tools` to launch `Modify Thread Tools` page.

Notice that there is no way to close with page. Only way is to use the `Update` option.

<img width="1266" alt="Image" src="https://github.com/user-attachments/assets/5e881264-264f-4ee3-910f-40509e99bda3" />

**Expected behavior**
Provide a way to close the page without having to use "Update" option.

cjellick:
all you have to do is click outside the modal right?

shweta8996:
I have verified on main and it is not working fine as expected.
There is no way to close with page. Only way is to use the Update option.
Please refer attachment:

https://github.com/user-attachments/assets/7aebb064-dd56-4846-bf51-e406eddb88bd

sangee2004:
We are able to click outside the modal to dismiss this page. But this modal pretty much occupies the entire screen and there is very little space to click outside to get the modal disnissed.

shweta8996:
I have verified this on main, and it is not working as expected. There is still no way to close the "Modify Thread Tools" page without using the Update option.
Please refer attachment:

![Image](https://github.com/user-attachments/assets/9678deab-96c5-43f6-a7bd-98b738a1b0a7)

ivyjeong13:
Weird, I'm seeing it on main!  
<img width="1517" alt="Image" src="https://github.com/user-attachments/assets/c7fe5b24-ebb8-4c46-b642-2ad698c52d1f" />
<img width="1461" alt="Image" src="https://github.com/user-attachments/assets/189d98dc-6e43-4b27-9b92-d82e60c1387a" />

ivyjeong13:
Tried a few more times and still seeing it! Moving to test for verification again, but lemme know if you don't see it. Can huddle and debug.

shweta8996:
I have verified on Obot main and it is working fine as expected.
![Image](https://github.com/user-attachments/assets/f31449c7-675a-4080-845c-d8bd4c152b1f)

PR #3110, opened 2025-05-22
fix: another way to close thread tools dialog

ivyjeong13:
Addresses #3050


ryu-man (commented):
after adding an mcp server, the configuration dialog pops up but it does not contains the (x) close button, please skip this if it is intended 

image

ryu-man (approved):

PR #3122, opened 2025-05-22
enhance: improve agent unconfigured mcp warning ux

njhale:
- **fix: add missing script tag in user ui**
- **fix: ensure chatbot user mcp config is limited to existing envs and headers**
- **enhance: implement unconfigured mcp ux for top-level agents**

### DIFF
--- a/ui/user/src/lib/components/edit/McpServers.svelte
+++ b/ui/user/src/lib/components/edit/McpServers.svelte
@@ -19,6 +19,7 @@
 	import McpSetupWizard from '$lib/components/mcp/McpSetupWizard.svelte';
 	import { getToolBundleMap } from '$lib/context/toolReferences.svelte';
 	import { DEFAULT_CUSTOM_SERVER_NAME } from '$lib/constants';
+	import { errors } from '$lib/stores';
 
 	interface Props {
 		project: Project;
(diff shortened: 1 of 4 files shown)


thedadams (approved):
Seems fine to me, but would appreciate a review from @ivyjeong13

ui/user/src/lib/components/edit/McpServers.svelte
```diff
@@ -83,36 +83,68 @@
...
+	async function hasLocalConfig(mcp: ProjectMCP): Promise<boolean> {
+		// Handle legacy tool bundles
+		if (mcp.catalogID && toolBundleMap.get(mcp.catalogID)) {
+			return localCredentials.some((cred) => cred.toolID === mcp.catalogID && cred.exists === true);
+		}
+
+		// Real MCP server, reveal any configured env headers
+		let envHeaders: Record<string, string> = {};
+		try {
+			envHeaders = await ChatService.revealProjectMCPEnvHeaders(
```
ivyjeong13, line 108: Wondering if it's a pain to make it so a user with chatbot gets "their" ProjectMCP mcps when using chatbot and `configured` matches against them or something if that makes sense lol But get why this needs to be called for each mcp!

ivyjeong13 (approved):

ui/user/src/lib/components/edit/McpServers.svelte
njhale: > Wondering if it's a pain to make it so a user with chatbot gets "their" ProjectMCP mcps when using chatbot and configured matches against them

If I understand you correctly, they already kinda do, but only for "real" mcpservers. For bundle tools, it always shows as configured. I _should_ probably just fix that in the backend instead of doing this here, but I'd like to follow up with that change after the summit.

Issue #3135, opened 2025-05-23
Model provider card has overflow issue when user click on reconfigure


ryu-man:
**Describe the bug**
When user open the model provider's page on large screen and click on reconfigure, the content of the card will goes over the flow

**To Reproduce**
1. Go to models providers page
2. Configure a model provider 
3. Click on the reconfigure button
4. See issue

sangee2004:
Reproduced the issue when testing with commit id - `v0.0.0-dev+a1bb89c9`
<img width="534" alt="Image" src="https://github.com/user-attachments/assets/0484ced0-703f-4447-95f1-809cb6d78401" />

<img width="976" alt="Image" src="https://github.com/user-attachments/assets/cd0c97e4-9a6d-4065-9a9c-4ede3ef11aa9" />

PR #3136, opened 2025-05-23
fix: model provider overflow issue on reconfigure

ryu-man:
Addresses #3135

### DIFF
--- a/ui/user/src/lib/components/ModelProviderCard.svelte
+++ b/ui/user/src/lib/components/ModelProviderCard.svelte
@@ -467,7 +467,7 @@
 						{/if}
 					</div>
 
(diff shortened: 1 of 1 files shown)


ui/user/src/lib/components/ModelProviderCard.svelte
```diff
@@ -467,7 +467,7 @@
 						{/if}
 					</div>
 
-					<div class="relative flex min-h-[180px] flex-1 flex-col">
+					<div class="md:mi relative flex min-h-[180px] flex-1 flex-col lg:min-h-[120px]">
```
ivyjeong13: `md:mi` Just checking if this is a misaddition? Don't recognize it as a tailwind class!

ryu-man: yep, just typo

PR #3146, opened 2025-05-27
chore: allow setting the model on a thread when creating it

g-linville:


pkg/api/handlers/projects.go
```diff
@@ -525,6 +527,31 @@ func (h *ProjectsHandler) CreateProjectThread(req api.Context) error {
...
+			return fmt.Errorf("failed to unmarshal request body: %w", err)
+		}
+
+		if bodyContents.Model != "" && bodyContents.ModelProvider != "" {
+			// Make sure that this model and model provider are valid on the project.
+			if slices.Contains(projectThread.Spec.Models[bodyContents.ModelProvider], bodyContents.Model) {
+				thread.Spec.Manifest.Model = bodyContents.Model
+				thread.Spec.Manifest.ModelProvider = bodyContents.ModelProvider
+			}
+		}
```
g-linville: Should I return an error here if the model is not set in the body, or if the model is not configured on the project? Right now we just ignore and fallback to using the default, but idk what the right behavior here would be.

thedadams: I say return a `400` if the model is not configured on the project (which should also take care of the model not being set).

g-linville: fixed @thedadams

pkg/api/handlers/projects.go
```diff
@@ -525,6 +527,31 @@ func (h *ProjectsHandler) CreateProjectThread(req api.Context) error {
...
+			ModelProvider string `json:"modelProvider"`
+		}
+
+		if err := json.Unmarshal(body, &bodyContents); err != nil {
+			return fmt.Errorf("failed to unmarshal request body: %w", err)
+		}
+
+		// Make sure that this model and model provider are valid on the project.
+		if !slices.Contains(projectThread.Spec.Models[bodyContents.ModelProvider], bodyContents.Model) {
+			return types.NewErrBadRequest("model %s from model provider %s is not configured for this project", bodyContents.Model, bodyContents.ModelProvider)
```
thedadams: nit:
```suggestion
			return types.NewErrBadRequest("model %q from model provider %q is not configured for this project", bodyContents.Model, bodyContents.ModelProvider)
```

PR #3157, opened 2025-05-28
feat: add APIs for MCP server prompts and resources

thedadams:
This change also refactors and simplifies the tools endpoints to be consistent with the prompts and resources endpoints.


ui/user/src/lib/services/chat/operations.ts
thedadams, line 1: I don't know why my IDE alphabetized these, but it is esthetically pleasing.

pkg/mcp/types.go
```diff
@@ -39,7 +39,7 @@ func expandEnvVars(text string, credEnv map[string]string) string {
 	})
 }
 
-func ToServerConfig(mcpServer v1.MCPServer, projectThreadName string, credEnv map[string]string, allowedTools []string) (ServerConfig, []string) {
+func ToServerConfig(mcpServer v1.MCPServer, projectThreadName string, credEnv map[string]string, allowedTools ...string) (ServerConfig, []string) {
```
njhale, line 42: +1 for using a variadic parameter for backwards compatibility

njhale (approved):

pkg/mcp/types.go
g-linville: +1

pkg/api/handlers/mcp.go
```diff
@@ -261,7 +204,242 @@ func (m *MCPHandler) DeleteServer(req api.Context) error {
...
+	if caps.Prompts == nil {
+		return types.NewErrHTTP(http.StatusFailedDependency, "MCP server does not support prompts")
+	}
+
+	resources, err := m.mcpSessionManager.ListPrompts(req.Context(), serverConfig)
+	if err != nil {
+		return fmt.Errorf("failed to list resources: %w", err)
+	}
+
+	return req.Write(resources)
```
g-linville:
```suggestion
	prompts, err := m.mcpSessionManager.ListPrompts(req.Context(), serverConfig)
	if err != nil {
		return fmt.Errorf("failed to list prompts: %w", err)
	}

	return req.Write(prompts)
```

pkg/api/handlers/mcp.go
```diff
@@ -427,11 +605,11 @@ func (m *MCPHandler) UpdateServer(req api.Context) error {
 
 	existing.Spec.Manifest = updated
 
-	if err := req.Update(&existing); err != nil {
+	if err = req.Update(&existing); err != nil {
 		return err
 	}
 
-	return req.Write(convertMCPServer(existing, nil, cred.Env))
+	return req.Write(convertMCPServer(existing, cred.Env))
```
g-linville, line 612: I want to make sure here that we are intending to still use the original `cred.Env` instead of `sharedCred.Env`, as this is a slight behavioral change (since `cred` is no longer overwritten with the shared credential, and instead remains the original, non-shared credential), and I'm not sure what the significance of it would be.

g-linville (changes requested):

pkg/api/handlers/mcp.go
thedadams: Yep. This was technically a bug. The `sharedCred` should only be used for chatbots. Updating the server is an agent-based thing, which should use the `cred`.

pkg/api/handlers/mcp.go
thedadams: Done.

PR #3170, opened 2025-05-28
chore: improve dev script process group cleanup

njhale:
Be more precise targeting service process groups to clean up and reduce log noise when killing process groups.


drpebcak (approved):
This does work, but it exits with an error code

g-linville (approved):

PR #3178, opened 2025-06-09
Improve MCPServer API configuration status

njhale:


njhale:
This is no longer relevant since we're not supporting non-mcpserver based tools anymore.

PR #3179, opened 2025-06-09
feat: allow admin to configure remote MCP catalogs

g-linville:


pkg/api/handlers/catalogs.go
```diff
@@ -0,0 +1,102 @@
...
+
+	if err := req.Create(catalog); err != nil {
+		return fmt.Errorf("failed to create catalog: %w", err)
+	}
+
+	go func() {
+		if err := h.toolRefHandler.ForceRefreshMCPCatalogs(context.Background(), req.Storage); err != nil {
+			log.Errorf("Failed to force refresh MCP catalogs: %v", err)
+		}
+	}()
```
g-linville: This was the best I could come up with to immediately refresh the catalogs when we create a new one. The way that we read the catalogs right now, we basically ditch everything we have locally, load all the catalogs, and then save that. I felt it would be overkill to refactor this to load individual catalogs at this time (as this would, among other things, involve tying specific `MCPServerCatalogEntries` to their particular `Catalog`), but we could revisit that in the future if we think it is worthwhile.

pkg/api/handlers/catalogs.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+func NewCatalogHandler(toolRefHandler *toolreference.Handler) *CatalogHandler {
+	return &CatalogHandler{
+		toolRefHandler: toolRefHandler,
+	}
+}
+
+// List returns all catalogs.
+func (h *CatalogHandler) List(req api.Context) error {
```
njhale: nit: quite a few of these methods don't need a receiver variable

```suggestion
func (*CatalogHandler) List(req api.Context) error {
```

pkg/api/handlers/catalogs.go
njhale: My only question here is whether force refresh should get triggered by a controller watching `v1.Catalog` instead of through a go-routine kicked off by the API.

pkg/api/handlers/catalogs.go
g-linville: Thanks. I'm gonna blame Claude for this one lol.

pkg/api/router/router.go
```diff
@@ -390,6 +395,13 @@ func Router(services *services.Services) (http.Handler, error) {
 	mux.HandleFunc("GET /api/mcp/catalog", mcp.ListCatalog)
 	mux.HandleFunc("GET /api/mcp/catalog/{mcp_server_id}", mcp.GetCatalogEntry)
 
+	// MCP Remote Catalogs
+	mux.HandleFunc("GET /api/catalogs", catalogs.List)
+	mux.HandleFunc("GET /api/catalogs/{catalog_id}", catalogs.Get)
+	mux.HandleFunc("POST /api/catalogs", catalogs.Create)
+	mux.HandleFunc("DELETE /api/catalogs/{catalog_id}", catalogs.Delete)
+	// mux.HandleFunc("POST /api/mcp/remote-catalogs/force-refresh", toolRefHandler.ForceRefreshMCPCatalogs)
```
njhale: Did you mean to remove this entirely?

pkg/api/handlers/catalogs.go
g-linville: Fixed

pkg/api/handlers/catalogs.go
g-linville: Yeah I am not sure either. I'm curious to hear what @thedadams thinks.

pkg/api/router/router.go
g-linville: Yes, thanks, fixed

pkg/storage/apis/obot.obot.ai/v1/catalog.go
```diff
@@ -0,0 +1,62 @@
...
+	Spec   CatalogSpec   `json:"spec,omitempty"`
+	Status CatalogStatus `json:"status,omitempty"`
+}
+
+type CatalogSpec struct {
+	URL string `json:"url,omitempty"`
+}
+
+type CatalogStatus struct {
+	LastSyncTime metav1.Time `json:"lastSynced,omitempty"`
```
StrongMonkey:
```suggestion
	LastSyncTime metav1.Time `json:"lastSyncTime,omitempty"`
```

pkg/storage/apis/obot.obot.ai/v1/catalog.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (in *Catalog) NamespaceScoped() bool {
+	return false
+}
+
+func (in *Catalog) GetColumns() [][]string {
+	return [][]string{
+		{"Name", "Name"},
+		{"URL", "Spec.URL"},
+		{"Last Synced", "{{ago .Status.LastSynced}}"},
```
StrongMonkey:
```suggestion
		{"Last Synced", "{{ago .Status.LastSyncTime}}"},
```

g-linville: Great catch. I think I changed the name of that field but forgot to update it here and in that JSON tag. I committed your suggestions.

StrongMonkey (approved):

pkg/api/handlers/catalogs.go
```diff
@@ -0,0 +1,132 @@
...
+			URL: u.String(),
+		},
+	}
+
+	if err := req.Create(catalog); err != nil {
+		return fmt.Errorf("failed to create catalog: %w", err)
+	}
+
+	go func() {
+		if err := h.toolRefHandler.ForceRefreshMCPCatalogs(context.Background(), req.Storage); err != nil {
```
thedadams: Can we have a timeout here, so we don't have a bunch of stuck go routines?

pkg/api/handlers/catalogs.go
thedadams: I would say a controller would be good. At least because of the fact that the controller will use a cached client.

pkg/api/handlers/catalogs.go
```diff
@@ -0,0 +1,132 @@
...
+	if err := req.Get(&catalog, req.PathValue("catalog_id")); err != nil {
+		return fmt.Errorf("failed to get catalog: %w", err)
+	}
+
+	if err := req.Delete(&catalog); err != nil {
+		return fmt.Errorf("failed to delete catalog: %w", err)
+	}
+
+	go func() {
+		if err := h.toolRefHandler.ForceRefreshMCPCatalogs(context.Background(), req.Storage); err != nil {
```
thedadams: Same comment about timeout, irrelevant if we switch to a controller.

pkg/api/router/router.go
```diff
@@ -390,6 +395,12 @@ func Router(services *services.Services) (http.Handler, error) {
 	mux.HandleFunc("GET /api/mcp/catalog", mcp.ListCatalog)
 	mux.HandleFunc("GET /api/mcp/catalog/{mcp_server_id}", mcp.GetCatalogEntry)
 
+	// MCP Remote Catalogs
+	mux.HandleFunc("GET /api/catalogs", catalogs.List)
```
thedadams: nit:
```suggestion
	mux.HandleFunc("GET /api/mcp-catalogs", catalogs.List)
```

pkg/controller/handlers/toolreference/github.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+func readGitHubCatalog(catalogURL string) ([]catalogEntryInfo, error) {
+	// Normalize the URL to ensure HTTPS
+	if !strings.HasPrefix(catalogURL, "http://") && !strings.HasPrefix(catalogURL, "https://") {
+		catalogURL = "https://" + catalogURL
+	}
+
+	// Validate protocol is HTTPS
+	if strings.HasPrefix(catalogURL, "http://") {
```
thedadams, line 23: nit: move this above the check above. Then we don't have to check if the URL starts with `http://`.

pkg/controller/handlers/toolreference/github.go
```diff
@@ -0,0 +1,152 @@
...
+				}
+			}
+			if len(patterns) > 0 {
+				catalogPatterns = patterns
+			}
+		}
+	}
+
+	// Get repository file listing using GitHub API
+	apiURL := fmt.Sprintf("https://api.github.com/repos/%s/%s/git/trees/%s?recursive=1", org, repo, branch)
```
thedadams, line 90: Can we look for a GitHub API key env var, so we don't get a bunch of rate-limiting errors here?

pkg/controller/handlers/toolreference/toolreference.go
```diff
@@ -434,11 +445,16 @@ func (h *Handler) readMCPCatalogDirectory(catalog string) ([]catalogEntryInfo, e
...
 
-func (h *Handler) readFromMCPCatalogs(ctx context.Context, c client.Client) error {
+func (h *Handler) readFromMCPCatalogs(ctx context.Context, c client.Client, catalogs []v1.Catalog) error {
 	var (
 		toAdd []client.Object
 		errs  []error
 	)
+
+	h.catalogLock.Lock()
+	defer h.catalogLock.Unlock()
```
thedadams: We shouldn't hold this lock during IO.
```suggestion
	h.catalogLock.Lock()
	mcpCatalogs := h.mcpCatalogs
	h.catalogLock.Unlock()
```

pkg/controller/handlers/toolreference/toolreference.go
```diff
@@ -910,3 +952,18 @@ func (h *Handler) CleanupModelProvider(req router.Request, _ router.Response) er
...
 }
+
+func (h *Handler) ForceRefreshMCPCatalogs(ctx context.Context, c client.Client) error {
+	var catalogList v1.CatalogList
+	if err := c.List(ctx, &catalogList); err != nil {
+		return fmt.Errorf("failed to list catalogs: %w", err)
+	}
+
+	go func() {
+		if err := h.readFromMCPCatalogs(context.Background(), c, catalogList.Items); err != nil {
```
thedadams: Same comment about timeout, irrelevant if we switch to a controller.

pkg/storage/apis/obot.obot.ai/v1/catalog.go
```diff
@@ -0,0 +1,62 @@
...
+type CatalogSpec struct {
+	URL string `json:"url,omitempty"`
+}
+
+type CatalogStatus struct {
+	LastSyncTime metav1.Time `json:"lastSyncTime,omitempty"`
+}
+
+func (in *Catalog) NamespaceScoped() bool {
+	return false
```
thedadams, line 28: If I understand correctly, this would be the only object we have that isn't namespace scoped. Is this necessary?

pkg/api/handlers/catalogs.go
g-linville: Gonna change this to use a controller instead

pkg/controller/handlers/toolreference/toolreference.go
g-linville: I think we still need to do it the way that I have it implemented here because of the `apply` at the end. We refresh the catalog by pulling everything and then overwriting everything that is saved. Implementing it the way you suggested could lead to a race condition like this:

1. Admin adds a new catalog and kicks off a scrape
2. Admin immediately deletes that catalog and kicks off another scrape
3. The second scrape job finishes before the first scrape job, so the MCP servers from the newly added and then deleted catalog remain present in Obot, even though that catalog is now gone

pkg/storage/apis/obot.obot.ai/v1/catalog.go
g-linville: Not really. I'll fix it

g-linville: Oh, Aliases are not namespace scoped. So is it okay if we do this? There's never a reason that we would need these to be namespaced.

pkg/api/handlers/catalogs.go
g-linville: Refactored into a kick pattern with a channel

pkg/api/router/router.go
g-linville: Fixed

pkg/controller/handlers/toolreference/github.go
g-linville: Fixed

pkg/controller/handlers/toolreference/github.go
g-linville: Fixed

pkg/controller/handlers/toolreference/toolreference.go
g-linville: Refactored, no longer needed a mutex.

g-linville:
Re-requested review from everyone because I refactored the way we kick off a catalog refresh. We don't set up goroutines anymore; instead we trigger it by sending on a channel.

pkg/api/handlers/catalogs.go
```diff
@@ -0,0 +1,129 @@
...
+
+	if err := req.Create(catalog); err != nil {
+		return fmt.Errorf("failed to create catalog: %w", err)
+	}
+
+	// Kick off a catalog refresh.
+	select {
+	case h.catalogRefreshKick <- struct{}{}:
+	default:
+	}
```
njhale: nit: should this also have a `case <-ctx.Context()`?

g-linville: Probably unnecessary. This `select` is non-blocking. It sends on the channel if it's available, and if it's not, the default case triggers and nothing happens, and then we return.

pkg/controller/controller.go
```diff
@@ -32,7 +32,7 @@ func New(services *services.Services) (*Controller, error) {
 		return nil, err
 	}
 
-	services.Router.PosStart(c.PostStart)
+	services.Router.PosStart(c.PostStart(services.MCPCatalogRefreshKick))
```
thedadams: nit: I probably would have made the `services.MCPCatalogRefreshKick` a field on the `Controller` type, but, since this is only used in the `PostStart` (which is only called once), I think this is fine.

pkg/services/config.go
```diff
@@ -542,6 +544,7 @@ func New(ctx context.Context, config Config) (*Services, error) {
 		RetentionPolicy:            retentionPolicy,
 		MCPCatalog:                 config.MCPCatalogs,
 		AllowedMCPDockerImageRepos: config.AllowedMCPDockerImageRepos,
+		MCPCatalogRefreshKick:      make(chan struct{}),
```
thedadams: nit: can we make this a function instead that does the right thing? That way callers can just call the function instead of
```go
select {
case services.MCPCatalogRefreshKick <- struct{}{}:
default:
}
```

g-linville: Good call

g-linville: I implemented this, but I wasn't sure how to do it without having both a channel and a function as fields on the `Services` object. Let me know what you think. The change is pushed.

thedadams (approved):

g-linville:
Putting this into draft because we are going to make some changes related to the structure of the catalog.

g-linville:
Closing this PR because the requirements have changed enough to warrant starting again on a new branch. Though I will likely reuse the GitHub-related code from this.

PR #3181, opened 2025-06-11
Enhance: support slack socket mode

StrongMonkey:
Support socket mode in slack to subscribe slack events via websocket. User will need to configure an extra app-level token when enabling socket mode in slack and supply it into obot.

### DIFF
--- a/pkg/controller/handlers/slackreceiver/slackreceiver.go
+++ b/pkg/controller/handlers/slackreceiver/slackreceiver.go
@@ -75,7 +75,7 @@ func CreateOAuthApp(req router.Request, _ router.Response) error {
 	return nil
 }
 
-func (s *Handler) SubscribeToSlackEvents(req router.Request, resp router.Response) error {
+func (s *Handler) SubscribeToSlackEvents(req router.Request, _ router.Response) error {
 	slackReceiver := req.Object.(*v1.SlackReceiver)
 
 	s.lock.RLock()
@@ -126,11 +126,11 @@ func (s *Handler) SubscribeToSlackEvents(req router.Request, resp router.Respons
 			}
 		}
 		err := socketmodeHandler.RunEventLoopContext(ctx)
-		if err != nil {
-			log.Errorf("error running event loop: %v", err)
-		}
 		s.lock.Lock()
-		delete(s.subscribed, slackReceiver.Name)
+		if cancel, ok := s.subscribed[slackReceiver.Name]; ok {
+			cancel()
+			delete(s.subscribed, slackReceiver.Name)
+		}
 		s.lock.Unlock()
 		if err != nil {
 			err = fmt.Errorf("failed to run event loop: %w", err)


ui/user/src/lib/components/integrations/slack/Slack.svelte
```diff
@@ -427,28 +429,51 @@
...
 		<div>
-			<h4 class="font-medium">Step 3: Enable Events</h4>
-			<ul class="my-2 list-disc space-y-3 pl-4 text-sm font-light text-gray-500">
+			<h4 class="font-medium">Step 4: Enable Events</h4>
+			<p class="text-sm text-gray-500">
+				You can subscribe to Slack events using one of two approaches:
+				<strong> Slack Events API</strong> or <strong>Socket Mode</strong>. Use the Events API if
+				you can expose a public endpoint to receive webhook events. If you're running in a secure or
+				internal environment where public URLs are not an option,
+				<strong>Socket Mode</strong> is recommended.
```
g-linville: nit

```suggestion
				Socket Mode is recommended.
```

ui/user/src/lib/components/integrations/slack/Slack.svelte
```diff
@@ -427,28 +429,51 @@
...
-					<div class="copy-pill">
-						<CopyButton text={'app_mention'} />
-						app_mention
-					</div>
-				</li>
-			</ul>
+			<p class="mt-4 text-sm text-gray-500">
+				Once configured, your app will receive real-time event payloads via the WebSocket connection
+				established using your App-Level Token.
+			</p>
```
g-linville: I think this needs to be moved up higher to where we talk about Socket Mode?

pkg/controller/handlers/slackreceiver/slackreceiver.go
```diff
@@ -47,3 +71,149 @@ func CreateOAuthApp(req router.Request, _ router.Response) error {
...
+
+	slackAppToken := cred.Env["APP_TOKEN"]
+
+	if slackAppToken == "" {
+		return nil
+	}
+
+	if s.subscribed[slackReceiver.Name] != nil {
+		return nil
+	}
```
g-linville: I recommend moving this up to the top of the function before we try to reveal the credential, as it could save us from some unnecessary `Reveal` calls.

pkg/controller/handlers/slackreceiver/slackreceiver.go
```diff
@@ -47,3 +71,149 @@ func CreateOAuthApp(req router.Request, _ router.Response) error {
...
+
+	slackClient := slack.New("", slack.OptionAppLevelToken(slackAppToken))
+
+	client := socketmode.New(
+		slackClient,
+	)
```
g-linville: nit

```suggestion
	client := socketmode.New(slackClient)
```

g-linville (changes requested):

pkg/controller/handlers/slackreceiver/slackreceiver.go
StrongMonkey: yeah, good point.

ui/user/src/lib/components/integrations/slack/Slack.svelte
StrongMonkey: yeah I am going to remove it. Think this paragraph is not needed.

pkg/controller/handlers/slackreceiver/slackreceiver.go
```diff
@@ -47,3 +71,146 @@ func CreateOAuthApp(req router.Request, _ router.Response) error {
...
+	socketmodeHandler := socketmode.NewSocketmodeHandler(client)
+
+	socketmodeHandler.Handle(socketmode.EventTypeConnecting, middlewareConnecting)
+	socketmodeHandler.Handle(socketmode.EventTypeConnectionError, middlewareConnectionError)
+	socketmodeHandler.Handle(socketmode.EventTypeConnected, middlewareConnected)
+
+	// Handle a specific event from EventsAPI
+	socketmodeHandler.HandleEvents(slackevents.AppMention, h.middlewareAppMentionEvent)
+
+	ctx, cancel := context.WithCancel(context.Background())
```
thedadams: You can use `req.Ctx` here and the context will be canceled when the controller shutsdown. This won't happen for `context.Background()`.

pkg/controller/handlers/slackreceiver/slackreceiver.go
```diff
@@ -47,3 +71,146 @@ func CreateOAuthApp(req router.Request, _ router.Response) error {
...
+	s.subscribed[slackReceiver.Name] = cancel
```
thedadams, line 118: This needs a lock to protect against concurrent reads and writes.

pkg/controller/handlers/slackreceiver/slackreceiver.go
```diff
@@ -47,3 +71,146 @@ func CreateOAuthApp(req router.Request, _ router.Response) error {
...
+
+	go func() {
+		if err := socketmodeHandler.RunEventLoopContext(ctx); err != nil {
+			fmt.Printf("error running event loop: %v", err)
+		}
+		delete(s.subscribed, slackReceiver.Name)
+		resp.RetryAfter(time.Second * 10)
```
thedadams: I don't think this will work after all the controllers return. There isn't anything looking at this anymore.

pkg/controller/handlers/slackreceiver/slackreceiver.go
```diff
@@ -47,3 +71,146 @@ func CreateOAuthApp(req router.Request, _ router.Response) error {
...
+	}()
+
+	return nil
+}
+
+func (s *Handler) UnsubscribeFromSlackEvents(req router.Request, _ router.Response) error {
+	slackReceiver := req.Object.(*v1.SlackReceiver)
+	s.subscribed[slackReceiver.Name]()
+
+	delete(s.subscribed, slackReceiver.Name)
```
thedadams: Same comment here about the lock.

pkg/controller/handlers/slackreceiver/slackreceiver.go
```diff
@@ -47,3 +71,146 @@ func CreateOAuthApp(req router.Request, _ router.Response) error {
...
+	slackClient.Ack(*evt.Request)
+
+	ev, ok := eventsAPIEvent.InnerEvent.Data.(*slackevents.AppMentionEvent)
+	if !ok {
+		return
+	}
+
+	if ev.BotID != "" {
+		return
+	}
```
thedadams: nit:
```suggestion
	if !ok || ev.BotID != "" {
		return
	}
```

pkg/controller/handlers/slackreceiver/slackreceiver.go
```diff
@@ -47,3 +71,146 @@ func CreateOAuthApp(req router.Request, _ router.Response) error {
...
+
+	var slackTriggers v1.SlackTriggerList
+	if err := h.storage.List(h.ctx, &slackTriggers, client.MatchingFields{
+		"spec.appID":  eventsAPIEvent.APIAppID,
+		"spec.teamID": eventsAPIEvent.TeamID,
+	}); err != nil {
+		fmt.Printf("failed to list slack triggers: %v", err)
```
thedadams: Instead of `fmt.Print` can we use the package logger in Obot?

pkg/controller/handlers/slackreceiver/slackreceiver.go
```diff
@@ -47,3 +71,146 @@ func CreateOAuthApp(req router.Request, _ router.Response) error {
...
+		}
+	}
+
+	if len(errs) > 0 {
+		fmt.Printf("failed to create workflow executions: %v", errors.Join(errs...))
+	}
+}
+
+func middlewareConnecting(_ *socketmode.Event, _ *socketmode.Client) {
+	log.Println("Connecting to Slack with Socket Mode...")
```
thedadams: Do we need these logs? They're probably useful for debugging.

If so, can we use the package logger in Obot instead of `log`?

thedadams (changes requested):

pkg/controller/handlers/slackreceiver/slackreceiver.go
StrongMonkey: added lock to protect read/write

pkg/controller/handlers/slackreceiver/slackreceiver.go
StrongMonkey: yeah, switched to use the log package. These log are useful for debugging.

pkg/controller/handlers/slackreceiver/slackreceiver.go
```diff
@@ -24,14 +24,17 @@ import (
...
 type Handler struct {
 	gptScript  *gptscript.GPTScript
+	lock       sync.RWMutex
 	subscribed map[string]context.CancelFunc
 	storage    storage.Client
 }
 
 func NewHandler(gptScript *gptscript.GPTScript, storage storage.Client) *Handler {
-	return &Handler{gptScript: gptScript, subscribed: make(map[string]context.CancelFunc), storage: storage}
+	return &Handler{gptScript: gptScript, subscribed: make(map[string]context.CancelFunc), storage: storage, lock: sync.RWMutex{}}
```
thedadams: nit: not necessary.

pkg/controller/handlers/slackreceiver/slackreceiver.go
thedadams: I looked into this and this won't work like you expect. If you really need to look at this again, you'll have to do something else.

pkg/controller/handlers/slackreceiver/slackreceiver.go
```diff
@@ -47,3 +74,151 @@ func CreateOAuthApp(req router.Request, _ router.Response) error {
...
+		resp.RetryAfter(time.Second * 10)
+	}()
+
+	return nil
+}
+
+func (s *Handler) UnsubscribeFromSlackEvents(req router.Request, _ router.Response) error {
+	slackReceiver := req.Object.(*v1.SlackReceiver)
+	s.lock.Lock()
+	s.subscribed[slackReceiver.Name]()
```
thedadams: Should we check that the function exists before calling it, just to be safe?

pkg/controller/handlers/slackreceiver/slackreceiver.go
StrongMonkey: Sorry missed this. I am going to update the status of the resource properly when connection fails so that it will reenqeue the object properly and retry in the goroutine. The goal is to provide a status error when connection to slack failed and controller retry properly.

pkg/controller/handlers/slackreceiver/slackreceiver.go
StrongMonkey: yeah, added the check.

pkg/controller/handlers/slackreceiver/slackreceiver.go
```diff
@@ -47,3 +74,165 @@ func CreateOAuthApp(req router.Request, _ router.Response) error {
...
+			if err := req.Client.Status().Update(req.Ctx, slackReceiver); err != nil {
+				log.Errorf("failed to update slack receiver status: %v", err)
+			}
+		}
+		err := socketmodeHandler.RunEventLoopContext(ctx)
+		if err != nil {
+			log.Errorf("error running event loop: %v", err)
+		}
+		s.lock.Lock()
+		delete(s.subscribed, slackReceiver.Name)
```
thedadams: We still want to call the cancel function here so the resources associated to the context can be cleaned up.

StrongMonkey: Oh, good point

pkg/controller/handlers/slackreceiver/slackreceiver.go
```diff
@@ -1,15 +1,42 @@
...
+type Handler struct {
+	gptScript  *gptscript.GPTScript
+	lock       sync.RWMutex
+	subscribed map[string]context.CancelFunc
+	storage    storage.Client
+}
+
+func NewHandler(gptScript *gptscript.GPTScript, storage storage.Client) *Handler {
+	return &Handler{gptScript: gptScript, subscribed: make(map[string]context.CancelFunc), storage: storage}
+}
```
g-linville, line 38: Shouldn't the `lock` be created here too?

pkg/controller/handlers/slackreceiver/slackreceiver.go
```diff
@@ -47,3 +74,168 @@ func CreateOAuthApp(req router.Request, _ router.Response) error {
...
+			delete(s.subscribed, slackReceiver.Name)
+		}
+		s.lock.Unlock()
+		if err != nil {
+			err = fmt.Errorf("failed to run event loop: %w", err)
+			slackReceiver.Status.SocketConnectError = err.Error()
+			if err := req.Client.Status().Update(req.Ctx, slackReceiver); err != nil {
+				log.Errorf("failed to update slack receiver status: %v", err)
+			}
+		}
```
g-linville: Can we combine these two `if err != nil` checks? It doesn't look like `err` gets changed in between them.

g-linville (approved):
just nits

pkg/controller/handlers/slackreceiver/slackreceiver.go
StrongMonkey: donnie mentioned that this is not necessary to initialize lock

pkg/controller/handlers/slackreceiver/slackreceiver.go
StrongMonkey: yeah, my bad. The first one can be moved down here

pkg/controller/handlers/slackreceiver/slackreceiver.go
g-linville: Oh interesting, I did not know that

Issue #3128, opened 2025-05-23
Chabot Configure Page - Input field's border disappear when entering the fields


sangee2004:
**Describe the bug**
Chabot Configure Page - Input field's border disappear when entering the fields

**To Reproduce**
1. Create an agent with github MCP server.
2. Enable chatbot
3. Choose `configure it for everyone` which will present user with the following configure page 

Notice that Input field's border disappear when entering any of the fields in this page
<img width="681" alt="Image" src="https://github.com/user-attachments/assets/c9cd0d93-722d-411e-a202-2576a44b4417" />

**Expected behavior**

**Screenshots**

**Additional context**

sangee2004:
This issue is addressed when testing with latest version - `v0.0.0-dev+a1bb89c9`
<img width="639" alt="Image" src="https://github.com/user-attachments/assets/ff069010-24e6-44fe-949e-0f448df0c9e8" />

PR #3182, opened 2025-06-11
fix: prevent chatbot config input border from being trimmed

njhale:
Addresses https://github.com/obot-platform/obot/issues/3128


StrongMonkey (approved):

ui/user/src/lib/components/edit/ChatbotConfig.svelte
```diff
@@ -409,7 +409,7 @@
 				users.
 			</p>
 
-			<div class="max-h-[60vh] overflow-y-auto">
+			<div class="default-scrollbar-thin min-h-0 w-full overflow-y-auto px-1">
```
ivyjeong13: Only thing is I don't think I could get the scrollbar when I put padding into the content to try and force content overflow but might be something on my end!

njhale:
@ivyjeong13 I removed `default-scrollbar-thin` and was able to trigger the scroll bar by forcing the content to overflow (decreasing the vertical size of the view/browser). Going to merge this.

Issue #2807, opened 2025-05-06
Chatbot - Disabling chatbot for an agent should result in deleting all  chatbots created from this agent.


sangee2004:
**Describe the bug**
Chatbot - Disabling chatbot for an agent should result in deleting all  chatbots created from this agent.

**To Reproduce**
1. As user1 , Create an agent.
2. Enable chatbot access
3. As another user  say user2 , launch chat bot using share link .
4.  As user1 , disable chatbot.

Notice that user2 is still able to use the chatbots and chat with it successfully.

**Expected behavior**
Disabling chatbot for an agent should result in deleting all  chatbots created from this agent.

njhale:
I believe this was fixed. Just tested this and refreshing the chatbot as `user2` leads to a 404 as expected. trying to reuse the old share link also doesn't work. After re-enabling chatbots as `user1` and hitting the new share link as `user2` leads to the creation of a new chatbot instance as well.

njhale:
> I believe this was fixed. Just tested this and refreshing the chatbot as user2 leads to a 404 as expected. trying to reuse the old share link also doesn't work. After re-enabling chatbots as user1 and hitting the new share link as user2 leads to the creation of a new chatbot instance as well.

I jumped the gun here. The share link no longer works, and using it again after re-enabling results in a new chatbot instance, but the old chatbot instance is still available via the home page and didn't get cleaned up. Thanks for catching that @sangee2004

sangee2004:
Tested with `v0.0.0-dev+a1bb89c9`.

I am still able to reproduce the issue mentioned in the bug.

The chatbot instance that was created continues to be present when the chatbot gets disabled. We expect this chatbot instance to get deleted .

sangee2004:
This issue is addressed when testing with latest builds. When disabling chatbot for agent, existing chatbot instances get deleted as expected.

PR #3188, opened 2025-06-12
fix: cleanup chatbot projects when chatbots are disabled

njhale:
Addresses https://github.com/obot-platform/obot/issues/2807

Note: Need to add tests for this, but works manually

### DIFF
--- a/pkg/controller/handlers/projects/projects.go
+++ b/pkg/controller/handlers/projects/projects.go
@@ -1,9 +1,13 @@
 package projects
 
 import (
+	"strings"
+
 	"github.com/obot-platform/nah/pkg/router"
 	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
+	"github.com/obot-platform/obot/pkg/system"
 	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
 type Handler struct{}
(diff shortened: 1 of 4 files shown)


pkg/controller/handlers/projects/projects.go
```diff
@@ -30,3 +34,41 @@ func (h *Handler) CopyProjectInfo(req router.Request, _ router.Response) error {
...
 }
+
+func (*Handler) CleanupChatbots(req router.Request, _ router.Response) error {
+	thread := req.Object.(*v1.Thread)
+	if !thread.Spec.Project || thread.Spec.ParentThreadName == "" {
+		return nil
+	}
+
+	// Bail out early if the thread is already being deleted
+	if thread.DeletionTimestamp != nil && !thread.DeletionTimestamp.IsZero() {
```
thedadams: Threads that are being deleted won't come through this handler.

pkg/controller/handlers/projects/projects.go
```diff
@@ -30,3 +34,41 @@ func (h *Handler) CopyProjectInfo(req router.Request, _ router.Response) error {
...
+		parentProjectID  = strings.Replace(parentThread.Name, system.ThreadPrefix, system.ProjectPrefix, 1)
+		projectShareName = system.GetProjectShareName(parentThread.Spec.UserID, parentProjectID)
+		threadShare      v1.ThreadShare
+		err              = req.Get(&threadShare, thread.Namespace, projectShareName)
+	)
+
+	if apierrors.IsNotFound(err) {
+		// Project share doesn't exist, delete the child project thread
+		return req.Delete(thread)
+	}
```
thedadams: nit: this is confusing. I would prefer the standard if err = ...; apierrors.IsNotFound(err) {...}

pkg/controller/handlers/projects/projects.go
njhale: fixed

pkg/controller/handlers/projects/projects.go
njhale: fixed

g-linville (approved):

PR #3196, opened 2025-06-13
chore: add community s3 mcp server

drpebcak:


catalog/aws-s3.json
```diff
@@ -0,0 +1,64 @@
...
+  "metadata": {
+    "categories": "SaaS & API Integrations, Verified"
+  },
+  "manifest": [
+    {
+      "env": [
+        {
+          "key": "AWS_REGION",
+          "name": "AWS Region",
+          "value": "<aws_region>",
```
njhale, line 17: nit: Is `value` used anywhere? Last time I checked, I think it wasn't part of the type this gets marshaled into

drpebcak: I don't think so, I just copied it from another definition we had. Although, maybe if I keep doing this we will eventually support passing a default value lol

njhale: I have the feeling @g-linville added something for this at some point

PR #3199, opened 2025-06-14
feat: add admin MCP catalog management

g-linville:
**What this PR does**: add an MCPCatalog type and make it manageable by the admin.
Each MCPCatalog contains a set of source URLs, allowed users, and MCPServerCatalogEntries (hereafter just "entries"). For each of the source URLs, a controller handler scrapes them and creates the entries, but admins can also manually CRUD entries in MCPCatalogs as well.

When a catalog entry is deleted (manually by an admin, or due to syncing with a source URL, or due to the deletion of the catalog it belongs to), we shut down and delete all of the MCP servers that were created from that entry. We do the same whenever the user permissions on the MCP catalog change: we look at all the servers created from the catalog and shut down and delete all of them that are no longer authorized to run.

Note that a catalog entry is uniquely identified by its `fullName` field. If this changes in the upstream source that the source URL points to, then the existing catalog entry will be deleted and a new one created in its place, shutting down any servers that were created from it.

**What this PR doesn't do**: enforce strict adherence to the catalog. While users will only be able to see the catalog entries that they have been granted access to, there are no restrictions preventing them from creating custom MCP servers or from modifying the URL, command, and arguments on an MCP server created from a catalog entry. This will come in a future change.

Also, tests. There are some tests for scraping a GitHub repo for catalog entries. The rest of this does not really have tests yet, since we don't yet have a framework for API routes or controller handlers.


apiclient/types/mcpserver.go
```diff
@@ -48,7 +49,7 @@ type MCPServer struct {
 	Configured             bool     `json:"configured"`
 	MissingRequiredEnvVars []string `json:"missingRequiredEnvVars,omitempty"`
 	MissingRequiredHeaders []string `json:"missingRequiredHeader,omitempty"`
-	CatalogID              string   `json:"catalogID"`
+	CatalogEntryID         string   `json:"catalogID"`
```
g-linville: Might be worthwhile changing the JSON tag on this. This isn't a catalog ID; it's a catalog entry ID.

pkg/api/handlers/catalogs.go
```diff
@@ -0,0 +1,385 @@
...
+		items = append(items, types.MCPCatalog{
+			Metadata: MetadataFrom(&item),
+			MCPCatalogManifest: types.MCPCatalogManifest{
+				DisplayName:    item.Spec.DisplayName,
+				SourceURLs:     item.Spec.SourceURLs,
+				AllowedUserIDs: item.Spec.AllowedUserIDs,
+			},
+			IsReadOnly: item.Spec.IsReadOnly,
+			LastSynced: *types.NewTime(item.Status.LastSyncTime.Time),
+		})
```
g-linville: TODO: consider writing a function to convert this, to DRY it up a bit

pkg/api/handlers/catalogs.go
```diff
@@ -0,0 +1,385 @@
...
+	for _, urlStr := range manifest.SourceURLs {
+		u, err := url.Parse(urlStr)
+		if err != nil {
+			return fmt.Errorf("invalid URL: %w", err)
+		}
+
+		if u.Scheme != "https" {
+			return fmt.Errorf("only HTTPS URLs are supported")
+		}
+	}
```
g-linville: Do we want to allow the admin to do file paths as well? If you work around this API handler by using kubectl to directly modify the catalogs, you can just point them to directories and that works without any issues.

pkg/api/handlers/catalogs.go
```diff
@@ -0,0 +1,385 @@
...
+	)
+	if manifest.Server.Command != "" {
+		hasCommand = true
+
+		if manifest.Server.Command == "docker" {
+			if len(manifest.Server.Args) == 0 || len(h.allowedDockerImageRepos) > 0 && !slices.ContainsFunc(h.allowedDockerImageRepos, func(s string) bool {
+				return strings.HasPrefix(manifest.Server.Args[len(manifest.Server.Args)-1], s)
+			}) {
+				return false, false, fmt.Errorf("docker command must be followed by a valid image name from one of the allowed repos (%s)", strings.Join(h.allowedDockerImageRepos, ", "))
+			}
```
g-linville: TODO: probably worthwhile to validate to make sure the command is one of ["docker", "uvx", "npx"], etc.

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -0,0 +1,467 @@
...
+
+		return nil
+	}
+
+	// I know we don't want to do apply anymore. But we were doing it before in a different place.
+	// Now we're doing it here. It's not important enough to change right now.
+	return apply.New(req.Client).WithOwnerSubContext(mcpCatalog.Name).Apply(req.Ctx, mcpCatalog, toAdd...)
+}
+
+func (h *Handler) readMCPCatalog(catalogName, sourceURL string) ([]client.Object, error) {
```
g-linville, line 75: Virtually all of the code from this function all the way down to the start of the next proper controller route was copied over from the ToolReference handler, which used to be responsible for setting up the MCPServerCatalogEntries. Very little of this is new logic.

pkg/services/config.go
```diff
@@ -376,6 +376,7 @@ func New(ctx context.Context, config Config) (*Services, error) {
 	}
 
 	apply.AddValidOwnerChange("otto-controller", "obot-controller")
+	apply.AddValidOwnerChange("mcpcatalogentries", "default")
```
g-linville: TODO: consider using the subcontext format `catalog-{catalog name}` rather than just the catalog name

pkg/api/handlers/catalogs.go
g-linville: done

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -0,0 +1,381 @@
...
+// Refresh refreshes a catalog to sync its entries.
+func (h *MCPCatalogHandler) Refresh(req api.Context) error {
+	catalogName := req.PathValue("catalog_id")
+
+	var catalog v1.MCPCatalog
+	if err := req.Get(&catalog, catalogName); err != nil {
+		return fmt.Errorf("failed to get catalog: %w", err)
+	}
+
+	if catalog.Annotations[v1.MCPCatalogSyncAnnotation] != "" {
```
njhale, line 63: shouldn't the nil check take place before this check?

apiclient/types/mcpserver.go
thedadams: Yeah, change it. There is probably some UI code that would also need to change.

pkg/api/handlers/mcp.go
```diff
@@ -49,21 +50,57 @@ func (m *MCPHandler) GetCatalogEntry(req api.Context) error {
...
+	if !req.UserIsAdmin() {
+		userCatalogAuthorizations, err := usercatalogauthorization.GetUserAuthorizationsForCatalog(req.Context(), req.Storage, req.User.GetUID(), entry.Spec.MCPCatalogName)
+		if err != nil {
+			return err
+		}
+
+		if len(userCatalogAuthorizations) == 0 {
+			return types.NewErrForbidden("user is not authorized to access this catalog entry")
+		}
+	}
```
thedadams: Can this be in the `authz` package? We only want to have authorization in the API layer if it can be added there.

pkg/api/handlers/mcp.go
```diff
@@ -49,21 +50,57 @@ func (m *MCPHandler) GetCatalogEntry(req api.Context) error {
...
 
-func (m *MCPHandler) ListCatalog(req api.Context) error {
+func (m *MCPHandler) ListEntriesForAllCatalogs(req api.Context) error {
 	var list v1.MCPServerCatalogEntryList
 	if err := req.List(&list); err != nil {
 		return err
 	}
 
-	items := make([]types.MCPServerCatalogEntry, 0, len(list.Items))
+	if req.UserIsAdmin() {
```
thedadams, line 74: Should this be?
```suggestion
	if !req.UserIsAdmin() {
```

pkg/api/handlers/mcp.go
```diff
@@ -49,21 +50,57 @@ func (m *MCPHandler) GetCatalogEntry(req api.Context) error {
...
+	if err != nil {
+		return err
+	}
+
+	// TODO(g-linville): would it be better to do a separate list for each catalog that the user is authorized for,
+	// rather than filtering out entries from the full list here?
+	authorizedCatalogs := map[string]struct{}{}
+	for _, authorization := range userCatalogAuthorizations {
+		authorizedCatalogs[authorization.Spec.MCPCatalogName] = struct{}{}
+	}
```
thedadams: As a counterpoint to myself, I think this is an example of authorization that can't be done easily in the `authz` package.

pkg/api/handlers/mcp.go
```diff
@@ -49,21 +50,57 @@ func (m *MCPHandler) GetCatalogEntry(req api.Context) error {
...
+
+	var entries []types.MCPServerCatalogEntry
 	for _, entry := range list.Items {
-		items = append(items, convertMCPServerCatalogEntry(entry))
+		if _, ok := authorizedCatalogs[entry.Spec.MCPCatalogName]; ok || entry.Spec.MCPCatalogName == "" {
+			// (Entries without a catalog likely come from legacy gptscript tool bundles.)
```
thedadams: If legacy tool bundles is what you are _really_ after here, you can check `entry.Spec.ToolReferenceName`.

apiclient/types/mcpserver.go
```diff
@@ -5,6 +5,7 @@ type MCPServerCatalogEntry struct {
 	CommandManifest   MCPServerCatalogEntryManifest `json:"commandManifest,omitzero"`
 	URLManifest       MCPServerCatalogEntryManifest `json:"urlManifest,omitzero"`
 	ToolReferenceName string                        `json:"toolReferenceName,omitzero"`
+	Editable          bool                          `json:"editable,omitempty"`
```
thedadams, line 8: Does it make sense to add the `CatalogName` here, too?

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -0,0 +1,381 @@
...
+
+	for _, urlStr := range manifest.SourceURLs {
+		u, err := url.Parse(urlStr)
+		if err != nil {
+			return fmt.Errorf("invalid URL: %w", err)
+		}
+
+		if u.Scheme != "https" {
+			return fmt.Errorf("only HTTPS URLs are supported")
+		}
```
thedadams: How do you feel about only erroring here if `u.Host != "localhost" || u.Host != "127.0.0.1"`?

I run MCP servers locally for testing all the time and connect to them via HTTP.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -0,0 +1,381 @@
...
+		return fmt.Errorf("failed to read catalog manifest: %w", err)
+	}
+
+	var catalog v1.MCPCatalog
+	if err := req.Get(&catalog, req.PathValue("catalog_id")); err != nil {
+		return fmt.Errorf("failed to get catalog: %w", err)
+	}
+
+	if catalog.Spec.IsReadOnly {
+		return fmt.Errorf("cannot update a read-only catalog")
```
thedadams: This error and the next few seem like they should be `400`-level errors. Return the error here will produce a `500`.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -0,0 +1,381 @@
...
+
+// Delete deletes a catalog.
+func (h *MCPCatalogHandler) Delete(req api.Context) error {
+	var catalog v1.MCPCatalog
+	if err := req.Get(&catalog, req.PathValue("catalog_id")); err != nil {
+		return fmt.Errorf("failed to get catalog: %w", err)
+	}
+
+	if catalog.Spec.IsReadOnly {
+		return fmt.Errorf("cannot delete a read-only catalog")
```
thedadams: Same comment here about `400`-level errors. Last comment I'll make about this, but we should look through this file and update as needed.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+
+	var (
+		hasCommand, hasURL bool
+	)
+	if manifest.Server.Command != "" {
+		hasCommand = true
+
+		if manifest.Server.Command == "docker" {
+			if len(manifest.Server.Args) == 0 || len(h.allowedDockerImageRepos) > 0 && !slices.ContainsFunc(h.allowedDockerImageRepos, func(s string) bool {
```
thedadams: Two comments here:
1. `len(manifest.Server.Args) == 0` should probably fail validation.
2. The other logic in this code that validates allowed Docker images treats an empty slice (either `nil` or length 0) as "no Docker images allowed." I am fine with changing that everywhere, but I think it should be consistent.

pkg/api/router/router.go
```diff
@@ -377,10 +378,22 @@ func Router(services *services.Services) (http.Handler, error) {
...
+	mux.HandleFunc("GET /api/mcp-catalogs", mcpCatalogs.List)
+	mux.HandleFunc("GET /api/mcp-catalogs/{catalog_id}", mcpCatalogs.Get)
+	mux.HandleFunc("POST /api/mcp-catalogs/{catalog_id}/refresh", mcpCatalogs.Refresh)
+	mux.HandleFunc("POST /api/mcp-catalogs", mcpCatalogs.Create)
+	mux.HandleFunc("PUT /api/mcp-catalogs/{catalog_id}", mcpCatalogs.Update)
+	mux.HandleFunc("DELETE /api/mcp-catalogs/{catalog_id}", mcpCatalogs.Delete)
+	mux.HandleFunc("GET /api/mcp-catalogs/{catalog_id}/entries", mcpCatalogs.ListEntriesForCatalog)
+	mux.HandleFunc("POST /api/mcp-catalogs/{catalog_id}/entries", mcpCatalogs.CreateEntry)
+	mux.HandleFunc("PUT /api/mcp-catalogs/{catalog_id}/entries/{entry_id}", mcpCatalogs.UpdateEntry)
+	mux.HandleFunc("DELETE /api/mcp-catalogs/{catalog_id}/entries/{entry_id}", mcpCatalogs.DeleteEntry)
```
thedadams, line 395: nit: can we make this consistent with what's above? My preference would be `mcp-catalogs`.

If we do make this change, we probably need to update some UI code, too.

pkg/controller/handlers/mcpcatalog/github.go
```diff
@@ -0,0 +1,182 @@
...
+			for line := range strings.SplitSeq(string(content), "\n") {
+				line = strings.TrimSpace(line)
+				if line != "" && !strings.HasPrefix(line, "#") {
+					patterns = append(patterns, line)
+				}
+			}
+			if len(patterns) > 0 {
+				catalogPatterns = patterns
+			}
+		}
```
thedadams: nit: I wonder if using a `bufio.Scanner` here would make it slightly more efficient?

pkg/controller/handlers/mcpcatalog/github.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	resp, err = http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list repository contents: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("failed to list repository contents: %s", resp.Status)
```
thedadams: nit: I've found it useful to read the body and include it in the error.

pkg/controller/handlers/mcpcatalog/github.go
```diff
@@ -0,0 +1,182 @@
...
+
+		resp, err := http.DefaultClient.Do(req)
+		if err != nil {
+			log.Warnf("Failed to get contents of %s: %v", item.Path, err)
+			continue
+		}
+		content, err := io.ReadAll(resp.Body)
+		resp.Body.Close()
+		if err != nil {
+			log.Warnf("Failed to read contents of %s: %v", item.Path, err)
```
thedadams, line 161: Same comment here about the body in the error message.

pkg/controller/handlers/usercatalogauthorization/operations.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/obot-platform/obot/pkg/storage"
+	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
+	"github.com/obot-platform/obot/pkg/system"
+	kclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func GetAuthorizationsForUser(ctx context.Context, c storage.Client, userID string) ([]v1.UserCatalogAuthorization, error) {
+	var authorizations v1.UserCatalogAuthorizationList
+	if err := c.List(ctx, &authorizations, kclient.InNamespace(system.DefaultNamespace), kclient.MatchingFields{
```
thedadams: nit: I know that everything lives in the `system.DefaultNamespace` right now, but is it possible to pass the namespace into these functions? That way, if we ever use namespaces, this logic wouldn't have to change.

Same for all of these functions.

pkg/controller/handlers/usercatalogauthorization/operations.go
```diff
@@ -0,0 +1,60 @@
...
+		"spec.mcpCatalogName": catalogName,
+	}); err != nil {
+		return nil, err
+	}
+	return authorizations.Items, nil
+}
+
+func GetUserAuthorizationsForCatalog(ctx context.Context, c storage.Client, userID string, catalogName string) ([]v1.UserCatalogAuthorization, error) {
```
thedadams: Can we add a comment that this function can't be called with a cached client? The cached clients don't work with multiple field selectors.

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -0,0 +1,467 @@
...
+				}
+			}
+		}
+
+		return nil
+	}
+
+	// I know we don't want to do apply anymore. But we were doing it before in a different place.
+	// Now we're doing it here. It's not important enough to change right now.
+	return apply.New(req.Client).WithOwnerSubContext(fmt.Sprintf("catalog-%s", mcpCatalog.Name)).Apply(req.Ctx, mcpCatalog, toAdd...)
```
thedadams: This will delete previous entries if `len(toAdd) == 0`, so you'd be able to remove the above logic.
```suggestion
	return apply.New(req.Client).WithOwnerSubContext(fmt.Sprintf("catalog-%s", mcpCatalog.Name)).WithPruneTypes(&v1.MCPServerCatalogEntry{}).Apply(req.Ctx, mcpCatalog, toAdd...)
```

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -0,0 +1,467 @@
...
+	}
+
+	// This would mean that existing is docker and newer is either npx or uvx.
+	return true
+}
+
+// DeleteUnauthorizedMCPServers deletes all MCP servers that are no longer authorized to run.
+// This can happen when a user has launched an MCP server that they used to have permission for,
+// but their access to the catalog that the server came from has been revoked.
+func (h *Handler) DeleteUnauthorizedMCPServers(req router.Request, _ router.Response) error {
```
thedadams, line 306: This does seem pretty inefficient.

Can we put the catalog name on the MCP server itself? Then this controller could be on the MCP server object, and it could get its `MCPCatalog` and check to see if it needs to delete itself.

What are your thoughts on this? Is that actually better?

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -0,0 +1,467 @@
...
+		return fmt.Errorf("failed to create default catalog: %w", err)
+	}
+
+	return nil
+}
+
+func (h *Handler) SetUpUserCatalogAuthorizations(req router.Request, _ router.Response) error {
+	mcpCatalog := req.Object.(*v1.MCPCatalog)
+
+	authorizationNames := map[string]struct{}{}
```
thedadams:
```suggestion
	authorizationNames := make(map[string]struct{}, len(mcpCatalog.Spec.AllowedUserIDs))
```

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -0,0 +1,467 @@
...
+		authorizationName := name.SafeHashConcatName(mcpCatalog.Name, userID)
+		if userID == "*" {
+			authorizationName = name.SafeHashConcatName(mcpCatalog.Name, "all-users")
+		}
+
+		authorizationNames[authorizationName] = struct{}{}
+
+		// See if this authorization already exists.
+		var existingAuthorization v1.UserCatalogAuthorization
+		if err := req.Client.Get(req.Ctx, router.Key(system.DefaultNamespace, authorizationName), &existingAuthorization); apierrors.IsNotFound(err) {
```
thedadams: There is a `create.IfNotExists` if you want to use it here.

pkg/storage/apis/obot.obot.ai/v1/mcpcatalog.go
```diff
@@ -0,0 +1,62 @@
...
+		{"Source URLs", "Spec.SourceURLs"},
+		{"Last Synced", "{{ago .Status.LastSyncTime}}"},
+	}
+}
+
+func (in *MCPCatalog) Get(field string) string {
+	switch field {
+	case "spec.sourceURLs":
+		return strings.Join(in.Spec.SourceURLs, "\n")
+	}
```
thedadams: It looks as though this field selector is not used. Is that an oversight? If not, can we remove it?

pkg/storage/apis/obot.obot.ai/v1/mcpservercatalogentry.go
```diff
@@ -17,9 +19,40 @@ type MCPServerCatalogEntry struct {
...
+}
+
+func (in *MCPServerCatalogEntry) Has(field string) bool {
+	return slices.Contains(in.FieldNames(), field)
+}
+
+func (in *MCPServerCatalogEntry) Get(field string) string {
+	switch field {
+	case "spec.toolReferenceName":
+		return in.Spec.ToolReferenceName
```
thedadams: It looks like this field selector isn't used either. Can it be removed?

pkg/system/ids.go
```diff
@@ -30,6 +30,7 @@ const (
 	SlackTriggerPrefix        = "st1"
 	UserDeletePrefix          = "ud1"
 	MCPServerPrefix           = "ms1"
+	CatalogPrefix             = "cat1"
```
thedadams: nit:
```suggestion
	MCPCatalogPrefix             = "mcat1"
```

pkg/api/handlers/mcpcatalogs.go
thedadams: Reading from a `nil` map is totally fine.

pkg/api/handlers/mcp.go
g-linville: No, this is correct as it is. For the admin, we just return the full list immediately. After this point, for a normal user, we filter down the list to what they are allowed to see.

pkg/api/handlers/mcp.go
g-linville: Yeah, I think I would rather keep all of the authz for this and the previous function here in this file rather than splitting it between the authz package and here.

pkg/api/handlers/mcp.go
g-linville: Thanks, idk why I didn't just do that lol

apiclient/types/mcpserver.go
g-linville: Good call, the UI will need to know that

pkg/api/handlers/mcpcatalogs.go
g-linville: These are source URLs for catalogs, not URLs on the MCP servers themselves. Catalog source URLs would most commonly be GitHub repos or a static JSON file hosted somewhere.

pkg/api/handlers/mcp.go

apiclient/types/mcpserver.go
g-linville: Fixed

pkg/api/handlers/mcpcatalogs.go

pkg/api/handlers/mcpcatalogs.go

pkg/api/handlers/mcpcatalogs.go

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
g-linville: this is fantastic

pkg/api/handlers/mcpcatalogs.go
njhale: TIL only writes panic on a nil map. Thanks.

pkg/api/router/router.go
g-linville: The way the pattern matching works I can't quite make this `mcp-catalogs`, as having `/api/mcp-catalogs/entries/{entry_id}` and `/api/mcp-catalogs/{catalog_id}/entries` panics due to ambiguity. So I think I will settle for `/api/all-mcp-catalogs` as the path for the two old ones.

g-linville: Fixed

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
g-linville: fixed

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
g-linville: I think we should revisit this after we implement what we talked about today, where the admin can create servers directly in a catalog.

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
g-linville: thanks

pkg/storage/apis/obot.obot.ai/v1/mcpcatalog.go
g-linville: Yep, I probably added this thinking I would need it and then didn't. Removing

pkg/storage/apis/obot.obot.ai/v1/mcpservercatalogentry.go
g-linville: Same for this one, removing

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
g-linville: fixed

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
g-linville: fixed

pkg/storage/apis/obot.obot.ai/v1/mcpcatalog.go
g-linville: fixed

pkg/storage/apis/obot.obot.ai/v1/mcpservercatalogentry.go
g-linville: fixed

pkg/system/ids.go
g-linville: fixed

pkg/api/handlers/mcpcatalogs.go
g-linville: no, you can read from a nil map. it just returns the zero value of the type

pkg/controller/handlers/usercatalogauthorization/operations.go
g-linville: done

pkg/controller/handlers/usercatalogauthorization/operations.go
g-linville: fixed

pkg/controller/handlers/mcpcatalog/github.go
g-linville: fixed

pkg/controller/handlers/mcpcatalog/github.go
g-linville: fixed

pkg/controller/handlers/mcpcatalog/github.go
g-linville: fixed

g-linville:
@njhale @thedadams this is ready for review again

pkg/api/handlers/mcpcatalogs.go
g-linville: sorry I just saw Donnie's reply on here. for some reason GitHub wasn't showing me that when I left my comment lol

g-linville:
I am going to wait until I have approvals to do the rebase to fix conflicts, because I don't want to mess up the review history.

pkg/api/handlers/mcp.go
```diff
@@ -42,28 +43,63 @@ func NewMCPHandler(gptscript *gptscript.GPTScript, mcpLoader *mcp.SessionManager
...
+	}
+
+	userCatalogAuthorizations, err := usercatalogauthorization.GetAuthorizationsForUser(req.Context(), req.Storage, req.Namespace(), req.User.GetUID())
+	if err != nil {
+		return err
+	}
+
+	// TODO(g-linville): would it be better to do a separate list for each catalog that the user is authorized for,
+	// rather than filtering out entries from the full list here?
+	authorizedCatalogs := map[string]struct{}{}
```
```suggestion
	authorizedCatalogs := make(map[string]struct{}, len(userCatalogAuthorizations)
```

pkg/controller/handlers/mcpcatalog/github.go
```diff
@@ -73,20 +74,19 @@ func readGitHubCatalog(catalogURL string) ([]toolreference.CatalogEntryInfo, err
...
+		scanner := bufio.NewScanner(resp.Body)
+		var patterns []string
+		for scanner.Scan() {
+			line := scanner.Text()
+			line = strings.TrimSpace(line)
+			if line != "" && !strings.HasPrefix(line, "#") {
+				patterns = append(patterns, line)
 			}
 		}
+		if len(patterns) > 0 {
```
thedadams: After the loop exits, you also need to check that `scanner.Err()` is `nil` or `io.EOF`.

pkg/controller/handlers/mcpcatalog/github.go
```diff
@@ -153,10 +154,13 @@ func readGitHubCatalog(catalogURL string) ([]toolreference.CatalogEntryInfo, err
 			continue
 		}
 		content, err := io.ReadAll(resp.Body)
-		resp.Body.Close()
+		defer resp.Body.Close()
```
thedadams: Don't defer inside of the loop. Since you're reading the entire contents right above, I think it's right to just close it now (as you had it before).

pkg/api/handlers/mcp.go
g-linville: fixed

pkg/controller/handlers/mcpcatalog/github.go
g-linville: fixed

pkg/controller/handlers/mcpcatalog/github.go
g-linville: fixed

thedadams (approved):

pkg/api/handlers/mcp.go
```diff
@@ -42,28 +43,63 @@ func NewMCPHandler(gptscript *gptscript.GPTScript, mcpLoader *mcp.SessionManager
...
+
+		return req.Write(types.MCPServerCatalogEntryList{Items: items})
+	}
+
+	userCatalogAuthorizations, err := usercatalogauthorization.GetAuthorizationsForUser(req.Context(), req.Storage, req.Namespace(), req.User.GetUID())
+	if err != nil {
+		return err
+	}
+
+	// TODO(g-linville): would it be better to do a separate list for each catalog that the user is authorized for,
```
njhale, line 88: I think either is fine at this point, but once the set of catalogs/entries increases, we'll probably want to do separate lists.

pkg/controller/handlers/usercatalogauthorization/operations.go
njhale: the diff is still showing in GitHub with `system.DefaultNamespace` for me; not sure if this is GitHub being slow or if the change to use the `namespace` parameter wasn't pushed

pkg/controller/handlers/usercatalogauthorization/operations.go
```diff
@@ -0,0 +1,61 @@
...
+	}); err != nil {
+		return nil, err
+	}
+
+	return append(authorizations.Items, allUsersAuthorizations.Items...), nil
+}
+
+func GetAuthorizationsForCatalog(ctx context.Context, c storage.Client, namespace, catalogName string) ([]v1.UserCatalogAuthorization, error) {
+	var authorizations v1.UserCatalogAuthorizationList
+	if err := c.List(ctx, &authorizations, kclient.InNamespace(system.DefaultNamespace), kclient.MatchingFields{
```
njhale: Same comment to use the `namespace` parameter here.

pkg/controller/handlers/usercatalogauthorization/operations.go
```diff
@@ -0,0 +1,61 @@
...
+	}); err != nil {
+		return nil, err
+	}
+	return authorizations.Items, nil
+}
+
+// GetUserAuthorizationsForCatalog cannot be called with a cached client, as cached clients do not support more than one field selector.
+func GetUserAuthorizationsForCatalog(ctx context.Context, c storage.Client, namespace, userID, catalogName string) ([]v1.UserCatalogAuthorization, error) {
+	var authorizations v1.UserCatalogAuthorizationList
+	if err := c.List(ctx, &authorizations, kclient.InNamespace(system.DefaultNamespace), kclient.MatchingFields{
```
njhale: Use `namespace` here as well.

PR #3202, opened 2025-06-16
enhance: use project scoped workspaces for all threads and tasks

njhale:
- **wip: make all workspaces project scoped**
- **wip: user ui changes for making all workspaces project scoped**

Addresses https://github.com/obot-platform/obot/issues/3212


njhale:
Going with a slightly different UX for this. Will open a separate PR shortly.

Issue #3209, opened 2025-06-17
Ability to show user info who sent message in the thread


StrongMonkey:
We want to track the user who sent message in the same thread. For example, multiple members get added into the same thread and are able to chat in the same thread. We want the ability to track message and the person who sent the message in the chat view.

sangee2004:
When testing with latest build, we are able to see the username being presented in the chat view. 

When admin user chats in the thread belonging to a regular user using `Chat Threads` view in admin ui, chat view presented to the user shows the admin user name for the chats initiated by the user as expected.


Some follow up issues (or enhancements) logged relating to friendly user names being made available for users:

https://github.com/obot-platform/obot/issues/3227
https://github.com/obot-platform/obot/issues/3228
https://github.com/obot-platform/obot/issues/3229

PR #3210, opened 2025-06-17
Enhance: Add friendly username to user and runID

StrongMonkey:
This PR tracks the displayName from user and add that into chat run and UI so that it can show who sent message in the same thread. 

Required by https://github.com/obot-platform/obot/issues/3209

https://github.com/obot-platform/obot/issues/3209


pkg/gateway/client/user.go
```diff
@@ -388,6 +408,14 @@ func (c *Client) decryptUser(ctx context.Context, user *types.User) error {
 		errs = append(errs, err)
 	}
 
+	decoded = make([]byte, base64.StdEncoding.DecodedLen(len(user.DisplayName)))
+	n, err = base64.StdEncoding.Decode(decoded, []byte(user.DisplayName))
+	if err == nil {
+		user.DisplayName = string(decoded[:n])
```
g-linville: I think we need to make a call to `transformer.TransformFromStorage` here.

StrongMonkey: good catch!

g-linville (approved):

PR #3223, opened 2025-06-18
chore: remove the ability to add custom tools to projects

njhale:
Remove the ability to add custom tools to projects.

Addresses https://github.com/obot-platform/obot/issues/3224


pkg/render/render.go
```diff
@@ -591,3 +573,12 @@ func oauthAppsByName(ctx context.Context, c kclient.Client, namespace string, oa
 
 	return result, nil
 }
+
+func tool(ctx context.Context, c kclient.Client, ns, name string) (string, []gptscript.ToolDef, error) {
```
StrongMonkey: is this used somewhere?

njhale: I removed some of these changes from the PR. Going to follow up with a more detailed cleanup afterwards, but going to just drop the functionality from the user ui and APIs for now.

PR #3230, opened 2025-06-18
Enhance: Add integratin test and framework

StrongMonkey:
This PR added basic integration test framework to obot. It launch a obot server with controller running, setting up mock model provider so that developer can write tests against api/controller code against full api.

Still actively adding more tests, this PR serves as a baseline for setting up groundwork and and test suite examples.

### DIFF
--- a/.github/workflows/integration-test.yaml
+++ b/.github/workflows/integration-test.yaml
@@ -1,0 +1,45 @@
+name: Integration Tests
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  integration-tests:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:14
+        env:
+          POSTGRES_USER: testuser
+          POSTGRES_PASSWORD: testpass
+          POSTGRES_DB: testdb
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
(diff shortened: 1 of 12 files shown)


go.mod
```diff
@@ -225,7 +228,7 @@ require (
 	github.com/oasdiff/yaml v0.0.0-20250309154309-f31be36b4037 // indirect
 	github.com/oasdiff/yaml3 v0.0.0-20250309153720-d2182401db90 // indirect
 	github.com/olekukonko/tablewriter v0.0.6-0.20230925090304-df64c4bbad77 // indirect
-	github.com/onsi/ginkgo/v2 v2.20.2 // indirect
+	github.com/onsi/ginkgo/v2 v2.20.2
```
g-linville, line 231: Does this PR need a `go mod tidy`? It's weird to see a non-indirect dependency among the indirect ones.

g-linville (approved):

go.mod
StrongMonkey: yeah, let me run that

StrongMonkey: nope, running go mod seems to be fine

thedadams: There is something weird happening with `go mod tidy` recently. This isn't supposed to happen. Also, there is a new `require` added above, which also isn't supposed to happen. Weird.

.github/workflows/integration-test.yaml
```diff
@@ -0,0 +1,45 @@
...
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  integration-tests:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:14
```
thedadams, line 14: We're using version 17 in other places. Should we use 17 here?

thedadams (approved):

.github/workflows/integration-test.yaml
StrongMonkey: yeah, i will go ahead and update that

PR #3233, opened 2025-06-18
feat: allow admin to create MCPServers inside MCPCatalogs

g-linville:
This PR allows the admin to configure MCPServers directly within MCPCatalogs, to be used by the same users who are authorized in that catalog.

I have some questions and comments on my own PR that I would like reviewers to take a look at.

### DIFF
--- a/apiclient/types/mcpserver.go
+++ b/apiclient/types/mcpserver.go
@@ -52,6 +52,7 @@ type MCPServer struct {
 	MissingRequiredHeaders  []string `json:"missingRequiredHeader,omitempty"`
 	CatalogEntryID          string   `json:"catalogEntryID"`
 	SharedWithinCatalogName string   `json:"sharedWithinCatalogName,omitempty"`
+	ConnectURL              string   `json:"connectURL,omitempty"`
 }
 
 type MCPServerList List[MCPServer]
--- a/apiclient/types/mcpserver.go
+++ b/apiclient/types/mcpserver.go
@@ -47,10 +47,11 @@ type MCPEnv struct {
 type MCPServer struct {
 	Metadata
 	MCPServerManifest
-	Configured             bool     `json:"configured"`
-	MissingRequiredEnvVars []string `json:"missingRequiredEnvVars,omitempty"`
-	MissingRequiredHeaders []string `json:"missingRequiredHeader,omitempty"`
-	CatalogEntryID         string   `json:"catalogEntryID"`
+	Configured              bool     `json:"configured"`
+	MissingRequiredEnvVars  []string `json:"missingRequiredEnvVars,omitempty"`
+	MissingRequiredHeaders  []string `json:"missingRequiredHeader,omitempty"`
+	CatalogEntryID          string   `json:"catalogEntryID"`
+	SharedWithinCatalogName string   `json:"sharedWithinCatalogName,omitempty"`
 }
 
 type MCPServerList List[MCPServer]
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -117,34 +117,51 @@ func convertMCPServerCatalogEntry(entry v1.MCPServerCatalogEntry) types.MCPServe
 }
 
 func (m *MCPHandler) ListServer(req api.Context) error {
-	t, err := getThreadForScope(req)
-	if err != nil {
-		return err
-	}
+	catalogID := req.PathValue("catalog_id")
 
-	topMost, err := projects.GetRoot(req.Context(), req.Storage, t)
-	if err != nil {
-		return err
+	var fieldSelector kclient.MatchingFields
+	if catalogID != "" {
+		fieldSelector = kclient.MatchingFields{
+			"spec.sharedWithinMCPCatalogName": catalogID,
+		}
+	} else {
+		t, err := getThreadForScope(req)
+		if err != nil {
+			return err
+		}
+
+		topMost, err := projects.GetRoot(req.Context(), req.Storage, t)
+		if err != nil {
+			return err
+		}
+
+		fieldSelector = kclient.MatchingFields{
+			"spec.threadName": topMost.Name,
+		}
 	}
 
 	var servers v1.MCPServerList
-	if err := req.List(&servers, kclient.MatchingFields{
-		"spec.threadName": topMost.Name,
-	}); err != nil {
+	if err := req.List(&servers, fieldSelector); err != nil {
 		return nil
 	}
 
-	project, err := getProjectThread(req)
-	if err != nil {
-		return err
-	}
-
 	credCtxs := make([]string, 0, len(servers.Items))
-	for _, server := range servers.Items {
-		credCtxs = append(credCtxs, fmt.Sprintf("%s-%s", project.Name, server.Name))
-		if project.IsSharedProject() {
-			// Add default credentials shared by the agent for this MCP server if available.
-			credCtxs = append(credCtxs, fmt.Sprintf("%s-%s-shared", server.Spec.ThreadName, server.Name))
+	if catalogID != "" {
+		for _, server := range servers.Items {
+			credCtxs = append(credCtxs, fmt.Sprintf("%s-%s", catalogID, server.Name))
+		}
+	} else {
+		project, err := getProjectThread(req)
+		if err != nil {
+			return err
+		}
+
+		for _, server := range servers.Items {
+			credCtxs = append(credCtxs, fmt.Sprintf("%s-%s", project.Name, server.Name))
+			if project.IsSharedProject() {
+				// Add default credentials shared by the agent for this MCP server if available.
+				credCtxs = append(credCtxs, fmt.Sprintf("%s-%s-shared", server.Spec.ThreadName, server.Name))
+			}
 		}
 	}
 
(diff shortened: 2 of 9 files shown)


pkg/api/handlers/mcp.go
```diff
@@ -213,29 +240,36 @@ func (m *MCPHandler) GetServer(req api.Context) error {
...
-		server v1.MCPServer
-		id     = req.PathValue("mcp_server_id")
+		server    v1.MCPServer
+		id        = req.PathValue("mcp_server_id")
+		catalogID = req.PathValue("catalog_id")
 	)
 
-	project, err := getProjectThread(req)
-	if err != nil {
+	if err := req.Get(&server, id); err != nil {
```
g-linville, line 250: Is it okay that this got moved to above where we check for the project thread? Or was it important that the project check happen prior to getting the server from the db?

pkg/api/handlers/mcp.go
```diff
@@ -432,13 +474,29 @@ func (m *MCPHandler) serverForAction(req api.Context) (v1.MCPServer, mcp.ServerC
 		id     = req.PathValue("mcp_server_id")
 	)
 
-	project, err := getProjectThread(req)
-	if err != nil {
+	if err := req.Get(&server, id); err != nil {
 		return server, mcp.ServerConfig{}, nmcp.ServerCapabilities{}, err
 	}
```
g-linville, line 480: Same question here.

pkg/api/handlers/mcp.go
```diff
@@ -504,27 +554,54 @@ func mergeMCPServerManifests(existing, override types.MCPServerManifest) types.M
...
+		server = v1.MCPServer{
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: system.MCPServerPrefix,
+				Namespace:    req.Namespace(),
+			},
+			Spec: v1.MCPServerSpec{
+				Manifest:                   input.MCPServerManifest,
+				SharedWithinMCPCatalogName: catalogID,
+				UserID:                     req.User.GetUID(),
+			},
```
g-linville: I set this up so that the admin must provide the full manifest, rather than referencing a catalog entry. Let me know if this needs to change.

pkg/api/handlers/mcp.go
```diff
@@ -829,17 +974,32 @@ func (m *MCPHandler) DeconfigureSharedServer(req api.Context) error {
...
 	}
 
-	project, err := getProjectThread(req)
-	if err != nil {
-		return err
+	// For servers that are in catalogs, this checks to make sure that a catalogID was provided and that it matches.
+	// For servers that are not in catalogs, this checks to make sure that no catalogID was provided. (Both are empty strings.)
+	if mcpServer.Spec.SharedWithinMCPCatalogName != catalogID {
+		return types.NewErrNotFound("MCP server not found")
+	}
```
g-linville: This is a perfect example of why I don't like reusing API route handler functions across multiple routes like I am doing in this PR.

Without this check, there is a danger that the user could abuse the `/assistants/{assistant_id}/projects/.../{mcp_server_id}/reveal` route to access the credential belonging to a server that is actually in a catalog that they have access to, when only the admin should be able to reveal credentials in catalog-based MCP servers. I find it kind of spooky tbh. I would be more in favor of just duplicating basically all of these handlers so that we can have a better separation of concerns, but there would certainly be a lot of duplicated code. I am open to doing this tomorrow if we think it's a better idea than what I have here.

apiclient/types/mcpserver.go
```diff
@@ -47,10 +47,11 @@ type MCPEnv struct {
...
 	MCPServerManifest
-	Configured             bool     `json:"configured"`
-	MissingRequiredEnvVars []string `json:"missingRequiredEnvVars,omitempty"`
-	MissingRequiredHeaders []string `json:"missingRequiredHeader,omitempty"`
-	CatalogEntryID         string   `json:"catalogEntryID"`
+	Configured              bool     `json:"configured"`
+	MissingRequiredEnvVars  []string `json:"missingRequiredEnvVars,omitempty"`
+	MissingRequiredHeaders  []string `json:"missingRequiredHeader,omitempty"`
+	CatalogEntryID          string   `json:"catalogEntryID"`
+	SharedWithinCatalogName string   `json:"sharedWithinCatalogName,omitempty"`
```
thedadams, line 54: Add a `ConnectURL` (not committed to the name) field here. You don't have to set it, but at least it will be there and Ivy can use it for displaying how to connect to the MCP server from other clients.

I will make sure it gets set to the right thing in my PR.

pkg/api/handlers/mcp.go
thedadams: This is fine.

pkg/api/handlers/mcp.go
```diff
@@ -255,19 +293,27 @@ func (m *MCPHandler) GetTools(req api.Context) error {
...
+	var allowedTools []string
 
-	thread, err = projects.GetFirst(req.Context(), req.Storage, thread, func(project *v1.Thread) (bool, error) {
-		return project.Spec.Manifest.AllowedMCPTools[server.Name] != nil, nil
-	})
-	if err != nil {
-		return fmt.Errorf("failed to get project: %w", err)
+	if server.Spec.SharedWithinMCPCatalogName != "" {
+		allowedTools = []string{"*"}
+	} else {
```
thedadams: micro-nit:
```suggestion
	if server.Spec.SharedWithinMCPCatalogName == "" {
```

pkg/api/handlers/mcp.go
thedadams: Please correct me if I am wrong, but I think that the below logic will allow an admin to set a catalog entry ID only and the manifest be set from there.

If I am wrong, then we should update it so that the project-based and catalog-based MCP servers can be created the same way. However, I think it works this way now.

pkg/api/handlers/mcp.go
thedadams: I think this is fine, and consistent with the same check in the other handlers. I understand that this check may have more substantial meaning, but it is perfectly reasonable to have it here.

pkg/api/handlers/mcp.go
```diff
@@ -1127,11 +1287,116 @@ func convertMCPServer(server v1.MCPServer, credEnv map[string]string) types.MCPS
...
+		for _, server := range list.Items {
+			credCtxs = append(credCtxs, fmt.Sprintf("%s-%s", catalog.Name, server.Name))
+		}
+
+		creds, err := m.gptscript.ListCredentials(req.Context(), gptscript.ListCredentialsOptions{
+			CredentialContexts: credCtxs,
+		})
+		if err != nil {
+			return fmt.Errorf("failed to list credentials: %w", err)
+		}
```
thedadams: What are your thoughts about collecting all the credentials for every server up front? I think it would be better to loop over the catalogs twice rather than make a bunch of `ListCredential` calls.

What are your thoughts?

pkg/api/handlers/mcp.go
njhale: This should be fine. We only need the project to ensure chatbot users don't delete MCP servers in the agent project later on in the method, and getting the mcp server first shouldn't affect that check.

pkg/api/handlers/mcp.go
njhale: this should be fine here too.

apiclient/types/mcpserver.go
njhale, line 54: nit: `SharedByCatalog` sounds more apt to me

pkg/api/handlers/mcp.go
g-linville: You're correct. I'll switch it over.

g-linville: Actually I just heard from Craig that servers within catalogs should always be created from scratch, and never from an existing catalog entry, so I will keep this as it currently is.

pkg/api/handlers/mcp.go
g-linville: Thanks. I didn't realize before that a `nil` allowedTools is the same as `[]string{"*"}`.

pkg/api/handlers/mcp.go
g-linville: Sounds good. Working on this now.

g-linville: Done.

pkg/api/handlers/mcp.go
g-linville: Done.

apiclient/types/mcpserver.go
g-linville: Done.

pkg/storage/apis/obot.obot.ai/v1/mcpserver.go
```diff
@@ -52,18 +55,24 @@ func (in *MCPServer) DeleteRefs() []Ref {
...
+	ThreadName string `json:"threadName,omitempty"`
+	// UserID is the user that created this server.
+	UserID string `json:"userID,omitempty"`
+	// SharedWithinMCPCatalogName contains the name of the MCPCatalog inside of which this server was directly created by the admin, if there is one.
+	SharedWithinMCPCatalogName string `json:"sharedWithinMCPCatalogName,omitempty"`
+	// MCPServerCatalogEntryName contains the name of the MCPServerCatalogEntry from which this MCP server was created, if there is one.
 	MCPServerCatalogEntryName string `json:"mcpServerCatalogEntryName,omitempty"`
-	ToolReferenceName         string `json:"toolReferenceName,omitempty"`
+	// ToolReferenceName contains the name of the legacy gptscript tool reference for this MCP server, if there is one.
+	ToolReferenceName string `json:"toolReferenceName,omitempty"`
```
njhale, line 75: 💯 for retroactive documentation

njhale (approved):

apiclient/types/mcpserver.go
g-linville: Yeah I struggled a bit with the name for this one. Tbh I'm not a big fan of `SharedByCatalog` because that implies to me that the catalog is the thing that is doing the sharing. `SharedInCatalog` could work. I guess I liked "within" for this because it strongly implies that only people who can do stuff in this catalog can do stuff with this server.

Issue #3232, opened 2025-06-18
No username shown in chat for github users that do not have a name set.


sangee2004:
**Describe the bug**
Empty username shown in chat for github users that do not have a name set.

**To Reproduce**
1. Enable github Auth provider in otto server deployment.
2. Log in as github user who does not have a user name setup

Only account Id is shown in the github accounts profile
<img width="340" alt="Image" src="https://github.com/user-attachments/assets/4050c9c9-9da5-4e7b-8350-fa94f594220a" />

3. Create an agent and chat with this agent
Notice that empty username is shown in the chat view and we see just `by` for the chat

<img width="785" alt="Image" src="https://github.com/user-attachments/assets/1c1c11cf-2984-438c-8fb3-39b1246abd78" />

**Expected behavior**
In this case we should be able to fall back to the `username`  field for the user which is the github id.

sangee2004:
With the latest build , we do not see `by` with empty username anymore in the chat. We show only time for the chat.

Tracking this issue for  showing the `username` field for the user which is the github id in this case.

sangee2004:
This issue is addressed when testing with latest build from main.

We fallback to showing the username field for the user when friendly username is not available for the users.

PR #3238, opened 2025-06-19
Enhance: fallback to github username, return displayName in api

StrongMonkey:
Fixed an issue in case github name is not set, we use github username as displayName

https://github.com/obot-platform/obot/issues/3232


cjellick (approved):
lint issue but otherwise good

PR #3245, opened 2025-06-20
Fix sse stdio mcp gateway

thedadams:


pkg/api/handlers/mcpgateway/handler.go
```diff
@@ -93,6 +93,25 @@ func (m *messageHandler) OnMessage(ctx context.Context, msg nmcp.Message) {
...
 	case "initialize":
+		id := msg.Session.ID()
+		context.AfterFunc(ctx, func() {
+			if err := m.handler.mcpSessionManager.CloseClient(context.Background(), m.serverConfig); err != nil {
+				log.Errorf("Failed to shutdown server %s: %v", m.mcpServer.Name, err)
+			}
+			m.handler.sessions.LoadAndDelete(id)
+		})
+
+		if client.Session.InitializeResult != nil {
```
thedadams, line 104: This is for when the downstream MCP server is SSE. The initialization happens on connect, so detect that we've already initialize and just respond instead of trying to send this request downstream.

g-linville (approved):

Issue #3248, opened 2025-06-20
Not able to configure `everything` server with "streamable http" transport.


sangee2004:
**Describe the bug**
Not able to configure `everything` server with "streamable http" transport.

**To Reproduce**
**Scenario 1** 
1. Create an agent
2. Add `everything` server with "streamable http" transport using `Obot hosted` option
<img width="772" alt="Image" src="https://github.com/user-attachments/assets/a87120b9-b8e9-4eba-b7c3-d39021679df5" />

Notice that `tools` call times out with error -  `524` error  with response being a html page with title -
```
<title>main.acornlabs.com | 524: A timeout occurred</title> 
```

<img width="779" alt="Image" src="https://github.com/user-attachments/assets/f007375d-de8a-4210-8a70-fc1111273ad2" />

**Scenario 2**
PreReq:
Start mcpserver locally using `npm run start:streamableHttp` 
Have this server exposed using ngrok running on http port 3001

 1. Create an agent
2. Add `everything` server with "streamable http" transport using `Remote` option with URL set to `<ngrok url>/mcp`

`tools` call fails with `500 Internal Server Error` with error `failed to create MCP client: failed to create MCP stdio client: failed to decode mcp send message response: failed to read stream response: %!w(<nil>)`

**Expected behavior**
In both cases , we expect `tools` call to succeed .

I am able to configure mcp server in both these ways using Inspector successfully.

thedadams:
@sangee2004: Scenario 2 is fixed by https://github.com/obot-platform/obot/pull/3252. Not sure how you want to handle moving the issue since it doesn't fix scenario 1.

sangee2004:
@thedadams Once I validate scenario 2 mentioned in the issue , I will close this issue and track a new one to track scenario 1 mentioned in this issue.

sangee2004:
Issue mentioned in scenario2 is addressed when testing with `v0.0.0-dev+d791a568`.

1. Start mcpserver locally using `npm run start:streamableHttp`
2. Have this server exposed using ngrok running on http port 3001

3. Create an agent
4. Add everything server with "streamable http" transport using Remote option with URL set to 1<ngrok url>/mcp`

Mcpserver gets added successfully and we are able to list all the available tools.

sangee2004:
Scenario 1 mentioned in this issue is tracked in a separate issue - https://github.com/obot-platform/obot/issues/3254

PR #3252, opened 2025-06-23
chore: bump gptscript and nanobot to pickup streamable http init fix

thedadams:
https://github.com/obot-platform/obot/issues/3248


thedadams:
Merging because the failed test are caused by a test repo being updated, unrelated to these changes.

PR #3256, opened 2025-06-24
one catalog and access control groups

g-linville:
See the Loom videos in Slack for more information.


apiclient/types/mcpcatalog.go
```diff
@@ -3,14 +3,12 @@ package types
 type MCPCatalog struct {
 	Metadata
 	MCPCatalogManifest
-	IsReadOnly bool `json:"isReadOnly,omitempty"`
```
g-linville, line 6: Now that there is only one catalog, we don't need to distinguish between read-only and not-read-only.

pkg/api/handlers/mcp.go
```diff
@@ -643,12 +655,31 @@ func (m *MCPHandler) CreateServer(req api.Context) error {
...
+				return types.NewErrForbidden("user does not have access to MCP server catalog entry")
+			}
+		}
+
 		var catalogEntry v1.MCPServerCatalogEntry
 		if err := req.Get(&catalogEntry, input.CatalogEntryID); err != nil {
 			return err
 		}
 
-		manifest, err := serverManifestFromCatalogEntryManifest(catalogEntry.Spec.CommandManifest, input.MCPServerManifest)
```
g-linville, line 651: There was a bug here where we were ignoring the `URLManifest` and always using the `CommandManifest`. I fixed that.

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -248,142 +246,30 @@ func (h *Handler) readMCPCatalogDirectory(catalog string) ([]types.MCPServerCata
 	return entries, nil
 }
 
-// DeleteUnauthorizedMCPServers deletes all MCP servers that are no longer authorized to run.
-// This can happen when a user has launched an MCP server that they used to have permission for,
-// but their access to the catalog that the server came from has been revoked.
-func (h *Handler) DeleteUnauthorizedMCPServers(req router.Request, _ router.Response) error {
```
g-linville, line 254: I will reintroduce this functionality in a follow-up.

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -248,142 +246,30 @@ func (h *Handler) readMCPCatalogDirectory(catalog string) ([]types.MCPServerCata
...
+			SourceURLs:  sourceURLs,
 		},
 	}); err != nil {
 		return fmt.Errorf("failed to create default catalog: %w", err)
 	}
 
 	return nil
 }
-
-func (h *Handler) SetUpUserCatalogAuthorizations(req router.Request, _ router.Response) error {
```
g-linville, line 343: `UserCatalogAuthorizations` were an old type that we used to check authz logic that is no longer needed.

pkg/services/config.go
```diff
@@ -370,6 +375,35 @@ func New(ctx context.Context, config Config) (*Services, error) {
...
+			acr := obj.(*v1.AccessControlRule)
+			return acr.Spec.MCPServerCatalogEntryNames, nil
+		},
+		"server-names": func(obj any) ([]string, error) {
+			acr := obj.(*v1.AccessControlRule)
+			return acr.Spec.MCPServerNames, nil
+		},
+	}); err != nil {
+		return nil, err
+	}
```
g-linville: Shoutout to Mr. Thedadams for telling me about this.

pkg/api/authz/authz.go
```diff
@@ -60,11 +61,6 @@ var staticRules = map[string][]string{
 		"GET /api/templates",
 		"GET /api/tool-references",
 
-		"GET /api/mcp/catalog",
-		"GET /api/mcp/catalog/{id}",
-
-		"/api/mcp/{mcp_server_id}",
```
g-linville, line 66: These routes don't exist anymore.

pkg/api/authz/authz.go
```diff
@@ -99,8 +95,6 @@ var staticRules = map[string][]string{
 		"GET /api/all-mcp-catalogs/servers/{mcp_server_id}",
 
 		"GET /oauth/callback/{oauth_auth_request}",
-
-		"/mcp-connect/{mcp_server_id}",
```
g-linville, line 103: I don't think we want to allow this? I think this would allow users to share connection URLs with each other.

pkg/api/handlers/mcp.go
```diff
@@ -129,7 +129,7 @@ func (m *MCPHandler) ListServer(req api.Context) error {
 		fieldSelector = kclient.MatchingFields{
 			"spec.sharedWithinMCPCatalogName": catalogID,
 		}
-	} else {
+	} else if req.PathValue("project_id") != "" {
```
njhale, line 132: Do any of the new routes, types, or UI components use projects at all? Will gateway even retain the concept of a project after it splits from chat? If not, I think it would be cleaner to define new handlers instead of overloading/maintaining compatibility with the old ones. I think this would make it easier to iterate and grok changes too.

cjellick (approved):

njhale (approved):

pkg/api/authz/authz.go
thedadams: Yes, thank you for removing it.

pkg/api/handlers/accesscontrolrules.go
```diff
@@ -0,0 +1,122 @@
...
+	rule := v1.AccessControlRule{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: system.AccessControlRulePrefix,
+			Namespace:    req.Namespace(),
+		},
+		Spec: v1.AccessControlRuleSpec{
+			DisplayName:                manifest.DisplayName,
+			UserIDs:                    manifest.UserIDs,
+			MCPServerCatalogEntryNames: manifest.MCPServerCatalogEntryIDs,
+			MCPServerNames:             manifest.MCPServerIDs,
```
thedadams: nit: similar to our other types, can the `spec` here have a `manifest` field that is of type `types.AccessControlRuleManifest`? Then you would be able to:
```suggestion
			Manifest: manifest,
```

pkg/api/handlers/accesscontrolrules.go
```diff
@@ -0,0 +1,122 @@
...
+
+	var existing v1.AccessControlRule
+	if err := req.Get(&existing, req.PathValue("access_control_rule_id")); err != nil {
+		return fmt.Errorf("failed to get access control rule: %w", err)
+	}
+
+	existing.Spec.DisplayName = manifest.DisplayName
+	existing.Spec.UserIDs = manifest.UserIDs
+	existing.Spec.MCPServerCatalogEntryNames = manifest.MCPServerCatalogEntryIDs
+	existing.Spec.MCPServerNames = manifest.MCPServerIDs
```
thedadams: Same here:
```suggestion
	existing.Spec.Manifest = manifest
```

pkg/api/handlers/accesscontrolrules.go
```diff
@@ -0,0 +1,122 @@
...
+
+func convertAccessControlRule(rule v1.AccessControlRule) types.AccessControlRule {
+	return types.AccessControlRule{
+		Metadata: MetadataFrom(&rule),
+		AccessControlRuleManifest: types.AccessControlRuleManifest{
+			DisplayName:              rule.Spec.DisplayName,
+			UserIDs:                  rule.Spec.UserIDs,
+			MCPServerCatalogEntryIDs: rule.Spec.MCPServerCatalogEntryNames,
+			MCPServerIDs:             rule.Spec.MCPServerNames,
+		},
```
thedadams: Finally:
```suggestion
		AccessControlRuleManifest: rule.Spec.Manifest,
```

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -72,78 +72,38 @@ func (h *MCPCatalogHandler) Refresh(req api.Context) error {
...
+			u, err := url.Parse(urlStr)
+			if err != nil {
+				return types.NewErrBadRequest("invalid URL: %v", err)
+			}
 
-		if u.Scheme != "https" {
-			return types.NewErrBadRequest("only HTTPS URLs are supported")
+			if u.Scheme != "https" {
+				return types.NewErrBadRequest("only HTTPS URLs are supported")
+			}
```
thedadams, line 104: The default catalog in the Docker container will have a `./catalog` as the `urlStr`, which will fail this check. Since we only have one catalog, I could see a user adding multiple `SourceURLs` to it. However, they would only be able to do that if they first removed `./catalog`.

pkg/api/handlers/mcp.go
```diff
@@ -54,22 +54,25 @@ func (m *MCPHandler) GetCatalogEntry(req api.Context) error {
 		return err
 	}
 
+	if entry.Spec.MCPCatalogName != system.DefaultCatalog {
+		return types.NewErrNotFound("MCP catalog entry not found")
+	}
```
thedadams, line 59: Main, at least, will have a bunch of stuff that isn't in the default catalog. Are we OK with essentially abandoning all of that?

pkg/api/handlers/serverinstances.go
```diff
@@ -0,0 +1,120 @@
...
+	return &ServerInstancesHandler{
+		acrHelper: acrHelper,
+		serverURL: serverURL,
+	}
+}
+
+func (h *ServerInstancesHandler) ListServerInstances(req api.Context) error {
+	var instances v1.MCPServerInstanceList
+	if err := req.List(&instances, kclient.MatchingFields{
+		"spec.userID": req.User.GetUID(),
```
thedadams: Is there an instance where we would want an admin to be able to list all server instances?

pkg/api/handlers/serverinstances.go
```diff
@@ -0,0 +1,120 @@
...
+
+	return req.Write(convertMCPServerInstance(instance, h.serverURL))
+}
+
+func (h *ServerInstancesHandler) CreateServerInstance(req api.Context) error {
+	var input struct {
+		MCPServerID string `json:"mcpServerID"`
+	}
+	if err := req.Read(&input); err != nil {
+		return fmt.Errorf("failed to read server name: %w", err)
```
thedadams: Any time we fail to read input, we should return `types.NewErrBadRequest`.

pkg/api/handlers/serverinstances.go
```diff
@@ -0,0 +1,120 @@
...
+		},
+		Spec: v1.MCPServerInstanceSpec{
+			UserID:         req.User.GetUID(),
+			MCPServerName:  input.MCPServerID,
+			MCPCatalogName: server.Spec.SharedWithinMCPCatalogName,
+		},
+	}
+
+	if err := req.Create(&instance); err != nil {
+		return fmt.Errorf("failed to create MCP server instance: %w", err)
```
thedadams: Should we detect when the server already exists here and return the proper HTTP status code (400 or 409 come to mind).

pkg/api/handlers/serverinstances.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	return req.WriteCreated(convertMCPServerInstance(instance, h.serverURL))
+}
+
+func (h *ServerInstancesHandler) DeleteServerInstance(req api.Context) error {
+	var instance v1.MCPServerInstance
+	if err := req.Get(&instance, req.PathValue("mcp_server_instance_id")); err != nil {
+		return fmt.Errorf("failed to get MCP server instance: %w", err)
+	}
```
thedadams: Our other `DELETE` APIs don't return a 404 if the object you're deleting is not found. I think we can just call `req.Delete` instead of getting first.

pkg/controller/handlers/accesscontrolrule/helper.go
```diff
@@ -0,0 +1,109 @@
...
+import (
+	"fmt"
+
+	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
+	"github.com/obot-platform/obot/pkg/system"
+	gocache "k8s.io/client-go/tools/cache"
+)
+
+type Helper struct {
+	acrInformer gocache.SharedIndexInformer
```
thedadams: nit: looks like we always call `GetIndexer()` on this. Can this just be the indexer instead of the informer?

pkg/controller/handlers/accesscontrolrule/helper.go
```diff
@@ -0,0 +1,109 @@
...
+func (h *Helper) GetAccessControlRulesForUser(namespace, userID string) ([]v1.AccessControlRule, error) {
+	acrs, err := h.acrInformer.GetIndexer().ByIndex("user-ids", userID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get access control rules for user: %w", err)
+	}
+
+	result := make([]v1.AccessControlRule, 0, len(acrs))
+	for _, acr := range acrs {
+		res := acr.(*v1.AccessControlRule)
+		if res != nil && res.Namespace == namespace {
```
thedadams: Same comment here and the places below.

It's not possible for `res == nil` here. If `acr == nil`, then `arc.(*v1.AccessControlRule)` will panic. If this is something we want to protect against, then we can do:
```suggestion
		res, ok := acr.(*v1.AccessControlRule)
		if ok && res.Namespace == namespace {
```

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
thedadams: Ultimately, I'm fine with you reintroducing this functionality in a follow up PR. However, that will mean that we'll have a bunch of Kubernetes deployments in main that will have to be manually deleted.

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -248,142 +246,30 @@ func (h *Handler) readMCPCatalogDirectory(catalog string) ([]types.MCPServerCata
...
-		if len(existing.Spec.SourceURLs) > 0 && existing.Spec.SourceURLs[0] != h.defaultCatalogPath {
-			existing.Spec.SourceURLs = []string{h.defaultCatalogPath}
-			if err := c.Update(ctx, &existing); err != nil {
-				return fmt.Errorf("failed to update default catalog: %w", err)
-			}
-		}
-		return nil
+	sourceURLs := []string{}
+	if h.defaultCatalogPath != "" {
+		sourceURLs = []string{h.defaultCatalogPath}
```
thedadams: nit:
```suggestion
		sourceURLs = append(sourceURLs, h.defaultCatalogPath)
```

pkg/controller/handlers/mcpserver/mcpserver.go
```diff
@@ -0,0 +1,36 @@
...
+var log = logger.Package()
+
+// DeleteOrphans deletes non-shared MCPServer that have no MCPServerInstances at least one hour after creation.
+func DeleteOrphans(req router.Request, resp router.Response) error {
+	server := req.Object.(*v1.MCPServer)
+
+	if server.Spec.ThreadName != "" || server.Spec.SharedWithinMCPCatalogName != "" {
+		return nil
+	} else if time.Since(server.CreationTimestamp.Time) < time.Hour {
+		resp.RetryAfter(time.Hour)
```
thedadams: nit:
```suggestion
	} else if since := time.Since(server.CreationTimestamp.Time); since < time.Hour {
		resp.RetryAfter(time.Hour - since)
```

pkg/controller/routes.go
```diff
@@ -216,18 +217,18 @@ func (c *Controller) setupRoutes() error {
...
-	root.Type(&v1.MCPCatalog{}).HandlerFunc(mcpCatalog.DeleteUnauthorizedMCPServers)
-	root.Type(&v1.MCPCatalog{}).HandlerFunc(mcpCatalog.SetUpUserCatalogAuthorizations)
 
 	// MCPServerCatalogEntry
 	root.Type(&v1.MCPServerCatalogEntry{}).HandlerFunc(cleanup.Cleanup)
 
 	// MCPServer
 	root.Type(&v1.MCPServer{}).HandlerFunc(cleanup.Cleanup)
+	root.Type(&v1.MCPServer{}).HandlerFunc(mcpserver.DeleteOrphans)
+	root.Type(&v1.MCPServer{}).FinalizeFunc(v1.MCPServerFinalizer, credentialCleanup.RemoveMCPCredentialsForProject)
```
thedadams: The same finalizer can't be used on multiple handlers. What will happen is:
1. Before running the first finalizer function, we check that the finalizer is on the object. It is so run the first function.
2. The first function will run successfully.
3. The finalizer gets removed.
4. Before running the second function,  we check that the finalizer is on the object. It is not, so the second function doesn't get run.

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
g-linville: Why would we have orphaned deployments? I figured we would just have MCPServer objects that stuck around even after being no longer authorized, and once we reintroduce the controller handler for it, they would get deleted, and so would the k8s deployments.

thedadams: Is that true for existing non-default catalogs? If it is, then please ignore my comment.

pkg/api/handlers/mcpcatalogs.go
g-linville: Good catch

pkg/api/handlers/mcp.go
g-linville: I think so, but I'm asking Craig to make sure

pkg/api/handlers/serverinstances.go
g-linville: Hmm...probably. I could make this a query parameter (that we only honor for the admin). Would that be good?

pkg/api/handlers/serverinstances.go
g-linville: Thanks, I need to pay more attention to these things

pkg/api/handlers/serverinstances.go
thedadams: I think that's good.

pkg/api/handlers/serverinstances.go
g-linville: Yes, good call

pkg/api/handlers/serverinstances.go
g-linville: Thanks, fixing it here and for AccessControlRules.

pkg/controller/handlers/accesscontrolrule/helper.go
g-linville: Yep

pkg/controller/handlers/accesscontrolrule/helper.go
g-linville: Thanks. Yeah I do want to prevent panics. I don't know if we could get into this situation or not though anyway

pkg/controller/routes.go
g-linville: Oh, thanks. I will consolidate them into a single function.

pkg/api/handlers/accesscontrolrules.go
g-linville: Good call

pkg/api/handlers/accesscontrolrules.go

pkg/api/handlers/accesscontrolrules.go

pkg/api/handlers/mcpcatalogs.go
g-linville: Fixed

pkg/api/handlers/serverinstances.go
g-linville: Fixed

pkg/api/handlers/serverinstances.go
g-linville: Fixed (with a 409)

pkg/api/handlers/serverinstances.go
g-linville: Fixed

pkg/controller/handlers/accesscontrolrule/helper.go
g-linville: Fixed

pkg/controller/handlers/accesscontrolrule/helper.go
g-linville: Fixed

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
g-linville: Fixed

pkg/controller/handlers/mcpserver/mcpserver.go

pkg/controller/routes.go
g-linville: Fixed

pkg/api/handlers/serverinstances.go
g-linville: Fixed

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
g-linville: I can implement it in a way so that we detect servers that were created from no-longer-existing catalogs and delete them.

pkg/api/handlers/accesscontrolrules.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// List returns all access control rules (admin only).
+func (*AccessControlRuleHandler) List(req api.Context) error {
+	var list v1.AccessControlRuleList
+	if err := req.List(&list); err != nil {
+		return fmt.Errorf("failed to list access control rules: %w", err)
+	}
+
+	var items []types.AccessControlRule
```
thedadams: nit:
```suggestion
	items := make([]types.AccessControlRule, 0, len(list.Items))
```

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -72,78 +74,38 @@ func (h *MCPCatalogHandler) Refresh(req api.Context) error {
...
-		return types.NewErrBadRequest("display name is required")
-	}
-
+	// The only field that can be updated is the source URLs.
 	for _, urlStr := range manifest.SourceURLs {
-		u, err := url.Parse(urlStr)
-		if err != nil {
-			return types.NewErrBadRequest("invalid URL: %v", err)
-		}
+		if urlStr != "" && urlStr != h.defaultCatalogPath {
```
thedadams, line 96: nit: This fixes the specific issue I pointed out, but what if I want to do some local testing of catalogs?

One suggestion: make the check below `if u.Host != "" && u.Host != "localhost" && u.Scheme != "https" {`.

This is a nit because I am not sure if this is something we'll want to do, and the fix would be fairly straightforward.

pkg/api/handlers/serverinstances.go
```diff
@@ -0,0 +1,131 @@
...
+	} else {
+		err = req.List(&instances, kclient.MatchingFields{
+			"spec.userID": req.User.GetUID(),
+		})
+	}
+	if err != nil {
+		return err
+	}
+
+	var convertedInstances []types.MCPServerInstance
```
thedadams: nit:
```suggestion
	convertedInstances := make([]types.MCPServerInstance, 0, len(instances.Items))
```

thedadams (approved):

pkg/api/handlers/mcpcatalogs.go
g-linville: I think I will hold off on this for now, until we get direction that this is something that we want to do.

pkg/api/handlers/accesscontrolrules.go

pkg/api/handlers/serverinstances.go
g-linville: Fixed

g-linville:
I am going to hold off on merging this until the UI changes are ready.

thedadams (approved):

thedadams (approved):

PR #3257, opened 2025-06-24
Add resource explorer to chat threads

njhale:
Adds a new component to the chat-well that enables discovering resources exposed by project MCP servers, downloading resources directly, and saving resources to files in the chat thread's workspace.

https://github.com/user-attachments/assets/52915ffa-fb81-442b-b92d-dc0ecc1a9a4d

This PR also includes the following distinct changes:

- Remove the ability to interact with MCP server resources from the sidenav. This capability is being moved to the chat-well and will be thread scoped.
- Add server side validation to ensure that `MCPServer` resources are created with manifests that have a:
  - non-empty (display) name
  - (display) name that is unique to the thread they belong to (this makes it easier to add more than one "custom" MCPServer to a project)
  - specify a non-empty URL or command (but not both)
- Move the toggle to enable/disable chatbots for a project back into the sidenav and display the share URL there when chatbots are enabled.
- Revamp the mcp server config forms:
  - intercepts server side validation errors and displays human readable
  error messages on the forms
  - makes form rendering more consistent
  - enables creating required custom headers
  - enables setting the name for a custom MCP server on create
  
Note: I'll have a Loom up later this morning to walk through the changes here


pkg/api/handlers/mcp.go
```diff
@@ -1409,3 +1413,35 @@ func (m *MCPHandler) GetServerFromCatalogs(req api.Context) error {
...
+		}); err != nil {
+			return fmt.Errorf("failed to list existing servers: %w", err)
+		}
+
+		for _, s := range servers.Items {
+			if s.Name != server.Name {
+				return types.NewErrBadRequest("MCP server with name %q already exists", server.Spec.Manifest.Name)
+			}
+		}
+	}
```
thedadams: This logic is a little confusing to me. The containing function is called just before we create the `MCPServer`. So, at this point, `server.Name == ""`. Therefore, this logic will always pass.

It is not clear to me what the intention was, so I am just bringing it up so we can adjust.

njhale: My intent was to add this validation to `*MCPHandler.UpdateServer` as well, which apparently didn't make it to my remote branch 🙃 

In that case, `s.Name` will already be set by the time this is called (as the last step before `req.Update`). That's what this check was added for; we don't want to return an error if the server with the matching display name (`s.spec.Manifest.Name`) is the one we're updating.

I'm going to add that back _and_ I'm going to add a comment to clarify.

ui/user/src/lib/components/mcp/RemoteMcpForm.svelte
```diff
@@ -29,225 +24,191 @@
...
-{/if}
+{/snippet}
 
-{#if !custom && (config.headers ?? []).length > 0}
-	<div class="flex flex-col gap-1">
-		<h4 class="text-base font-semibold">Headers Configuration</h4>
-		{@render showConfigHeaders('default')}
-	</div>
-{/if}
+{#snippet keyValueSection(
```
njhale, line 75: This snippet could probably be factored out and made into a shared dependency of `HostedMcpForm.svelte` and `RemoteMcpForm.svelte`, but the snippet itself will still be ugly.

I couldn't really decompose these forms into nicely composable building blocks, which makes me feel like we need to revisit the UX to see if we can come up with a more uniform UX to reduce the number of variants it needs to satisfy.

pkg/api/handlers/mcp.go
thedadams: Keying off the display name seems off to me. Can we key off of `spec.mcpServerCatalogEntryName` instead?

ui/user/src/lib/components/mcp/RemoteMcpForm.svelte
ivyjeong13: For the ui/ux overhaul, I've changed the RemoteMcpForm & HostedMcpForm to be used on the IT admin server setup side so it doesn't get used in chat (chat has an MCPCredentials instead) so there may be some conflict here.

njhale: Ah, so does that mean we are dropping support for creating a new custom configuration from the "Chat UI" after all?

ivyjeong13: That was my initial impression, but I saw your thread on UI and tagged Craig just in case to make sure!

pkg/api/handlers/mcp.go
thedadams: Based on the Loom that Craig shared, I don't think we need this. Is the intention here just to help the user to not create servers with the same display name so they're not confused?

njhale: >  Is the intention here just to help the user to not create servers with the same display name so they're not confused?

yup, it was making it hard to differentiate between custom configurations since, in main, they all had the same name (`My Custom Server`). I figured we'd want to have some hard validation to ensure that within a given scope -- which in this PR is the project thread -- so we could do things like sort and search resources by the display name too.

pkg/api/handlers/mcp.go
```diff
@@ -1445,3 +1453,38 @@ func (m *MCPHandler) GetServerFromCatalogs(req api.Context) error {
...
+			"spec.threadName":    server.Spec.ThreadName,
+			"spec.manifest.name": server.Spec.Manifest.Name,
+		}); err != nil {
+			return fmt.Errorf("failed to list existing servers: %w", err)
+		}
+
+		for _, s := range servers.Items {
+			// On update, the server we're validating will be returned in the list. Ignore it and continue checking for duplicates.
+			// On create, s.Name should be empty at this point which will (correctly) cause this check to pass and fail validation.
+			if s.Name != server.Name {
```
thedadams, line 1483: I am not a fan of this, but I'm OK with it if it is a requirement.

njhale:
PR is now moot based on the shift in design. Closing out.

PR #3264, opened 2025-06-25
fix: user-ui breaking changes with mcp catalog structure

ivyjeong13:
* updates to user-ui based on backend structure breaking changes for mcp catalog entry/server

### DIFF
--- a/ui/user/src/lib/services/chat/types.ts
+++ b/ui/user/src/lib/services/chat/types.ts
@@ -350,13 +350,14 @@ export interface ProjectMCPList {
 	items: ProjectMCP[];
 }
 
-export interface ProjectMCP extends MCPServer {
+export interface ProjectMCP {
 	id: string;
 	catalogEntryID?: string;
 	configured?: boolean;
 	deleted?: boolean;
 	type: string;
 	connectURL?: string;
+	manifest: MCPServer;
 }
 
 export interface Credential {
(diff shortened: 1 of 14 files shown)


ui/user/src/lib/services/chat/types.ts
```diff
@@ -692,7 +690,5 @@ export interface MCPCatalogServer extends MCPCatalogEntryServerManifest {
 	createdAt: string;
 	updatedAt: string;
 	type: string;
-	metadata?: {
-		categories?: string;
-	};
+	manifest: MCPCatalogEntryServerManifest;
```
g-linville: I'm not sure that this is correct. Iirc, this type has the `fixedURL` and `hostname` fields, whereas an actual MCPServer (not a catalog entry), which is what I believe this type represents, only has `url`.

Here is the type definition for the MCPServerCatalogEntryManifest: https://github.com/obot-platform/obot/blob/9b5253fe5ab7852e87318e489091892634b89333/apiclient/types/mcpserver.go#L13

And for the MCPServerManifest: https://github.com/obot-platform/obot/blob/9b5253fe5ab7852e87318e489091892634b89333/apiclient/types/mcpserver.go#L47

ivyjeong13: :+1: Oh yes, I missed changing that to MCPCatalogServerManifest after realizing! Will change that!

g-linville (approved):

PR #3275, opened 2025-06-27
feat: admin IT / business user MCP servers UI/UX Part 2

ivyjeong13:
Updates UI/UX to the next set of changes for IT admin / business users setting up & using MCP servers


ui/user/src/lib/services/chat/operations.ts
```diff
@@ -1379,3 +1420,111 @@ export async function bootstrapLogin(token: string) {
...
+}: {
+	mcpServerID?: string;
+	mcpCatalogID?: string;
+}): Promise<MCPServerInstance> {
+	const response = (await doPost('/mcp-server-instances', {
+		mcpServerID,
+		mcpCatalogID
+	})) as MCPServerInstance;
+	return response;
+}
```
g-linville: The `mcpCatalogID` will be ignored here. We auto-populate that on the backend if the MCPServer pointed to by `mcpServerID` belongs to a catalog.

PR #3276, opened 2025-06-30
enhance: add controller handlers to delete unauthorized MCPServers and MCPServerInstances

g-linville:


pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -273,3 +278,114 @@ func (h *Handler) SetUpDefaultMCPCatalog(ctx context.Context, c client.Client) e
...
+	for _, server := range mcpServers.Items {
+		if server.Spec.ToolReferenceName != "" || server.Spec.ThreadName != "" || server.Spec.SharedWithinMCPCatalogName != "" {
+			// For legacy gptscript tools, project-scoped servers, and multi-user servers created by the admin, we don't need to check them.
+			continue
+		}
+
+		user, err := h.gatewayClient.UserByID(req.Ctx, server.Spec.UserID)
+		if err != nil {
+			return fmt.Errorf("failed to get user %s: %w", server.Spec.UserID, err)
+		}
```
njhale, line 310: Does the server get cleaned up by another controller if the user no longer exists? If not, should we clean it up here.

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -273,3 +278,114 @@ func (h *Handler) SetUpDefaultMCPCatalog(ctx context.Context, c client.Client) e
...
+	for _, instance := range mcpServerInstances.Items {
+		if instance.Spec.MCPCatalogName == "" {
+			// This instance points to a single-user server, so we don't need to worry about it.
+			continue
+		}
+
+		user, err := h.gatewayClient.UserByID(req.Ctx, instance.Spec.UserID)
+		if err != nil {
+			return fmt.Errorf("failed to get user %s: %w", instance.Spec.UserID, err)
+		}
```
njhale, line 370: same question here

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
g-linville: Good call. I need to update the UserDelete controller route to delete the servers that belong to the user. I'll take care of that shortly.

g-linville: This is fixed now

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
g-linville: the instances will get deleted when the servers get deleted because they have a DeleteRef to the server

njhale (approved):

thedadams (approved):

PR #3278, opened 2025-06-30
enhance: lock down server creation better

g-linville:
This makes it so that the user can only deploy servers from catalog entries that they have access to, and they cannot override any of the configuration fields on them (other than the URL, for remote servers with only a `hostname` configured).

### DIFF
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -598,7 +598,7 @@ func serverForActionWithCapabilities(req api.Context, gptClient *gptscript.GPTSc
 	return server, serverConfig, caps, err
 }
 
-func serverManifestFromCatalogEntryManifest(entry types.MCPServerCatalogEntryManifest, input types.MCPServerManifest) (types.MCPServerManifest, error) {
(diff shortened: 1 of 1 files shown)


pkg/api/handlers/mcp.go
```diff
@@ -730,8 +732,11 @@ func (m *MCPHandler) CreateServer(req api.Context) error {
 		server.Spec.Manifest = manifest
 		server.Spec.ToolReferenceName = catalogEntry.Spec.ToolReferenceName
 		server.Spec.UnsupportedTools = catalogEntry.Spec.UnsupportedTools
-	} else {
+	} else if req.UserIsAdmin() {
+		// If the user is an admin, they can create a server with a manifest that is not in the catalog.
 		server.Spec.Manifest = input.MCPServerManifest
+	} else {
+		return types.NewErrBadRequest("catalogEntryID is required")
```
njhale, line 739: nit: I wish we could reject this in the authorizers

njhale (approved):

thedadams (approved):

PR #3288, opened 2025-07-01
wip: acrs use auth provider groups

njhale:


njhale:
Recent changes are causing this to fail on rebase. I'll fix this up shortly, but figured I'd open it to review in the meantime.

ui/user/src/lib/components/admin/SearchUsersAndGroups.svelte
```diff
@@ -7,60 +7,100 @@
...
+		filteredUsers = [];
+		filteredGroups = [];
 	}
 
 	let { onAdd, filterIds }: Props = $props();
 
-	function getFilteredData(users?: OrgUser[]) {
-		if (!users) {
-			return [];
+	$effect(() => {});
```
ivyjeong13: can drop unused effect here!

njhale: thanks for catching that! seems like an errant tab complete snuck in there

pkg/gateway/types/group.go
```diff
@@ -0,0 +1,14 @@
...
+
+// Group represents a group that users can belong to in an auth provider.
+type Group struct {
+	ID                    string    `json:"id" gorm:"primaryKey;unique"`
+	AuthProviderName      string    `json:"authProviderName" gorm:"primaryKey;index:idx_group_auth_provider"`
+	AuthProviderNamespace string    `json:"authProviderNamespace" gorm:"primaryKey;index:idx_group_auth_provider"`
+	Name                  string    `json:"name"`
+	IconURL               *string   `json:"iconURL"`
+	CreatedAt             time.Time `json:"createdAt"`
+	UpdatedAt             time.Time `json:"updatedAt"`
```
njhale, line 13: does this info need to be encrypted at rest?

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -279,6 +281,48 @@ func (h *Handler) SetUpDefaultMCPCatalog(ctx context.Context, c client.Client) e
...
+	identities, err := h.gatewayClient.FindIdentitiesForUser(ctx, gatewayUser.ID)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get identities: %w", err)
+	}
+
+	var groupIDs []string
+	var authProviderName, authProviderNamespace string
+
+	if len(identities) > 0 {
+		// Use the first identity (there should only be one active at a time)
```
thedadams: I don't think this is quite right because of the way we do tokens. Do we loose something if we loop over the identities that get returned?

pkg/gateway/client/auth.go
```diff
@@ -32,10 +33,30 @@ func (u UserDecorator) AuthenticateRequest(req *http.Request) (*authenticator.Re
 		return nil, false, nil
 	}
 
+	// TODO(njhale): Can we do this in the proxy and pull out completed groups here?
```
thedadams: Can this TODO be removed?

pkg/gateway/client/auth.go
```diff
@@ -32,10 +33,30 @@ func (u UserDecorator) AuthenticateRequest(req *http.Request) (*authenticator.Re
...
+	var (
+		authProviderName      = firstValue(resp.User.GetExtra(), "auth_provider_name")
+		authProviderNamespace = firstValue(resp.User.GetExtra(), "auth_provider_namespace")
+		groupInfos            = auth.GroupInfosFromContext(req.Context())
+		authProviderGroups    = make([]types.Group, 0, len(groupInfos))
+	)
+	if authProviderName != "" && authProviderNamespace != "" {
+		for _, groupInfo := range auth.GroupInfosFromContext(req.Context()) {
```
thedadams:
```suggestion
		for _, groupInfo := range groupInfos {
```

pkg/gateway/client/identity.go
```diff
@@ -277,6 +277,18 @@ func (c *Client) ensureIdentity(ctx context.Context, tx *gorm.DB, id *types.Iden
...
+		for i := range id.AuthProviderGroups {
+			if err := tx.Save(&id.AuthProviderGroups[i]).Error; err != nil {
+				return nil, fmt.Errorf("failed to save group %s: %w", id.AuthProviderGroups[i].ID, err)
+			}
+		}
+
+		if err := tx.Model(id).Association("AuthProviderGroups").Replace(id.AuthProviderGroups); err != nil {
+			return nil, fmt.Errorf("failed to associate groups with identity: %w", err)
+		}
+	}
```
thedadams, line 288:
```suggestion
	for i := range id.AuthProviderGroups {
		if err := tx.Save(&id.AuthProviderGroups[i]).Error; err != nil {
			return nil, fmt.Errorf("failed to save group %s: %w", id.AuthProviderGroups[i].ID, err)
		}
	}

	if err := tx.Model(id).Association("AuthProviderGroups").Replace(id.AuthProviderGroups); err != nil {
		return nil, fmt.Errorf("failed to associate groups with identity: %w", err)
	}
```

pkg/gateway/client/user.go
```diff
@@ -41,12 +44,33 @@ func (c *Client) UserFromToken(ctx context.Context, token string) (*types.User,
...
+			return err
+		}
+
+		// Get the identity and associated groups
+		identity := new(types.Identity)
+		err := tx.Preload("AuthProviderGroups").Where(
+			"user_id = ? AND auth_provider_namespace = ? AND auth_provider_name = ?",
+			tkn.UserID, namespace, name,
+		).First(identity).Error
+
```
thedadams:
```suggestion
```

pkg/gateway/client/user.go
```diff
@@ -315,6 +339,91 @@ func (c *Client) fetchUserProfile(ctx context.Context, authProviderURL, accessTo
...
+	if authProviderURL != "" {
+		u, err := url.Parse(authProviderURL + "/obot-list-auth-groups")
+		if err == nil {
+			if nameFilter != "" {
+				q := u.Query()
+				q.Set("name", nameFilter)
+				u.RawQuery = q.Encode()
+			}
+
+			req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
```
StrongMonkey, line 352: Errors are not handled here

pkg/gateway/types/group.go
StrongMonkey: chatgpt told me that group data is not qualified as PII

pkg/gateway/types/identity.go
```diff
@@ -14,4 +14,6 @@ type Identity struct {
 	IconURL               string    `json:"iconURL"`
 	IconLastChecked       time.Time `json:"iconLastChecked"`
 	Encrypted             bool      `json:"encrypted"`
+
+	AuthProviderGroups []Group `json:"groups" gorm:"many2many:group_memberships;joinForeignKey:AuthProviderName,AuthProviderNamespace,HashedProviderUserID;joinReferences:ID,AuthProviderName,AuthProviderNamespace"`
```
StrongMonkey, line 18: are those gorm tags legit?

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
njhale: This was an AI flub. I thought I had removed it. It should just use the current active identity for the user.

pkg/gateway/client/auth.go
njhale: yes

pkg/gateway/types/identity.go
njhale: yes, it generates a reference table under the hood and lets you pull out the groups with the associated identities

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
thedadams: What I am saying is that there may not be one active identity for the user because of the way we do auth tokens. Can we just loop over the identities?

njhale: ah, I see. yeah, I'll make the change.

pkg/gateway/client/user.go
njhale: I wouldn't consider this error to be terminal. I'll still want it to return results from the cached groups in the database even when it can't make this request.

StrongMonkey: ok, probably better to log it. i think it is quite common that `/obot-list-auth-groups` will return error, since it is api that we manage to make request to external auth provider api.

StrongMonkey (approved):

PR #3296, opened 2025-07-01
enhance: add routes for admin to list server instances

g-linville:
This adds two new API routes that the admin can use to list all MCPServerInstances that have been created from a multi-user MCPServer or from a single-user/remote MCPServerCatalogEntry.


g-linville:
Converted this to draft because I intend to move some of the fields for the MCPServerInstance from spec to status. Some of them should just be populated by the controller rather than being done at the API layer.

g-linville:
This is ready for review now.

pkg/api/handlers/serverinstances.go
```diff
@@ -122,10 +136,49 @@ func (h *ServerInstancesHandler) DeleteServerInstance(req api.Context) error {
...
+		"spec.mcpServerCatalogEntryName": req.PathValue("entry_id"),
+		"spec.mcpCatalogName":            req.PathValue("catalog_id"),
+	}); err != nil {
+		return err
+	}
+
+	convertedInstances := make([]types.MCPServerInstance, 0, len(instances.Items))
+	for _, instance := range instances.Items {
+		convertedInstances = append(convertedInstances, convertMCPServerInstance(instance, h.serverURL))
+	}
```
njhale, line 160: nit: might be worth it to extract this as a helper method that has a variadic argument for instances

njhale (approved):

pkg/controller/handlers/mcpserverinstance/mcpserverinstance.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"github.com/obot-platform/nah/pkg/router"
+	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Migrate makes sure that all spec fields are set properly.
+func Migrate(req router.Request, _ router.Response) error {
+	instance := req.Object.(*v1.MCPServerInstance)
+
```
thedadams, line 12: Can we put an if-statement here that short-circuits if this instance doesn't need to be migrated?

thedadams (changes requested):

pkg/controller/handlers/mcpserverinstance/mcpserverinstance.go
g-linville: I couldn't do it at the top because it depends on what's in the `server` object, but I just added a couple short circuits farther down.

PR #3298, opened 2025-07-01
feat: auth-providers & bootstrap login to user-ui facing admin

ivyjeong13:
* be able to configure auth providers in user-ui
* be able to login with bootstrap user through /v2/admin
* used vibe coding to include allowing /v2/admin route be accessible for unauthorized user if bootstrapper is enabled to be able to log in


pkg/api/authz/ui.go
```diff
@@ -19,14 +25,32 @@ var uiResources = []string{
...
 	vars, match := a.uiResources.Match(req)
 	if !match {
 		return false
 	}
 	if vars("assistant") == "api" {
 		return false
 	}
+
+	// Allow unauthenticated users to access /v2/admin only if bootstrap is enabled
+	if req.URL.Path == "/v2/admin" {
```
ivyjeong13: Vibe coded so hoping to get a backend engineer to take a look and make sure this is aight 😄 
Was going for the /v2/admin route (but not the subroutes) could be accessible while bootstrap is enabled!

pkg/bootstrap/bootstrap.go
```diff
@@ -266,3 +266,8 @@ func (b *Bootstrap) bootstrapEnabled(ctx context.Context) (bool, error) {
 	}
 	return true, nil
 }
+
+// Exported version for use in authz
+func (b *Bootstrap) BootstrapEnabled(ctx context.Context) (bool, error) {
+	return b.bootstrapEnabled(ctx)
+}
```
g-linville: If the function is needed in another package, I would rather just export the existing one rather than defining a new one here.

pkg/api/authz/authz.go
```diff
@@ -119,15 +120,21 @@ type Authorizer struct {
 	apiResources *pathMatcher
 	uiResources  *pathMatcher
 	acrHelper    *accesscontrolrule.Helper
+	Bootstrapper interface {
+		BootstrapEnabled(ctx context.Context) (bool, error)
+	}
```
g-linville: This field doesn't need to be exported.

```suggestion
	bootstrapper interface {
		BootstrapEnabled(ctx context.Context) (bool, error)
	}
```

pkg/api/authz/ui.go
g-linville: this lgtm

g-linville (changes requested):
Just a couple suggestions related to the Go code.

ivyjeong13:
> Just a couple suggestions related to the Go code.

Appreciated @g-linville ! Making the updates!

pkg/api/authz/ui.go
```diff
@@ -19,14 +25,25 @@ var uiResources = []string{
...
 	vars, match := a.uiResources.Match(req)
 	if !match {
 		return false
 	}
 	if vars("assistant") == "api" {
 		return false
 	}
+
+	// Allow all users to access /v2/admin and /v2/admin/
+	if req.URL.Path == "/v2/admin" || req.URL.Path == "/v2/admin/" {
```
ivyjeong13, line 38: Changed this to be similar to admin-ui where a non-authenticated or authenticated non-admin user can visit and they get the appropriate 'you need to login as an admin' experience!

cjellick (approved):

PR #3305, opened 2025-07-02
server updates

g-linville:


pkg/api/handlers/serverinstances.go
```diff
@@ -182,3 +185,71 @@ func (h *ServerInstancesHandler) AdminListServerInstancesForServerInCatalog(req
...
+			}
+
+			if currentURL.Hostname() != entry.Spec.URLManifest.Hostname {
+				// Set the URL to empty. The user will have to go update it before they can use this server again.
+				server.Spec.Manifest.URL = ""
+			}
+		}
+	}
+
+	if err := req.Update(&server); err != nil {
```
g-linville, line 252: TODO: Shut down existing servers before we update here. Borrow the logic from `mcp.UpdateServer` to do this.

PR #3311, opened 2025-07-03
feat: add third-party MCP server OAuth

thedadams:
After the user authenticates with the MCP gateway, the gateway will check the third-party MCP server for OAuth, and redirect the user to authenticate if necessary. These credentials are stored locally and passed to the third-party MCP server for each client request.


pkg/controller/handlers/cleanup/credentials.go
```diff
@@ -185,3 +188,7 @@ func (c *Credentials) RemoveMCPCredentials(req router.Request, resp router.Respo
 
 	return nil
 }
+
+func (c *Credentials) RemoveMCPServerInstanceToken(req router.Request, _ router.Response) error {
```
g-linville: Is this unused? I can't seem to find a reference to it.

pkg/api/handlers/mcpgateway/oauth/authorize.go
```diff
@@ -232,7 +235,101 @@ func (h *handler) callback(req api.Context) error {
...
+			return nil
+		}
+	}
+
 	redirectWithAuthorizeResponse(req, oauthAppAuthRequest, code)
+
+	return nil
+}
+
+func (h *handler) oauthCallback(req api.Context) error {
```
g-linville, line 301: So this is the callback used for second level OAuth, while the above callback function is used for first level OAuth, right? Would be good to note that in some comments if I am right about this.

pkg/api/handlers/mcpgateway/sessionstore.go
```diff
@@ -0,0 +1,143 @@
...
+			return nil, false, fmt.Errorf("failed to decode session state: %w", err)
+		}
+
+		sess, err = nmcp.NewExistingServerSession(req.Context(), sessionState, s.handler)
+		if err != nil {
+			return nil, false, err
+		}
+
+		s.sessionCache.Store(sessionID, sess)
+		s.mcpSessionCache.Store(sessionID, mcpSess)
```
g-linville, line 96: I believe this call is redundant.

pkg/api/handlers/mcpgateway/oauth/token.go
```diff
@@ -28,12 +29,41 @@ func (h *handler) token(req api.Context) error {
...
+				Description: "client_id is required",
+			})
+		}
+
+		clientID, err = url.QueryUnescape(clientID)
+		if err != nil {
+			return types.NewErrBadRequest("%v", Error{
+				Code:        ErrInvalidRequest,
+				Description: "client_id is invalid",
+			})
```
g-linville, line 63: Does the ClientSecret need to be unescaped as well?

g-linville (commented):
just some comments and questions

pkg/controller/handlers/cleanup/credentials.go
thedadams: Yes, apparently I added a different handler for this. Removed.

pkg/api/handlers/mcpgateway/oauth/authorize.go
thedadams: Added.

pkg/api/handlers/mcpgateway/sessionstore.go
thedadams: Thanks. Removed the one above.

pkg/api/handlers/mcpgateway/oauth/token.go
thedadams: It does not. The issue is that I put a colon in the client ID, which gets escaped.

PR #3323, opened 2025-07-07
chore: refactor mcpserverinstances

g-linville:
This refactors MCPServerInstances so that they are used **only** for shared (multi-user) servers, and for nothing else.

So now, for single-user/remote servers, the user only has to create an MCPServer, and that will have the connect URL on it immediately. There is no need (and in fact, you aren't allowed) to create an MCPServerInstance for it.

### DIFF
--- a/pkg/api/authz/mcpid.go
+++ b/pkg/api/authz/mcpid.go
@@ -1,0 +1,35 @@
+package authz
+
+import (
+	"net/http"
+	"strings"
+
+	"github.com/obot-platform/nah/pkg/router"
+	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
+	"github.com/obot-platform/obot/pkg/system"
+	"k8s.io/apiserver/pkg/authentication/user"
+)
+
+func (a *Authorizer) checkMCPID(req *http.Request, resources *Resources, user user.Info) (bool, error) {
+	if resources.MCPID == "" {
+		return true, nil
+	}
+
+	if strings.HasPrefix(resources.MCPID, system.MCPServerInstancePrefix) {
+		var mcpServerInstance v1.MCPServerInstance
+		if err := a.storage.Get(req.Context(), router.Key(system.DefaultNamespace, resources.MCPID), &mcpServerInstance); err != nil {
+			return false, err
+		}
+
+		return mcpServerInstance.Spec.UserID == user.GetUID(), nil
+	}
+
+	var mcpServer v1.MCPServer
+	if err := a.storage.Get(req.Context(), router.Key(system.DefaultNamespace, resources.MCPID), &mcpServer); err != nil {
+		return false, err
+	}
+
+	// For servers, we only allow it if the user owns the server.
+	// Shared servers should be interacted with using MCPServerInstances.
+	return mcpServer.Spec.UserID == user.GetUID(), nil
+}
--- a/pkg/api/authz/resources.go
+++ b/pkg/api/authz/resources.go
@@ -14,11 +14,11 @@ var apiResources = []string{
 	"GET    /api/all-mcp-catalogs/servers/{mcpserver_id}/resources/{resource_uri}",
 	"GET    /api/all-mcp-catalogs/servers/{mcpserver_id}/prompts",
 	"GET    /api/all-mcp-catalogs/servers/{mcpserver_id}/prompts/{prompt_name}",
-	"GET    /oauth/callback/{oauth_request_id}/{mcp_server_instance_id}",
-	"GET    /oauth/mcp/callback/{oauth_request_id}/{mcp_server_instance_id}",
-	"GET    /mcp-connect/{mcp_server_instance_id}",
-	"POST   /mcp-connect/{mcp_server_instance_id}",
-	"DELETE /mcp-connect/{mcp_server_instance_id}",
+	"GET    /oauth/callback/{oauth_request_id}/{mcp_id}",
+	"GET    /oauth/mcp/callback/{oauth_request_id}/{mcp_id}",
+	"GET    /mcp-connect/{mcp_id}",
+	"POST   /mcp-connect/{mcp_id}",
+	"DELETE /mcp-connect/{mcp_id}",
 	"GET    /api/assistants",
 	"GET    /api/assistants/{assistant_id}",
 	"GET    /api/assistants/{assistant_id}/projects",
@@ -154,11 +154,13 @@ var apiResources = []string{
 	"GET    /api/mcp-server-instances/{mcp_server_instance_id}",
 	"POST   /api/mcp-server-instances",
 	"DELETE /api/mcp-server-instances/{mcp_server_instance_id}",
+	"DELETE /api/mcp-server-instances/{mcp_server_instance_id}/oauth",
 	"GET    /api/mcp-servers",
 	"GET    /api/mcp-servers/{mcpserver_id}",
 	"POST   /api/mcp-servers",
 	"PUT    /api/mcp-servers/{mcpserver_id}",
 	"DELETE /api/mcp-servers/{mcpserver_id}",
+	"DELETE /api/mcp-servers/{mcpserver_id}/oauth",
 	"POST   /api/mcp-servers/{mcpserver_id}/configure",
 	"POST   /api/mcp-servers/{mcpserver_id}/deconfigure",
 	"POST   /api/mcp-servers/{mcpserver_id}/reveal",
@@ -215,14 +217,16 @@ var apiResources = []string{
 }
 
 type Resources struct {
-	AssistantID            string
-	ProjectID              string
-	ThreadID               string
-	ThreadShareID          string
-	TemplateID             string
-	TaskID                 string
-	MCPServerID            string
-	MCPServerInstanceID    string
+	AssistantID         string
+	ProjectID           string
+	ThreadID            string
+	ThreadShareID       string
+	TemplateID          string
+	TaskID              string
+	MCPServerID         string
+	MCPServerInstanceID string
+	// MCPID is the ID of either an MCPServer or an MCPServerInstance. It is used for interaction with the MCP gateway.
+	MCPID                  string
 	RunID                  string
 	WorkflowID             string
 	PendingAuthorizationID string
@@ -253,6 +257,7 @@ func (a *Authorizer) evaluateResources(req *http.Request, vars GetVar, user user
 		WorkflowID:             vars("workflow_id"),
 		MCPServerID:            vars("mcpserver_id"),
 		MCPServerInstanceID:    vars("mcp_server_instance_id"),
+		MCPID:                  vars("mcp_id"), // this will be either a server ID or a server instance ID
 		PendingAuthorizationID: vars("pending_authorization_id"),
 		ThreadShareID:          vars("share_public_id"),
 		ToolID:                 vars("tool_id"),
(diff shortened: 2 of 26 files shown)


pkg/api/handlers/serverinstances.go
```diff
@@ -89,20 +89,16 @@ func (h *ServerInstancesHandler) CreateServerInstance(req api.Context) error {
 		if !hasAccess {
 			return types.NewErrNotFound("MCP server not found")
 		}
-	} else if server.Spec.UserID != req.User.GetUID() {
```
g-linville: ServerInstances are now supported only on shared servers.
For non-shared servers, users are expected to just use the ConnectURL that exists on them already.

pkg/controller/handlers/cleanup/credentials.go
```diff
@@ -142,6 +142,10 @@ func (c *Credentials) removeMCPCredentialsForProject(req router.Request, _ route
 func (c *Credentials) RemoveMCPCredentials(req router.Request, resp router.Response) error {
 	mcpServer := req.Object.(*v1.MCPServer)
 
+	if err := c.gatewayClient.DeleteMCPOAuthToken(req.Ctx, req.Object.GetName()); err != nil {
+		return err
+	}
```
g-linville, line 147: This is part of the finalizer for MCPServerInstances, and now that we can have OAuth tokens for MCPServers too, I added the cleanup for it here.

pkg/gateway/client/mcpoauthtoken.go
```diff
@@ -20,9 +20,9 @@ var mcpOAuthTokenGroupResource = schema.GroupResource{
 	Resource: "mcpoauthtokens",
 }
 
-func (c *Client) GetMCPOAuthToken(ctx context.Context, mcpServerInstanceName string) (*types.MCPOAuthToken, error) {
+func (c *Client) GetMCPOAuthToken(ctx context.Context, mcpID string) (*types.MCPOAuthToken, error) {
 	token := new(types.MCPOAuthToken)
-	err := c.db.WithContext(ctx).Where("mcp_server_instance = ?", mcpServerInstanceName).First(token).Error
+	err := c.db.WithContext(ctx).Where("mcp_id = ?", mcpID).First(token).Error
```
g-linville, line 25: TODO: make sure that "mcp_id" is indeed how GORM will name this field.

g-linville: I have confirmed that yes, it does.

pkg/api/authz/resources.go
```diff
@@ -154,11 +154,13 @@ var apiResources = []string{
 	"GET    /api/mcp-server-instances/{mcp_server_instance_id}",
 	"POST   /api/mcp-server-instances",
 	"DELETE /api/mcp-server-instances/{mcp_server_instance_id}",
+	"DELETE /api/mcp-server-instances/{mcp_server_instance_id}/oauth",
```
g-linville, line 157: This is an existing route that hadn't been added here. I assume we want to allow users to clear the oauth credential on their own MCPServerInstances (and MCPServers), right?

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -362,8 +362,7 @@ func (h *Handler) DeleteUnauthorizedMCPServerInstances(req router.Request, _ rou
 
 	// Iterate through each MCPServerInstance and make sure it is still allowed to exist.
 	for _, instance := range mcpServerInstances.Items {
-		if !instance.DeletionTimestamp.IsZero() {
-			// This server is already being deleted, so we don't have to delete it again.
+		if instance.Spec.MCPCatalogName != system.DefaultCatalog || !instance.DeletionTimestamp.IsZero() {
```
g-linville: This is future-proofing by skipping the loop iteration if the instance is outside the default catalog. This cannot happen currently, since everything is in the default catalog.

pkg/api/authz/resources.go
thedadams: Yes, thank you. I keep forgetting to add the routes here, because I think that we catch all paths with path variables. Not sure why I always think that way.

pkg/api/handlers/mcp.go
```diff
@@ -32,15 +33,19 @@ type MCPHandler struct {
...
 	mcpSessionManager *mcp.SessionManager
 	acrHelper         *accesscontrolrule.Helper
+	serverURL         string
+	gatewayClient     *gateway.Client
 }
 
 var envVarRegex = regexp.MustCompile(`\${([^}]+)}`)
 
-func NewMCPHandler(gptscript *gptscript.GPTScript, mcpLoader *mcp.SessionManager, acrHelper *accesscontrolrule.Helper) *MCPHandler {
+func NewMCPHandler(gptscript *gptscript.GPTScript, mcpLoader *mcp.SessionManager, acrHelper *accesscontrolrule.Helper, serverURL string, gatewayClient *gateway.Client) *MCPHandler {
```
thedadams: nit: I recently learned that the `api.Context` has its own gateway client, so we don't have to pass this here. Not a big deal since we do pass it in other places.

pkg/api/handlers/mcp.go
```diff
@@ -485,13 +490,11 @@ func (m *MCPHandler) GetPrompt(req api.Context) error {
 	})
 }
 
-func ServerFromMCPServerInstance(req api.Context, gptClient *gptscript.GPTScript) (v1.MCPServer, mcp.ServerConfig, error) {
+func ServerFromMCPServerInstance(req api.Context, gptClient *gptscript.GPTScript, instanceID string) (v1.MCPServer, mcp.ServerConfig, error) {
```
thedadams: Similarly, the `api.Context` has its own gptscript client.

pkg/api/authz/mcpid.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/obot-platform/obot/pkg/system"
+	"k8s.io/apiserver/pkg/authentication/user"
+)
+
+func (a *Authorizer) checkMCPID(req *http.Request, resources *Resources, user user.Info) (bool, error) {
+	if resources.MCPID == "" {
+		return true, nil
+	}
+
+	// MCP ID is the ID of either an MCPServer or an MCPServerInstance.
```
njhale: nit: seems like this comment should be with the field definition on the `Resources` struct

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
thedadams: Is there a reason why we would want to skip this logic if the instance is outside of the default catalog? I don't see any logic in the loop that would break in that case.

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -362,8 +362,7 @@ func (h *Handler) DeleteUnauthorizedMCPServerInstances(req router.Request, _ rou
...
 			continue
 		}
```
thedadams, line 365: I can only comment on this line because it's part of the diff, but I believe lines 369 through 381 can be removed now, correct?

pkg/gateway/client/mcpoauthtoken.go
```diff
@@ -231,5 +231,5 @@ func (c *Client) decryptMCPOAuthToken(ctx context.Context, token *types.MCPOAuth
 }
 
 func mcpOAuthTokenCtx(token *types.MCPOAuthToken) value.Context {
-	return value.DefaultContext(fmt.Sprintf("%s/%s", mcpOAuthTokenGroupResource.String(), token.MCPServerInstance))
+	return value.DefaultContext(fmt.Sprintf("%s/%s", mcpOAuthTokenGroupResource.String(), token.MCPID))
```
thedadams, line 234: This means that none of the existing tokens in main will work after this goes in. That's fine, but there may be a few issues to resolve once this goes in.

pkg/api/handlers/serverinstances.go
```diff
@@ -89,20 +89,16 @@ func (h *ServerInstancesHandler) CreateServerInstance(req api.Context) error {
...
+	} else {
```
thedadams: nit: this else block can be removed if the check on line 84 is negated.

pkg/api/handlers/mcpgateway/oauth/authorize.go
```diff
@@ -237,34 +238,46 @@ func (h *handler) callback(req api.Context) error {
...
+		mcpServer       v1.MCPServer
+		mcpServerConfig mcp.ServerConfig
+		err             error
+	)
+
+	if strings.HasPrefix(mcpID, system.MCPServerInstancePrefix) {
+		mcpServer, mcpServerConfig, err = handlers.ServerFromMCPServerInstance(req, h.gptClient, mcpID)
+	} else {
+		mcpServer, mcpServerConfig, err = handlers.ServerForActionWithID(req, h.gptClient, mcpID)
+	}
```
thedadams: nit: this logic is duplicated in a couple places. Does that make it worth adding another function for this?

thedadams (commented):
Great job! Only very minor comments.

pkg/gateway/client/mcpoauthtoken.go
g-linville: Yeah I got permission from Craig to just fully wipe Main and start over there. Though I don't think I have the permissions to do that on my own.

pkg/api/authz/mcpid.go
g-linville: Good call

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
g-linville: Well, only because the ACRs are not catalog scoped and we only apply them to the default catalog. Obviously it does not matter right now, but that's why I called this future-proofing, but I think I am convinced that there's not a point in doing this, so I'll just remove it.

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
g-linville: Yes you are correct, thanks

pkg/api/handlers/mcpgateway/oauth/authorize.go
g-linville: Tbh I would rather not make this a function, I kinda like how explicit it is about what we're doing. If we end up needing to do it more times, then I would be willing, but I don't think we will, as this kind of "is it an instance or a server?" question only needs to be answered when doing MCP gateway stuff.

pkg/api/authz/mcpid.go
g-linville: Fixed

pkg/api/handlers/mcp.go
g-linville: fixed

pkg/api/handlers/mcp.go
g-linville: fixed

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
g-linville: fixed

pkg/api/handlers/serverinstances.go
g-linville: fixed

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
thedadams: The access control rules not being catalog scoped was the part I was missing. Thanks for that clarification.

njhale (approved):

g-linville:
@thedadams @njhale re-requested review because I added migrations. I am not going to rebase to fix conflicts until after I have approvals, so that I don't wipe your review history and you can see the changes since you last approved.

njhale (approved):

PR #3327, opened 2025-07-07
fix: look for subject-based selectors when determining user access

thedadams:


pkg/controller/handlers/accesscontrolrule/helper.go
```diff
@@ -74,13 +74,18 @@ func (h *Helper) GetAccessControlRulesForMCPServerCatalogEntry(namespace, entryN
...
 
-	result := make([]v1.AccessControlRule, 0, len(acrs))
-	for _, acr := range acrs {
+	acrsForSubject, err := h.acrIndexer.ByIndex("subject-selectors", selector)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get access control rules for selector: %w", err)
+	}
+
+	result := make([]v1.AccessControlRule, 0, len(acrsForResource)+len(acrsForSubject))
+	for _, acr := range append(acrsForResource, acrsForSubject...) {
```
g-linville: I think we need to have this split into `GetAccessControlRulesForResourceSelector` and `GetAccessControlRulesForSubjectSelector` functions. In the function right below this one (`UserHasAccessToMCPServer`), we call this function to try to find ACRs that target all Resources (i.e. `selector` is `*`). This change would make this also return ACRs that grant access to all users, even if they do not also grant access to all resources.

thedadams: I understand what you're saying, and I will definitely separate them out, but the fix was to the "function right below." So that will have to change to incorporate these changes.

g-linville: I don't see how these changes could have fixed the function below. Here is the current pseudocode for the function below:

```
find all ACRs that grant access to all resources
for each of those ACRs:
  if the user is a subject on it, return true (access granted)

find all ACRs that grant access to the specific server in question
for each of those ACRs:
  if the user is a subject on it, return true (access granted)

return false (access denied)
```

With the change that you made, this is the pseudocode representation:

```
find all ACRs that grant access to all resources OR that allow all users, regardless of the resources configured on it
for each of those ACRs:
  if the user is a subject on it, return true (access granted)

// the rest is the same
```

This is a logic error. We could find an ACR that has literally zero servers on it but gives access to all users, and then return true as a result of that, granting a user access to an MCP server that they're not actually supposed to have access to. That's the issue here.

thedadams: OK. I will make the change and request your review again.

thedadams: Resolving because any continued conversation would probably need to happen in a different thread.

g-linville (approved):

PR #3332, opened 2025-07-07
enhance: add select ACR step in MCP server creation

ivyjeong13:
* adds 'Select Access Control Rule' as last optional step in MCP server creation workflow


ui/user/src/lib/actions/tooltip.svelte.ts
```diff
@@ -3,19 +3,35 @@ import popover from '$lib/actions/popover.svelte';
 interface TooltipOptions {
 	text: string;
 	disablePortal?: boolean;
-	maxWidthClass?: string;
+	classes?: string[];
+	placement?:
```
njhale: nit: is it possible to define a type for this and reuse it in `InfoTooltop.svelte`?

```
export type TooltipPlacement =
	| 'top'
	| 'top-start'
	| 'top-end'
	| 'bottom'
	| 'bottom-start'
	| 'bottom-end'
	| 'left'
	| 'left-start'
	| 'left-end'
	| 'right'
	| 'right-start'
	| 'right-end';
```

njhale (approved):

ui/user/src/lib/actions/tooltip.svelte.ts
ivyjeong13: Oh yes, definitely!

njhale: sweet! thanks

pkg/api/router/router.go
```diff
@@ -422,6 +422,7 @@ func Router(services *services.Services) (http.Handler, error) {
 
 	// MCPServerCatalogEntries (admin only, for single-user and remote MCP servers)
 	mux.HandleFunc("GET /api/mcp-catalogs/{catalog_id}/entries", mcpCatalogs.ListEntriesForCatalog)
+	mux.HandleFunc("GET /api/mcp-catalogs/{catalog_id}/entries/{entry_id}", mcpCatalogs.GetEntry)
```
ivyjeong13, line 425: Added this so I could get the single entry for the mcp-servers/c/{catalog_entry_id} route to edit a single catalog entry. Wanted to make sure this is aight!

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -133,6 +133,28 @@ func (h *MCPCatalogHandler) ListEntriesForCatalog(req api.Context) error {
...
+	if err := req.Get(&entry, entryName); err != nil {
+		return fmt.Errorf("failed to get entry: %w", err)
+	}
+
+	if entry.Spec.MCPCatalogName != catalogName {
+		return types.NewErrBadRequest("entry does not belong to catalog")
+	}
+
+	return req.Write(convertMCPServerCatalogEntry(entry))
+}
```
g-linville: This looks mostly good, just need to remove the unnecessary getting of the catalog object itself:


```suggestion
func (h *MCPCatalogHandler) GetEntry(req api.Context) error {
	catalogName := req.PathValue("catalog_id")
	entryName := req.PathValue("entry_id")

	var entry v1.MCPServerCatalogEntry
	if err := req.Get(&entry, entryName); err != nil {
		return fmt.Errorf("failed to get entry: %w", err)
	}

	if entry.Spec.MCPCatalogName != catalogName {
		return types.NewErrBadRequest("entry does not belong to catalog")
	}

	return req.Write(convertMCPServerCatalogEntry(entry))
}
```

ivyjeong13: Thanks Grant! : ) Updated!

Issue #3284, opened 2025-07-01
Usage statistics and audit logging of mcp servers


cjellick:
Arguably usage statistics and audit logging could be broken out into two separate features, but I am expecting that the data will come from the same source, so I am lumping them together.

We have to provide the ability to capture, display, and query calls being made to MCP servers through the gateway.

Here are the fields/values I think we should be capturing. Unless otherwise stated, expect that each can be search or filtered on in the UI (and thus needs to support efficent querying and indexing)
- Timestamp
- User
- MCP Server
- Client (if available) like VS Code vs Claude Desktop vs Obot Chat
- Client IP address
- MCP Server, like "firecrawl" or "PayPal"
- API call type, meaning the different types of calls available via the MCP spec - init, tool list, tool call, resource list, resource get, etc
- There is an "identifier" bit as well, but it differs per API call type. For example, for a tool call, the identifier would be the name of the tool, but for resource get it would be the resource URI. I'm not sure how you want to handle that.
- "Request Body" - again this is specific to the api call type. The main thing i am thining of here is that for a tool call it would be the parameters passed to the tool call. I'm ok if you have fields specific to each api type. I think these "body" type fields can just be stored and presented as json blobs.
- Response - yeah, the entire response body.
- There is also the category of "calls" that are initiated from the server, like elicitations and sampling.

Usage statistics are like live roll up metrics. Right now in my wireframes I have just "Monthly Tool Calls". We probably want to go a little further than that, but not by much. I think if we have an API that does the following, we'll have the flexibility we need in the API:
For an MCP server, for a give timespan ranging from down to a minute to up to the past 90 days, let me see:
- total number of tool calls
- total number of users who have used it
- the number of calls for each tool, organize by tool name
I'm open to other ideas.

sangee2004:
`Audit Logs` page is now available to admin users which shows Audit entries for MCP servers.

<img width="1119" height="741" alt="Image" src="https://github.com/user-attachments/assets/97718bac-bbdb-493b-811e-7148c3b88192" />

@thedadams @ivyjeong13 Issues found when testing this feature is tracked as separate issues:

https://github.com/obot-platform/obot/issues/3397 
https://github.com/obot-platform/obot/issues/3398
https://github.com/obot-platform/obot/issues/3399
https://github.com/obot-platform/obot/issues/3418
https://github.com/obot-platform/obot/issues/3431
https://github.com/obot-platform/obot/issues/3432

sangee2004:
[...]

PR #3340, opened 2025-07-08
feat: MCP auditing

thedadams:
This change adds auditing through the MCP gateway. As requests come in from the user's client, the messages and their responses are stored for querying. Similarly, messages from the server to the user's client as stored.

Two APIs are added for querying: the audit logs directly and some basic stats on the MCP server usage. A caller can use query parameters to control the data that is returned in these APIs.

Issue: https://github.com/obot-platform/obot/issues/3284


pkg/api/authz/resources.go
```diff
@@ -19,6 +19,7 @@ var apiResources = []string{
 	"GET    /mcp-connect/{mcp_server_instance_id}",
 	"POST   /mcp-connect/{mcp_server_instance_id}",
 	"DELETE /mcp-connect/{mcp_server_instance_id}",
+	"GET    /api/mcp-audit-logs/stats/{mcp_server_instance_id}",
```
StrongMonkey: will that change giving that we are only using mcp server instance for shared server? I don't have the full context, just saw your discussion with grant

thedadams: Yep. Good catch. I will update.

apiclient/types/mcpauditlog.go
```diff
@@ -0,0 +1,75 @@
+package types
+
+import "encoding/json"
+
+// MCPAuditLog represents an audit log entry for MCP API calls
+type MCPAuditLog struct {
+	ID                   uint            `json:"id"`
+	CreatedAt            Time            `json:"createdAt"`
+	UserID               string          `json:"userId"`
+	MCPID                string          `json:"mcpID"`
```
StrongMonkey, line 10:
```suggestion
	MCPID                string          `json:"mcpId"`
```

apiclient/types/mcpauditlog.go
```diff
@@ -0,0 +1,75 @@
...
+	Name    string `json:"name"`
+	Version string `json:"version"`
+}
+
+// MCPAuditLogList represents a list of MCP audit logs
+type MCPAuditLogList List[MCPAuditLog]
+
+// MCPUsageStats represents usage statistics for MCP servers
+type MCPUsageStats struct {
+	MCPID                string                 `json:"mcpID"`
```
StrongMonkey, line 45:
```suggestion
	MCPID                string                 `json:"mcpId"`
```

pkg/gateway/types/mcpauditlog.go
```diff
@@ -0,0 +1,133 @@
...
+
+	types2 "github.com/obot-platform/obot/apiclient/types"
+)
+
+// MCPAuditLog represents an audit log entry for MCP API calls
+type MCPAuditLog struct {
+	ID                   uint            `json:"id" gorm:"primaryKey"`
+	CreatedAt            time.Time       `json:"createdAt" gorm:"index"`
+	UserID               string          `json:"userId" gorm:"index"`
+	MCPID                string          `json:"mcpID" gorm:"index"`
```
StrongMonkey, line 15:
```suggestion
	MCPID                string          `json:"mcpId" gorm:"index"`
```

StrongMonkey (approved):
seeing a mixture of ID and Id in datastruture, i think we should all standardize it over `Id`

thedadams:
@StrongMonkey We've typically used `ID` in these situations. The reason there is `Id` is because of Claude Code.

I agree that I need to make them the same. Are you against using `ID` in this case?

StrongMonkey:
@thedadams nope, I am good with `ID` if that's the conversion we use.

PR #3344, opened 2025-07-08
fix: ensure Kubernetes objects are delete when MCP servers are removed

thedadams:


pkg/api/handlers/mcpgateway/handler.go
```diff
@@ -123,7 +123,6 @@ func (m *messageHandler) OnMessage(ctx context.Context, msg nmcp.Message) {
 		}
 	}()
 
-	m.serverConfig.Scope = msg.Session.ID()
```
thedadams, line 126: This was messing up the scope used to cleanup the server.

g-linville (approved):

PR #3361, opened 2025-07-10
feat: add k8s information to MCP server details

g-linville:


pkg/api/handlers/mcp.go
```diff
@@ -1577,3 +1578,116 @@ func (m *MCPHandler) ClearOAuthCredentials(req api.Context) error {
...
+		return err
+	}
+
+	defer logs.Close()
+
+	// Set up Server-Sent Events headers
+	req.ResponseWriter.Header().Set("Content-Type", "text/event-stream")
+	req.ResponseWriter.Header().Set("Cache-Control", "no-cache")
+	req.ResponseWriter.Header().Set("Connection", "keep-alive")
+	req.ResponseWriter.Header().Set("Access-Control-Allow-Origin", "*")
```
g-linville: Is this header necessary? AI added it, but it seems to me like we shouldn't need to do CORS stuff here.

pkg/mcp/loader.go
```diff
@@ -70,10 +75,16 @@ func NewSessionManager(ctx context.Context, defaultLoader *gmcp.Local, opts Opti
 		})); err != nil {
 			log.Warnf("failed to create MCP namespace, namespace must exist for MCP deployments to work: %v", err)
 		}
+
+		clientset, err = kubernetes.NewForConfig(config)
+		if err != nil {
+			return nil, err
+		}
```
g-linville, line 82: Sadly, the controller-runtime client that we typically use can't really interact with events, so we had to do a clientset here just for that.

pkg/api/handlers/mcp.go
thedadams: It is not necessary. I believe it gets added by the CORS middleware we use.

pkg/mcp/loader.go
thedadams: You meant logs, right? Just making sure I understand the issue here.

pkg/api/handlers/mcp.go
```diff
@@ -1577,3 +1578,124 @@ func (m *MCPHandler) ClearOAuthCredentials(req api.Context) error {
...
+	req.ResponseWriter.Header().Set("Access-Control-Allow-Headers", "Cache-Control")
+
+	flusher, ok := req.ResponseWriter.(http.Flusher)
+	if !ok {
+		return fmt.Errorf("response writer does not support flushing")
+	}
```
thedadams: Do we want to error here? Or continue without worrying about flushing?

pkg/api/handlers/mcp.go
```diff
@@ -1577,3 +1578,124 @@ func (m *MCPHandler) ClearOAuthCredentials(req api.Context) error {
...
+		buf := make([]byte, 1024)
+		for {
+			select {
+			case <-req.Context().Done():
+				return
+			default:
+				n, err := logs.Read(buf)
+				if n > 0 {
+					// Split by lines to send each line as separate event
+					lines := strings.Split(string(buf[:n]), "\n")
```
thedadams: Since you are reading 1024 bytes at a time, it is possible that this will break up long lines. Or reading could end in the middle of a line. Probably not a huge deal, but it would create weird line breaks for anything that is reading this.

I would suggest using a `bufio.Scanner` for reading instead because it will handle lines correctly and we won't have to worry about it here.

pkg/api/handlers/mcp.go
```diff
@@ -1577,3 +1578,124 @@ func (m *MCPHandler) ClearOAuthCredentials(req api.Context) error {
...
+			fmt.Fprintf(req.ResponseWriter, "event: ping\ndata: keep-alive\n\n")
+			flusher.Flush()
+		case logLine, ok := <-logChan:
+			if !ok {
+				fmt.Fprintf(req.ResponseWriter, "event: ended\ndata: Log stream ended\n\n")
+				flusher.Flush()
+				return nil
+			}
+			// Escape newlines in log data and send as SSE event
+			escapedLine := strings.ReplaceAll(logLine, "\n", "\\n")
```
thedadams: You already split on new lines when you passed the log line. Is this necessary in that case?

pkg/mcp/details.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	id := sessionID(serverConfig)
+
+	var deployment appsv1.Deployment
+	if err := sm.client.Get(ctx, client.ObjectKey{Name: id, Namespace: sm.mcpNamespace}, &deployment); err != nil {
+		return nil, fmt.Errorf("failed to get deployment %s: %w", id, err)
+	}
+
+	pods, err := sm.clientset.CoreV1().Pods(sm.mcpNamespace).List(ctx, metav1.ListOptions{
```
thedadams: nit: can we use the client here and only use the clientset for logs?

pkg/api/handlers/mcp.go
StrongMonkey: Yeah we already have cors here https://github.com/obot-platform/obot/commit/9de6d0d198e4514e5fa891c300820c142b6880d1#diff-104a11712684e520f0affa95e6a053faeb713306a10f69ee99b9729cc9dd1996R44

and in prod it should probably not doing * for `Access-Control-Allow-Origin`

pkg/mcp/details.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func (sm *SessionManager) StreamServerLogs(ctx context.Context, serverConfig ServerConfig) (io.ReadCloser, error) {
+	if !sm.KubernetesEnabled() {
+		return nil, fmt.Errorf("kubernetes is not enabled")
+	}
+
+	id := sessionID(serverConfig)
+
+	var deployment appsv1.Deployment
```
StrongMonkey, line 91: what if the mcp server is not backed a k8s deployment, like the remote ones? should it gracefully return without an error?

pkg/api/handlers/mcp.go
thedadams: We use `*` for that for the MCP gateway. We can talk about it if you'd like. Craig and I went down that rabbit hole a couple weeks ago.

StrongMonkey: Is it because of the MCP client that are making CORS request from a different domain? I think I also have * set for oauth proxy at some point but I thought it should not be needed.

pkg/mcp/loader.go
g-linville: Logs too I think, but I did mean events. I don't think I really checked for logs.

pkg/api/handlers/mcp.go
g-linville: Yeah you're right; I can just make it conditional

pkg/api/handlers/mcp.go
g-linville: Good call

pkg/api/handlers/mcp.go
g-linville: Nope, not necessary

pkg/mcp/details.go
g-linville: well there are no details to get if that is the case, so I have it return a 404. I figured that made the most sense.

pkg/mcp/details.go
g-linville: Fixed

pkg/api/handlers/mcp.go

pkg/api/handlers/mcp.go

pkg/api/handlers/mcp.go

pkg/mcp/details.go
g-linville: ohhh wait I just realized I misunderstood your question. I need to handle this situation too. I'll push a change to do that

g-linville: fixed

thedadams (approved):

pkg/mcp/loader.go
thedadams: It looks like you are only using this client set for logs. The client is being used to get/list events.

g-linville: Sorry, my memory was incorrect. Yes, I was able to use the main controller-runtime client to handle the events and only wound up needing the clientset for logs.

When I was originally working on events, AI told me I needed to use the clientset, but I figured out a way to not use it, but then when it came to the logs I realized I did need to use it there. Then when responding here I forgot the details of that and thought I was using it for events too, when I wasn't.

PR #3363, opened 2025-07-10
enhance: add audit logs ui/ux & Filters page

ivyjeong13:
* adds audit ui/ux
* adds Filters page (empty data, backend not yet impl. For demo presentation per Will/Sheng request)
* also does package dependency upgrades with inclusion of chart package (eslint being one of them, mostly for each key lint fixes)
* resolve lint errors from package upgrades of above 
* adds server/server instance route for server instances of a multi/single/remote (currently showing empty data as backend not yet impl.)

### DIFF
--- a/ui/user/src/lib/components/admin/AuditDetails.svelte
+++ b/ui/user/src/lib/components/admin/AuditDetails.svelte
@@ -1,5 +1,12 @@
 <script lang="ts">
-	import { ChevronsLeft, ChevronsRight, LoaderCircle } from 'lucide-svelte';
+	import {
+		ArrowDownWideNarrow,
+		ChevronDown,
+		ChevronsLeft,
+		ChevronsRight,
+		ListFilter,
+		LoaderCircle
+	} from 'lucide-svelte';
 	import Table from '../Table.svelte';
 	import { type Snippet } from 'svelte';
 	import {
@@ -17,6 +24,7 @@
 	import BarGraph from '../graph/BarGraph.svelte';
 	import { afterNavigate } from '$app/navigation';
 	import { twMerge } from 'tailwind-merge';
+	import Search from '../Search.svelte';
 
 	interface Props {
 		mcpId?: string;
@@ -42,6 +50,7 @@
 
 	let currentPage = $state(0);
 	let limit = $state(100);
+	let search = $state('');
 
 	let usersMap = $derived(new Map(users.map((u) => [u.id, u])));
 	let graphView = $state<'calls' | 'tools' | 'resources' | 'prompts'>('calls');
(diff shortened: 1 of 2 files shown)


.github/workflows/user-ui.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: "20.17.0"
+          node-version: "22.12.0"
```
thedadams, line 24: For my own understanding: why did we need to drop the version here?

ui/user/src/lib/components/admin/FilterForm.svelte
```diff
@@ -0,0 +1,244 @@
...
+				onclick={() => {
+					goto('/v2/admin/filters');
+				}}
+			>
+				Cancel
+			</button>
+			<button
+				class="button-primary text-sm disabled:opacity-75"
+				onclick={async () => {
+					// TODO:
```
g-linville, line 183: Is there something that should be added here?

ui/user/src/lib/components/admin/FilterForm.svelte
```diff
@@ -0,0 +1,244 @@
...
+	onAdd={async (mcpCatalogEntryIds, mcpServerIds) => {
+		filter.servers = [...(filter.servers ?? []), ...mcpCatalogEntryIds, ...mcpServerIds];
+	}}
+/>
+
+<Confirm
+	msg="Are you sure you want to delete this filter?"
+	show={deletingFilter}
+	onsuccess={async () => {
+		// TODO:
```
g-linville, line 241: Same here

ui/user/src/routes/v2/admin/mcp-servers/s/[id]/instance/[msi_id]/+page.svelte
```diff
@@ -0,0 +1,76 @@
+<script lang="ts">
```
Comment (g-linville) [inline ui/user/src/routes/v2/admin/mcp-servers/s/[id]/instance/[msi_id]/+page.svelte:1]: This file is fine for now with mock data, but I did want to mention that getting the k8s details for server instances is not a supported operation, since server instances basically just represent a connection to a shared server. So getting k8s details is limited to non-remote MCPServer IDs (with prefix `ms1`).

thedadams (approved):

ui/user/src/lib/components/admin/FilterForm.svelte
ivyjeong13: Not currently, this is something Will & Sheng wanna show for the demo but the backend isnt set up yet!

ui/user/src/lib/components/admin/FilterForm.svelte
ivyjeong13: Same thing!

ui/user/src/routes/v2/admin/mcp-servers/s/[id]/instance/[msi_id]/+page.svelte
Comment (ivyjeong13) [reply ui/user/src/routes/v2/admin/mcp-servers/s/[id]/instance/[msi_id]/+page.svelte:1]: Thanks Grant! Im thinking this will list the instance but have a note or section saying its displaying a shared server detail or something like that!

.github/workflows/user-ui.yml
ivyjeong13: I had to upgrade some packages and it ended up requiring bumping node up too!

Issue #3294, opened 2025-07-01
Takes 10-12 secs to show the `Overview` page/`Capabilities` page (with tools) for  MCP servers due to `tools` call taking a long time to respond. The same delay is also noted when connecting to MCP servers from clients.


sangee2004:
**Describe the bug**
Takes 10-12 secs to show the `Overview` page for Multi-User and Remote MCP servers.

Pre Req - Make sure to have a Multi-User and Remote MCP server configured.

**To Reproduce**
1. As admin ,  from `MCP Servers` page click on a Multi-User  or Remote server entry.

User is presented with `Overview` page which takes about 10 -12 secs to show the data relating to tools 

Revisiting the same  `MCP Servers` entry again , results in the same delay. This behavior is seen for  Remote MCP servers as well.

https://www.loom.com/share/fd47cb6955b74b98b5ec83a30fd83eb3?sid=0cf93a73-2fd0-44ce-b266-cb1cc1a16d9b

2. As any regular user ,  from `MCP Servers` page click  on any MCP Server from `Connected MCP Servers` (This can also be a Single-User type)

User is presented with `Capabilities` page which takes about 10 -12 secs to load the data relating to tools.

Revisiting the same  `MCP Servers` entry again  , `Capabilities` page gets rendered quickly since no api call to `tools` is made.

When I refresh the page and revisit  the same  `MCP Servers` entry again  , page would again take 10-12 seconds to show data relating to tools.

https://www.loom.com/share/19298131a1454137a37b309dea2e1c51?sid=260e1d3a-7aa9-4c5a-be7e-034ffca37507

3. The same delay is also seen when connecting to the MCP servers from VS code.

It takes about 12-15 secs when  connecting to an MCP server from VS code. User is taken to https://main.acornlabs.com/mcp-servers page after oauth workflow and then is stuck in this state for 12 - 15 secs before he is given the "You are signed in and now can close the window` 

**Expected behavior**
There should not be a delay when presenting the  tools in `Overview` page/`Capabilities` page except may be for the very first time.

thedadams:
Some notes for this issue:

1. When first deploying the MCP server (for multi or single user), we do expect a delay when listing tools because we need to deploy the Kubernetes objects and wait for them to be accessible. After the initial deploy, the list should be much faster.
2. There is an issue with GitMCP that could delay the listing of tools. I verified that it is an issue with them and not us. For example, Zapier always seems to be quick.

sangee2004:
[...]

PR #3364, opened 2025-07-10
chore: simplify tool loading

thedadams:
Instead of going back and forth from nanobot to gptscript back to nanobot, just use nanobot. In other words, don't use the Load method and call ListTools on the client itself.

Additionally, don't wait before making calls to determine if the pods are ready.

Issue: https://github.com/obot-platform/obot/issues/3294


pkg/mcp/loader.go
```diff
@@ -190,13 +190,15 @@ func (sm *SessionManager) ensureDeployment(ctx context.Context, server ServerCon
 				return gmcp.ServerConfig{}, fmt.Errorf("failed to parse MCP server URL: %w", err)
 			}
 
+			fmt.Printf("Checking that MCP server URL %s is not a localhost URL\n", server.URL)
```
StrongMonkey: is that for debug or do we want to log these actually?

pkg/mcp/loader.go
```diff
@@ -190,13 +190,15 @@ func (sm *SessionManager) ensureDeployment(ctx context.Context, server ServerCon
...
 			// LookupHost will properly detect IP addresses.
 			addrs, err := net.DefaultResolver.LookupHost(ctx, u.Hostname())
 			if err != nil {
 				return gmcp.ServerConfig{}, fmt.Errorf("failed to resolve MCP server URL hostname: %w", err)
 			}
 
 			for _, addr := range addrs {
+				fmt.Printf("Checking %s\n", addr)
```
StrongMonkey: same above

pkg/mcp/loader.go
thedadams: I was using these for debugging. I removed some, but forgot these. Thanks for catching.

g-linville (approved):
lgtm once the logs that Daishan pointed out are removed

pkg/mcp/loader.go
thedadams: Fixed.

pkg/mcp/loader.go
thedadams: Fixed.

StrongMonkey (approved):

PR #3371, opened 2025-07-11
enhance: add chat option for user connected mcp servers

ivyjeong13:
**temporary** For the upcoming demo, will be swapped out once Darren's chat is ready to switch over to.

* adds chat option for users' connected mcp servers 
Loom: https://www.loom.com/share/a2a37f8401aa4922a182089c03ac33ce

### DIFF
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -147,24 +147,13 @@ func (m *MCPHandler) ListServer(req api.Context) error {
 		// List servers scoped to the user.
 		fieldSelector = kclient.MatchingFields{
 			"spec.userID": req.User.GetUID(),
+			"spec.threadName": "",
 		}
 	}
 
 	var servers v1.MCPServerList
 	if err := req.List(&servers, fieldSelector); err != nil {
 		return nil
-	}
-
-	// Filter out servers that are tied to projects when listing user-deployed servers
-	if catalogID == "" && req.PathValue("project_id") == "" {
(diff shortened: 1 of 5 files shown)


pkg/api/handlers/mcp.go
```diff
@@ -155,6 +155,18 @@ func (m *MCPHandler) ListServer(req api.Context) error {
 		return nil
 	}
 
+	// Filter out servers that are tied to projects when listing user-deployed servers
+	if catalogID == "" && req.PathValue("project_id") == "" {
```
ivyjeong13: Added this to filter out project MCP servers I was seeing but mostly vibe coded my way through this so lemme know if this logic doesn't make sense!

pkg/api/handlers/mcp.go
```diff
@@ -147,6 +147,7 @@ func (m *MCPHandler) ListServer(req api.Context) error {
 		// List servers scoped to the user.
 		fieldSelector = kclient.MatchingFields{
 			"spec.userID": req.User.GetUID(),
+			"spec.threadName": "",
```
ivyjeong13, line 150: Updated! Thanks @thedadams!

thedadams (approved):

pkg/api/handlers/mcp.go
g-linville: Oh sorry, I think I was supposed to do this but forgot.

g-linville (approved):

PR #3383, opened 2025-07-14
chore: bump nanobot to 4948aef5dc74 for lots of fixes

thedadams:
This version of nanobot changes the context that is passed to message handlers. This change also creates a goroutine instead of relying on the context AfterFunc function for closing clients.


StrongMonkey (approved):

njhale (approved):
nit: based on the new pseudo-version here, the PR title/final commit name should be updated to `4948ae...`

PR #3386, opened 2025-07-14
chore: add pod events to k8s details for an mcp server

g-linville:


pkg/mcp/details.go
```diff
@@ -55,14 +68,16 @@ func (sm *SessionManager) GetServerDetails(ctx context.Context, serverConfig Ser
 	}
 
 	var mcpEvents []types.MCPServerEvent
-	for _, event := range events.Items {
+	for _, event := range append(deploymentEvents.Items, podEvents...) {
```
thedadams: Should we sort these based on creation time?

thedadams (approved):
Should

pkg/mcp/details.go
g-linville: Yes, good call, I will add that

g-linville: Fixed

PR #3392, opened 2025-07-14
fix: stop deleting non-shared MCP servers without MCP server instances

thedadams:
These servers no longer have MCP server instances, so we shouldn't be deleting them.

Issue: https://github.com/obot-platform/obot/issues/3393


drpebcak (approved):

StrongMonkey (approved):

cjellick (approved):

g-linville (approved):
Sorry, I totally missed this when I refactored how we do serverinstances!

Issue #3297, opened 2025-07-01
Managing MCP Server updates


cjellick:
Initial implementation is manual updates triggered by the admin, not automatic

Admin should be able to upgrade servers individually or in batches, but batches should just be the UI allowing the admin to select more than one server at a time and submit an upgrade for it

Any config change for the server constitutes an upgrade. In fact, we should use the word “Update” rather than “Upgrade”

If an update introduces a new required configuration parameter, then existing servers should be stopped and await the user to input the new required config.

The system must show an admin what servers have updates available. So, the flow should look like this:
1. On the MCP Server page, we should show an update is available.
2. When the admin clicks on that mcp server and navigates to the server instances page, the table of server instances should show what instances have an update available and the admin should be able to filter down to just those instances.
3. The admin should able to click to select or multi-select to pick serves to update. They they click an update button and the update process begins
4. For any individual server, the admin should be able to see a diff or comparison of the server’s current config vs what it would be updated to
5. The system should have a visual indicator for servers that are in the process of updating

There is no “rollback” mechanism. To fix a bad update, the admin needs to update the mcp server config again and trigger another update.

g-linville:
Some notes from me after the backend implementation of this:

- I didn't implement batches at the backend level. If we decide we need that, I can add it. The API route to update an individual server is `POST /api/mcp-servers/{mcp_server_id}/trigger-update` (no body). For batches, we could do something like `POST /api/update-mcp-servers` with a body format like `{"ids":["ms1asdfasdf", "ms1abcdefg"]}`. Let me know if this is something we want.
- For showing a diff of the configuration, while we detect configuration drift and set `needsUpdate: true` on the server, we don't generate any kind of diff in the backend that the UI can just fetch and display. We can do this if we want, but it is also possible for all of that logic to live in the UI, as all it needs to do is `GET /api/mcp-servers/{mcp_server_id}` and compare it to `GET /api/all-mcp-catalogs/entries/{entry_id}`, and that entry ID lives on the server object so you know what to compare it to.
- Right now, `needsUpdate: true` remains on the server if it has a URL that does not match the hostname on the catalog entry (for remote cases). `needsURL: true` is also set. If we want, I can make sure that `needsUpdate` is set to `false` instead, if that would make things easier for the UI. Just let me know.

sangee2004:
This feature is available when testing in latest build

Following is the workflow that was tested for single and remote MCP  servers.
[...]

PR #3395, opened 2025-07-15
feat: give admin the ability to update servers

g-linville:
for https://github.com/obot-platform/obot/issues/3297

### DIFF
--- a/pkg/controller/handlers/mcpserver/mcpserver.go
+++ b/pkg/controller/handlers/mcpserver/mcpserver.go
@@ -35,22 +35,15 @@ func DetectDrift(req router.Request, _ router.Response) error {
 		return err
 	}
 
-	server.Status.NeedsUpdate = drifted
-	return req.Client.Status().Update(req.Ctx, server)
+	if server.Status.NeedsUpdate != drifted {
+		server.Status.NeedsUpdate = drifted
+		return req.Client.Status().Update(req.Ctx, server)
+	}
+	return nil
 }
 
 func configurationHasDrifted(serverManifest types.MCPServerManifest, entryManifest types.MCPServerCatalogEntryManifest) (bool, error) {
-	drifted := !utils.SlicesEqualIgnoreOrder(serverManifest.Env, entryManifest.Env) ||
-		serverManifest.Command != entryManifest.Command ||
-		!slices.Equal(serverManifest.Args, entryManifest.Args) ||
-		!utils.SlicesEqualIgnoreOrder(serverManifest.Headers, entryManifest.Headers)
-
-	if drifted {
-		return true, nil
-	}
-
-	// Now check on the URL.
-
+	// First, check on the URL or hostname.
 	if entryManifest.FixedURL != "" && serverManifest.URL != entryManifest.FixedURL {
 		return true, nil
 	}
@@ -67,5 +60,11 @@ func configurationHasDrifted(serverManifest types.MCPServerManifest, entryManife
 		}
 	}
 
-	return false, nil
+	// Check the rest of the fields to see if anything has changed.
+	drifted := serverManifest.Command != entryManifest.Command ||
+		!slices.Equal(serverManifest.Args, entryManifest.Args) ||
+		!utils.SlicesEqualIgnoreOrder(serverManifest.Env, entryManifest.Env) ||
+		!utils.SlicesEqualIgnoreOrder(serverManifest.Headers, entryManifest.Headers)
+
+	return drifted, nil
 }


pkg/api/handlers/mcp.go
```diff
@@ -1718,3 +1725,140 @@ func (m *MCPHandler) StreamServerLogs(req api.Context) error {
...
+	} else {
+		server.Spec.Manifest.Metadata = entry.Spec.URLManifest.Metadata
+		server.Spec.Manifest.Name = entry.Spec.URLManifest.Name
+		server.Spec.Manifest.Description = entry.Spec.URLManifest.Description
+		server.Spec.Manifest.Icon = entry.Spec.URLManifest.Icon
+		server.Spec.Manifest.Headers = entry.Spec.URLManifest.Headers
+		server.Spec.Manifest.Env = entry.Spec.URLManifest.Env
+
+		server.Spec.Manifest.Command = ""
+		server.Spec.Manifest.Args = nil
```
g-linville, line 1830: TODO: deduplicate some of this

apiclient/types/mcpserver.go
```diff
@@ -84,3 +90,71 @@ type MCPServerTool struct {
 	Enabled     bool              `json:"enabled"`
 	Unsupported bool              `json:"unsupported,omitempty"`
 }
+
+// EqualMCPHeaders compares two slices of MCPHeader regardless of order
+func EqualMCPHeaders(a, b []MCPHeader) bool {
```
g-linville: If this function (and the ones below it) should be in a different package, let me know. To me, it seemed most logical to put them here.

pkg/api/handlers/mcp.go
```diff
@@ -559,6 +559,10 @@ func ServerForActionWithID(req api.Context, tokenService *jwt.TokenService, id s
 		return server, mcp.ServerConfig{}, err
 	}
 
+	if server.Spec.NeedsURL {
+		return server, mcp.ServerConfig{}, types.NewErrBadRequest("mcp server %s needs to update its URL", server.Name)
+	}
```
g-linville, line 568: I wasn't sure where all this check should go. We don't want to attempt to run a server if it is in a `needsURL` state. This is one of the places that I put it.

pkg/api/handlers/mcp.go
```diff
@@ -1718,3 +1725,139 @@ func (m *MCPHandler) StreamServerLogs(req api.Context) error {
...
+
+func (m *MCPHandler) UpdateURL(req api.Context) error {
+	var mcpServer v1.MCPServer
+	if err := req.Get(&mcpServer, req.PathValue("mcp_server_id")); err != nil {
+		return fmt.Errorf("failed to get server: %w", err)
+	}
+
+	if mcpServer.Spec.SharedWithinMCPCatalogName != "" {
+		return types.NewErrBadRequest("cannot update the URL for a multi-user MCP server; use the UpdateServer endpoint instead")
+	}
```
g-linville, line 1741: For now, multi-user servers still aren't created from the catalog.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -184,8 +188,10 @@ func (h *MCPCatalogHandler) CreateEntry(req api.Context) error {
 	}
 
 	if hasCommand {
+		manifest.Headers = nil
 		entry.Spec.CommandManifest = manifest
 	} else if hasURL {
+		manifest.Args = nil
```
g-linville, line 194: I noticed a bug where remote catalog entries manually created from the UI would have `args: [""]` rather than an empty array, so this accounts for that by setting it to nil.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -278,6 +286,51 @@ func (h *MCPCatalogHandler) DeleteEntry(req api.Context) error {
...
+	}); err != nil {
+		return fmt.Errorf("failed to list servers: %w", err)
+	}
+
+	var items []types.MCPServer
+	for _, server := range list.Items {
+		var credCtx string
+		if server.Spec.SharedWithinMCPCatalogName != "" {
+			credCtx = fmt.Sprintf("%s-%s", server.Spec.SharedWithinMCPCatalogName, server.Name)
+		} else {
```
g-linville, line 319: For now, there shouldn't be any shared servers that were created from catalog entries, but I put this here for the future.

pkg/api/handlers/mcpgateway/handler.go
```diff
@@ -83,6 +83,10 @@ func (h *Handler) StreamableHTTP(req api.Context) error {
 		return fmt.Errorf("failed to get mcp server config: %w", err)
 	}
 
+	if mcpServer.Spec.NeedsURL {
+		return fmt.Errorf("mcp server %s needs to update its URL", mcpServer.Name)
+	}
```
g-linville: This is another place that I put this check. Might as well skip setting up all of the gateway stuff here if we know the server isn't ready to be used.

pkg/controller/routes.go
```diff
@@ -229,7 +229,7 @@ func (c *Controller) setupRoutes() error {
 
 	// MCPServer
 	root.Type(&v1.MCPServer{}).HandlerFunc(cleanup.Cleanup)
-	root.Type(&v1.MCPServer{}).HandlerFunc(mcpserver.DeleteOrphans)
```
g-linville: This handler is being deleted in another PR, I replaced it here but it doesn't really matter. There will be a merge conflict either way.

pkg/gateway/types/activity.go
```diff
@@ -1,3 +1,4 @@
+//nolint:revive
```
g-linville, line 1: The `revive` part of golangci-lint really doesn't like `types` as a package name.

pkg/mcp/details.go
```diff
@@ -25,6 +26,10 @@ func (sm *SessionManager) GetServerDetails(ctx context.Context, serverConfig Ser
 
 	var deployment appsv1.Deployment
 	if err := sm.client.Get(ctx, client.ObjectKey{Name: id, Namespace: sm.mcpNamespace}, &deployment); err != nil {
+		if apierrors.IsNotFound(err) {
+			return types.MCPServerDetails{}, fmt.Errorf("mcp server %s is not running", id)
+		}
```
g-linville, line 32: This and the other change in this file are unrelated to the main goal of this PR; I just figured we might as well handle this error better.

pkg/mcp/loader.go
```diff
@@ -198,6 +198,10 @@ func (sm *SessionManager) ensureDeployment(ctx context.Context, server ServerCon
 	}
 
 	if server.Command == "" || !sm.KubernetesEnabled() {
+		if server.URL == "" {
+			return gmcp.ServerConfig{}, fmt.Errorf("MCP server %s needs to update its URL", serverName)
+		}
```
g-linville: We don't have access to the full `MCPServer` object here to check the `.Spec.NeedsURL` field, so I just check for the empty URL, since we set it to empty when there is a URL-hostname conflict.

apiclient/types/mcpserver.go
```diff
@@ -84,3 +90,71 @@ type MCPServerTool struct {
...
+	if len(a) != len(b) {
+		return false
+	}
+
+	// Create a map to track which elements in b have been matched
+	matched := make([]bool, len(b))
```
thedadams: It's more efficient to make this a `map[MCPHeader]struct{}` and only loop through each array once.

apiclient/types/mcpserver.go
```diff
@@ -84,3 +90,71 @@ type MCPServerTool struct {
...
+}
+
+// EqualMCPEnvs compares two slices of MCPEnv regardless of order
+func EqualMCPEnvs(a, b []MCPEnv) bool {
+	if len(a) != len(b) {
+		return false
+	}
+
+	// Create a map to track which elements in b have been matched
+	matched := make([]bool, len(b))
```
thedadams: Same comment.

apiclient/types/mcpserver.go
```diff
@@ -84,3 +90,71 @@ type MCPServerTool struct {
...
+		a.Description == b.Description &&
+		a.Key == b.Key &&
+		a.Sensitive == b.Sensitive &&
+		a.Required == b.Required
+}
+
+// equalMCPEnv compares two MCPEnv structs for equality
+func equalMCPEnv(a, b MCPEnv) bool {
+	return equalMCPHeader(a.MCPHeader, b.MCPHeader) && a.File == b.File
+}
```
thedadams: You can just do `==` to compare these: https://go.dev/play/p/oipQx0NS6WZ

pkg/mcp/loader.go
```diff
@@ -198,6 +198,10 @@ func (sm *SessionManager) ensureDeployment(ctx context.Context, server ServerCon
 	}
 
 	if server.Command == "" || !sm.KubernetesEnabled() {
+		if server.URL == "" {
```
thedadams: It looks like it's possible that `server.Command != ""` here, which would mean that there are instances where `server.URL == ""` is valid. So should this be
```suggestion
		if server.Command == "" && server.URL == "" {
```

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
```diff
@@ -0,0 +1,22 @@
+package mcpservercatalogentry
+
+import (
+	"github.com/obot-platform/nah/pkg/router"
+	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
+)
+
+func RemoveArgsOnRemoteEntries(req router.Request, _ router.Response) error {
```
thedadams, line 8: Do we need a controller handler for this? Does correcting the bug when we apply fix the issue?

pkg/api/handlers/mcp.go
```diff
@@ -1718,3 +1725,139 @@ func (m *MCPHandler) StreamServerLogs(req api.Context) error {
...
+	// Shutdown the server, even if there is no credential
+	if err := m.removeMCPServer(req.Context(), server, server.Spec.UserID, cred.Env); err != nil {
+		return err
+	}
+
+	if err := req.Update(&server); err != nil {
+		return err
+	}
+
+	return req.Write(nil)
```
```suggestion
	return nil
```

pkg/api/handlers/mcpgateway/handler.go
thedadams: I like the idea of keeping this check in one area. You added it to `handlers.ServerForActionWithID`. If you also added it to `handlers.ServerFromMCPServerInstance`, then this check would be centralized to that file and we wouldn't need it here.

pkg/controller/handlers/mcpserver/mcpserver.go
```diff
@@ -34,3 +37,63 @@ func DeleteOrphans(req router.Request, resp router.Response) error {
 
 	return nil
 }
+
+func UpdateStatus(req router.Request, _ router.Response) error {
```
thedadams: nit:
```suggestion
func DetectDrift(req router.Request, _ router.Response) error {
```

apiclient/types/mcpserver.go
g-linville: Thanks

pkg/mcp/loader.go
g-linville: Great catch, thank you!!

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
g-linville: Sadly, we do need it. The bug happens when you create one manually. So we never end up reapplying, and have to fix all of the existing ones with this handler.

The only reason it matters is because, if the `args: [""]` remains on remote catalog entries, then every server created from them will always have `needsUpdate: true` since the args are different between the two, even though they don't matter.

pkg/api/handlers/mcpgateway/handler.go
g-linville: As long as you are confident that that will suffice, then that sounds good to me.

apiclient/types/mcpserver.go
g-linville: Discussed in Slack and fixed this with a generic solution

apiclient/types/mcpserver.go
g-linville: Fixed

apiclient/types/mcpserver.go
g-linville: Fixed

pkg/mcp/loader.go

pkg/api/handlers/mcp.go

pkg/api/handlers/mcpgateway/handler.go
g-linville: Fixed

pkg/controller/handlers/mcpserver/mcpserver.go

g-linville:
Holding off on rebasing to fix conflicts until I have approvals, so that I don't destroy the review history.

g-linville:
Nvm, @thedad

pkg/controller/handlers/mcpserver/mcpserver.go
```diff
@@ -0,0 +1,71 @@
...
+		drifted, err = configurationHasDrifted(server.Spec.Manifest, entry.Spec.CommandManifest)
+	} else {
+		drifted, err = configurationHasDrifted(server.Spec.Manifest, entry.Spec.URLManifest)
+	}
+	if err != nil {
+		return err
+	}
+
+	server.Status.NeedsUpdate = drifted
+	return req.Client.Status().Update(req.Ctx, server)
```
thedadams: Two comment here:
1. I believe that the Update call here is unnecessary because the status of the object will get persisted after all handlers are run.
2. If you do want to make this Update call, then you should only do it if it is necessary.
```suggestion
	if server.Status.NeedsUpdate != drifted {
		server.Status.NeedsUpdate = drifted
		return req.Client.Status().Update(req.Ctx, server)
	}
	return nil
```

pkg/controller/handlers/mcpserver/mcpserver.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func configurationHasDrifted(serverManifest types.MCPServerManifest, entryManifest types.MCPServerCatalogEntryManifest) (bool, error) {
+	drifted := !utils.SlicesEqualIgnoreOrder(serverManifest.Env, entryManifest.Env) ||
+		serverManifest.Command != entryManifest.Command ||
+		!slices.Equal(serverManifest.Args, entryManifest.Args) ||
+		!utils.SlicesEqualIgnoreOrder(serverManifest.Headers, entryManifest.Headers)
```
thedadams: nit: do the easy thing first. In fact, unless there is a reason not to, I say we do everything below this before doing these harder things.
```suggestion
	drifted := serverManifest.Command != entryManifest.Command ||
		!slices.Equal(serverManifest.Args, entryManifest.Args) ||
		!utils.SlicesEqualIgnoreOrder(serverManifest.Env, entryManifest.Env) ||
		!utils.SlicesEqualIgnoreOrder(serverManifest.Headers, entryManifest.Headers)
```

pkg/controller/handlers/mcpserver/mcpserver.go
g-linville: Fixed (decided to use your suggested code)

pkg/controller/handlers/mcpserver/mcpserver.go
g-linville: Fixed (moved this to the bottom of the function and rearranged the checks as you suggested)

pkg/api/router/router.go
```diff
@@ -410,6 +410,8 @@ func Router(services *services.Services) (http.Handler, error) {
 	mux.HandleFunc("GET /api/mcp-servers/{mcp_server_id}/resources/{resource_uri}", mcp.ReadResource)
 	mux.HandleFunc("GET /api/mcp-servers/{mcp_server_id}/prompts", mcp.GetPrompts)
 	mux.HandleFunc("GET /api/mcp-servers/{mcp_server_id}/prompts/{prompt_name}", mcp.GetPrompt)
+	mux.HandleFunc("POST /api/mcp-servers/{mcp_server_id}/update-url", mcp.UpdateURL)
+	mux.HandleFunc("POST /api/mcp-servers/{mcp_server_id}/trigger-update", mcp.TriggerUpdate)
```
njhale, line 414: Is the intent for the trigger-update endpoint to be admin only?

g-linville: Yes

Issue #3307, opened 2025-07-02
Catalog - source URLs added after an invalid source URL do not get processed.


sangee2004:
**Describe the bug**
Catalog - source URLs added after an invalid source URL do not get processed.

**To Reproduce**
1. As admin, add MCP servers using `Add servers from Git` and provide an invalid Source URL ( `https://github.com/obot-platform/obot` in my case )
2. Add MCP servers again using valid Source URL.

Notice that valid Source URLs added in step 2 are not getting processed.

I see errors relating to invalid Source URL from step1 , after adding  MCP servers using`valid Source URL` 
```
[server] time="2025-07-02T12:48:43-07:00" level=error msg="error syncing 'default/default': handler obot-controller obot.obot.ai/v1, Kind=MCPCatalog: [routes.go:219] failed to read catalog https://github.com/obot-platform/obot : failed to read GitHub catalog https://github.com/obot-platform/obot : failed to list repository contents: 404 Not Found {"message":"Not Found","documentation_url":"https://docs.github.com/rest/git/trees#get-a-tree","status":"404"}, requeuing"
[server] time="2025-07-02T12:48:43-07:00" level=error msg="error syncing 'default/default': handler obot-controller obot.obot.ai/v1, Kind=MCPCatalog: [routes.go:219] failed to read catalog https://github.com/obot-platform/obot : failed to read GitHub catalog https://github.com/obot-platform/obot : failed to list repository contents: 404 Not Found {"message":"Not Found","documentation_url":"https://docs.github.com/rest/git/trees#get-a-tree","status":"404"}, requeuing"
```
**Expected behavior**
Should be able to process Source URLS that gets added after an  invalid source URL.

g-linville:
I just merged the fix for this.

sangee2004:
This issue is addressed when testing with latest build from `main`

1. As admin, add MCP servers using Add servers from Git and provide an invalid Source URL ( https://github.com/obot-platform/obot in my case )
2. Add MCP servers again using valid Source URL.

After couple of `refresh` attempts , I am able to see catalog entries from Source URL added in step2 get synced to the catalog as expected.

Issue #3306, opened 2025-07-02
Catalog - No indication provided to the user when adding catalog with invalid source url fails.


sangee2004:
**Describe the bug**
No indication provided to the user when adding catalog with invalid source url fails.

**To Reproduce**
1. As admin, add MCP servers using `Add servers from Git` and provide an invalid Source URL ( `https://github.com/obot-platform/obot` in my case )

Notice that user is not presented with any error message.

POST call made to `api/mcp-catalogs/default`  success with `200 OK` status code.

Following errors are seen in server logs:

```
[server] time="2025-07-02T12:48:43-07:00" level=error msg="error syncing 'default/default': handler obot-controller obot.obot.ai/v1, Kind=MCPCatalog: [routes.go:219] failed to read catalog https://github.com/obot-platform/obot : failed to read GitHub catalog https://github.com/obot-platform/obot : failed to list repository contents: 404 Not Found {"message":"Not Found","documentation_url":"https://docs.github.com/rest/git/trees#get-a-tree","status":"404"}, requeuing"
[server] time="2025-07-02T12:48:43-07:00" level=error msg="error syncing 'default/default': handler obot-controller obot.obot.ai/v1, Kind=MCPCatalog: [routes.go:219] failed to read catalog https://github.com/obot-platform/obot : failed to read GitHub catalog https://github.com/obot-platform/obot : failed to list repository contents: 404 Not Found {"message":"Not Found","documentation_url":"https://docs.github.com/rest/git/trees#get-a-tree","status":"404"}, requeuing"
```

**Expected behavior**
User should be presented with an error message.

cjellick:
This wil require front and backend, I think. But @g-linville im assigning to you. I think whatever object represents the git source should have a `status` or at least a an `error` field that can convey these details. I DO NOT think the request to submit the URL needs to be rejected, just that eventually we should see a status in the UI that indicates there's been a problem.

g-linville:
Adding @ivyjeong13 as an assignee because this one is ready for UI work now.

g-linville:
The UI change for this was merged earlier today. If a source URL fails to sync, it will turn yellow in the UI, and there is a thing you can either hover over or click on (I don't remember) that will provide more details.

sangee2004:
When testing with latest build , I am presented with following error message when clicking on the Catalog URL which has an indication that it is in error state

[...]

PR #3400, opened 2025-07-15
fix: log and display errors when we fail to read a git source

g-linville:
We just want to log an error and move on to the next source URL, rather than returning an error and ignoring the remaining source URLs.

This change also surfaces the errors for source URLs that failed.

for https://github.com/obot-platform/obot/issues/3307
for https://github.com/obot-platform/obot/issues/3306


g-linville:
Sorry for prematurely assigning reviewers. I'm going to add onto this PR to knock out a second, related bug.

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -49,11 +49,16 @@ func (h *Handler) Sync(req router.Request, resp router.Response) error {
...
+
 	for _, sourceURL := range mcpCatalog.Spec.SourceURLs {
 		objs, err := h.readMCPCatalog(mcpCatalog.Name, sourceURL)
 		if err != nil {
-			return fmt.Errorf("failed to read catalog %s: %w", sourceURL, err)
+			log.Errorf("failed to read catalog %s: %v", sourceURL, err)
+			mcpCatalog.Status.SyncErrors[sourceURL] = err.Error()
+			continue
 		}
+		delete(mcpCatalog.Status.SyncErrors, sourceURL)
```
StrongMonkey, line 61: I think we don't need to add this line, since you always reinitilize the map at line 52

g-linville: There is technically one niche situation where this would be useful, which is if the same `source URL` is on the catalog twice, and it failed the first loop iteration but succeeded on the second. The user shouldn't be doing that, but imo it doesn't really do any harm to leave this `delete` here.

PR #3403, opened 2025-07-15
fix: set environment variables in nanobot config

thedadams:
Nanobot will only pass the environment variable set in the configuration to the command. This change will pass the environment variables in that file.

This change also adds the annotations to the pod template so that any change in the configuration will redeploy the pod.

### DIFF
--- a/pkg/mcp/loader.go
+++ b/pkg/mcp/loader.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/util/homedir"
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/yaml"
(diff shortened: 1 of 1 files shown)


pkg/mcp/loader.go
```diff
@@ -21,6 +21,7 @@ import (
 	"github.com/obot-platform/obot/logger"
 	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
 	"github.com/obot-platform/obot/pkg/wait"
+	yamlv3 "gopkg.in/yaml.v3"
```
thedadams: This package has been deprecated. Testing a switch to `sigs.k8s.io/yaml`.

StrongMonkey (approved):

drpebcak (approved):

PR #3404, opened 2025-07-15
enhance: add support for yaml catalog manifests

njhale:
Add support for consuming catalog entries from YAML.


pkg/controller/handlers/mcpcatalog/github.go
```diff
@@ -12,6 +12,7 @@ import (
 	"strings"
 
 	"github.com/obot-platform/obot/apiclient/types"
+	"gopkg.in/yaml.v3"
```
thedadams: As of April 25th, 2025, this package has been deprecated. Should we switch to `sigs.k8s.io/yaml`?

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -235,14 +237,14 @@ func (h *Handler) readMCPCatalogDirectory(catalog string) ([]types.MCPServerCata
 				return nil, fmt.Errorf("failed to read nested catalog directory %s: %w", file.Name(), err)
 			}
 			entries = append(entries, nestedEntries...)
-		} else if strings.HasSuffix(file.Name(), ".json") {
+		} else if catalogFilePattern.MatchString(file.Name()) {
```
StrongMonkey: nit: we can just check for suffix to avoid overhead of regexp.

StrongMonkey (approved):

pkg/controller/handlers/mcpcatalog/github.go
njhale: sure thing. thanks for catching that!

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
njhale: switched to `strings.HasSuffix`

PR #3416, opened 2025-07-16
chore: prebuild base image to avoid blocking failures

drpebcak:


drpebcak:
note that the docker build will fail until i push this new image for the first time.

thedadams (approved):

Dockerfile
```diff
@@ -1,10 +1,14 @@
...
 ARG ENTERPRISE_IMAGE=cgr.dev/chainguard/wolfi-base:latest
+ARG BASE_IMAGE=cgr.dev/chainguard/wolfi-base
 
-FROM cgr.dev/chainguard/wolfi-base AS base
-
-RUN apk add --no-cache gcc=14.2.0-r13 go make git nodejs npm pnpm
+FROM ${BASE_IMAGE} AS base
+ARG BASE_IMAGE
+RUN if [ "${BASE_IMAGE}" = "cgr.dev/chainguard/wolfi-base" ]; then \
+        echo "from inside if" > /logic; \
```
g-linville: Is this a debug thing that was left behind?

drpebcak: yes it was, good catch!

PR #3417, opened 2025-07-16
feat: add support for provider auth groups

njhale:
Add the ability for admins to use auth provider sourced groups to gate user access to MCP servers.

Addresses part of https://github.com/obot-platform/obot/issues/3484

### DIFF
--- a/pkg/auth/auth.go
+++ b/pkg/auth/auth.go
@@ -1,0 +1,25 @@
+package auth
+
+import (
+	"context"
+)
+
+// GroupInfo represents information about a user group from an authentication provider
+type GroupInfo struct {
+	ID      string  `json:"id"`
+	Name    string  `json:"name"`
+	IconURL *string `json:"iconURL"`
+}
+
+type authProviderURLKey struct{}
+
+// ContextWithProviderURL adds the auth provider URL to the context
+func ContextWithProviderURL(ctx context.Context, url string) context.Context {
+	return context.WithValue(ctx, authProviderURLKey{}, url)
+}
+
+// ProviderURLFromContext retrieves the auth provider URL from the context
+func ProviderURLFromContext(ctx context.Context) string {
+	url, _ := ctx.Value(authProviderURLKey{}).(string)
+	return url
+}
--- a/pkg/controller/handlers/mcpcatalog/mcpcatalog.go
+++ b/pkg/controller/handlers/mcpcatalog/mcpcatalog.go
@@ -23,6 +23,7 @@ import (
 	"github.com/obot-platform/obot/pkg/system"
 	"github.com/obot-platform/obot/pkg/validation"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kuser "k8s.io/apiserver/pkg/authentication/user"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
 )
@@ -341,12 +342,12 @@ func (h *Handler) DeleteUnauthorizedMCPServers(req router.Request, _ router.Resp
 			continue
 		}
 
-		user, err := h.gatewayClient.UserByID(req.Ctx, server.Spec.UserID)
+		user, err := h.getUserInfoForAccessControl(req.Ctx, server.Spec.UserID)
 		if err != nil {
-			return fmt.Errorf("failed to get user %s: %w", server.Spec.UserID, err)
+			return fmt.Errorf("failed to get user info for %s: %w", server.Spec.UserID, err)
 		}
 
-		if user.Role.HasRole(types.RoleAdmin) {
+		if user.role.HasRole(types.RoleAdmin) {
 			// Don't delete servers created by admins.
 			continue
 		}
@@ -360,7 +361,7 @@ func (h *Handler) DeleteUnauthorizedMCPServers(req router.Request, _ router.Resp
 			continue
 		}
 
-		hasAccess, err := h.accessControlRuleHelper.UserHasAccessToMCPServerCatalogEntry(server.Spec.UserID, server.Spec.MCPServerCatalogEntryName)
+		hasAccess, err := h.accessControlRuleHelper.UserHasAccessToMCPServerCatalogEntry(user, server.Spec.MCPServerCatalogEntryName)
 		if err != nil {
 			return fmt.Errorf("failed to check if user %s has access to catalog entry %s: %w", server.Spec.UserID, server.Spec.MCPServerCatalogEntryName, err)
 		}
@@ -400,17 +401,17 @@ func (h *Handler) DeleteUnauthorizedMCPServerInstances(req router.Request, _ rou
 			continue
 		}
 
-		user, err := h.gatewayClient.UserByID(req.Ctx, instance.Spec.UserID)
+		user, err := h.getUserInfoForAccessControl(req.Ctx, instance.Spec.UserID)
 		if err != nil {
 			return fmt.Errorf("failed to get user %s: %w", instance.Spec.UserID, err)
 		}
 
-		if user.Role.HasRole(types.RoleAdmin) {
+		if user.role.HasRole(types.RoleAdmin) {
 			// Don't delete instances created by admins.
 			continue
 		}
 
-		hasAccess, err := h.accessControlRuleHelper.UserHasAccessToMCPServer(instance.Spec.UserID, instance.Spec.MCPServerName)
+		hasAccess, err := h.accessControlRuleHelper.UserHasAccessToMCPServer(user, instance.Spec.MCPServerName)
 		if err != nil {
 			return fmt.Errorf("failed to check if user %s has access to MCP server %s: %w", instance.Spec.UserID, instance.Spec.MCPServerName, err)
 		}
@@ -424,4 +425,37 @@ func (h *Handler) DeleteUnauthorizedMCPServerInstances(req router.Request, _ rou
 	}
 
 	return nil
+}
+
+// userInfo is a wrapper around kuser.Info that includes the user's role.
+type userInfo struct {
+	kuser.Info
+	role types.Role
+}
+
+// getUserInfoForAccessControl gets user info needed for access control checks
+func (h *Handler) getUserInfoForAccessControl(ctx context.Context, userID string) (*userInfo, error) {
+	gatewayUser, err := h.gatewayClient.UserByID(ctx, userID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get user %s: %w", userID, err)
+	}
+
+	// Get all provider auth groups for the user.
+	groupIDs, err := h.gatewayClient.ListGroupIDsForUser(ctx, gatewayUser.ID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list user group IDs: %w", err)
+	}
+
+	return &userInfo{
+		Info: &kuser.DefaultInfo{
+			Name:   gatewayUser.Username,
+			UID:    fmt.Sprintf("%d", gatewayUser.ID),
+			Groups: []string{},
+			Extra: map[string][]string{
+				// Omit the auth provider namespace and name since groupIDs may include groups from multiple auth providers.
+				"auth_provider_groups": groupIDs,
+			},
+		},
+		role: gatewayUser.Role,
+	}, nil
 }
--- a/pkg/gateway/client/auth.go
+++ b/pkg/gateway/client/auth.go
@@ -32,13 +32,14 @@ func (u UserDecorator) AuthenticateRequest(req *http.Request) (*authenticator.Re
 		return nil, false, nil
 	}
 
-	gatewayUser, err := u.client.EnsureIdentity(req.Context(), &types.Identity{
+	identity := &types.Identity{
 		Email:                 firstValue(resp.User.GetExtra(), "email"),
 		AuthProviderName:      firstValue(resp.User.GetExtra(), "auth_provider_name"),
 		AuthProviderNamespace: firstValue(resp.User.GetExtra(), "auth_provider_namespace"),
 		ProviderUsername:      resp.User.GetName(),
 		ProviderUserID:        resp.User.GetUID(),
-	}, req.Header.Get("X-Obot-User-Timezone"))
+	}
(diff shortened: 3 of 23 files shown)


pkg/gateway/client/auth.go
```diff
@@ -32,10 +33,29 @@ func (u UserDecorator) AuthenticateRequest(req *http.Request) (*authenticator.Re
 		return nil, false, nil
 	}
 
+	var (
+		authProviderName      = firstValue(resp.User.GetExtra(), "auth_provider_name")
+		authProviderNamespace = firstValue(resp.User.GetExtra(), "auth_provider_namespace")
```
g-linville: We should reuse these vars later in this function when calling `EnsureIdentity`.

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -404,3 +406,44 @@ func (h *Handler) DeleteUnauthorizedMCPServerInstances(req router.Request, _ rou
...
+	)
+
+	for _, identity := range identities {
+		authProviderName = identity.AuthProviderName
+		authProviderNamespace = identity.AuthProviderNamespace
+
+		for _, group := range identity.AuthProviderGroups {
+			groupIDs = append(groupIDs, group.ID)
+		}
+	}
```
g-linville: I'm a bit confused by this loop. We are collecting all of the group IDs, but only the provider name and namespace of the last identity in the slice. Is that the intended behavior? Why is preference given specifically to the last one in the slice, as opposed to any of the others?

njhale: Originally, I had it break at the first one with groups. I changed it at Donnie's recommendation the first time I put this code in, but in retrospect, you're right, don't think it makes a ton of sense. I will say that the group names themselves are fully qualified with the provider name as a prefix, so the provider name/namespace are kind of moot

thedadams: The issue with the previous implementation was that it assumed that the identity related to this user was the first one and would take the groups from that. There is an issue with that because of the way the tokens are handled in Obot. We really need to take the groups from all identities to make sure we get them all.

As far as the auth provider name and namespace: I believe those will be the same for the two identities I am talking about.

pkg/gateway/client/identity.go
```diff
@@ -277,6 +277,16 @@ func (c *Client) ensureIdentity(ctx context.Context, tx *gorm.DB, id *types.Iden
 		}
 	}
 
+	for i := range id.AuthProviderGroups {
+		if err := tx.Save(&id.AuthProviderGroups[i]).Error; err != nil {
+			return nil, fmt.Errorf("failed to save group %s: %w", id.AuthProviderGroups[i].ID, err)
+		}
+	}
+
+	if err := tx.Model(id).Association("AuthProviderGroups").Replace(id.AuthProviderGroups); err != nil {
```
thedadams: I think this deserves a comment. Not sure what is going on here.

njhale:
Found a problem with this. I think it's related to how these changes interact with unauthorized server cleanup. Need to fix the issue before merging.

pkg/gateway/client/identity.go
njhale: done

njhale:
> Found a problem with this. I think it's related to how these changes interact with unauthorized server cleanup. Need to fix the issue before merging.

Fixed this issue, but I'm still concerned about the foreign key issues caused by the many-to-many relationship between Identities and Groups. So I'm going to do away with the foreign key constraints and explicitly manage a join table. Should have the change up soon.

pkg/gateway/client/group.go
```diff
@@ -0,0 +1,320 @@
...
+
+// fetchGroups fetches the groups that the owner of the access token is a member of from the auth provider.
+func (*Client) fetchGroups(ctx context.Context, authProviderURL, accessToken, authProviderNamespace, authProviderName string) ([]types.Group, error) {
+	// Fetch groups from the auth provider
+	providerGroups := []auth.GroupInfo{}
+	u, err := url.Parse(authProviderURL + "/obot-list-user-auth-groups")
+	if err == nil {
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
+		if err == nil && accessToken != "" {
+			if accessToken := accesstoken.GetAccessToken(ctx); accessToken != "" {
```
njhale: Going to fix this, it should be the `accessToken` passed in here.

pkg/proxy/proxy.go
```diff
@@ -356,10 +359,19 @@ func (p *Proxy) authenticateRequest(req *http.Request) (*authenticator.Response,
...
+	// Why? Right now, token auth only hits the cached groups and never updates them.
+	// This is bad because if there are upstream changes to the user's groups -- e.g. the user is removed from a github team --
+	// we won't kick them out of the ACR group until they hit the UI again, so connected external mcp clients would stay connected.
+	// Alternatively, the caching could happen on the auth provider's side. The key to this approach
+	// would be mapping Obot Identities/AuthTokens to tokens cached on the auth provider's side, but that feels insecure.
+	*req = *req.WithContext(accesstoken.ContextWithAccessToken(req.Context(), ss.AccessToken))
+
+	// TODO(njhale): we can't rely on this being set when AuthToken requests are authenticated.
+	// I think we'll need to call the manager's URLForAuthProvider() and plumb it through to the pkg/gateway/client somehow.
+	*req = *req.WithContext(auth.ContextWithProviderURL(req.Context(), p.url))
```
njhale, line 374: These comments outline the biggest outstanding issues.

njhale: I'm going to address this as a follow up.

ui/user/src/lib/components/admin/SearchUsers.svelte
```diff
@@ -7,60 +7,98 @@
...
+		loading = false;
+		searchNames = '';
 		selectedUsers = [];
+		users = [];
+		filteredUsers = [];
+		filteredGroups = [];
 	}
 
-	let { onAdd, filterIds }: Props = $props();
+	$effect(() => {
```
ivyjeong13: If this effect is to trigger after search,  I think you can move it to happen as part of the 'onChange' that happens with <Search>

something like:
```
	async function search() {
        // move loading here
		loading = true;
		filteredUsers =
			searchNames.length > 0
				? users.filter(
						(user) =>
							user.email.toLowerCase().includes(searchNames.toLowerCase()) ||
							user.username.toLowerCase().includes(searchNames.toLowerCase())
					)
				: users;

		filteredGroups = await AdminService.listGroups(
			searchNames.length > 0 ? { query: searchNames } : undefined
		);
		loading = false;
	}
	const handleSearch = debounce(() => {
        // adding debounce for while user is typing
		if (searchNames === '') {
			users = [];
			filteredGroups = [];
			return;
		}

		search();
	}, 500);

    // ...
			<Search
				class="dark:bg-surface1 dark:border-surface3 shadow-inner dark:border"
				onChange={(val) => {
					searchNames = val;
					handleSearch();
				}}
				placeholder="Search by user name, email, or group name..."
			/>
```

pkg/gateway/client/group.go
```diff
@@ -0,0 +1,318 @@
...
+			if err == nil {
+				if accessToken := accesstoken.GetAccessToken(ctx); accessToken != "" {
+					req.Header.Set("Authorization", "Bearer "+accessToken)
+				}
+
+				resp, err := http.DefaultClient.Do(req)
+				if err == nil {
+					defer resp.Body.Close()
+					if resp.StatusCode == http.StatusOK {
+						_ = json.NewDecoder(resp.Body).Decode(&providerGroups)
```
g-linville, line 59: Why do we want to ignore this error?

pkg/gateway/client/identity.go
```diff
@@ -272,20 +279,40 @@ func (c *Client) ensureIdentity(ctx context.Context, tx *gorm.DB, id *types.Iden
 		id.HashedProviderUserID = hash.String(id.ProviderUserID)
 
 		// Copy so we don't have to decrypt again
```
g-linville: This comment should be deleted.

g-linville (commented):
I reviewed the backend only. Thank you for all of the helpful comments you included in the code!

pkg/gateway/client/group.go
```diff
@@ -0,0 +1,318 @@
...
+// This allows admins to discover groups that authenticated users belong to for auth providers
+// limited group search capabilities; e.g. there's not an effective way to perform a fuzzy search for
+// GitHub teams or orgs by name.
+func (c *Client) ListAuthGroups(ctx context.Context, authProviderURL, authProviderNamespace, authProviderName, nameFilter string) ([]types.Group, error) {
+	// Fetch groups from the auth provider
+	providerGroups := []auth.GroupInfo{}
+	if authProviderURL != "" {
+		u, err := url.Parse(authProviderURL + "/obot-list-auth-groups")
+		if err == nil {
+			if nameFilter != "" {
```
StrongMonkey, line 43: Do we intentionally ignore errors here?

pkg/gateway/client/group.go
```diff
@@ -0,0 +1,318 @@
...
+}
+
+// fetchGroups fetches the groups that the owner of the access token is a member of from the auth provider.
+func (*Client) fetchGroups(ctx context.Context, authProviderURL, accessToken, authProviderNamespace, authProviderName string) ([]types.Group, error) {
+	// Fetch groups from the auth provider, ignore errors so that auth providers that don't yet
+	// implement group support don't block the user from logging in.
+	providerGroups := []auth.GroupInfo{}
+	u, err := url.Parse(authProviderURL + "/obot-list-user-auth-groups")
+	if err == nil {
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
```
StrongMonkey, line 297: Same here, looks like the error from request is not handled here

pkg/gateway/client/group.go
njhale: ~Actually, for a failure to parse the URL, we should probably return an error. I'll update this. Thanks!~

on second thought, I think it's probably better to keep ignoring this error. This is called by the frontend during user/group search and I think it's better to fail silently and return any cached groups.

pkg/gateway/client/group.go
njhale: We ignore this error b/c even if the auth provider fails to return groups for whatever reason, we can still want to return the cached groups to the admin. I'll add a comment here elaborating.

pkg/gateway/client/group.go
njhale: There's a comment a few lines above this that explains the reasoning here.

pkg/proxy/proxy.go
StrongMonkey: Just a different thought:

We can also perioperiodically refresh groups every x hours, and provide a way to refresh group manually for all the users(if admin want to reflect group changes immediately after they changed in auth provider). This will effectively kick out all users that are already invalid.

(I think this is what we used to do in rancher).

StrongMonkey: (By doing this we don't have to get user groups everytime they are making request)

njhale (commented):
Addressed all feedback; requesting re-review

ui/user/src/lib/components/admin/SearchUsers.svelte
njhale: @ivyjeong13 should be fixed now.

pkg/auth/auth.go
```diff
@@ -0,0 +1,28 @@
...
+	return context.WithValue(ctx, authProviderURLKey{}, url)
+}
+
+// ProviderURLFromContext retrieves the auth provider URL from the context
+func ProviderURLFromContext(ctx context.Context) string {
+	url, ok := ctx.Value(authProviderURLKey{}).(string)
+	if !ok {
+		return ""
+	}
+	return url
```
thedadams: micro-nit:
```suggestion
	url, _ := ctx.Value(authProviderURLKey{}).(string)
	return url
```

pkg/gateway/client/group.go
```diff
@@ -0,0 +1,323 @@
...
+//
+// It supports fuzzy finding group names using on the given nameFilter.
+// It queries the auth provider for "live" group search from the auth provider, then combines the
+// results with cached groups from the database.
+// This allows admins to discover groups that authenticated users belong to for auth providers
+// limited group search capabilities; e.g. there's not an effective way to perform a fuzzy search for
+// GitHub teams or orgs by name.
+func (c *Client) ListAuthGroups(ctx context.Context, authProviderURL, authProviderNamespace, authProviderName, nameFilter string) ([]types.Group, error) {
+	// Fetch groups from the auth provider
+	providerGroups := []auth.GroupInfo{}
```
thedadams: nit:
```suggestion
	var providerGroups []auth.GroupInfo
```

pkg/gateway/client/group.go
```diff
@@ -0,0 +1,323 @@
...
+		if nameFilter != "" {
+			query = query.Where("LOWER(name) LIKE LOWER(?)", "%"+nameFilter+"%")
+		}
+
+		if err := query.Find(&dbGroups).Error; err != nil {
+			return nil, fmt.Errorf("failed to fetch groups from database: %w", err)
+		}
+	}
+
+	groups := make(map[string]types.Group)
```
thedadams:
```suggestion
	groups := make(map[string]types.Group, len(dbGroups))
```

pkg/gateway/client/identity.go
```diff
@@ -28,16 +28,23 @@ var (
...
+			return err
+		}
 
-	for i := range identities {
-		if err := c.decryptIdentity(ctx, &identities[i]); err != nil {
-			return nil, fmt.Errorf("failed to decrypt identity: %w", err)
+		for i := range identities {
+			if err := c.decryptIdentity(ctx, &identities[i]); err != nil {
+				return fmt.Errorf("failed to decrypt identity: %w", err)
+			}
```
thedadams: If we can avoid it, we shouldn't decrypt inside a transaction. Is there a reason to do it here?

njhale: No, not that I'm aware of. I think this was an artifact of refactoring this code several times. I'll fix it. Thanks for catching that.

thedadams (approved):

Issue #3398, opened 2025-07-15
Audit Logs - Unique User count always shows up as 1 


sangee2004:
**Describe the bug**
Audit Logs - Unique User count always shows up as "1"

**To Reproduce**
1. As  multiple regular/admin user, connect to few MCP severs and chat with them.
2. As admin user , got to `Audit Logs` page
Notice that `Unique User` count always shows as 1 even though we have audit log entries from 3 different users. 

<img width="1345" height="947" alt="Image" src="https://github.com/user-attachments/assets/420d836a-9a0d-4e12-be10-96261eb8f7c2" />

**Expected Behavior:**
 Unique User count should have the correct value.

sangee2004:
@thedadams `Unique User` count seems to take into account only admin users and regular users.

sangee2004:
@thedadams I am testing with `v0.0.0-dev+cac5ed2b` .

I am still not seeing the right value for user count being reflected in `Audit Logs` page.

Also wanted to confirm if the stats is for the last 24 hours ?  Even if it is for the last 24 hours , I have 2 unique users that have done `tool/call` in the last 24 hours and 3 unique users if we also include other types of connection like `tool/list`

sangee2004:
`Total Calls` is also not showing the same value. This is a note for me to check that as well when this issue gets addressed.

sangee2004:
This issue is addressed when testing with latest build from `main`.Both `Unique Users` and `Total Calls` reflect the right values for the last hour.

PR #3430, opened 2025-07-17
fix: pull correct unique user and call counts for stats

thedadams:
Issue: https://github.com/obot-platform/obot/issues/3398


pkg/gateway/client/mcpauditlog.go
```diff
@@ -99,17 +99,17 @@ func (c *Client) GetMCPUsageStats(ctx context.Context, opts MCPUsageStatsOptions
 			MCPServerCatalogEntryName string
 		}
 
+		if err := tx.Select("COUNT(*) AS total_calls, COUNT(DISTINCT user_id) AS unique_users").Scan(&callsAndUsers).Error; err != nil {
```
thedadams, line 102: Yes, the order does matter here.

g-linville (approved):

Issue #3272, opened 2025-06-26
Updates done to existing catalog entries are not reflected in the catalog entries after catalog refresh. These changes are reflected only when catalog delete changes are synced.


sangee2004:
**Describe the bug**
Updates done to existing catalog entries are not reflected in the catalog entries after catalog refresh. These changes are reflected only when catalog delete changes are synced.

**To Reproduce**
1. Create a MCP catalog .
2. Add MCP Servers using Add servers from Git.
3. This results in all existing entries from git location to get synced to catalog and list is presented to the user.
4. Update an existing MCP Server entry in  github location. I changed the name and some environment variable's  key, name and description. 
5. Use "Refresh Catalog" option to fetch the changes done in step2.

Notice that the changes do not get reflected in the UI. Even a manual refresh of the page does not get the changes.

6. Delete an existing MCP Server entry in  github location 
7. Use "Refresh Catalog" option.
8. Manually refresh the page

Now changes from step 4 (updates done to existing catalog entries) and step 6 ( deletion of catalog entries) will get reflected.

**Note** - Adding a new entry instead of deletion in step 6 will not result in catalog entry updates to be fetched either.

**Expected behavior**
Updates done to existing catalog entries should be reflected in the catalog entries after catalog refresh.

g-linville:
This bug appears to be fixed. A couple things I noticed though that might cause false alarm:

1. Make sure you have `GITHUB_AUTH_TOKEN` set so that you don't get rate limited.
2. After clicking the refresh button in the UI, the UI immediately fetches the list of catalog entries again. It does this before the backend has had time to scrape the GitHub repo, and so any deleted entries will still appear. Wait a couple more seconds and then refresh the page, and you will see that it is gone.

g-linville:
(I also saw this work for adding new entries to the repo.)

g-linville:
Actually, moving this back to In Progress, because I think there might be an issue with the updating of existing entries.

g-linville:
Never mind, updating existing entries works, it just takes a little bit longer for the changes to appear for some reason, and I am not entirely sure why.

sangee2004:
@g-linville Updates done to catalog entries are not getting reflected after I do a `refresh` from UI.

Even when I attempt to manually refresh the page at which point UI queries the catalog entries, i dont see the updates being reflected in the UI.

[...]

PR #3435, opened 2025-07-17
enhance: catalog source URLs: use go-git to clone

g-linville:
This will hopefully help us to process catalog repos faster, compared to making tons of HTTP requests. Hopefully the rate limiting will be less severe, too.

This should help with https://github.com/obot-platform/obot/issues/3272

### DIFF
--- a/pkg/controller/handlers/mcpcatalog/github.go
+++ b/pkg/controller/handlers/mcpcatalog/github.go
@@ -2,25 +2,86 @@ package mcpcatalog
 
 import (
 	"bufio"
+	"encoding/json"
 	"fmt"
 	"io"
+	"net/http"
 	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/plumbing"
-	"github.com/go-git/go-git/v5/plumbing/transport/http"
+	githttp "github.com/go-git/go-git/v5/plumbing/transport/http"
 	"github.com/obot-platform/obot/apiclient/types"
 	"sigs.k8s.io/yaml"
 )
 
 var githubToken = os.Getenv("GITHUB_AUTH_TOKEN")
 
+// GitHubRepoInfo represents the repository information from GitHub API
+type GitHubRepoInfo struct {
+	Size int `json:"size"` // Size in KB
+}
+
 func isGitHubURL(catalogURL string) bool {
 	u, err := url.Parse(catalogURL)
 	return err == nil && u.Host == "github.com"
+}
+
+// checkRepoSize checks the repository size using GitHub API before cloning
+func checkRepoSize(org, repo string, maxSizeMB int) error {
+	apiURL := fmt.Sprintf("https://api.github.com/repos/%s/%s", org, repo)
+
+	// Create HTTP client with timeout
+	client := &http.Client{
+		Timeout: 30 * time.Second,
+	}
+
+	// Create request
+	req, err := http.NewRequest("GET", apiURL, nil)
+	if err != nil {
+		return fmt.Errorf("failed to create API request: %w", err)
+	}
+
+	// Add authentication if token is available
+	if githubToken != "" {
+		req.Header.Set("Authorization", "Bearer "+githubToken)
+	}
+
+	// Make the request
+	resp, err := client.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to fetch repository info: %w", err)
+	}
+	defer resp.Body.Close()
+
+	// Check response status
+	if resp.StatusCode == 404 {
+		return fmt.Errorf("repository not found: %s/%s", org, repo)
+	}
+	if resp.StatusCode == 403 {
+		return fmt.Errorf("access denied to repository: %s/%s (may be private or rate limited)", org, repo)
+	}
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("GitHub API returned status %d for repository %s/%s", resp.StatusCode, org, repo)
+	}
+
+	// Parse response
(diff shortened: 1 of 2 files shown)


pkg/controller/handlers/mcpcatalog/github.go
```diff
@@ -10,18 +10,140 @@ import (
...
 	return err == nil && u.Host == "github.com"
 }
 
+// checkRepoSize checks the repository size using GitHub API before cloning
+func checkRepoSize(org, repo string, maxSizeMB int) error {
+	apiURL := fmt.Sprintf("https://api.github.com/repos/%s/%s", org, repo)
+
+	// Create HTTP client with timeout
+	client := &http.Client{
+		Timeout: 30 * time.Second,
```
thedadams: nit: seems like a long timeout.

pkg/controller/handlers/mcpcatalog/github.go
```diff
@@ -10,18 +10,140 @@ import (
...
+
+	// Make the request
+	resp, err := client.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to fetch repository info: %w", err)
+	}
+	defer resp.Body.Close()
+
+	// Check response status
+	if resp.StatusCode == 404 {
```
thedadams: nit (similar for the ones below):
```suggestion
	if resp.StatusCode == http.StatusNotFound {
```

pkg/controller/handlers/mcpcatalog/github.go
```diff
@@ -10,18 +10,140 @@ import (
...
+		return fmt.Errorf("repository not found: %s/%s", org, repo)
+	}
+	if resp.StatusCode == 403 {
+		body, _ := io.ReadAll(resp.Body)
+		if len(body) > 0 {
+			return fmt.Errorf("access denied to repository: %s/%s - %s", org, repo, string(body))
+		}
+		return fmt.Errorf("access denied to repository: %s/%s (may be private or rate limited)", org, repo)
+	}
+	if resp.StatusCode != 200 {
```
thedadams: Not sure how much value we get out of separating the status codes above or if this is enough.

pkg/controller/handlers/mcpcatalog/github.go
```diff
@@ -10,18 +10,140 @@ import (
...
+}
+
+// validateBranchName validates that the branch name doesn't contain suspicious characters
+func validateBranchName(branch string) error {
+	if branch == "" {
+		return fmt.Errorf("branch name cannot be empty")
+	}
+
+	// Check for path traversal attempts and other suspicious characters
+	if strings.Contains(branch, "..") || strings.Contains(branch, "/") ||
```
thedadams: Branch names can have `/` in them. For example: `fix/something-i-broke`.

pkg/controller/handlers/mcpcatalog/github.go
g-linville: agreed, changing it to 5

pkg/controller/handlers/mcpcatalog/github.go
g-linville: Yeah I think you are right. I will just have one check to capture all non-200 codes

pkg/controller/handlers/mcpcatalog/github.go
g-linville: Thanks! I need to get better at challenging Claude's assumptions about things like this, lol

pkg/controller/handlers/mcpcatalog/github.go
g-linville: Fixed

pkg/controller/handlers/mcpcatalog/github.go
g-linville: Fixed

pkg/controller/handlers/mcpcatalog/github.go
g-linville: Fixed

pkg/controller/handlers/mcpcatalog/github.go
g-linville: Fixed

pkg/controller/handlers/mcpcatalog/github.go
```diff
@@ -53,29 +159,59 @@ func readGitHubCatalog(catalogURL string) ([]types.MCPServerCatalogEntryManifest
 	branch := "main"
 	if len(parts) > 2 {
 		branch = parts[2]
```
thedadams: Now, that we are correctly allowing branch names to have `/` in them, I think this needs to change, too.

g-linville: you're right, fixed

thedadams (approved):

StrongMonkey:
@g-linville do we do anything regards to lastSyncTime so that we don't sync multiple times within the same period? last time i checked(i easily ran into the rate limit issue), the sync controller was run multiple times in a row when starting, which easily blow up the github limit.

g-linville:
@StrongMonkey I'm hesitant to try to gate this around the last sync time. I would rather just let the handler run as often as it is supposed to. This PR changes it so that we only make two requests out to GitHub: one to check the repo size, and one to clone. (I'm not sure how "rate limit-intensive" the clone operation is; for all I know it could count the same as several API requests.) I think this will be an improvement but I am not 100% sure. If it is still terrible, then I would consider trying to stop this handler from running so often.

StrongMonkey:
Ok, I think it is still a good idea to gate that based on sync time so that we prevent unnecessary sync process running(in this case, multiple clone process will be running at the start due to controller re-enqueuing, which seems unnecessary and quite heavy).  But up to your call.

StrongMonkey (approved):

Issue #3301, opened 2025-07-01
Deleting MCP server does not result in MCP server being removed from Access Controls that they are added to.


sangee2004:
**Describe the bug**
Deleting MCP server does not result in MCP server being removed from Access Controls that they are added to.

**To Reproduce**
1. As admin , create an Access Control say `AC1` with users  say `user1` and `user2`  and MCP servers say `server1` and `server2`
2. As user1 , get connection url for MCP server - `server1`.
3. From VS code , use the connection url from step2 to create MCP server entry.
4. As admin user , delete MCP server - `server1`

Notice that , MCP server entry - `server1` continues to be present in the Access Control - `AC1`.
Call made by UI to `/api/access-control-rules/<ac id> returns the deleted MCP server entry as well,.
In the UI , this is shown as empty entry

<img width="926" alt="Image" src="https://github.com/user-attachments/assets/9456705e-6904-4628-af4d-ee4e92eeccc6" />

**Expected behavior**
Deleted MCP Servers should be removed from all the  from Access Controls that they are part of.

g-linville:
I just merged the fix for this. When a user is deleted, we now remove them from all the access control rules that they were on, and when a server is deleted, it also gets pruned from the ACRs.

sangee2004:
This issue is addressed when testing with latest build.

Deleting MCP servers/users  results in them being removed from Access Controls that they are added to as expected.

PR #3445, opened 2025-07-18
fix: access control rules: prune nonexistent resources

g-linville:
for https://github.com/obot-platform/obot/issues/3301

### DIFF
--- a/pkg/controller/handlers/accesscontrolrule/accesscontrolrule.go
+++ b/pkg/controller/handlers/accesscontrolrule/accesscontrolrule.go
@@ -1,0 +1,52 @@
+package accesscontrolrule
+
+import (
+	"fmt"
+
+	"github.com/obot-platform/nah/pkg/router"
+	"github.com/obot-platform/obot/apiclient/types"
+	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+)
+
+type Handler struct {
+	acrHelper *Helper
+}
+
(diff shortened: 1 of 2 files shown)


g-linville:
Tests are only failing due to annoying GH rate limits.

pkg/controller/handlers/accesscontrolrule/accesscontrolrule.go
```diff
@@ -0,0 +1,52 @@
...
+
+func (h *Handler) PruneDeletedResources(req router.Request, _ router.Response) error {
+	acr := req.Object.(*v1.AccessControlRule)
+
+	// Make sure each resource still exists, and remove it if it is gone.
+	newResources := make([]types.Resource, 0, len(acr.Spec.Manifest.Resources))
+	for _, resource := range acr.Spec.Manifest.Resources {
+		switch resource.Type {
+		case types.ResourceTypeMCPServerCatalogEntry:
+			mcpservercatalogentry := &v1.MCPServerCatalogEntry{}
```
thedadams: nit: since we don't actually need this, it can be declared outside of the loop so we only have to allocate memory once. Same for the `mcpserver` below.

thedadams (approved):

pkg/controller/handlers/accesscontrolrule/accesscontrolrule.go
g-linville: Good call, fixed

thedadams (approved):

StrongMonkey (approved):

Issue #3302, opened 2025-07-02
Allowed to get connection url even when required parameters are not entered.


sangee2004:
**Describe the bug**
Allowed to get connection url even when required parameters are not entered.

**To Reproduce**
1. Add a Single user type MCP server  entry that has required parameter (FireCrawl in my case) to access control  and give access to say `user1`
2. As user1 , use `connect to server` option .
3. When prompted to enter Firecrawl api key , do not enter any value and click on `Launch`.

Notice that the `connection url` is created. When used with VS code ,we get the following error

```
400 status sending message to <connection url> will attempt to fall back to legacy SSE
2025-07-01 17:15:59.110 [info] Connection state: Error 400 status connecting to <connection url> as SSE: missing required config: FIRECRAWL_API_KEY
```

https://www.loom.com/share/bdaa6446e5d8426cbfc3f40d333a8d89

**Expected behavior**
User should be forced to enter required parameters before they can launch the server.

g-linville:
This is really just a frontend change, but it's a small one, so I'm confident I can do it pretty well.

g-linville:
The UI will now require you to provide the required parameters before you can move on. Just merged that fix.

sangee2004:
This issue is addressed when testing with latest build.
User is presented with the following page when launch is clicked without entering the values for required parameters

<img width="716" height="247" alt="Image" src="https://github.com/user-attachments/assets/99a4d590-3706-43fd-b99c-922bf9934764" />

Once a value is entered and `Launch` is clicked then connection url is generated as expected.

PR #3446, opened 2025-07-18
fix: ui: single-user/remote server config dialog: disable launch button until all required fields are filled

g-linville:
for https://github.com/obot-platform/obot/issues/3302

### DIFF
--- a/ui/user/src/routes/mcp-servers/+page.svelte
+++ b/ui/user/src/routes/mcp-servers/+page.svelte
@@ -71,6 +71,7 @@
 	let showServerInfo = $state(false);
 
 	let search = $state('');
+	let highlightedFields = $state<Set<string>>(new Set());
 	let serverInstancesMap = $derived(
 		new Map(
 			userServerInstances.map((instance) => [
@@ -240,6 +241,7 @@
 				parent: connectToEntry?.entry
 			};
 			connectToEntry = undefined;
+			clearHighlights();
 			configDialog?.close();
 			connectDialog?.open();
 		} else if (connectToServer?.server) {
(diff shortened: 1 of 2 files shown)


njhale (approved):

ivyjeong13 (approved):
:+1: Looks good to me! I was thinking, if it helps the user in the future, they could click the Launch and it highlights the required fields they need to fill in? Something like: 

image

ui/user/src/routes/mcp-servers/+page.svelte
```diff
@@ -342,6 +344,78 @@
...
+
+		highlightedFields = fieldsToHighlight;
+	}
+
+	function clearHighlights() {
+		highlightedFields = new Set();
+	}
+
+	// Clear highlights when values change for sensitive inputs
+	$effect(() => {
```
ivyjeong13: nit: Can just wait till the end to call highlightedFields = new Set(highlightedFields) to trigger svelte reactivity

```
        function updateHighlightedFields(
		config: typeof connectToEntry | typeof userConfiguredServerToEdit
	) {
		if (!config) return;

		[...config.envs ?? [], ...config.headers ?? []].forEach((field) => {
			if (field.required && !field.value) {
				highlightedFields.delete(field.key);
			}
		});

		if (config.url && highlightedFields.has('url-manifest-url')) {
			highlightedFields.delete('url-manifest-url');
		}

		highlightedFields = new Set(highlightedFields);
	}

	// Clear highlights when values change for sensitive inputs
	$effect(() => {
		if (userConfiguredServerToEdit || connectToEntry) {
			updateHighlightedFields(userConfiguredServerToEdit ?? connectToEntry);
		}
	});
```

ui/user/src/routes/mcp-servers/+page.svelte
```diff
@@ -822,11 +896,28 @@
 			{@render configureForm(connectToEntry, connectToEntry.entry)}
 			<div class="flex justify-end">
 				<button
-					class="button-primary"
-					disabled={connectToEntry.envs?.some((env) => env.required && !env.value) ||
+					class={`button-primary ${
```
ivyjeong13: nit: maybe a shared function like this for checking isDisabled?
```
	function missingRequiredFields(
		config: typeof connectToEntry | typeof userConfiguredServerToEdit
	) {
		if (!config) return false;

		if ('entry' in config && config.entry.urlManifest && !config.url) {
			return true;
		}

		return [...(config.envs ?? []), ...(config.headers ?? [])].some(
			(field) => field.required && !field.value
		);
	}
```

```
<ResponsiveDialog
	bind:this={editUserConfiguredServerDialog}
   // ...

	{#if userConfiguredServerToEdit}

// ...
			<button
				class={`button-primary ${
					missingRequiredFields(userConfiguredServerToEdit)
						? 'cursor-not-allowed opacity-50'
						: ''
				}`}
				onclick={async (e) => {
					if (!userConfiguredServerToEdit) return;

                                         const isDisabled = missingRequiredFields(userConfiguredServerToEdit);
					if (isDisabled) {
```

```
<ResponsiveDialog bind:this={configDialog} animate="slide" onClose={clearHighlights}>
// ...
	{#if connectToEntry}
// ...
<button
					class={`button-primary ${
						missingRequiredFields(connectToEntry)
							? 'cursor-not-allowed opacity-50'
							: ''
					}`}
					onclick={(e) => {
						if (!connectToEntry) return;
                                                  const isDisabled = missingRequiredFields(connectToEntry);
						if (isDisabled) {
							e.preventDefault();
```

ui/user/src/routes/mcp-servers/+page.svelte
ivyjeong13: Actually, wondering if the $effect should be removed and maybe we can add an `onInput` prop callback on the `SensitiveInput` component instead? edit: Then SensitiveInput could use the same logic as the non-sensitive `input` element

ui/user/src/routes/mcp-servers/+page.svelte
ivyjeong13: Also realizing actually, you could probably drop the `cursor-not-allowed opacity-50` classes since the button can be clicked.

ui/user/src/routes/mcp-servers/+page.svelte
g-linville: Fixed. I added the suggested prop to SensitiveInput and removed the whole $effect

ui/user/src/routes/mcp-servers/+page.svelte
g-linville: Fixed, created the function and removed the unneeded style classes

Issue #3433, opened 2025-07-17
Copy-to-Clipboard Button Fails in Oauth Providers


cloudnautique:
**Describe the bug**
When you click on the copy callback url button on `/v2/admin/auth-providers` flow the copy button does not work. 

<img width="842" height="299" alt="Image" src="https://github.com/user-attachments/assets/c6819f0e-a3e8-48ce-8955-d6c4c3942de0" />

**To Reproduce**
1. Go to `/v2/admin/auth-providers`
2. Click configure on one of the providers.
3. Copy the callback URL

**Expected behavior**
URL is added to the clip board so it can be pasted.

njhale:
Looks like this may be an issue caused by not using HTTPS (see https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard). Looking for a way to avoid using `navigator.clipboard` in these scenarios and gracefully degrade.

njhale:
Spoke with @cloudnautique about this and we settled on showing the copy button _only_ in secure contexts. Will update the PR after I finish up some more critical work.

cjellick:
@ivyjeong13 you want to re-impl this? it got stalled, figured it may be easier if you just pick it up.

shwetapsdet:
Verified it on Obot main and it's working as expected 

https://www.loom.com/share/03ed563dc14f4527a869850778284e77

PR #3447, opened 2025-07-18
fix: disable copy button when clipboard is unavailable

njhale:
`navigator.clipboard` is unavailable in insecure contexts (like
non-localhost http). This results in the copy button not working.

Detect this scenario and disable the copy button.

Addresses https://github.com/obot-platform/obot/issues/3433

Note: I'm not sure this is actually how we want to handle this. I can
try another method to make this work in these contexts, but wanted to
gather feedback on this approach first.

### DIFF
--- a/ui/user/src/lib/components/CopyButton.svelte
+++ b/ui/user/src/lib/components/CopyButton.svelte
@@ -28,6 +28,12 @@
 	let buttonTextToShow = $state(buttonText);
 	const COPIED_TEXT = 'Copied!';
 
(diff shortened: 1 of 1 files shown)


g-linville (approved):

ivyjeong13 (approved):
Looks like navigator.clipboard is appropriate, https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard says it's only available in secure contexts. Was wondering are there cases where we would want to avoid copying in a secure context?

PR #3448, opened 2025-07-18
chore: work with mcp-remote

thedadams:
For some reason, after going through the OAuth process, mcp-remote is re-requesting the oauth-authorization-server metadata, but using the wrong URL. This change serves the same metadata from a different path to make mcp-remote happy.

### DIFF
--- a/pkg/api/handlers/wellknown/handler.go
+++ b/pkg/api/handlers/wellknown/handler.go
@@ -17,7 +17,9 @@ func SetupHandlers(baseURL string, config services.OAuthAuthorizationServerConfi
 	}
 
 	mux.HandleFunc("GET /.well-known/oauth-protected-resource/mcp-connect/{mcp_id}", h.oauthProtectedResource)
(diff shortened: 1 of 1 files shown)


cjellick (approved):

pkg/api/handlers/wellknown/handler.go
```diff
@@ -18,6 +18,7 @@ func SetupHandlers(baseURL string, config services.OAuthAuthorizationServerConfi
 
 	mux.HandleFunc("GET /.well-known/oauth-protected-resource/mcp-connect/{mcp_id}", h.oauthProtectedResource)
 	mux.HandleFunc("GET /.well-known/oauth-authorization-server/{mcp_id}", h.oauthAuthorization)
+	mux.HandleFunc("GET /.well-known/oauth-authorization-server/mcp-connect/{mcp_id}", h.oauthAuthorization)
```
g-linville, line 22: imo a comment should be left here explaining why we did this

thedadams: Done.

PR #3460, opened 2025-07-21
chore: gate github tests behind a check for the token

g-linville:
We get rate limited in our PRs way too much. ~~This section of the code is pretty isolated and hardly ever touched, so there's really not a need to run this test on every PR anyway.~~

I added the GitHub token to the environment when running tests in order to help prevent this.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -25,3 +25,5 @@ jobs:
       - name: Run Tests
         run: |
           make test
+        env:
+          GITHUB_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


thedadams:
We have access to `secrets.GITHUB_TOKEN` in the action. Can we add `GITHUB_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}` to the action definition instead?

g-linville:
@thedadams good call, I fixed it

g-linville:
(and I will make sure the commit message I use matches this, since the PR title no longer does)

thedadams (approved):

StrongMonkey (approved):

StrongMonkey:
~@g-linville actually, are we running this on pull requests? I think the secret are not exposed in pull request due to security reason.~ nvm, donnie explained

thedadams:
The `secrets.GITHUB_TOKEN` is a special secret that is passed to every action, I think.
https://docs.github.com/en/actions/concepts/security/github_token

Issue #3221, opened 2025-06-17
409 errors when starting a new thread.


sangee2004:
**Describe the bug**
409 errors when starting a new thread.

Note - This issue does not reproduce always. I saw it happen 3 times when starting a new thread for this agent.
**To Reproduce**
1. As regular/admin user, connect to any MCP server from `Available MCP Servers` list.
2. Chat with this MCP sever

Notice that a 409 error is presented to the user 

<img width="1505" height="854" alt="Image" src="https://github.com/user-attachments/assets/a1853930-c214-4192-9c97-1554bd84660e" />

**Expected behavior**
There should be no 409 errors presented to the user.

sangee2004:
I have not been seeing the `409` issues when stating new thread when testing with latest build.

PR #3467, opened 2025-07-21
Fix; fix 409 when updating thread

StrongMonkey:
after creating thread and updating it, we are likely to run into conflict when controller doing updating. So when client updates thread it should not log error and should retry 409 properly.

https://github.com/obot-platform/obot/issues/3221

### DIFF
--- a/ui/user/src/lib/components/edit/ThreadModelSelector.svelte
+++ b/ui/user/src/lib/components/edit/ThreadModelSelector.svelte
@@ -118,26 +118,56 @@
 		isUpdatingModel = true;
 
 		try {
-			const updatedThread = await updateThread(project.assistantID, project.id, {
-				...threadType,
-				model: model,
-				modelProvider: provider
-			});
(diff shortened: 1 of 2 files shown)


ui/user/src/lib/components/edit/ThreadModelSelector.svelte
```diff
@@ -118,26 +118,56 @@
...
 		try {
-			const updatedThread = await updateThread(project.assistantID, project.id, {
-				...threadType,
-				model: model,
-				modelProvider: provider
-			});
+			let retryCount = 0;
+			const maxRetries = 5;
+
+			while (retryCount < maxRetries) {
```
g-linville: pico-nit


```suggestion
			const maxTries = 5;

			while (retryCount < maxTries) {
```

(this variable is also referenced farther down)

g-linville (approved):

thedadams (approved):

Issue #3291, opened 2025-07-01
Gateway webhooks


cjellick:
We want to implement a webhook system for the gateway so that requests going to (or coming from) MCP servers can be validated by them.

The inspiration for this is very much K8s's validating and mutating webhooks.  I'd like us to draw from their design, as I believe it is well thought out.

I want phase 1 of this feature to just be validating. I don't want to implement mutating until we see demand for it.

Basic sample use case: scan tool call input for secret keys using an open source project like
- https://github.com/trufflesecurity/trufflehog
- https://github.com/gitleaks/gitleaks

Reject the tool call a secret is detected.

Initial implementation of this coudl be JUST for tool calls, but ultimately I'd like to see that anything passing through the gateway could go through a configured webhook (meaning resource GETs, elicitations, sampling, etc).

When configuring a wehbook, the author should be able to specify a filter or selector such that not every single interaction goes through every single webhook. For example, they should be able to say a webhook is for tool call requests and even specify that it applies to 1, many, or all tool calls by name. Meaning, maybe they can set an array of tool names and/or have an easy to specify all tools, like a wild card.

thedadams:
Here is an example webhook I used for testing: https://github.com/thedadams/obot-webhook-example

The keys are:
1. It logs when it receives a valid request.
2. It validates the signature header.
3. It checks the payload for potential secrets.

sangee2004:
Filters feature is available when testing with latest build from main.

Able to create filters and see it work as expected with the example webhook application from https://github.com/obot-platform/obot/issues/3291#issuecomment-3133545818.

Issues found when testing this feature are tracked in separate issues
https://github.com/obot-platform/obot/issues/3662
https://github.com/obot-platform/obot/issues/3663
https://github.com/obot-platform/obot/issues/3664
https://github.com/obot-platform/obot/issues/3666
https://github.com/obot-platform/obot/issues/3672
https://github.com/obot-platform/obot/issues/3673

PR #3473, opened 2025-07-22
feat: add support for MCP webhooks and filters

thedadams:
Webhooks can be configured at the catalog entry or MCP server level. The '*' can be used to indicate that a webhook should be called for all MCP servers or catalog entries. Finally, webhooks can be configured for some or all MCP methods as well as some or all call identifiers (tool, prompt, or resource names).

Issue: https://github.com/obot-platform/obot/issues/3291


pkg/api/handlers/mcpwebhookvalidation.go
```diff
@@ -0,0 +1,180 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: system.MCPWebhookValidationPrefix,
+			Namespace:    req.Namespace(),
+		},
+		Spec: v1.MCPWebhookValidationSpec{
+			Manifest: manifest,
+		},
+	}
+
+	if err := req.Create(&validation); err != nil {
```
StrongMonkey, line 73: This doesn't seem to return the **generated name** from system after creating to the client, if we care about that.

pkg/api/handlers/mcpwebhookvalidation.go
```diff
@@ -0,0 +1,180 @@
...
+		return fmt.Errorf("failed to create mcp webhook validation: %w", err)
+	}
+
+	if err := req.GPTClient.CreateCredential(req.Context(), gptscript.Credential{
+		Context:  system.MCPWebhookValidationCredentialContext,
+		ToolName: validation.Name,
+		Type:     gptscript.CredentialTypeTool,
+		Env:      secretCred,
+	}); err != nil {
+		_ = req.Delete(&validation)
```
StrongMonkey, line 83: should we handle error here?

pkg/api/handlers/mcpwebhookvalidation.go
```diff
@@ -0,0 +1,180 @@
...
+
+	if err = req.Delete(&validation); err != nil {
+		return fmt.Errorf("failed to delete mcp webhook validation: %w", err)
+	}
+
+	return req.Write(convertMCPWebhookValidation(validation, secretCred.Env))
+}
+
+func convertMCPWebhookValidation(validation v1.MCPWebhookValidation, secrets map[string]string) types.MCPWebhookValidation {
+	for key, secret := range secrets {
```
StrongMonkey: Do we want to actually expose the secret in api?

StrongMonkey (commented):
a few comments

pkg/api/handlers/mcpwebhookvalidation.go
thedadams: It seems to return the `Name` field on the object. Just tested to make sure.

pkg/api/handlers/mcpwebhookvalidation.go
thedadams: No, the intention is just to try to cleanup the validation if we can't create the credential.

pkg/api/handlers/mcpwebhookvalidation.go
thedadams: Every time I don't expose it, I am eventually asked to expose it. So I just did it to start.

What are your thoughts?

StrongMonkey: I would avoid exposing that unless client really needs the secret

StrongMonkey (approved):

pkg/api/handlers/mcpwebhookvalidation.go
thedadams: I will be changing this after our standup discussion.

thedadams: This is removed now.

StrongMonkey (approved):

Issue #1783, opened 2025-02-19
UI: for GitHub users, show username from identity rather than user


g-linville:
If you set up both GitHub and Google for authentication, and then log in with Google first followed by GitHub (with accounts that have the same email address), you reach a state where the UI displays your username (since you are logged in with GitHub), but that username is actually the Google username, which is just a long string of numbers.

This is happening because, when the user is first created in the database, the username from the provider is saved and never changed.

So in order to fix this, we need to use the user's username from their current identity (i.e. GitHub in the case I described above), rather than the username attached to the user in the database.

thedadams:
I believe this is a backend issue. @ivyjeong13: if you feel the same way after investigating, then please throw it back over.

ivyjeong13:
@thedadams I do believe the database part is backend but there is a small frontend change that needs to happen (similar to admin-ui) so it shows username instead of email. Will make that fix and can throw this back over!

ivyjeong13:
Merged in frontend fix, kicking back for backend.

StrongMonkey:
The displayName will show user's actual friendly name in auth provider after the fix. We will also be updating username properly in backend but this won't impact how we shows displayName in UI.

sangee2004:
Tested with latest build from `main` and this issue is resolved.

1. Had both gmail and github providers configured.
2. Logged in as a user with gmail provider
 
Display Name shows the  user's actual friendly name set for the gmail account as expected:

<img width="399" height="376" alt="Image" src="https://github.com/user-attachments/assets/ded23e23-8cd6-48c2-8cfa-f6f36d2805d6" />


2. Logged in with github account that has email set to the  same gmail user as step 2 using github provider.


Display Name shows the  user's actual friendly name set for the github account as expected:

<img width="333" height="181" alt="Image" src="https://github.com/user-attachments/assets/0678bd3a-c32a-47cd-8f81-2e8df70c811d" />

PR #3486, opened 2025-07-22
Fix: properly updating username which switching provider, show displayname in UI

StrongMonkey:
We should be properly updating username when switching auth provider to sign in. Also since we have displayName support in backend, we should just use it to show it UI.

https://github.com/obot-platform/obot/issues/1783

### DIFF
--- a/pkg/gateway/client/identity.go
+++ b/pkg/gateway/client/identity.go
@@ -220,6 +220,12 @@ func (c *Client) ensureIdentity(ctx context.Context, tx *gorm.DB, id *types.Iden
 				userChanged = true
 			}
 
+			if user.Username != id.ProviderUsername {
+				user.Username = id.ProviderUsername
+				user.HashedUsername = hash.String(user.Username)
+				userChanged = true
+			}
+
 			// Update the verified email status if needed.
 			// This can happen in two cases:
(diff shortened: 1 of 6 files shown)


ivyjeong13 (approved):

pkg/gateway/client/identity.go
```diff
@@ -220,6 +220,12 @@ func (c *Client) ensureIdentity(ctx context.Context, tx *gorm.DB, id *types.Iden
 				userChanged = true
 			}
 
+			if user.Username != id.ProviderUsername {
+				user.Username = id.ProviderUsername
+				user.HashedUsername = hash.String(user.Username)
+				userChanged = true
+			}
```
thedadams, line 227: Curious about this change. The original intention was that a user should be able to change their username and we just use the username from the provider to start with.

What's the intention now?

StrongMonkey: If user changed the auth provider to login, the identify will also change but the new username is not reflected on the user. For exmaple, when user login with google, they got assigned with google username which is a random string. then later they login with github(with the same email account), when we update existing user to reflect the latest identity, we will need to update google's username to github's. This is from my understanding that username should get changed when user changed their auth provider(in this case, when email is the same and we merge accounts).

Issue #3317, opened 2025-07-03
Refactor command section of server to be specific runtimes


cjellick:
The current launch command data structure field is too open-ended and limits our ability to enforce consistent runtime behavior and support future enhancements. This issue proposes a new structured approach centered around a runtime field that impacts the expected configuration

Problem:

The existing approach of an open command and args fields mixes all launch details into a freeform command, making it difficult to:
- Validate runtime-specific configurations
- Enforce constraints for different execution environments
- Add support for future protocols or runtime types
- Track packages for upgrade paths (e.g., via npm)

Proposal:

Introduce a `runtime` field with specific supported values: npx, uvx, or containerized. The configuration fields required or allowed will vary based on the runtime:

For npx and uvx:
- The user will specify the package and optionally arguments
- State in the UI that stdio is the only supported protocol

For containerized:
- The user will specify the image, command, args, port, and path
- State in the UI that only http+streaming (or whatever the official name of the protocol is) and SSE are supported
- For containerized runtimes, the system will assume the provided image is self-contained and listening on the specified port and path. These will be exposed as Kubernetes services
- As part of this change, we should open the containerized runtime up to any images. Right now it is locked down and the admin has to set an env var to whitelist specific images (or organizations like `acorn-io`). Since this is now creating entries is fully under the control of the admin, they should be able to put any image they want in there.

For any of the above, I'm flexible on the fields we expose. Just shoot for the fields that allow for the minimal viable version of each runtime and give us the same level of functionality we currently have.

This change should apply to both single-user and multi-user MCP server configurations and both backend data structures and corresponding UI elements need to reflect this change

Future considerations:
- Later, the package field can be used to detect packages or images that have upgrades available "upstream" (outside of obot)
- For now, only stdio needs to be supported for npx and uvx, but later we could go as far as letting the user select a protocol for npx/uvx which would impact how we set up the supergateway

The above write up was mostly generated by AI  from the transcript of the this loom
https://www.loom.com/share/f1bf97ad43fe43ccbd9b5c6be2bf560e

https://www.figma.com/board/ATA31aCJyA2JVYucsy6rVv/Obot?node-id=0-1&p=f&t=nCBb3UfJxyXFj5W1-0

sangee2004:
All changes requested for the command section is available when testing with latest build.

PR #3499, opened 2025-07-23
chore: refactor manifests to use runtimes

g-linville:
for https://github.com/obot-platform/obot/issues/3317

See the Loom demo in Slack.


pkg/storage/apis/obot.obot.ai/v1/mcpservercatalogentry.go
```diff
@@ -54,8 +54,7 @@ func (in *MCPServerCatalogEntry) DeleteRefs() []Ref {
 }
 
 type MCPServerCatalogEntrySpec struct {
-	CommandManifest   types.MCPServerCatalogEntryManifest `json:"commandManifest,omitzero"`
-	URLManifest       types.MCPServerCatalogEntryManifest `json:"urlManifest,omitzero"`
```
g-linville: There was never any good reason to have two separate manifests.

thedadams: The reason there were two separate manifests is so that we could have MCP servers that had remote and STDIO versions, but were still considered the "same" MCP server catalog entry.

apiclient/types/mcpserver.go
```diff
@@ -108,3 +161,145 @@ type ProjectMCPServer struct {
 }
 
 type ProjectMCPServerList List[ProjectMCPServer]
+
+// RuntimeValidationError represents a validation error for runtime-specific configuration
+type RuntimeValidationError struct {
+	Runtime string
```
thedadams: nit:
```suggestion
Runtime Runtime
```

apiclient/types/mcpserver.go
```diff
@@ -108,3 +161,145 @@ type ProjectMCPServer struct {
...
+		Runtime:     catalogEntry.Runtime,
+		Env:         catalogEntry.Env,
+	}
+
+	// Handle runtime-specific mapping
+	switch catalogEntry.Runtime {
+	case RuntimeUVX:
+		if catalogEntry.UVXConfig == nil {
+			return serverManifest, RuntimeValidationError{
+				Runtime: string(RuntimeUVX),
```
thedadams: If you take my above suggestion, then we won't have to cast here and below.

apiclient/types/mcpserver.go
```diff
@@ -108,3 +161,145 @@ type ProjectMCPServer struct {
...
+			return serverManifest, RuntimeValidationError{
+				Runtime: string(RuntimeRemote),
+				Field:   "remoteConfig",
+				Message: "either fixedURL or hostname must be specified in catalog entry",
+			}
+		}
+
+		// Copy headers from catalog entry
+		remoteConfig.Headers = catalogEntry.RemoteConfig.Headers
+		serverManifest.RemoteConfig = remoteConfig
```
thedadams, line 279: Is it worth setting Env to nil in this case?

pkg/api/handlers/mcp.go
```diff
@@ -321,7 +320,7 @@ func (m *MCPHandler) CheckOAuth(req api.Context) error {
 		return types.NewErrNotFound("MCP server not found")
 	}
 
-	if server.Spec.Manifest.URL != "" {
+	if server.Spec.Manifest.RemoteConfig != nil && server.Spec.Manifest.RemoteConfig.URL != "" {
```
thedadams: Previously, this was only checking that this was a "remote" MCP server. I think we would want to keep the same intent (and below).

```suggestion
	if server.Spec.Manifest.Runtime == types.RuntimeRemote {
```

pkg/api/handlers/mcp.go
```diff
@@ -1130,22 +1118,53 @@ func addExtractedEnvVars(server *v1.MCPServer) {
...
-		extracted[v] = struct{}{}
-	}
 
-	// Extract variables from args
-	for _, arg := range server.Spec.Manifest.Args {
-		for _, v := range extractEnvVars(arg) {
-			extracted[v] = struct{}{}
-		}
-	}
+	switch server.Spec.Manifest.Runtime {
```
thedadams, line 1152: nit: maybe some of this could be simplified by using the switch to get the `command` and the `args` and then constructing the `extracted` map more generically.

Same for the below change.

pkg/api/handlers/projectmcp.go
```diff
@@ -224,7 +224,7 @@ func (p *ProjectMCPHandler) CheckOAuth(req api.Context) error {
 		return err
 	}
 
-	if server.Spec.Manifest.URL != "" {
+	if server.Spec.Manifest.Runtime == types.RuntimeRemote && server.Spec.Manifest.RemoteConfig != nil && server.Spec.Manifest.RemoteConfig.URL != "" {
```
thedadams: Same comment here about maintaining the intention and below.

pkg/controller/routes.go
```diff
@@ -232,13 +232,11 @@ func (c *Controller) setupRoutes() error {
 
 	// MCPServerCatalogEntry
 	root.Type(&v1.MCPServerCatalogEntry{}).HandlerFunc(cleanup.Cleanup)
-	root.Type(&v1.MCPServerCatalogEntry{}).HandlerFunc(mcpservercatalogentry.RemoveArgsOnRemoteEntries)
 	root.Type(&v1.MCPServerCatalogEntry{}).HandlerFunc(mcpservercatalogentry.EnsureUserCount)
 
 	// MCPServer
 	root.Type(&v1.MCPServer{}).HandlerFunc(cleanup.Cleanup)
 	root.Type(&v1.MCPServer{}).HandlerFunc(mcpserver.DetectDrift)
-	root.Type(&v1.MCPServer{}).HandlerFunc(mcpserver.MigrateProjectMCPServers)
```
thedadams, line 241: Was this removal intentional? If so, can we remove the function, too?

pkg/mcp/k8s.go
```diff
@@ -0,0 +1,365 @@
...
+	"github.com/obot-platform/nah/pkg/name"
+	otypes "github.com/obot-platform/obot/apiclient/types"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	kclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func (sm *SessionManager) k8sObjectsForUVXOrNPX(server ServerConfig, key, serverName string) ([]kclient.Object, error) {
```
thedadams, line 17: Can we further consolidate the common logic here?

pkg/mcp/loader.go
```diff
@@ -125,7 +135,11 @@ func (sm *SessionManager) CloseClient(ctx context.Context, server ServerConfig)
 	if len(pods.Items) != 0 {
 		// If the pod was removed, then this won't do anything. The session will only get cleaned up when the server restarts.
 		// That's better than the alternative of having unusable sessions that users are still trying to use.
-		if err = sm.local.ShutdownServer(gmcp.ServerConfig{URL: fmt.Sprintf("http://%s.%s.svc.%s/sse", id, sm.mcpNamespace, sm.mcpClusterDomain), Scope: pods.Items[0].Name}); err != nil {
+		path := "sse"
+		if server.ContainerPath != "" {
+			path = strings.TrimPrefix(server.ContainerPath, "/")
+		}
```
thedadams: This is leftover from when we used supergateway. This path can be empty unless specified by the server.

```suggestion
		path := server.ContainerPath
```

pkg/mcp/loader.go
```diff
@@ -262,208 +266,46 @@ func (sm *SessionManager) ensureDeployment(ctx context.Context, server ServerCon
...
 	}
 
 	u := fmt.Sprintf("http://%s.%s.svc.%s", id, sm.mcpNamespace, sm.mcpClusterDomain)
-	podName, err := sm.updatedMCPPodName(ctx, u, id)
+	podName, err := sm.updatedMCPPodName(ctx, u, id, server)
 	if err != nil {
 		return gmcp.ServerConfig{}, err
 	}
 
+	fullURL := fmt.Sprintf("%s/sse", u)
```
thedadams: Similar comment here about dropping `sse`.

pkg/mcp/loader.go
```diff
@@ -492,20 +334,91 @@ func (sm *SessionManager) updatedMCPPodName(ctx context.Context, url, id string)
...
+						if err == nil {
+							req.Header.Set("Mcp-Session-Id", sessionID)
+							_, _ = http.DefaultClient.Do(req)
+						}
+					}
+					break
+				}
+			}
+
+			if err != nil || (resp != nil && resp.StatusCode != http.StatusOK) {
```
thedadams: If the err is not nil, then we don't have to try SSE. I think continuing in that case is best (and I would do it above so we don't have to do the `err == nil` check.

pkg/mcp/loader.go
```diff
@@ -492,20 +334,91 @@ func (sm *SessionManager) updatedMCPPodName(ctx context.Context, url, id string)
...
+				// Fallback to trying SSE.
+				req, err := http.NewRequest(http.MethodGet, url, nil)
+				if err != nil {
+					return "", fmt.Errorf("failed to create request: %w", err)
+				}
+				req.Header.Set("Accept", "text/event-stream")
+
+				resp, err := http.DefaultClient.Do(req)
+				if err == nil {
```
thedadams: Same comment here about the error: if it is not nil, then just continue.

pkg/mcp/loader.go
```diff
@@ -492,20 +334,91 @@ func (sm *SessionManager) updatedMCPPodName(ctx context.Context, url, id string)
...
+					defer resp.Body.Close()
```
thedadams: Deferring inside of a for loop is typically not a good idea because the deferred functions won't be called until the function returns. The typical fix is to create a helper function.

pkg/mcp/loader.go
```diff
@@ -492,20 +334,91 @@ func (sm *SessionManager) updatedMCPPodName(ctx context.Context, url, id string)
...
+					if resp.StatusCode == http.StatusOK {
+						ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+						defer cancel()
+
+						// Start looking for an event with "endpoint".
+						scanner := bufio.NewScanner(resp.Body)
+					scannerLoop:
+						for scanner.Scan() {
```
thedadams: After the for-loop exits, typically you'd check `scanner.Err()`.

pkg/mcp/loader.go
```diff
@@ -262,208 +266,46 @@ func (sm *SessionManager) ensureDeployment(ctx context.Context, server ServerCon
 				}
 			}
 		}
-		// Either we aren't deploying to Kubernetes, or this is a URL-based MCP server (so there is nothing to deploy to Kubernetes).
+		// Either we aren't deploying to Kubernetes, or this is a remote MCP server (so there is nothing to deploy to Kubernetes).
```
thedadams, line 265: Do we care about containerized application working when we aren't running in Kubernetes? My interpretation of these changes is that it won't work.

apiclient/types/mcpserver.go
g-linville: I don't think so, because I think remotes can still use Env to templatize some of the parameters with `${}` syntax.

pkg/controller/routes.go
g-linville: Yes, and I already removed the function. There is no need for migration code anymore when everything will be destroyed anyway.

pkg/mcp/loader.go
g-linville: No, I don't think we do.

g-linville: That would require a lot more engineering work that we, at the very least, don't have time for right now.

pkg/mcp/k8s.go
g-linville: Do you mean consolidating the functions that set up k8s objects for UVX/NPX and for Containerized? Tbh I would rather keep them completely separate, as I think that would make it easier to iterate on.

pkg/mcp/loader.go
g-linville: What should we check it for? I'm not sure that we really care what it is. It's the end of the loop execution anyway, so I think it makes the most sense to just ignore it.

pkg/api/handlers/mcp.go
```diff
@@ -349,7 +348,7 @@ func (m *MCPHandler) GetOAuthURL(req api.Context) error {
 	}
 
 	var u string
-	if server.Spec.Manifest.URL != "" {
+	if server.Spec.Manifest.RemoteConfig != nil && server.Spec.Manifest.RemoteConfig.URL != "" {
```
```suggestion
	if server.Spec.Manifest.Runtime == types.RuntimeRemote {
```

pkg/api/handlers/mcp.go
```diff
@@ -1231,34 +1189,79 @@ func addExtractedEnvVarsToCatalogEntry(entry *v1.MCPServerCatalogEntry) {
...
-				entry.Spec.URLManifest.Env = append(entry.Spec.URLManifest.Env, types.MCPEnv{
-					MCPHeader: types.MCPHeader{
-						Name:        v,
-						Key:         v,
-						Description: "Automatically detected variable",
-						Sensitive:   true,
-						Required:    true,
-					},
-				})
+	switch entry.Spec.Manifest.Runtime {
```
thedadams, line 1205: Similar comment here about simplifying this logic as you did above.

pkg/mcp/types.go
```diff
@@ -100,6 +105,105 @@ func ServerToServerConfig(mcpServer v1.MCPServer, scope string, credEnv map[stri
 	return serverConfig, missingRequiredNames, nil
 }
 
+func ServerToServerConfig(mcpServer v1.MCPServer, scope string, credEnv map[string]string, allowedTools ...string) (ServerConfig, []string, error) {
+	if string(mcpServer.Spec.Manifest.Runtime) == "" {
+		return legacyServerToServerConfig(mcpServer, scope, credEnv, allowedTools...)
+	}
```
thedadams, line 111: Such MCP servers are getting removed by your controller, so I say this can be removed, too.

pkg/mcp/types.go
```diff
@@ -100,6 +105,105 @@ func ServerToServerConfig(mcpServer v1.MCPServer, scope string, credEnv map[stri
...
+
+	serverConfig := ServerConfig{
+		ServerConfig: gmcp.ServerConfig{
+			Env:          make([]string, 0, len(mcpServer.Spec.Manifest.Env)),
+			Headers:      make([]string, 0),
```
thedadams: nit: we never use this allocated memory.
```suggestion
```

pkg/api/handlers/mcp.go
```diff
@@ -1231,34 +1189,79 @@ func addExtractedEnvVarsToCatalogEntry(entry *v1.MCPServerCatalogEntry) {
...
+	}
+
+	// Add any new vars to the manifest's Env list
+	for v := range extracted {
+		if _, exists := existing[v]; !exists {
+			entry.Spec.Manifest.Env = append(entry.Spec.Manifest.Env, types.MCPEnv{
+				MCPHeader: types.MCPHeader{
+					Name:        v,
+					Key:         v,
+					Description: "Automatically detected variable",
```
StrongMonkey: what does Automatically detected variable mean?

pkg/gateway/types/identity.go
```diff
@@ -1,3 +1,4 @@
+//nolint:revive
```
StrongMonkey, line 1: any reason to add this?

ui/user/src/lib/components/mcp/MyMcpServers.svelte
```diff
@@ -250,9 +250,8 @@
 		error = undefined;
 		saving = true;
 		try {
-			const isCatalogEntry =
-				'urlManifest' in selectedEntryOrServer || 'commandManifest' in selectedEntryOrServer;
-			if (isCatalogEntry) {
+			console.log('selected: ', JSON.stringify(selectedEntryOrServer));
```
StrongMonkey: debug log

pkg/api/handlers/mcp.go
g-linville: This isn't new, just moved from somewhere else.

When we detect things inside of `${}` in some of the configuration fields, we interpolate those as environment variables, and have to add them to the server manifest as something for the user to fill in.

pkg/gateway/types/identity.go
g-linville: The `revive` part of golangci-lint hates the package name `types` and annoyingly throws a warning about it.

ui/user/src/lib/components/mcp/MyMcpServers.svelte
g-linville: Thanks

pkg/mcp/types.go
g-linville: This is actually needed for the finalizer to complete properly.

g-linville:
@thedadams as for the style issues you mentioned, the search box is like that on main too, so I'm not gonna try to figure out how to fix it, since it is not obvious to me. I am fixing the other two style problems.

thedadams:
One more round of UI things.

When I create a single-user MCP server and go to the tools tab, I am greeted with an error. I don't believe we should try to look at the tools for such a server because it isn't running until someone connects to it.

Screenshot 2025-08-01 at 09 09 58

Also, in that above screenshot, the arrows near the top for the tabs don't work. It is worth noting that the arrows do work for multi-user MCP servers. We can also list the tools in that case, so the two issues might be related.

Secondly, when I try to connect to the single-user MCP server from the "My Connectors" page, I am first greeted by the same tools error. After I connect to the server, everything works as expected.

Screenshot 2025-08-01 at 09 11 46

The last thing to note is that both of these are issues for remote servers, too.

thedadams:
I submitted a fix to address these last few UI issues I found. I think we're good to go here.

StrongMonkey (approved):

Issue #3381, opened 2025-07-14
Add k8s info in user-ui


ivyjeong13:
When an admin clicks a Single or Multi-User server in a MCP Server's 'Server Instances', they can see the k8s info and log details. Use API endpoints created by Grant, from https://github.com/obot-platform/obot/pull/3361

sangee2004:
@ivyjeong13 I am able to see the `Server Instances` tab for `Single User` and `Remote` MCP servers being created by users.  We are not presented with `User` info like we do for `Multi User` MCP servers.

`Single User` MCP server( `Remote` MCP server also shows the same behavior)
<img width="1157" height="389" alt="Image" src="https://github.com/user-attachments/assets/2b3267f8-cbc2-4f3a-97d2-c79415c20a27" />

`Multi User` MCP server
<img width="1171" height="309" alt="Image" src="https://github.com/user-attachments/assets/ca22f9d6-a469-4223-9da4-c3ce0b97decd" />

Reopening to follow up on this.

ivyjeong13:
Just need to address the bug Sangeeth posted above.

cjellick:
In addition to the above bug, a few observations for improvements:
1. I want the logs to be "live". I think it would be sufficient if we just refreshed every 5 seconds, but if we can actually stream, that would be even better. Bare minimum, a manual refresh button for the logs.
2. I looked at a few servers and didnt see any events. I am suspicious of that. We should investigate if we are pulling events correclty or not.
3. I'd like a redeploy or restart button (that just deletes the pod and lets the deployment recreate it.

ivyjeong13:
> In addition to the above bug, a few observations for improvements:
> 
> 1. I want the logs to be "live". I think it would be sufficient if we just refreshed every 5 seconds, but if we can actually stream, that would be even better. Bare minimum, a manual refresh button for the logs.
> 2. I looked at a few servers and didnt see any events. I am suspicious of that. We should investigate if we are pulling events correclty or not.
> 3. I'd like a redeploy or restart button (that just deletes the pod and lets the deployment recreate it.

For #1, for the event logs the endpoint is an EventStream we're registering to so it should be updating as messages come in!

StrongMonkey:
We do pull events correcly but events are usually cleaned up after 60 minutes so likely you won't see anything after 1 hours.

<img width="1494" height="464" alt="Image" src="https://github.com/user-attachments/assets/921cd6b9-3399-48d9-a862-f18654e220fe" />

[...]

PR #3510, opened 2025-07-23
Feat: Add deployment restart button

StrongMonkey:
This PR added an endpoint to restart k8s deployment, and also buttons in frontend to be able to refresh k8s details, events and logs. 

https://github.com/obot-platform/obot/issues/3381

### DIFF
--- a/pkg/mcp/loader.go
+++ b/pkg/mcp/loader.go
@@ -150,6 +150,9 @@ func (sm *SessionManager) ShutdownServer(ctx context.Context, server ServerConfi
 
 // RestartK8sDeployment restarts the Kubernetes deployment by deleting its pods, which will trigger a restart.
 func (sm *SessionManager) RestartK8sDeployment(ctx context.Context, server ServerConfig) error {
+	if server.Command == "" {
+		return nil
+	}
 	id := sessionID(server)
 
 	var deployment appsv1.Deployment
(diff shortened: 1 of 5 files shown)


pkg/mcp/loader.go
```diff
@@ -148,6 +148,29 @@ func (sm *SessionManager) ShutdownServer(ctx context.Context, server ServerConfi
 	return nil
 }
 
+// RestartK8sDeployment restarts the Kubernetes deployment by deleting its pods, which will trigger a restart.
+func (sm *SessionManager) RestartK8sDeployment(ctx context.Context, server ServerConfig) error {
+	id := sessionID(server)
```
thedadams, line 156:
```suggestion
	if server.Command == "" {
		return nil
	}
	id := sessionID(server)
```

thedadams (changes requested):

pkg/mcp/loader.go
StrongMonkey: Updated. In which case server.command is empty and why should we avoid it?

thedadams: If `server.Command` is empty, then this is a remote MCP server which doesn't have a corresponding deployment.

thedadams (approved):

PR #3521, opened 2025-07-24
Feat: add allowModels in base agent

StrongMonkey:


pkg/gateway/server/llmproxy.go
```diff
@@ -52,7 +54,7 @@ func (s *Server) llmProxy(req api.Context) error {
 
 	// If the model string is different from the model, then we need to look up the model in our database to get the
 	// correct model and model provider information.
-	if modelProvider == "" || modelStr != token.Model {
+	if modelProvider == "" || modelStr != token.Model || isModelAllowedByAssistant(req.Context(), req.Storage, token, modelStr) {
```
thedadams: The name of the function seems to indicate that it is trying to prevent this request from going through if the assistant doesn't allow the model. But that's not what is happening here. This is just checking to see if we need to look up the model and model provider information. This is sort of leftover from a "feature" where we were going to allow users to provide their own API keys for model providers.

Additionally, wouldn't we also want to look at the project to see if it allowed the model?

Regardless, I think it's safe to remove this check because we would have caught this in the invoker.

StrongMonkey: let me double check on this. The method is not accurate, but it is checking if user sets the model from thread that is from allowedModel in baseAgent so that we don't look up cred in project scope. Otherwise things will fail because it can't look up project scope. But let me check if there is a better way

thedadams: I guess I don't understand why `modelProvider == "" || modelStr != token.Model` is true if we are just using a model that is allowed by the assistant. I would expect that `modelStr == token.Model` at least so that we go into this block.

StrongMonkey: Added another condition so that we don't go back to look up project creds because it will fail. 

where you set model in thread, both `modelProvider == ""` `modelStr != token.Model` appear to be false though so it fallback to look up project scope.

StrongMonkey: Talked offline, modelProvider on the client should be set to empty if it is using system's model. Removed check on the llmproxy code now.

thedadams (approved):

Issue #3464, opened 2025-07-21
Audit Logs - Search String gets cleared when traversing across pages.


sangee2004:
**Describe the bug**
Audit Logs - Search String gets cleared when traversing across pages.

**To Reproduce**
1. As admin user, go to `Audit Logs`
2. Search for any specific MCP server by typing the MCP server name in Search field.
3. Go to the next page.

Notice that the entered  Search String gets cleared when traversing across pages.

https://www.loom.com/share/45ac4bcbdcb04ccaac76371070ec5901?sid=297f5774-2be6-4ef5-8e99-73328ca0596e

**Expected behavior**
Search String should not get cleared when traversing across pages.

sangee2004:
This issue is addressed when testing with latest builds.

PR #3525, opened 2025-07-24
fix: ui render on pagination page change

ryu-man:
Addresses #3464

- Move promise await to the script section
- Use an absolute container to display the loading spinner

### DIFF
--- a/ui/user/src/lib/components/admin/audit-logs/AuditDetails.svelte
+++ b/ui/user/src/lib/components/admin/audit-logs/AuditDetails.svelte
@@ -29,6 +29,8 @@
 	import AuditLogDetails from './AuditLogDetails.svelte';
 	import AuditFilters from './AuditFilters.svelte';
 	import { goto } from '$app/navigation';
+	import { delay } from '$lib/utils';
+	import { fade } from 'svelte/transition';
 
 	interface Props {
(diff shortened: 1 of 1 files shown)


ivyjeong13:
I think what we may want is a user typing into Search actually triggers a whole new fetch call to get data and applies the necessary filters. Cause right now, Search is just filtering out rows that don't match based on the current page data (like the user gets 2 pages with 100 limit, but we only show 10 of them and they have another page?)

Was thinking something along the lines of DataDog search, where a user would have to tie the search to a string value we can filter by text (like displayName or mcpId, etc.)

image

cjellick:
yeah, what @ivyjeong13 said is correct.

ryu-man:
> I think what we may want is a user typing into Search actually triggers a whole new fetch call to get data and applies the necessary filters. Cause right now, Search is just filtering out rows that don't match based on the current page data (like the user gets 2 pages with 100 limit, but we only show 10 of them and they have another page?)
> 
> Was thinking something along the lines of DataDog search, where a user would have to tie the search to a string value we can filter by text (like displayName or mcpId, etc.)
> 
> image

@cjellick @ivyjeong13 Should I implement this behavior in this PR or in a separate one since the issue is related to the search value being cleared on the page?

Issue #3346, opened 2025-07-09
Adding description in the Overview section does not get reflected when entering the configuration tab.


sangee2004:
**Describe the bug**
Adding description in the Overview section does not get reflected when entering the configuration tab.

**To Reproduce**
1. Create a MCP server entry with no description.
2. Select this MCP server entry from `MCP Servers` page
3. When in the "Overview" tab , enter some description.

Notice that there is no `update` option here.
 
4. Click on `Configuration` tab.

Notice that the description that was entered in the `Overview` tab is not reflected here.

Go to `Overview` tab  and then come back to  `Configuration` tab. 

Notice that you will see the changes reflected in  `Configuration` tab now.

https://www.loom.com/share/9fd118026bdb4999b1988ea958bc0a4d?sid=4cec0c13-69fc-4b44-bec5-fd4fecc54be7

**Expected behavior**
Changes added to the `Overview` tab should get reflected in the `Configuration` tab. 
Also do we need an `Update` button in  `Overview` tab similar to `Configuration` tab?

sangee2004:
Tested with latest build from `main`

If there is no description for an MCP server, accessing the `Overview` tab presents the following message:
<img width="805" height="251" alt="Image" src="https://github.com/user-attachments/assets/5f89fefa-4522-4744-8776-fc33bf16efd6" />
If there is a description set for an MCP server, accessing the `Overview` tab presents the description but it is not editable.

PR #3532, opened 2025-07-24
fix: remove support for editing mcp server descriptions from overview

njhale:
Simplify the admin UX for editing MCP server descriptions by:
- removing the ability to edit descriptions in the Overview tab
- make the Configuration tab the only place descriptions may be edited
- adding some text directing admins to use the Configuration tab when
  the description is empty

https://github.com/user-attachments/assets/e4c2e4ad-4257-4e5e-a4aa-fc950b42908c

https://github.com/user-attachments/assets/3b10173c-fca9-44f5-95f1-051e4b18f61e

Addresses https://github.com/obot-platform/obot/issues/3346

### DIFF
--- a/ui/user/src/lib/components/mcp/McpServerInfo.svelte
+++ b/ui/user/src/lib/components/mcp/McpServerInfo.svelte
@@ -5,27 +5,24 @@
 		type MCPServerPrompt,
 		type McpServerResource,
 		ChatService,
-		AdminService,
 		type Project
 	} from '$lib/services';
-	import type { MCPCatalogEntry, MCPCatalogServerManifest } from '$lib/services/admin/types';
-	import { CircleCheckBig, CircleOff, Info, LoaderCircle, Pencil, RefreshCcw } from 'lucide-svelte';
+	import type { MCPCatalogEntry } from '$lib/services/admin/types';
+	import { CircleCheckBig, CircleOff, Info, LoaderCircle, RefreshCcw } from 'lucide-svelte';
 	import { twMerge } from 'tailwind-merge';
 	import McpServerTools from './McpServerTools.svelte';
 	import { formatTimeAgo } from '$lib/time';
 	import { responsive } from '$lib/stores';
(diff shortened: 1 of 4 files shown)


ui/user/src/lib/components/mcp/McpServerInfo.svelte
```diff
@@ -16,16 +15,14 @@
 	import { responsive } from '$lib/stores';
 	import { toHTMLFromMarkdownWithNewTabLinks } from '$lib/markdown';
 	import { tooltip } from '$lib/actions/tooltip.svelte';
-	import MarkdownTextEditor from '../admin/MarkdownTextEditor.svelte';
 	import { onDestroy } from 'svelte';
 
 	interface Props {
 		entry: MCPCatalogEntry | MCPCatalogServer;
-		editable?: boolean;
```
njhale, line 24: From what I could tell, the only component that used `McpServerInfo` in "description edit mode" was `McpServerEntryForm`, which also has the configuration tab for editing, so this is just dropping dead code.

ui/user/src/lib/components/mcp/McpServerInfo.svelte
```diff
@@ -261,41 +242,14 @@
...
-			{/if}
-		{:else if description}
+		{#if description}
 			<div class="milkdown-content">
 				{@html toHTMLFromMarkdownWithNewTabLinks(description)}
 			</div>
+		{:else}
+			<p class="text-md text-center font-light text-gray-500 italic">
+				{descriptionPlaceholder}
+			</p>
```
njhale, line 252: Note: A decent side effect of this is that we now get the `No description available` instead of an empty box in every other context where `McpServerInfo` is being used; e.g. chat and `/mcp-servers`

Image

cjellick (approved):

njhale:
Seems like some components may have changed since I submitted this PR. Taking into draft to resolve.

Issue #3482, opened 2025-07-22
Audit Log - Search filters are not applied to stats.


sangee2004:
**Describe the bug**
Audit Log - Search filters are not applied to stats.

**To Reproduce**
1. Go to Audit Log
2. Click on` Filter Logs` option and Select a value from `By User` and Apply Filters (any filter can be used)

Notice that the stats presented at the top of the graph does not account for the search filter applied

<img width="1136" height="587" alt="Image" src="https://github.com/user-attachments/assets/ca441aaa-7f37-407f-93a3-23f3a7d67939" />

Api call made for fetching stats by UI  does not pass userid as query parameter - `/api/mcp-stats?start_time=2025-07-15T20%3A23%3A36.307Z&end_time=2025-07-22T20%3A23%3A36.307Z`

**Expected behavior**
Stats should be presented for the search filters.

cjellick:
@thedadams does the stats endpoint have the same filters as audit logs?

thedadams:
@cjellick It doesn't have the same filters as the audit logs. Right now, you can filter by date and MCP server ID. I can add user ID. Any others?

sangee2004:
This issue is not seen anymore when testing with latest build.

Timeline graphs that is presented in the `Audit Logs` page works as expected when filters are applied to it.

PR #3534, opened 2025-07-24
pass filters to `listAuditLogUsageStats`

ryu-man:
Addresses #3482


ryu-man:
@cjellick It looks like the stat endpoint does not support all the filters of the logs endpoint. Currently, I managed to make the results synced when filtering by MCP name.

Issue #3321, opened 2025-07-04
Catalog - Adding/deleting/updating catalog entries in multi server catalog json file from github does not get synced to MCP catalog on Manual Refresh.


sangee2004:
**Describe the bug**
Adding/deleting/updating catalog entries in multi catalog json file in github does not get synced to MCP catalog on Manual Refresh.

**To Reproduce**
1. Add MCP Servers using `Add servers from Git` from  a github repository that has multiple server entries in 1 json file say `catalog.json`

```
[
      {
    "name": "sangee2 test1",
    "repoURL": "https://example.com/3",
    "description": "Test 1",
    "metadata": {
      "categories": "Retrieval & Search"
    },
    "env": [],
    "args": [
      "--version"
    ],
    "command": "npx"
  },
      {
    "name": "sangee2 test2",
    "repoURL": "https://example.com/3",
    "description": "Test 2",
    "metadata": {
      "categories": "Retrieval & Search"
    },
    "env": [],
    "args": [
      "--version"
    ],
    "command": "npx"
  }
]
```
2. Do a `Refresh` from `MCP Servers` page
3. This results in both MCP Server entries to be synced as expected.
4. From github, modify `catalog.json` and include a new MCP Server entry  say- `sangee3 test3` 
5. From obot server, do a `Refresh` from `MCP Servers` page

New MCP Server entry that was added in step 4 is not reflected in the `MCP Servers` list
Attempting a manual refresh does not also result in new entry being fetched.

6.  From github, modify `catalog.json` and delete an existing MCP Server entry  say- `sangee2 test2`
7. From obot server, do a `Refresh` from `MCP Servers` page

MCP Server entry that was deleted in step 6 is not reflected in the `MCP Servers` list
Attempting a manual refresh does not also result in entry being removed.

8.  From github, modify `catalog.json` and modify an existing MCP Server entry  say rename- `sangee1 test1` entry  and modify name to `sangee1 test1updated`
9. From obot server, do a `Refresh` from `MCP Servers` page

MCP Server entry that was updated  in step 8 is not reflected in the `MCP Servers` list
Attempting a manual refresh does not also result in entry being updated.

**Expected behavior**
All changes made to multi server catalog json files from github should get synced to `MCP Servers` list

StrongMonkey:
[...]

PR #3540, opened 2025-07-24
Fix: prevent catalog from refreshing within a hour

StrongMonkey:
We should prevent catalog from syncing within the same hour(unless there is a force resync event). Also, stop applying if we hit any error so we don't accidently delete all the catalogs.

https://github.com/obot-platform/obot/issues/3321

### DIFF
--- a/pkg/controller/handlers/mcpcatalog/mcpcatalog.go
+++ b/pkg/controller/handlers/mcpcatalog/mcpcatalog.go
@@ -47,8 +47,17 @@ func New(allowedDockerImageRepos []string, defaultCatalogPath string, gatewayCli
 
 func (h *Handler) Sync(req router.Request, resp router.Response) error {
 	mcpCatalog := req.Object.(*v1.MCPCatalog)
-	toAdd := make([]client.Object, 0)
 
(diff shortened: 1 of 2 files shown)


pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -67,11 +76,27 @@ func (h *Handler) Sync(req router.Request, resp router.Response) error {
...
 	// TODO(g-linville): make this configurable.
 	resp.RetryAfter(time.Hour)
 
+	// Don't run apply if there are sync errors
+	if len(mcpCatalog.Status.SyncErrors) > 0 {
+		errors := []string{}
+		for url, err := range mcpCatalog.Status.SyncErrors {
+			errors = append(errors, fmt.Sprintf("%s: %s", url, err))
+		}
+		log.Warnf("Skipping apply for catalog %s due to sync errors, errors: %v", mcpCatalog.Name, strings.Join(errors, ", "))
```
thedadams: nit: we already logged these above, so can we remove this one or the other one?

StrongMonkey: Ah, good point

thedadams (approved):

g-linville (approved):

Issue #3470, opened 2025-07-21
feat: admin v2 Setting up Default Chat


ivyjeong13:
Admin can set the default models, available models, and default system prompt

cjellick:
Was talking to will and we should also expose Introduction, name, and icon. We will want to "brand" the chat experience you get when you log in for the first time with these.

sangee2004:
When testing with latest build from main `Chat Configuration` shows the following fields that can be set


<img width="1360" height="941" alt="Image" src="https://github.com/user-attachments/assets/8ef0eefc-aac4-4cea-9a33-9bc2457c8451" />


`Allowed Models` set here, gets used in the chat view as expected.

Introduction and icons set here dont get used in the chat view. There are more changes getting done in the chat view which is tracked in https://github.com/obot-platform/obot/issues/3548.


Closing this issue for now.

PR #3544, opened 2025-07-25
feat: chat configuration admin page

ivyjeong13:
Addresses #3470 

Related Loom: https://www.loom.com/share/d6dbda4f0df442c6ae75f0708153e9e0

### DIFF
--- a/ui/user/src/lib/components/Layout.svelte
+++ b/ui/user/src/lib/components/Layout.svelte
@@ -106,7 +106,6 @@
 		if (responsive.isMobile) {
 			layout.sidebarOpen = false;
 		}
-		console.log(window.location.pathname);
 	});
 
 	initLayout();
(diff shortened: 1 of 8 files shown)


cjellick:
Looking good! Here's some feedback:
1. We need both instructions and introduction
2. Somewhat related, I opened this issue regarding the introduction message and how the initial chat view should look: https://github.com/obot-platform/obot/issues/3548. Give that a read and watch so that you can understand how the introduction will be used. 
3. We should not have the Model Providers section at all on this page. Model Providers on the base agent is an unused feature at this point. We should only have the allowed models bit and your dropdown of models to choose from should be driven by whatever the admin has configured on the Model Providers page.
4. We need to be able set the default model as well.
5. For openAI, do some sensible sorting of models. I think in the past this has worked to alphabetical, but bubble any starting with `gpt-` to the top, as they are the most common ones to be picked

cjellick (changes requested):
see above comment.

PR #3561, opened 2025-07-25
chore: doc revamp

drpebcak:


docs/docs/50-configuration/04-model-providers.md
```diff
@@ -1,8 +1,9 @@
 # Model Providers
 
+@TODO ensure this is still accurate
```
drpebcak: I left this in here - currently i dont see a way to define defaults for the different model types. is that gone for good? what happens if i DONT configure OpenAI?

docs/docs/01-overview.md
```diff
@@ -5,19 +5,19 @@ slug: /
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
-Obot is an open source AI assistant platform offered as a SaaS and self-hosted deployments. The Obot SaaS platform is built on top of the same open source technology that you can deploy yourself, offering flexibility in how you access and control your AI assistants.
+Obot is an open source AI assistant platform that can be deployed self-hosted in the cloud or on-prem.
```
cloudnautique: This is probably going to go through several iterations. We should be leading with the MCP Gateway, Directory, and Chat.

drpebcak: Fair, @cjellick  mentioned the same thing. He's going to put together some thoughts and then I'll be restructuring this to focus on those 3 areas.

Issue #3562, opened 2025-07-25
`403 /model-providers: forbidden` error seen when chatting.


sangee2004:
**Describe the bug**
`403 /model-providers: forbidden` error seen when chatting.

**To Reproduce**
1. As regular user, Connect with any MCP server.
2. Chat with it.

Notice that user is presented with `403 /model-providers: forbidden` error 

<img width="1068" height="244" alt="Image" src="https://github.com/user-attachments/assets/71dd49d6-b01e-42c7-99a3-cea536f9ce99" />

`/api/model-providers` call made by UI  gets `403 Forbidden` errors 

**Expected behavior**
There should be no errors presented to the user when chatting.

sangee2004:
This issue is not seen anymore when testing with latest builds.

PR #3565, opened 2025-07-25
Fix: allow authed group to get model providers

StrongMonkey:
https://github.com/obot-platform/obot/issues/3562

We need to get model provider in order to show properly model dropdown from chat views. I scanned through the model provider api list and there is no sensetive information, so we should be fine to expose that to regular users.


cjellick:
Why is this needed now when it wasn't before?

StrongMonkey:
@cjellick before we rely on project level model provider which is visiable on the project, now we list the model provider globally.

cjellick (approved):
ok, makes sense. LGTM

Issue #3543, opened 2025-07-25
Issues seen when configuring model provider.


sangee2004:
**Describe the bug**
Issues seen when configuring model providers.

Pre Req - Auth enabled setup
**To Reproduce**
1. Log in as bootstrap user.
2.  Access `Model Providers` page.
3. Configure OpenAI model provider.
Once model provider is configured notice that we are not presented with a flow to set `Default Models` automatically.
4. Click on  `Set Default Models` 

Notice that the models dont show up in the drop down selection list.

5. Manually refresh the page.

After this accessing `Set Default Models` page shows the models in the dropdown values.
But they do not preselect the `suggested` values .

https://www.loom.com/share/6af9df2fc8564c729c6da5ee01d6a9ea?sid=fba0a4e7-cc01-46f6-8ed9-39307163ca7f

**Expected behavior**
Right after model provider is configured in step3 , user should be presented with the  flow to set `Default Models` . In the `Default Model Aliases` page ,  suggested defaults should already be preselected for all entries.

sangee2004:
This issue is addressed when testing with latest builds.

User is presented with the following page to set default models with suggested model pre selected as expected.

<img width="766" height="591" alt="Image" src="https://github.com/user-attachments/assets/8f31e76e-bdc9-4e11-932f-2619d2ef724f" />

PR #3570, opened 2025-07-28
fix: empty default model aliases dropdowns

njhale:
Modify the `v2/admin/model-providers` UI route such that:
- The `DefaultModels` component is automatically opened after configuring a model provider IFF the provider adds models that can be used for unconfigured aliases
- Poll until the providers models are back-populated before checking if we should automatically open the default model aliases dialog (w/ a 10 second timeout)

Modify the `DefaultModels` component such that:
- The set of available providers is plumbed down as a property
- The `Set Default Models` button is disabled when no models are available
- The `Save` button is disabled until the selected models differ from the current default model aliases
- When the current default model alias for a usage isn't set, the suggested model is automatically selected from available models

Addresses https://github.com/obot-platform/obot/issues/3543

https://github.com/user-attachments/assets/6595f509-2723-4ab3-b669-668f0e0fb23d

### DIFF
--- a/ui/user/src/lib/components/admin/DefaultModels.svelte
+++ b/ui/user/src/lib/components/admin/DefaultModels.svelte
@@ -10,11 +10,10 @@
 	import { onMount } from 'svelte';
 	import ResponsiveDialog from '../ResponsiveDialog.svelte';
 	import { AdminService } from '$lib/services';
-	import { getAdminModels } from '$lib/context/admin/models.svelte';
 	import Select from '../Select.svelte';
 	import { LoaderCircle } from 'lucide-svelte';
 
-	const adminModels = getAdminModels();
+	let { availableModels }: { availableModels: Model[] } = $props();
 	let dialog = $state<ReturnType<typeof ResponsiveDialog>>();
 	let defaultModelAliases = $state<DefaultModelAlias[]>([]);
 	let sortedModelAliases = $derived(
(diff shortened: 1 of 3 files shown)


ui/user/src/lib/components/admin/DefaultModels.svelte
```diff
@@ -33,10 +39,46 @@
...
 		defaultModelAliases = await AdminService.listDefaultModelAliases();
 	});
 
+	function setSuggestedModels() {
+		if (!defaultModelAliases.length || !availableModels.length) return;
+
+		const suggestedChanges: Partial<Record<ModelAlias, string>> = {};
+		for (const modelAlias of defaultModelAliases) {
+			// Only suggest if no model is currently set
+			if (!modelAlias.model || modelAlias.model === '') {
```
thedadams: I've always thought that an empty string, even if it the variable could be undefined, would still pass `!modelAlias.model`. More for my own understanding, but, if I'm wrong, please let me know.

ui/user/src/lib/components/admin/DefaultModels.svelte
```diff
@@ -68,6 +110,30 @@
...
 	}
 
+	function getSelectedModel(modelAlias: DefaultModelAlias, activeModelOptions: Model[]) {
+		// If there's a pending change, use that
+		if (changes?.[modelAlias.alias]) {
+			return changes[modelAlias.alias];
+		}
+
+		// If a model is already set, use it
+		if (modelAlias.model && modelAlias.model !== '') {
```
thedadams: Same question here.

ui/user/src/lib/components/admin/DefaultModels.svelte
njhale: You're 100% correct. I've simplified the predicate (and inverted it to reduce nesting.

ui/user/src/lib/components/admin/DefaultModels.svelte
njhale: Also correct here. Fixed.

cjellick (approved):

Issue #3555, opened 2025-07-25
Drop project files from project configuration page


cjellick:
Will, Sheng, and I decided it was too confusing to try to explain, so we are just going to punt on it for now. Nick, removing it from the UI is enough, but if there's an easy way to disable it (without completing deleting the code because we might bring it back), do that as well so that threads dont ever "accidentally" write to the project file space.

<img height="500" alt="Image" src="https://github.com/user-attachments/assets/6966bd58-f908-40fe-a919-c2e770fa660f" />

njhale:
> do that as well so that threads dont ever "accidentally" write to the project file space.

@cjellick I had to remove some tool parameters and refactor a bit of the context in the `workspace-files` tool to ensure the LLM doesn't trend towards using project scoped files. The changes are minimal though.

cjellick:
ok, thanks

sangee2004:
When testing with latest builds - `v0.0.0-dev+cca60b3b` and tools - `obot-platform/tools: 46eebc9`

I dont see project files being exposed in Project Configuration page as expected.

But  I am still allowed to add resource to project and able to chat with the agent and list files from workspace


<img width="635" height="787" alt="Image" src="https://github.com/user-attachments/assets/30c326c7-8a5a-4029-bda8-f867fe9072b9" />

<img width="953" height="546" alt="Image" src="https://github.com/user-attachments/assets/d564bd98-2e24-4a16-8320-1298c2becb17" />


@njhale Should we not be allowing resources to get added to projects anymore ?

njhale:
@cjellick I missed this part. what do you want to do here? seems like we want to move it down to the chat well ala https://github.com/obot-platform/obot/pull/3257

njhale:
@sangee2004 just merged some changes that disable adding resources to the project workspace and enable adding them to thread workspaces. Check out the description of [this PR](https://github.com/obot-platform/obot/pull/3590) for details.

sangee2004:
We are now presented with `Add to thread files` option  (Files-> `Add from Connector`) for adding resource files to a chat thread.

<img width="783" height="836" alt="Image" src="https://github.com/user-attachments/assets/b778e5d8-dec9-4bf9-b0b0-97abd17806f3" />


There is no option to add resource files to a project anymore.

PR #3571, opened 2025-07-28
chore: disable project files configuration

njhale:
Remove project scoped files from project configuration in the user UI.
Leave the components to make it easier to revive the feature should we
decide to down the road.

Depends on https://github.com/obot-platform/tools/pull/717
Addresses https://github.com/obot-platform/obot/issues/3555


thedadams (approved):
nit: I know the issue said do it without deleting code, but this change is so small and easily reverted that I would be in favor of deleting rather than commenting. At least the component is still there and we can easily get it back.

cjellick (approved):
Have the same nit as donnie, but not enough of a problem to not merge it.

njhale:
@cjellick @thedadams removed instead of commenting.

cjellick (approved):

cjellick:
Approved again

PR #3572, opened 2025-07-28
enhance: add MCP webhooks/filters UI

thedadams:
Based on feedback as the UI was being implemented, this change also includes changes to the data structures for webhook filters.

Issue: https://github.com/obot-platform/obot/issues/3291

Loom: https://www.loom.com/share/6248475bcaab40568b9c104defad7fa9?sid=a6003e5a-fa5d-4e97-ab76-8818414a8dc5


ui/user/src/lib/context/admin/mcpFilters.svelte.ts
```diff
@@ -0,0 +1,42 @@
...
+import { getContext, hasContext, setContext } from 'svelte';
+
+const Key = Symbol('admin-mcp-filters');
+
+export interface AdminMcpFiltersContext {
+	filters: MCPFilter[];
+	loading: boolean;
+}
+
+export function getAdminMcpFilters() {
```
ivyjeong13: I'm thinking you might not need a context for mcpFilters! Context is nice for centralized information that descendant components can get data from without prop drilling, etc. (like mcp servers and entries gets used subcomponents and pages

It looks like it's used on the highest page level only for both places it's referenced (and I think one you can just axe it since it navigates to a new route on update!)

ui/user/src/routes/v2/admin/filters/+page.svelte
```diff
@@ -41,14 +45,15 @@
...
+	async function navigateAfterCreated() {
+		showCreateFilter = false;
+		// Refresh the filters list to ensure we have the latest data
+		await fetchMcpFilters(mcpFiltersContext);
+	}
 
 	const duration = PAGE_TRANSITION_DURATION;
 	onMount(async () => {
-		fetchMcpServerAndEntries(defaultCatalogId);
+		await Promise.all([fetchMcpServerAndEntries(defaultCatalogId), fetchMcpFilters()]);
```
ivyjeong13: Since you initally called listMcpFilters in the +page.ts, looks like this will do another to listMcpFilters in fetchMcpFilters(). I'm thinking you can drop the context and just call listMcpFilters when you need to -- looks like, in this case, it'd be for after creation or deletion to refresh the list. Something like:

```
	let filters = $state(initialFilters);
	let loading = $state(false);

// ...

	async function refresh() {
		loading = true;
		filters = await AdminService.listMCPFilters();
		loading = false;
	}

	async function navigateAfterCreated() {
		showCreateFilter = false;
		// Refresh the filters list to ensure we have the latest data
		refresh();
	}

// ...

					<div class="relative flex items-center gap-4">
						{#if loading}
							<LoaderCircle class="size-4 animate-spin" />
						{/if}
						{@render addFilterButton()}
					</div>

// ..

<Confirm
	msg="Are you sure you want to delete this filter?"
	show={!!filterToDelete}
	onsuccess={async () => {
		if (!filterToDelete) return;
		await AdminService.deleteMCPFilter(filterToDelete.id);
		filterToDelete = undefined;
		await refresh();
	}}
	oncancel={() => (filterToDelete = undefined)}
/>
```

ui/user/src/routes/v2/admin/filters/[id]/+page.svelte
```diff
@@ -0,0 +1,63 @@
...
+	});
+</script>
+
+<Layout>
+	<div class="my-4 h-full w-full" in:fly={{ x: 100, duration }} out:fly={{ x: -100, duration }}>
+		<FilterForm
+			{filter}
+			onUpdate={() => {
+				fetchMcpFilters(mcpFiltersContext);
+				goto('/v2/admin/filters');
```
Comment (ivyjeong13) [inline ui/user/src/routes/v2/admin/filters/[id]/+page.svelte:42]: Trying to think if the fetchMcpFilters is necessary here? The goto should navigate to that route and will fire calls for +page.ts or onMount for that +page.svelte

njhale (approved):

cjellick (approved):
👍

assuming you fixed the little things you mentioned in the loom (ie cancel button), LGTM

Issue #3547, opened 2025-07-25
Make `https://<host>/chat` go to chat


cjellick:
As far as I can tell, we dont have an easy canonical url path that will always drop you into chat. We want this so that it can be easily bookmarked/remembered/shared. it is fine if it redirects to a project.

sangee2004:
With latest build from `main` , https://<host>/chat takes user to chat view.

PR #3573, opened 2025-07-28
chore: add new chat redirect

njhale:
Add a new `/chat` route to the UI that redirects to chat:
- with the newest user project if available (i.e. `/o/[newest_project_id])
- with a new project if the user has no pre-existing projects

Addresses https://github.com/obot-platform/obot/issues/3547


ui/user/src/routes/chat/+page.svelte
```diff
@@ -0,0 +1,11 @@
...
+	<title>Chat - Obot</title>
+</svelte:head>
+
+<div class="flex h-svh w-svw flex-col items-center justify-center">
+	<div class="flex items-center justify-center">
+		<div class="animate-pulse">
+			<p class="text-base font-semibold">Redirecting...</p>
+		</div>
+	</div>
+</div>
```
njhale: @ivyjeong13 is even this necessary since the redirect is happening on page load?

pkg/api/authz/ui.go
```diff
@@ -17,6 +17,7 @@ var uiResources = []string{
 	"GET /landing/images/",
 	"GET /_app/",
 	"GET /{assistant}",
+	"GET /chat",
```
njhale, line 20: The `/{assistant}` pattern above will match this, but I don't want `chat` conflate the two, so I added the more specific path explicitly here.

Does order actually matter in terms of match precedence here?

ui/user/src/routes/chat/+page.ts
```diff
@@ -0,0 +1,29 @@
...
+			lastProjectId = newProject.id;
+		}
+	} catch (err) {
+		// Handle redirecting to login, showing unauthorized error, etc.
+		handleRouteError(err, '/chat', profile.current);
+		return;
+	}
+
+	// Redirect to project
+	throw redirect(302, `/o/${lastProjectId}`);
```
njhale, line 28: When a production build, this will run in the browser; IIUC, for dev mode, this will run server-side given our configuration. @ivyjeong13 could you confirm/refute these claims?

ui/user/src/routes/chat/+page.svelte
ivyjeong13: If everything is happening in page.ts, you shouldn't need this!

ui/user/src/routes/chat/+page.ts
ivyjeong13: Yea that is correct, ssr gets set on layout.ts and its tied to dev environment. Unfortunately haven't had a chance to dig in on why it's fighting on production side since Darren last attempted.

ui/user/src/routes/chat/+page.svelte
njhale: Cool, just tested it out with a dev and prod build and it seemed to work just fine. Thanks!

ui/user/src/routes/chat/+page.ts
njhale: yeah, looking more closely, for production we generate a static site and serve the files from go, so these would need to run client side if at all

ivyjeong13 (approved):
:+1: user-ui changes LGTM

cjellick (approved):

Issue #3520, opened 2025-07-24
Not able to chat when project has MCP servers with same tool calls


sangee2004:
**Describe the bug**
Not able to chat when project has MCP servers with same tool calls 

**To Reproduce**
1. Connect to Gmail MCP server
2. Chat with this MCP server using chat option.
3. Add Outlook MCP server to this project using `Add MCP Server` option
4. Send some chat message

Notice that following error message is presented to the user
```
failed to stream: failed to load program: line inline:0: duplicate tool name [list_email] in inline found at lines 0 and 0
```

<img width="933" height="230" alt="Image" src="https://github.com/user-attachments/assets/72d9fc2b-15aa-44fd-9e35-ed2cce46cc27" />

**Expected behavior**
We should be able to chat  .

cjellick:
Hmph. This is an interesting problem @thedadams. Any idea how we can handle collisions like this better?

thedadams:
I would have to think about how do to this without exposing internal details, but I think it is possible. I would need to spend some time with it.

sangee2004:
This issue is addressed when testing with latest builds.

Able to chat in a project with both `outlook` and `gmail` MCP servers and use `list emails` tool calls from both the MCP servers  successfully to retrieve data.

PR #3576, opened 2025-07-28
fix: namespace MCP tools when chatting to allow duplicate tool names

thedadams:
Issue: https://github.com/obot-platform/obot/issues/3520

### DIFF
--- a/pkg/mcp/mcp.go
+++ b/pkg/mcp/mcp.go
@@ -52,22 +52,18 @@ func (sm *SessionManager) GPTScriptTools(ctx context.Context, tokenService *jwt.
 	var toolNames []string
 
 	for _, tool := range tools.Tools {
-		if !allToolsAllowed && !slices.Contains(allowedTools, tool.Name) {
-			continue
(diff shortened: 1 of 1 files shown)


pkg/mcp/mcp.go
```diff
@@ -52,22 +52,18 @@ func (sm *SessionManager) GPTScriptTools(ctx context.Context, tokenService *jwt.
...
-
-		schemaData, err := json.Marshal(tool.InputSchema)
-		if err != nil {
-			return nil, fmt.Errorf("failed to marshal tool input schema: %w", err)
+		if !allToolsAllowed && !slices.Contains(allowedTools, tool.Name) {
+			continue
 		}
 
-		if err := json.Unmarshal(schemaData, &schema); err != nil {
+		toolName := mcpServerDisplayName + " -> " + tool.Name
```
njhale, line 63: my gut instinct is that this would break something, but if it works 🤷

cjellick (approved):

PR #3581, opened 2025-07-28
fix: use correct logic for needsURL and checking for hostname drift

g-linville:
~~I thought these used to be mutually exclusive, but apparently they never were.~~

I updated the logic for this. We want to check for hostname drift if the server hasn't yet been marked with `needsURL`, because that would mean that there was a real change in the catalog that the admin has not yet applied. If `needsURL` is true, then there is no need to check for hostname drift, because we are waiting for the user to supply the new URL already.

### DIFF
--- a/pkg/controller/handlers/mcpserver/mcpserver.go
+++ b/pkg/controller/handlers/mcpserver/mcpserver.go
@@ -2,6 +2,7 @@ package mcpserver
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


g-linville:
whoops, I forgot to update tests. I'll take care of that in a bit

thedadams (approved):

njhale (approved):

PR #3590, opened 2025-07-28
enhance: add thread scoped resources to chat well 

njhale:
- Disable the `Add to Project` button in the side-nav's MCP Server resources dialog by default.     
  - I added a property to the component that will allow us to quickly enable it should we   decide to resuscitate project workspaces down the road. 
  - I didn't disable the dialog entirely b/c having a place to view and download an MCP server's resources at the project scope still seems reasonably helpful.
- Support adding, removing, downloading, and viewing thread scoped resources to the chat well.

https://github.com/user-attachments/assets/375d9acf-8442-45b6-ab2f-199a8e78718d

Follow up for https://github.com/obot-platform/obot/issues/3555

### DIFF
--- a/ui/user/src/lib/components/mcp/McpResources.svelte
+++ b/ui/user/src/lib/components/mcp/McpResources.svelte
@@ -1,0 +1,448 @@
+<script lang="ts">
+	import {
+		ChatService,
+		type McpServerResource,
+		type Project,
+		type ProjectMCP,
+		type File
+	} from '$lib/services';
+	import { getProjectMCPs, type ProjectMcpItem } from '$lib/context/projectMcps.svelte';
+	import {
+		ChevronRight,
+		LoaderCircle,
+		HardDrive,
+		X,
+		Search,
+		Download,
+		ChevronsRight,
+		Server
+	} from 'lucide-svelte';
+	import { DEFAULT_CUSTOM_SERVER_NAME } from '$lib/constants';
+	import { responsive, errors } from '$lib/stores';
+	import { tooltip } from '$lib/actions/tooltip.svelte';
+	import { clickOutside } from '$lib/actions/clickoutside';
+	import { poll } from '$lib/utils';
+
+	interface Props {
+		project: Project;
+		threadID?: string;
+		currentThreadFiles?: File[];
+	}
+
+	type ServerResources = {
+		mcp: ProjectMcpItem;
+		resources: McpServerResource[];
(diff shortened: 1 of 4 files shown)


njhale:
Going to put this into draft while I make some changes Craig requested.

ui/user/src/lib/components/mcp/McpResources.svelte
```diff
@@ -0,0 +1,451 @@
...
+		loadingFiles = false;
+	}
+
+	function getMcpFQN(mcp: ProjectMCP) {
+		return (mcp.name ?? '') + '-' + mcp.id;
+	}
+
+	function getFilename(mcp: ProjectMCP, resourceName: string, mimeType: string) {
+		const extension = fileExtensions[mimeType] ?? mimeType.split('/')?.[1] ?? 'txt';
+		const filename = `obot-${getMcpFQN(mcp)}-resource-${resourceName}.${extension}`;
```
njhale, line 134: Not crazy about the naming, but we need something unique per MCP-server/instance to key off so we can discover existing resource files in the workspace.

More of a long-term idea, but it would be interesting if we had a "directory" in the workspace reserved for these types of files, and they were immutable to the `workspace-files` tool and stamped with the hash of the content so we can tell when they've changed compared to what the MCP server is returning. We could then offer an "update file" option that gets show when drift is detected.

ui/user/src/lib/components/mcp/McpResources.svelte
```diff
@@ -0,0 +1,451 @@
...
+					serverResources.resources.length > 0
+			)
+	);
+
+	function fetchProjectResources() {
+		loading = true;
+		projectResources = [];
+		for (const mcp of projectMcps) {
+			if (!mcp.authenticated) {
+				projectResources.push({
```
njhale: we're just omitting resources from MCP servers that still haven't been authenticated yet instead of getting fancy.

ivyjeong13 (approved):
:+1:

njhale:
Note:
- changed the user-facing verbiage from `MCP server` to `Connector` on these changes.
- merging this as-is, but I spoke with Craig and we eventually want to follow up with a change that hides the new button when no resources are available. I didn't include it in this PR because the current implementation pulls in MCP servers when the button is pressed; we'd need to change it so that its set of MCP servers is updated whenever the set of project MCP servers is updated

Issue #3593, opened 2025-07-28
After deleting default model from the base agent , new chat continues to show the deleted model as default for chats.


sangee2004:
**Describe the bug**
After deleting default model from the base agent , new chat continues to show the deleted model as default for chats.

**To Reproduce**
1. As admin , have a model - `gpt-4o` set as default in `Chat Configuration` 
2. As regular user, Chat with any connected MCP server. 
Notice that the default model gets picked as `gpt-4o`  as expected.
3. As admin , delete `gpt-4o`  from allowed model llist

<img width="991" height="398" alt="Image" src="https://github.com/user-attachments/assets/f5f9094b-cbc1-4f88-a904-2c4a24df3b40" />
4. As regular user, Chat with any connected MCP server. 
5. Notice that the deleted model in step 3 is still getting used as default model.

<img width="932" height="357" alt="Image" src="https://github.com/user-attachments/assets/0cf966bd-f4ad-4cd3-9dd5-612e98496f55" />

**Expected behavior**
1. Should the admin user not be allowed to delete the default model in this case ? should he be forced to pick another model as default before he can be allowed to remove the current default model from allowed list.

2. If we do allow for not having to pick a default for allowed model for `Chat configuration` , then the default model set for the installation when  configuring model providers should be used by the chat as default model .

StrongMonkey:
@sangee2004 i pushed a fix where user should not allow to delete a model that has been set as default. Talked with craig and we agree the behavior is a bit odd so we are disabling that.

sangee2004:
Admin user is not provided the ability to delete default models now.

Issue #3592, opened 2025-07-28
Model set for the chat does not not get used ( default model set for the base agent gets used instead)


sangee2004:
**Describe the bug**
Model set for the chat does not not get used ( default model set for the base agent gets used insead)

**To Reproduce**
1. As admin , set  one of the `Claude` models as the default model in the `Allowed Models` and have few openai models in the allowed model list.

<img width="1414" height="887" alt="Image" src="https://github.com/user-attachments/assets/5c9e843c-412a-4d17-afa2-304306970731" />

2. As regular user ,  chat with any connected MCP servers.
3. Select  any openai model  (from the models drop down) and chat with

<img width="1060" height="874" alt="Image" src="https://github.com/user-attachments/assets/45539555-fd45-4b58-9728-acda94714888" />

From the admin view , inspect the chat thread and get the chat completion calls made.

Notice that the claude modelName gets used - `modelName": "m1-anthropic-model-provider-3f1e86a6a2d3bb8dca958f9387b9f-0c0d0` 

snippet of chat completion call:

```
        {
[...]

PR #3601, opened 2025-07-29
Fix: fix thread model, use model id instead

StrongMonkey:
We should use model id to set on the model allow list. We don't set alias anymore for models, so it has to use id to look up model correctly.

@ivyjeong13 Sorry I told you the wrong thing.. I realize there is a bug in backend that would prevent model name from working in backend. So switched to use id now :-)

https://github.com/obot-platform/obot/issues/3592
https://github.com/obot-platform/obot/issues/3593

### DIFF
--- a/pkg/thread/threadmodel.go
+++ b/pkg/thread/threadmodel.go
@@ -48,7 +48,7 @@ func GetModelAndModelProviderForThread(ctx context.Context, c kclient.Client, th
 	model := thread.Spec.Manifest.Model
 
 	// If it wasn't set on the thread, try to find a parent project that has it set.
-	if modelProvider == "" || model == "" {
+	if model == "" {
 		project, err := projects.GetFirst(ctx, c, thread, func(thread *v1.Thread) (bool, error) {
 			return thread.Spec.DefaultModelProvider != "" && thread.Spec.DefaultModel != "", nil
 		})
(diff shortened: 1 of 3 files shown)


pkg/thread/threadmodel.go
```diff
@@ -48,7 +48,7 @@ func GetModelAndModelProviderForThread(ctx context.Context, c kclient.Client, th
 	model := thread.Spec.Manifest.Model
 
 	// If it wasn't set on the thread, try to find a parent project that has it set.
-	if modelProvider == "" || model == "" {
```
StrongMonkey, line 51: @thedadams since we are not setting model provider, we should prevent it from falling back to default model when it is empty. I missed that when addressing your comment last time.

ui/user/src/routes/v2/admin/chat-configuration/+page.svelte
```diff
@@ -264,17 +264,19 @@
...
-										onclick={() => {
-											if (!baseAgent) return;
-											baseAgent.allowedModels = baseAgent.allowedModels?.filter(
-												(modelName) => modelName !== d.name
-											);
-										}}
-									>
-										Remove
-									</button>
+									{#if !d.isDefault}
```
StrongMonkey, line 267: @ivyjeong13 I added so that we don't allow user to remove a model that has been set as default to prevent weird behavior like https://github.com/obot-platform/obot/issues/3593

Issue #3469, opened 2025-07-21
feat: admin v2 Chat Thread Monitoring


ivyjeong13:
Basically bringing Chat Threads into v2/admin from legacy admin-ui

cjellick:
@StrongMonkey - lets see if you can handle this one. I need a page in the new admin app that allows the admin to see chat threads. it should be a table view. It should allow me to filter/search by the columns. I should be able to click a thread as an admin and "go into it" to see the entire chat history. the actual chat view can be bascially the same view as what the user would see.

sangee2004:
`Chat Threads` is now available in the admin view. This view provides list all all chat threads from all users sorted by thread create time.

<img width="1133" height="362" alt="Image" src="https://github.com/user-attachments/assets/d2c80208-e94b-43ae-a0de-b202a4f1734d" />

Able to select any entry and view the entire chat thread 

We are allowed to filter by following filters

<img width="1136" height="231" alt="Image" src="https://github.com/user-attachments/assets/75d8dfb7-d520-4469-b22b-02945a37d03e" />
 

There is also a generic search field that searches by Thread Name , Username and User Email. But not by project.

@StrongMonkey @cjellick Needed input on the following observations:

1. In the chat history view, do we want to provide the ability to provide debug information for each chat like we do in our legacy admin chat thread view ?
2. In `Chat threads` view , there is no ability to sort the entries by Name, User Name, Project, Created fields. Do we need to support sorting by any of these fields.
3.  Do we need to present both UserName and User email  in `Chat threads` view like we do now ?
4. Generic search field searches by Thread Name , Username and User Email. But not by Project. Do we want to support searching by project as well?
5. There is no support for pagination in this `Chat threads`. Do we want to support pagination.

StrongMonkey:
Here is my take:
>In the chat history view, do we want to provide the ability to provide debug information for each chat like we do in our legacy admin chat thread view ?

I assume we would want that, can be done later.

>In Chat threads view , there is no ability to sort the entries by Name, User Name, Project, Created fields. Do we need to support sorting by any of these fields.

I think it is fine since we have search and filter functionality. Sorting by created ago makes the most sense for now.

>Do we need to present both UserName and User email in Chat threads view like we do now ?

I like it. Those are identiable information for the thread(and thread creator)

>Generic search field searches by Thread Name , Username and User Email. But not by Project. Do we want to support searching by project as well?

It is probably fine. I think Thread Name , Username and User Emai should be enough.

[...]

PR #3604, opened 2025-07-29
Enhance: chat view in admin

StrongMonkey:
https://github.com/obot-platform/obot/issues/3469

This PR added chat thread view in admin page so that admin can see all the thread and click on it to see chat history.

Demo:  https://www.loom.com/share/2f56831ce7fd4db9a6ff777015ed9615


ui/user/src/routes/v2/admin/chat-threads/+page.ts
```diff
@@ -0,0 +1,5 @@
+import type { PageLoad } from './$types';
```
ivyjeong13: If page.ts is not needed, you can just delete this!

ivyjeong13 (approved):
Some minor styling things in dark mode -- The inputs and boxes don't match with the rest of the app that uses it.
Screenshot 2025-07-29 at 2 54 46 PM

Could use the `Select` component to keep input consistency for dropdowns in the Chat Thread filters
Screenshot 2025-07-29 at 2 53 17 PM

Chat Thread has blue background
Screenshot 2025-07-29 at 2 53 28 PM

Pulled down and tested :+1: Styling things are non-blockers

StrongMonkey:
@ivyjeong13 thank you! will fix those

StrongMonkey:
image
image
@ivyjeong13 should be fixed

cjellick:
Screenshot 2025-07-30 at 6 21 06 AM
I feel like the bottom line here should not have a rounded corner. If that is too hard to fix, we can have ivy address it later, but let me know

cjellick:
Drop the refresh button. move Chat Threads up in the left nav to just above Chat Configuration

cjellick:
@ivyjeong13 was your intent/vision in calling out the color issues to have the chat history be black on black like this? I suspect it wasn't. It looks awkward harsh.
Screenshot 2025-07-30 at 6 27 04 AM

Also @StrongMonkey the chat history is now in a fixed scrollable window. That is not acceptable it should just use the page's main scrollable area.

I think if we dropped the "Chat History" title , the chat area white border, and the inner scrollbars all together, I would probably be satisfied.

cjellick:
see this thread - https://acorn-io.slack.com/archives/C07P0EZR917/p1753883406768759

Takeaways - let's simplify this:
1. Drop the entire header section with info
2. Drop the "Chat History" header
3. Drop the border around the chat and ensure that it just uses the main page's scrollable area
4. Just use a single background color for this page (like the chat app does) - white in light mode and black in dark mode. We don't need a separate outer background color like the other admin pages have.

cjellick (changes requested):
see above comments

Issue #3549, opened 2025-07-25
New user feature overview carousel


cjellick:
When a user logs in for the first time, we want to give them an overview of the major features in the app. Right now, I am thinking it is a carousel of screenshots with a bit of text underneath...Steam was the first example that came to mind:

<img width="500" height="1000" alt="Image" src="https://github.com/user-attachments/assets/05f02976-8f74-489b-8396-de93d632f959" />

I don't have any good copy or content yet, but we want to highly the major features of chat (rag, memory, projects, mcp servers) and the fact that there isa separate mcp gateway app.

cjellick:
I don't expect this to get done next week...i expect it to kind of be a last minute thing.

cjellick:
Will and Sheng decided they dont want to do this. They want these features to be exposed/handled by the marketing site.

PR #3605, opened 2025-07-29
feat: feature carousel dialog for first visit user

ivyjeong13:
Addresses #3549 

NOTE: contents in the carousel are placeholder data, getting what should be in there from Will & Craig


cjellick:
Will and Sheng decided they dont want to do this. They want these features to be exposed/handled by the marketing site.

I am going to go ahead and close this pr @ivyjeong13 

... there is a decent chance that they ask for a similar feature in the future, so maybe keep the branch around just in case :-)

Issue #3541, opened 2025-07-24
Revamp audit log functionality


cjellick:
Watch the loom and check out the figma whiteboard for an in-depth review of what I want.

https://www.loom.com/share/91d208da5fc946fd916126ecb181b1de
https://www.figma.com/board/c9m0OZHOE4GLAtEkQVxuz2/Audit-Logs?node-id=0-1&t=yZ85pwb9NsPl8WVh-1

Below is a pretty accurate summary of the loom, but you should watch the loom to have context. 

**Search Functionality**
* Replace existing **client-side search** with **server-side search**
  * Responsible: **Donnie**.
  * Until implemented, **Abdel** can leave the search box as-is (non-functional).

**Time Picker Enhancements**
* Implement shortcut options in time picker:
  * Last hour
  * Last 6 hours
  * Last 24 hours
  * Last 3 days
  * Last 7 days
  * Custom range - this should give the user precision down to the minute
 

* **Granularity** (AKA how many bars should be on the graph and how much time should each bar cover)
  * If the timeframe is a minute or less, granularity of the timeline graph should be 1 second
  * For anything over a minute, we should shoot to always have a consistent number of bars on the graph. Let's shoot for 100 bars. This means you'll have to do some math to figure out how much time each bar should cover, but @thedadams is a former math teacher ... he can help with that.

**Filters System**
* Allow **any column in the table** to be used as a filter.
* Update backend to support **multi-select filters** (e.g., filter on both `tools list` and `tools call` simultaneously).
* Add a **“Clear All Filters”** button to the filter side panel.
* Retain existing filter UX (side popover + apply button) as acceptable.
* Ensure **every column is indexed** on the backend to enable filtering.

**Graph Behavior**
* The only thing we should be graphing on this page is the timeline of audit log events.
* **Do not call the Graphs API** on the audit logs page. Derive graph data **entirely from audit log results**.
* Display timeline that corresponds to selected time range and granularity

**Timeline Visualization**
* Display single-color bars (no need for log-level colorization like Google viewer).
* Each bar reflects number of calls per time interval from fetched data.

**Pagination Strategy**
* Set default **results per page = 10,000**. If more than 10,000 results:
  * Show a **“Next Page”** button and clicking it will reload/rerender both the timeline and the table data
  * Clicking it reloads both:

**Client-side Rendering Strategy**
* First pass: render **all 10,000 rows at once**.
[...]

PR #3616, opened 2025-07-29
feat: revamp `/audit-logs` page

ryu-man:
Addresses #3541 

- Implement the new UI structure
- Implement stacked bar chart, show the num of log type per date
- Create `AuditLogCalendar` and implement quick actions
- Make filter select supports multiple mode
- Make everything dependant on one source of truth


ui/user/src/routes/v2/admin/audit-logs/+page.svelte
```diff
@@ -1,95 +1,164 @@
...
+
+	const numberOfPages = $derived(Math.ceil(auditLogsTotalItems / pageLimit));
+	const pageOffset = $derived(pageIndex * pageLimit);
+
+	const remoteAuditLogs = $derived(
+		(auditLogsResponse?.items ?? []).map(({ createdAt, ...restProps }) => ({
+			...restProps,
+			createdAt: new Date(createdAt.slice(0, -6)) + 'Z'
+		}))
 	);
```
thedadams: @ryu-man When I change these lines back to this, everything works as expected for me:
```suggestion
	const remoteAuditLogs = $derived(auditLogsResponse?.items ?? []);
```

Image

thedadams: Here is a better example where I am filtering down to a single minute:
Image

ryu-man: Apparently there is an issue because that does not work for me! If you see the results in the table in the image below, there are logs from `Jul 30 2025 10:00:00 PM EDT` to `Jul 30 2025 11:00:00 PM EDT`, but I have to select the start date to be ` Jul 30, 2025, 06:00:00 PM EDT` to get the results!

image

cjellick:
Feedback:

- tools/call rows show always have an identifier set and they dont.  i do see the appropriate data in the response: "callIdentifier": "tavily-crawl"
- let's drop the Client Version column. it doesnt seem to be populated and i really dont care about it right now
- The colors in the bar graph are too ... bright. I like the color coding, but can you make the colors a bit more muted?
- I find it annoying that the Client column takes up so much width. Would it be possible to cut that one off after like 15 characters and then you have to hover to see the rest?
- The horizontal scrolling inside the table isnt going to fly. Let's just make the content of the page much wider. Reduce the side gutters to some minimal amount, like 10 pixels. But it's important that the other pages stay their current width. @ivyjeong13 any thoughts here? Abdel mentioned that it might be hard to do this without affecting other pages.

cjellick (changes requested):
see above

my_commits.txt
```diff
@@ -0,0 +1,322 @@
+commit f693ebc55edb8d312ac6a59b87d892586af5e0ed
```
ivyjeong13: Curious what's this file for?

ui/user/src/lib/components/Select.svelte
```diff
@@ -106,18 +175,40 @@
...
+
+					if (isSelected) {
+						values.delete(key);
+					} else {
+						values.add(key);
+					}
+
+					selected = values.values().toArray().toReversed().join(',');
+
+					console.log(selected);
```
ivyjeong13: can remove console.logs

ui/user/src/lib/components/admin/audit-logs/AuditFilters.svelte
```diff
@@ -1,128 +1,126 @@
...
+		] as FilterKey[]
+	).reduce((acc, filterId) => {
+		acc[filterId] = {
+			property: filterId,
+			label: getFilterDisplayLabel?.(filterId) ?? filterId.replace(/_(\w)/, ' $1'),
+			get selected() {
+				return filters?.[filterId] ?? '';
+			},
+			set selected(v) {
+				filters[filterId] = v ?? '';
```
ivyjeong13, line 65: Is this the `filters` from `props`? If the child is modifying it back up, shouldn't it be a $bindable?

ui/user/src/lib/components/admin/audit-logs/AuditFilters.svelte
```diff
@@ -136,37 +134,25 @@
...
-					/>
-				</div>
-			{/if}
+		{#each filterInputsAsArray as filterInput, index (filterInput.property)}
+			<AuditFilter
+				filter={filterInput}
+				onSelect={(_, value) => {
+					filterInput.selected = value ?? '';
+				}}
+				onClear={(_, value) => {
```
ivyjeong13: nit: If this is only called on clear all, maybe makes more sense the prop is called `onClearAll`?

my_commits.txt
ryu-man: Oops, that's an old file that accidentally goes with the commits.

ui/user/src/routes/admin/audit-logs/+page.svelte
```diff
@@ -1,241 +1,467 @@
...
 	);
 
-	let sort = $derived(
-		currentFilters.sortBy && currentFilters.sortOrder
-			? {
-					sortBy: currentFilters.sortBy,
-					sortOrder: currentFilters.sortOrder as 'asc' | 'desc'
+	const fragmentedAuditLogs = $derived(remoteAuditLogs.slice(fragmentSliceStart, fragmentSliceEnd));
+
+	const users = new Map<string, OrgUser>();
```
ivyjeong13, line 58: Curious how come `users` is not in a `$state`? I imagine it should be reactive to update to show the name for a filter value.

ui/user/src/lib/components/admin/audit-logs/AuditFilters.svelte
ryu-man: Not necessarily, because the `filters` prop is calculated from URL search params and is used temporarily to hold the changes, and once the user hits the apply button, those changes will be pushed to the URL, and the `filters` prop will be recalculated again.

ui/user/src/lib/components/admin/audit-logs/AuditFilters.svelte
ryu-man: makes sense

ui/user/src/routes/admin/audit-logs/+page.svelte
ryu-man: Well, the `users` Map is used only for cache purposes to help show the user name instead of the user ID in the applied filters. Filter values are fetched in the `AuditFilters` component.

ui/user/src/lib/components/admin/audit-logs/AuditFilter.svelte
```diff
@@ -0,0 +1,53 @@
...
+</script>
+
+<div class={twMerge('mb-2 flex flex-col gap-1', !options.length && 'opacity-50')}>
+	<div class="flex items-center justify-between">
+		<label for={filter.property} class="text-md font-light">
+			By {filter.label}
+		</label>
+
+		{#if filter.selected}
+			<button
```
ivyjeong13, line 25: nit: Thinking about the multi select experience, and it's hard to see at a glance what options have been selected for the filters

example

Was thinking something like this instead? Maybe don't need the pills at the top if we don't allow a collapse/expand option, but everytime you select from the dropdown, it'll add as a list to below the input and you can quickly see and quickly remove a filter instead of having to go through the dropdown. Opening the dropdown should still be like as it is, having the checks of what you've already selected

ivyjeong13:
> Feedback:
> 
> * tools/call rows show always have an identifier set and they dont.  i do see the appropriate data in the response: "callIdentifier": "tavily-crawl"
> * let's drop the Client Version column. it doesnt seem to be populated and i really dont care about it right now
> * The colors in the bar graph are too ... bright. I like the color coding, but can you make the colors a bit more muted?
> * I find it annoying that the Client column takes up so much width. Would it be possible to cut that one off after like 15 characters and then you have to hover to see the rest?
> * The horizontal scrolling inside the table isnt going to fly. Let's just make the content of the page much wider. Reduce the side gutters to some minimal amount, like 10 pixels. But it's important that the other pages stay their current width. @ivyjeong13 any thoughts here? Abdel mentioned that it might be hard to do this without affecting other pages.

For the last bullet point, should be quick if you add a prop `classes` to the `Layout` component that specifically affects that well

```
<!-- Layout.svelte -->
	interface Props {
		// ...
		classes?: {
			well?: string;
		};
	}

	const { children, showUserLinks, onRenderSubContent, hideSidebar, whiteBackground, classes }: Props =
		$props();
```

```
<!-- audit logs +page.svelte -->
<Layout classes={{ well: 'w-[..] max-w-[..]` }}> '
```

```
// ... Layout.svelte

let {

			<div class="flex h-full flex-col items-center justify-center p-4 md:px-8">
				<div class={twMerge("h-full w-full max-w-(--breakpoint-xl)", classes?.well)}>
					{@render children()}
				</div>
			</div>
```

And go from there?

ui/user/src/routes/admin/audit-logs/+page.svelte
ivyjeong13: Sounds like it needs to be reactive then! If users or filter value changes, then we want to react to it and have the DOM display the user name for that filter value.

ui/user/src/lib/components/admin/audit-logs/AuditFilters.svelte
ivyjeong13:
```
<AuditFilters
			onClose={handleRightSidebarClose}
			filters={{ ...searchParamFilters }}
			{fetchUserById}
			{getFilterDisplayLabel}
		/>
```

Ahh I see, filters is a shallow copied searchParamFilters. I think what you have works but usually, if it's like this, I think svelte best practices you shouldn't directly modify a prop that's not $bindable, treat it like one-way binding (similar to React)

```
// AuditFilters.svelte
        interface Props {
		filters: AuditLogURLFilters;
		// etc.
	}

	let { filters: initialFilters, ... etc }: Props = $props();
        let filters = $state(initialFilters);

        $effect(() => {
            filters = initialFilters; <-- if prop changes, update filters
        });
```

ui/user/src/lib/components/admin/audit-logs/AuditFilter.svelte
ryu-man: Right, but I think the only downside of this approach is that it requires much vertical space, so using many multi-select filters will result in a long scrollable container. Maybe something more compact like below will do the trick?

image

ivyjeong13: How would it look with session ID? edit: Mostly wondering as I fiddle with filters and have selected two long name selections like `Obot MCP Chat (via nanobot v0.0.0-dev)` for Client Name or `04b22ec8-a9bb-41d6-b262-0fb80511afc0` or Session ID. Can you click the 'x' on them and can read multiple, etc.

ivyjeong13: Also this is non-blocking nit, just a thought while messing with the branch locally

ivyjeong13:
This is suppperrr nit but: 
Screenshot 2025-08-01 at 10 36 35 AM

Can the icon size and font size be the same for the buttons that are side by side? edit: also probably no need for the tooltip `Filter Logs` if the button itself has the word `Filters` on it

ivyjeong13:
Screenshot 2025-08-01 at 10 48 58 AM

Not sure if getting this to display in mm/dd format would look a little better once the graph gets into days/months? But think making the content wider for this page might help with this too, so just a thought.

ui/user/src/lib/components/admin/audit-logs/AuditFilter.svelte
ryu-man: This is how Session ID selected options will look.

image

I am also thinking of making the slideover a little bit wider!

ryu-man: I also think that truncating long content and displaying the fill value in a tooltip would work! What do you think about that?

ryu-man:
> Screenshot 2025-08-01 at 10 48 58 AM
> Not sure if getting this to display in mm/dd format would look a little better once the graph gets into days/months? But think making the content wider for this page might help with this too, so just a thought.

Yeah, I thought about that too. I will think about it again after changing the layout width.

ui/user/src/routes/admin/audit-logs/+page.svelte
ryu-man: the filter values are already reactive, the users map is initialized after navigate, so whenever a filter value is changed it will lookup in the map and get the name of the user

cjellick:
@ryu-man your screen shot is broke

ui/user/src/routes/admin/audit-logs/+page.svelte
ivyjeong13: Screenshot 2025-08-01 at 1 17 24 PM

Hmmm For me, it's still showing 2 instead of the user name

ui/user/src/lib/components/admin/audit-logs/AuditFilter.svelte
ivyjeong13: Nahh, I don't like the sound of putting a tooltip in options that are in the dropdown/input. Making the filter side panel larger is probably okay.

ui/user/src/lib/components/admin/audit-logs/AuditFilters.svelte
ryu-man: Updating a state in an $effect rune is also not recommended by the Svelte team. But I think using a $derived would work, since it is writable in the latest versions.

ui/user/src/routes/admin/audit-logs/+page.svelte
ryu-man: That's weird. That is how it looks like on my side

image

cjellick:
@thedadams is going to change our client to be `Obot Chat (...)` rather than `Obot MCP Chat (...`.

I would like the character cutoff for this column to be whatever makes that cutoff "nice", meaning I want it to look lik:
`Obot Chat...`.

ui/user/src/lib/components/admin/audit-logs/AuditFilters.svelte
ryu-man: https://github.com/ryu-man/obot/blob/88092a807a1a61518d90af6871b984842d14eb88/ui/user/src/lib/components/admin/audit-logs/AuditFilters.svelte

cjellick (approved):

cjellick:
approved just from a general look and feel perspective

Issue #3613, opened 2025-07-29
Rename `Threads` to `Chat` in chat view.


sangee2004:
Rename `Threads` to `Chat` in chat view.

Tracking this requirement  - https://github.com/obot-platform/obot/issues/3535#issuecomment-3119987590

We are still shown `Threads`

<img width="244" height="523" alt="Image" src="https://github.com/user-attachments/assets/c011a18a-700a-4f82-b243-52aa091e35b8" />

sangee2004:
We now show `Chats` in chat view as expected.

<img width="292" height="539" alt="Image" src="https://github.com/user-attachments/assets/f9867830-da12-4049-8b24-9755a4ed0bb2" />

Issue #3580, opened 2025-07-28
"No Prompt Found" shows up before prompts are listed when clicking on `Add prompts`


sangee2004:
**Describe the bug**
"No Prompt Found" shows up before prompts are listed when clicking on `Add prompts`

**To Reproduce**
1. Chat with MCP server that has prompts like `Everything` MCP server.
2. Click on `Add prompts` 

Notice that "No Prompt Found" shows up before prompts are listed when clicking on `Add prompts` 

https://www.loom.com/share/c8cedf62712644198e436d776cdca166?sid=5b1da16b-02d5-4409-834f-165c0ced5a38

sangee2004:
This issue is not seen anymore when launching prompt list using "/" .

Issue #3578, opened 2025-07-28
Standardize placement of chat well buttons


cjellick:
Our three chat well buttons all place their pop-up menus different. Can we consistently place them above their respective buttons and left aligned with their buttons?

Here screenshots of the current state, for reference:

<img width="300" height="274" alt="Image" src="https://github.com/user-attachments/assets/8faaad3c-8e56-43aa-9320-efb782f076af" />
<img width="300" height="374" alt="Image" src="https://github.com/user-attachments/assets/571a9354-8872-4c26-9a2a-c3bfec85d89b" />
<img width="300" height="380" alt="Image" src="https://github.com/user-attachments/assets/b7d7eee3-a618-40bd-8fa2-456d3a5a4f32" />

And here's the placement and alignment I want:

<img width="300" height="444" alt="Image" src="https://github.com/user-attachments/assets/41c62a68-c527-4ea2-9b6c-1a3a817def2c" />

cjellick:
Hit the breaks on this one too! As mentioned, we're going to drop the prompts button all together, BUT ALSO, let's drop the tools button (wrench) as well. We JUST want MCP servers (and the ability to enable/disable individual tools) in the left nav, dont need it down here.

So, we should ONLY show the Files button down here. Now, I do still want that one to have the menu aligned as described above.

sangee2004:
We only have `Files` option in the chat well now.

It is aligned as expected now

<img width="927" height="367" alt="Image" src="https://github.com/user-attachments/assets/9080f761-27b2-4365-81e8-31240bb54e56" />

Issue #3577, opened 2025-07-28
Drop prompts and tools buttons from chat well


cjellick:
**Edit: I am rug pulling this issue and completely changing the requirements**

~Right now the prompt button in chat is a `+` and it's the first button. I think this will confuse users because the plus is too generic and it being the first button in the row is too prominent for a feature that is relatively low in usage.~

~I want to address these concerns by change the icon to a chat bubble (or a chat bubble with a small + if you can find/make an appopriate one) and move it to be the third button, rather than the first.~

~Finally, when a user clicks and none of the servers have prompts and we present the message "No prompts found" - let's change that message to "No prompts available"~

We want to completely drop both the prompts and tools buttons from the chat well.

cjellick:
Hit the breaks! I discussed with @will-chan and we just want to drop prompts as a button here all together. You can already get to prompts by hitting `/`, so we JUST want to do that. the only tweak i would like to see is that when i do the `/` menu, i should be able to navigate the prompts using my arrow keys

cjellick:
@ivyjeong13 - im not going to open a separate issue for deleting the chat prompt button. i am just going to reuse this one. I will update the title.

sangee2004:
`Tools` and `Prompts` are not present in chat well anymore.
I am not able to to work with prompts when using "/" to launch prompt list which is tracked in https://github.com/obot-platform/obot/issues/3644

PR #3617, opened 2025-07-29
fix: prompts in chat related bugfixes

ivyjeong13:
Addresses #3574 
Addresses #3577 
Addresses #3578 
Addresses #3580 
Addresses #3613


cjellick:
dang it! i was too late! now we just want to completely drop the prompts and tools button from the chat well. we will just use `/` for accessing prompts (its a lesser used feature anyway).

ivyjeong13:
> dang it! i was too late! now we just want to completely drop the prompts and tools button from the chat well. we will just use `/` for accessing prompts (its a lesser used feature anyway).

:smile: No worries! Will be straightforward to drop them and will just make the changes in this branch

Issue #3359, opened 2025-07-10
Not allowed to get connection url to same remote MCP server(with different url) /Single user MCP server multiple times from `Available Connectors`


sangee2004:
**Describe the bug**
Not allowed to get connection url to same remote MCP server multiple times.

**To Reproduce**
**Scenario 1:**

1. Connect to `GitMCP` MCP (Remote)server with url say `https://gitmcp.io/obot-platform/obot`

Noice that after this, `GitMCP` server is not available in the `Available MCP Servers` list anymore. There is no way to use `GitMCP` server  again with a different url.

<img width="998" height="323" alt="Image" src="https://github.com/user-attachments/assets/be7e2bc9-3145-45ff-a7cb-edfa6db14d00" />

**Expected Behavior:**
Should be allowed to get connection url with `GitMCP` server  again for a different url.

**Scenario 2:**
1. Connect to `MySQL` MCP server (Single user)

Notice that after this , `MySQL` MCP server  is not available in the `Available MCP Servers` list anymore. There is no way to use `MySQL` MCP server  again with different Database credentials.

**Expected Behavior:**
We should be allowed to get connection url with ``MySQL` MCP server  again for a different DB credentials.

cjellick:
I'm good with the UX. Can you change "Edit Name" to "Rename"

cjellick:
Actually one other thing: in the chat modal - where you've now had to add the `...` button, it looks odd next to the "Add to chat" little icon, but i dont really like that icon and it's weird that its the one thing i actually CANT click on (this isnt you, this is the current behavior). Can you just drop that "Add to chat" icon in the top right?

njhale:
Moving this to Stalled Out until usage graphs are in.

sangee2004:
I am now allowed to connect to multiple instances of the same MCP server.

I was able to add `GitMCP` MCP Server for 2 different urls and was able to chat in the project and get documentation relating to each of the urls successfully.

I am also able to rename the MCP sever names (from the default ones that get assigned )
<img width="251" height="342" alt="Image" src="https://github.com/user-attachments/assets/1d7281e1-4143-4473-89e2-b5d6ce7cbfd2" />

Logged  this issue for improving the workflow around user choosing to connect to a connector from Available Connectors list when it is already present in Enabled Connectors list - https://github.com/obot-platform/obot/issues/3798

PR #3631, opened 2025-07-30
enhance: support multiple instances of a catalog entry

njhale:
Make the following changes to the UI:
- Support connecting to a single/remote catalog entry multiple times
- Support renaming a connected single/remote MCP server
- Drop support for editing the configuration of connected multi-user MCP server instances

Addresses https://github.com/obot-platform/obot/issues/3359

Loom: https://www.loom.com/share/3b61b564c9724520b3d328ebf7a620f3?sid=b55b516f-7731-4811-9a3f-3ef1aa1cccb6

Signed-off-by: Nick Hale <4175918+njhale@users.noreply.github.com>

### DIFF
--- a/ui/user/src/lib/components/chat/McpServerSetup.svelte
+++ b/ui/user/src/lib/components/chat/McpServerSetup.svelte
@@ -187,7 +187,7 @@
 							loadData(true);
 						}}
 					>
-						{#snippet connectedServerCardAction(d)}
+						{#snippet connectedServerCardAction(d: ConnectedServer)}
 							{@const requiresUpdate = requiresUserUpdate(d)}
 							<button
 								disabled={requiresUpdate}
--- a/ui/user/src/lib/components/mcp/CatalogEditNameForm.svelte
+++ b/ui/user/src/lib/components/mcp/CatalogEditNameForm.svelte
@@ -1,0 +1,107 @@
+<script lang="ts">
+	import { LoaderCircle, Server } from 'lucide-svelte';
+	import ResponsiveDialog from '../ResponsiveDialog.svelte';
+	import { ChatService } from '$lib/services';
+	import type { ConnectedServer } from './MyMcpServers.svelte';
+	import { errors } from '$lib/stores';
+
+	interface Props {
+		editingServer?: ConnectedServer;
+		onUpdateConfigure?: () => void;
+	}
+
+	let { editingServer, onUpdateConfigure }: Props = $props();
+
+	let dialog = $state<ReturnType<typeof ResponsiveDialog>>();
+	let newName = $state('');
+	let originalName = $state('');
+	let saving = $state(false);
+
+	export function open() {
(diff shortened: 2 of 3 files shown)


g-linville:
One potential issue with this: we can't just update the name on the manifest, as that will get wiped out if the admin triggers an update on that server. Unless we want to never update names. Your call @cjellick.

g-linville:
Also this technically exploits the bug I'm fixing here https://github.com/obot-platform/obot/pull/3633 so we will need a different approach.

Thanks for reminding me about that, lol.

StrongMonkey (approved):

njhale:
Moving to draft while I work out the UX and backend changes necessary for this to work.

apiclient/types/mcpserver.go
```diff
@@ -156,11 +164,13 @@ type MCPServerTool struct {
 
 type ProjectMCPServerManifest struct {
 	MCPID string `json:"mcpID"`
+	Alias string `json:"alias,omitempty"`
 }
 
 type ProjectMCPServer struct {
 	Metadata
 	ProjectMCPServerManifest
+	Alias       string `json:"alias,omitempty"`
```
g-linville: We only need `Alias` defined in one of these two locations.

g-linville (changes requested):
Just one minor change needed in the backend.

apiclient/types/mcpserver.go
njhale: Thanks for catching that. Fixed!

pkg/api/handlers/projectmcp.go
```diff
@@ -41,14 +41,17 @@ func NewProjectMCPHandler(mcpLoader *mcp.SessionManager, acrHelper *accesscontro
...
 		Metadata:                 MetadataFrom(projectServer),
 		ProjectMCPServerManifest: projectServer.Spec.Manifest,
 		Name:                     mcpServer.Spec.Manifest.Name,
 		Description:              mcpServer.Spec.Manifest.Description,
 		Icon:                     mcpServer.Spec.Manifest.Icon,
 		UserID:                   projectServer.Spec.UserID,
 	}
+	pmcp.Alias = mcpServer.Spec.Alias
+
+	return pmcp
```
thedadams, line 54:
```suggestion
	return types.ProjectMCPServer{
Metadata:                 MetadataFrom(projectServer),
ProjectMCPServerManifest: projectServer.Spec.Manifest,
Name:                     mcpServer.Spec.Manifest.Name,
Description:              mcpServer.Spec.Manifest.Description,
Icon:                     mcpServer.Spec.Manifest.Icon,
UserID:                   projectServer.Spec.UserID,
Alias:                    mcpSever.Spec.Alias,
}
```

njhale: Since this is in the embedded ProjectMCPServerManifest field, I get a linting error if I try to instantiate a `ProjectMCPServer` literal like this.

Issue #3489, opened 2025-07-22
Handle the legacy admin view better


cjellick:
I don't really want to completely delete the legacy admin view just yet ... it is kind of a nice fallback while we are still transitioning the product, but here is what I want:

1. Move it from `/admin` to `/legacy-admin`
2. Move the new admin from `/v2/admin` to just `/admin`. While you're doing this, fix `/v2/admin/`. Right now, if you hit that url specifically, you get a plain html page with just 4 or 5 links. It should just redirect to the admin mcp servers page I suppose
3. Put a undismissable warning banner at the top of every legacy admin page that says `Warning: this is Obot's legacy admin view and it has been deprecated. Use to /admin instead.` and make that a /admin a clickable link if you can.

cjellick:
@njhale i assigned this to you since you recently did work around redirects and what not.

sangee2004:
When testing with latest build , when accessing obot-server url (https://<host domain> as admin , user is directed to `/admin/mcp-servers` page.  All links from admin pages are served as `/admin` routes.

Legacy admin pages are accessible at `/legacy-admin` . All legacy admin pages come with following banner which has a link to `/admin` page.

<img width="1598" height="450" alt="Image" src="https://github.com/user-attachments/assets/72971305-a1da-4506-b479-47a917b7d6e8" />

 @StrongMonkey Just to confirm, when I access `/v2/admin` , I get a 404 page not found error. Wanted to confirm is it is expected that /`v2/admin` routes are not reachable anymore.

PR #3632, opened 2025-07-30
Enhance: rename route from v2/admin to admin

StrongMonkey:
- rename admin -> legacy-admin
- rename v2/admin -> admin
- add warning banner to /admin

https://github.com/obot-platform/obot/issues/3489


ui/user/static/admin/assets/anthropic_icon.svg
cjellick, line 1: why do these need moved?

StrongMonkey: our system tools like google-auth-provider and other model provider still use icon link from /admin like this https://github.com/obot-platform/tools/blob/main/google-auth-provider/tool.gpt#L12. Moving from /admin -> /legact-admin will break those. So I just moved the asset to new admin to avoid breaking those icon urls.

Issue #3280, opened 2025-06-30
Stability, scalability, and throughput of MCP Gateway


cjellick:
I want to make sure that MCP Gateway can handle a substantial amount of load and that it can stably handle requests over a long period of time.

I also want to make sure it can be horizontally scaled just by increase the scale of the obot deployment.

Here is a first pass at some success criteria. I'm firing from the hip on these, so im happy to adjust:
- handle 10k concurrent connections. scale up granduallly on this. i want to see us succeessfully handle 1, 10, 100, 1000 first
- 1,000 requests per second, where requests are MCP server tool calls
- large streaming requests (returned from tool calls or GETs to resources) should not block or slow it down. So lets pump some big responses through and make sure it stays functional, like 1 gig

These are just a start, see if you can think of anything.

thedadams:
This meta issue is being superseded by specific issues now.

PR #3634, opened 2025-07-30
fix: handle empty responses from MCP servers

thedadams:
An empty response most likely means that the MCP server failed to start or crashed for some reason. Report such errors to the user in the UI.

Part of: https://github.com/obot-platform/obot/issues/3280

Screenshot 2025-07-30 at 10 08 20


ivyjeong13 (approved):
ui lgtm :+1:

Issue #3628, opened 2025-07-30
Fresh install should have access control rule for allowing `Everything` and `Everyone`


sangee2004:
**Describe the bug**
Fresh install should have access control rule for allowing `Everything` and `Everyone`

**To Reproduce**
1. Install obot using helm chart
2. Once obot server is accessible , notice that no access control rules are present.

**Expected behavior**
Access control rule for allowing `Everything` and `Everyone` should be present by default .

sangee2004:
Access control rule for allowing Everything and Everyone is present by default in fresh obot installs.

PR #3648, opened 2025-07-30
chore: add a default access control rule for everything and everyone

thedadams:
Issue: https://github.com/obot-platform/obot/issues/3628


cjellick:
this will "do the right thing" if the admin deletes or modifies the rule right? the right thing being, leave it untouched and not pave it.

Issue #3582, opened 2025-07-28
Helm install - Command to fetch bootstrap password presents the password with "%" appended to the password.


sangee2004:
**Describe the bug**
Helm install - Command to fetch bootstrap password presents the password with "%" appended to the password.

**To Reproduce**
1. Install obot using helm chart
Following information is presented
```
% helm install obot chart/ --values chart/myvalues.yaml
NAME: obot
LAST DEPLOYED: Tue Jul 29 15:41:01 2025
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
Thank you for installing obot.

Your release is named obot.

To learn more about the release, try:

  $ helm status obot
  $ helm get all obot

You can retrieve the bootstrap token by running:
  $ kubectl get secret -n default obot-config -ojson | jq -r .data.OBOT_BOOTSTRAP_TOKEN | base64 -d
```
2. Try to fetch the bootstrap password using the suggested command
Notice that the bootstrap password is presented with an  extra "%" appended to the password.
```
kubectl get secret -n default obot-config -ojson | jq -r .data.OBOT_BOOTSTRAP_TOKEN | base64 -d
lEyQ2vPvmZAEu1C0iiKvw2xoZTL1xmkb% 
```

**Expected behavior**
Provide bootstrap password without the extra "%" .

cjellick:
i dont know who to assign this to, so @StrongMonkey you're it/

StrongMonkey:
I think this is a zsh issue....when the output doesn't have new line. it appends % at the end of the output..

sangee2004:
@StrongMonkey Just to confirm , we will not be providing a fix for this issue ?

When I test with bash, there is no new line when bootstrap password is presented

```
bash-3.2$ kubectl get secret -n default obot-config -ojson | jq -r .data.OBOT_BOOTSTRAP_TOKEN | base64 -d
Qg8Qjn7xJYqmQGZUXmAb2NcxumATCHx8bash-3.2$ 
```

StrongMonkey:
We can have a better command to print the bootstrap token with new line to give better user experience. I will fix it in our chart.

sangee2004:
This issue is addressed when testing with latest version from `main` for both zsh and bash.

PR #3665, opened 2025-07-30
Fix: fix helm install command to fetch token

StrongMonkey:
The previous helm command doesn't add new line at the end. this will produce weird output when user is using zsh. So add a new line to improve user experience.

https://github.com/obot-platform/obot/issues/3582


drpebcak (approved):
This is pretty standard behavior, but there's no harm in appending the newline.

StrongMonkey:
@drpebcak can you force merge it? It doesn't trigger build so I can't merge

Issue #3451, opened 2025-07-18
Catalog - Allowed to add duplicate Git source Urls. Deleting the duplicate source url results in all catalog entries(from the existing catalog url) to get deleted.


sangee2004:
**Describe the bug**
Catalog - Allowed to add duplicate Git source Urls. Deleting the duplicate source url results in all catalog entries(from the existing catalog url) to get deleted.

**To Reproduce**
1. Using `Add severs from Git` option add a source url - say `https://github.com/sangee2004/mymcp-catalog` in my case.
2. Wait for all mcp server entries from Git source to get synced.
3. Using `Add severs from Git` option add the same  source url `https://github.com/sangee2004/mymcp-catalog` from step1.
Even after multiple `refresh` , I dont see duplicate catalog entries.

4. Delete one of the duplicate source Git Source URLs.

Click on` Refresh` 

Notice that both git source urls entries get deleted and all catalog entries relating to the git source url gets deleted.

https://www.loom.com/share/ec5c9647e7f34f3db2862ef859ef35fa?sid=36e9418d-38f5-4658-bce6-c6d35efbbc6c

**Expected behavior**
User should not be allowed to add duplicate Git source Url.

sangee2004:
We are not allowed to add duplicate Git Source urls anymore.

Following error message is presented to this user

<img width="545" height="362" alt="Image" src="https://github.com/user-attachments/assets/f0e33d69-b333-474d-b398-942462368563" />

This error message itself can be improved. Will log a different issue to track it.

PR #3683, opened 2025-07-31
Fix: Don't allow duplicate URL in catalog server

StrongMonkey:
https://github.com/obot-platform/obot/issues/3451

### DIFF
--- a/pkg/api/handlers/mcpcatalogs.go
+++ b/pkg/api/handlers/mcpcatalogs.go
@@ -105,6 +105,17 @@ func (h *MCPCatalogHandler) Update(req api.Context) error {
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/api/handlers/mcpcatalogs.go
```diff
@@ -105,6 +105,17 @@ func (h *MCPCatalogHandler) Update(req api.Context) error {
 		}
 	}
 
+	// Check for duplicate URLs
+	seen := make(map[string]bool)
```
thedadams: nit:
```suggestion
	seen := make(map[string]struct{}, len(manifest.SourceURLs))
```

tybalex (approved):

Issue #3685, opened 2025-07-31
Provide better user experience when chatting for the first time after connecting to an MCP sever were it takes ~15 secs for the response to be received.


sangee2004:
**Describe the bug**
Provide better user experience when chatting for the first time after connecting to an MCP sever were it takes ~15 secs for the response to be received.

**To Reproduce**
1. Connect to a single user MCP server from `Available Connector` list
2. Notice that it take a very long time close to 15 -20 seconds for the first chat to respond .

See this video after 35 seconds . I hit a 403 issue initially.

https://www.loom.com/share/5065e228354f43048a5a7756fce0fb3b?sid=01e5f9c7-b144-4ec3-b4ea-5740fdf50db6

**Expected behavior**
Can we provide better user experience when chatting for the first time and user had to wait for 15-20 secs ? May be we can provide some text indicating that we are getting the connector deployed and  would be ready for chatting in a bit.

cjellick:
@thedadams any idea why the first chat can take so long? is that just bc the server is spinning up?

thedadams:
Yes, in this case it is deployment Kubernetes objects and waiting for them to be ready so we can list the tools.

sangee2004:
User is now presented with following page that shows progress for configuring and Initializing connector deployment after which is when the `Chat` option is presented to the user.

<img width="914" height="396" alt="Image" src="https://github.com/user-attachments/assets/3d0f4b5e-357c-40f1-b329-ef7b64210ff1" />

Once in chat view, it takes 3 -4 secs to get the response for the chat message.

PR #3715, opened 2025-08-01
enhance: add an MCP server launch API

thedadams:
This API simply ensures that the MCP server is launched, running, and responding to requests. This should only be used for non-remote MCP servers because it doesn't handle OAuth. Use the /check-oauth or /oauth-url APIs for those.

Also, this change moves /check-oauth to a POST request because that's probably what it should have been the whole time.

Part of https://github.com/obot-platform/obot/issues/3685


cjellick:
> Also, this change moves /check-oauth to a POST request because that's probably what it should have been the whole time.

And just to double check - it looks like you handle changing the frontend to POST as well, so that this can go in without breaking anything. is that right?

cjellick (approved):

thedadams:
>And just to double check - it looks like you handle changing the frontend to POST as well, so that this can go in without breaking anything. is that right?

Yes. Plus those functions aren't used anywhere in the UI right now. So, double-safe.

Issue #3719, opened 2025-08-02
Audit Logs - Searching with special characters (`tools/call`) or empty space (`DuckDuckGo Search`) finds no match.


sangee2004:
**Describe the bug**
Audit Logs - Searching with special characters (`tools/call`) or empty space (`DuckDuckGo Search`) finds no match.

**To Reproduce**
1. Go to 'Audit Logs'

<img width="1030" height="697" alt="Image" src="https://github.com/user-attachments/assets/d8e3633a-d78a-4432-ac40-1bacf28884e3" />

2. Search for all tools call by using  `tools/call`
Notice no matched entries are returned

<img width="1085" height="638" alt="Image" src="https://github.com/user-attachments/assets/806379bd-6489-487c-a377-a6b8df1e8697" />

This the query parameter used for the api call  `&query=tools%252Fcall`
3. Search for strings with empty space
Notice no matched entries are returned

<img width="1017" height="555" alt="Image" src="https://github.com/user-attachments/assets/4db83298-ce00-4bd1-8755-a7b0a0d4bf07" />

This the query parameter used for the api call `&query=%2522DuckDuckGo%2520Search%2522`

**Expected behavior**
Should be able to search with special characters/space in search string.

sangee2004:
This issue is addressed when testing with latest build.

PR #3727, opened 2025-08-03
fix: stop encode audit log query parameter

thedadams:
Issue: https://github.com/obot-platform/obot/issues/3719

### DIFF
--- a/ui/user/src/routes/admin/audit-logs/+page.svelte
+++ b/ui/user/src/routes/admin/audit-logs/+page.svelte
@@ -106,7 +106,7 @@
 		end_time: timeRangeFilters.endTime?.toISOString() ?? '',
 		limit: pageLimit,
 		offset: pageOffset,
-		query: encodeURIComponent(query)
+		query: query
 	});
 
 	afterNavigate(() => {


ui/user/src/routes/admin/audit-logs/+page.svelte
```diff
@@ -106,7 +106,7 @@
 		end_time: timeRangeFilters.endTime?.toISOString() ?? '',
 		limit: pageLimit,
 		offset: pageOffset,
-		query: encodeURIComponent(query)
+		query: query
```
ivyjeong13, line 109: Havent checked if the query in url params is encoded but wondering if it should be decodeUriComponent instead?

thedadams: Hmm. Not sure what you are asking, but I will try to give more detail to see if it helps.

The issue is that we were double-encoding the query. So, when it got to the server, we would decode it, but it was still encoded.

I believe the issue was that we were encoding here, but `buildQueryString`, which is used when we send the request to the server for this, also encodes the query parameters.

I did verify that the query parameters are still encoded when the request is made to the backend, even with this change.

ivyjeong13: Ah okay! If you're decoding server side then nevermind! Just wanted to make sure if we needed to decode it after getting from URL params from client side but sounds like its not the case. Thanks Donnie!

ryu-man:
@thedadams I think this is what's causing the issue. https://github.com/ryu-man/obot/blob/95382bacb21a05ea690f1d3d7e22bbb495f749b3/ui/user/src/routes/admin/audit-logs/%2Bpage.svelte#L65
The query is not being decoded when it is retrieved from the URL.

ryu-man (approved):

PR #3734, opened 2025-08-04
fix: alleviate block when deleting users when using sqlite

thedadams:
When we use sqlite, we have a single database connection. Also, sqlite's locking is not as sophisticated as others. So, when we're deleting a user in a transaction, that locks the database. Then, we're blocked when we try to clean up other resources associated with the user from inside the transaction.

This change alleviates that blocking by cleaning up all user-based resources in the user-delete controller.

### DIFF
--- a/pkg/controller/handlers/cleanup/user.go
+++ b/pkg/controller/handlers/cleanup/user.go
@@ -1,7 +1,9 @@
 package cleanup
 
 import (
+	"context"
 	"errors"
+	"fmt"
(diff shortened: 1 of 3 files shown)


g-linville (approved):
We don't care about deleting sessions anymore?

njhale (approved):

thedadams:
@g-linville They are already being deleted in the user-delete controller.

Issue #3711, opened 2025-08-01
Audit Logs - `404` errors relating to `user not found` presented when trying to show audit logs relating to deleted users.


sangee2004:
**Describe the bug**
Audit Logs - `404` errors relating to `user not found` presented when trying to show audit logs relating to deleted users.

**To Reproduce**
1. As `user1` , connect to any MCP server and make some chats relating to tool calls
2. As admin, go to `Audit Logs` to see the logs relating to this MCP server
3. Delete `user1`.
4. go to `Audit Logs` 

Notice `404` errors relating to `user not found` presented when trying to show audit logs relating to deleted users.

<img width="1159" height="784" alt="Image" src="https://github.com/user-attachments/assets/fb8bd885-e36d-42f8-a533-3e941799e7a1" />

**Expected Behavior:**
User should not be presented with  `404` errors relating to `user not found` error messages.

5. Try to use the Filter - `By User` . This also shows `404 errors` which prevents the user drop down from being shown to the user.
 
<img width="557" height="871" alt="Image" src="https://github.com/user-attachments/assets/afaae745-f1e4-4a0e-becc-7cd00c2e23ea" />

**Expected Behavior:**
User should not be presented with  `user not found` error message and User dropdown should be presented .

ryu-man:
@sangee2004 Could you provide more explanation on the expected behavior? I don't quite understand why to show users filter dropdown if a user is not found. Selecting only existing users does not sound like good UX.

ryu-man:
I am having some errors when deleting a user!

Response Message:
```
failed to delete user: the server was unable to return a response in the time allotted, but may still be processing the request (get threadauthorizations.obot.obot.ai)
```

<img width="1422" height="404" alt="Image" src="https://github.com/user-attachments/assets/50bc4a8b-eb64-4449-b9ad-67a03fcd3463" />

sangee2004:
@sangee2004 Could you provide more explanation on the expected behavior? I don't quite understand why to show users filter dropdown if a user is not found. Selecting only existing users does not sound like good UX. ==> @ryu-man when one of the users is deleted , we should still show all other users in the the `By User`  filter  drop down list . Currently `By User` drop down gets disabled with no user dropdown provided when we have deleted user.

ryu-man:
@sangee2004 Thanks, I got it now

sangee2004:
This issue is not seen anymore when testing with latest build.

[...]

PR #3735, opened 2025-08-04
fix: hide user not found errors and make it possible to filter by user after user is deleted

ryu-man:
Addresses #3711 

- Add the ability to disable logging errors when fetching by user ID.
- trim user ID
- Cache the fetch user promise for 1 minute

### DIFF
--- a/ui/user/src/routes/admin/audit-logs/+page.svelte
+++ b/ui/user/src/routes/admin/audit-logs/+page.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
 	import { fade, slide } from 'svelte/transition';
 	import { X, ChevronLeft, ChevronRight } from 'lucide-svelte';
-	import { throttle } from 'es-toolkit';
+	import { delay, throttle } from 'es-toolkit';
 	import { page } from '$app/state';
 	import { afterNavigate, goto } from '$app/navigation';
 	import { type DateRange } from '$lib/components/Calendar.svelte';
@@ -175,17 +175,40 @@
 		}
 	}
 
-	async function fetchUserById(id: string) {
-		const cache = users.get(id);
+	let fetchUserPromises: Map<string, Promise<OrgUser>> = new Map<string, Promise<OrgUser>>();
 
-		if (cache) {
-			return cache;
+	async function fetchUserById(id: string): Promise<OrgUser | undefined> {
+		// Input validation
+		if (!id || typeof id !== 'string' || id.trim() === '') {
+			console.warn('fetchUserById: Invalid user ID provided:', id);
+			return undefined;
 		}
 
-		const remote = await getUser(id);
-		users.set(id, remote);
+		const trimmedId = id.trim();
 
-		return remote;
+		if (users.has(trimmedId)) {
+			return users.get(trimmedId);
+		}
+
+		if (!fetchUserPromises.has(trimmedId)) {
+			fetchUserPromises.set(trimmedId, getUser(trimmedId, { dontLogErrors: true }));
+		}
+
+		const fetchUserPromise = fetchUserPromises.get(trimmedId)!;
+		try {
+			const remote = await fetchUserPromise;
+
+			users.set(trimmedId, remote);
+
+			// Invalidate promise cache after 1 minute
+			delay(1000 * 60).then(() => {
+				fetchUserPromises.delete(trimmedId);
+			});
+
+			return remote;
+		} catch (_) {
+			return undefined;
+		}
 	}
 
 	function getFilterDisplayLabel(key: keyof AuditLogURLFilters) {
@@ -222,7 +245,7 @@
 		}
 
 		if (label === 'user_id') {
-			return (await fetchUserById(value + '')).displayName;
+			return (await fetchUserById(value + ''))?.displayName;
 		}
 
 		return Promise.resolve(value + '');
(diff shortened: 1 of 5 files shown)


ui/user/src/routes/admin/audit-logs/+page.svelte
```diff
@@ -175,17 +175,40 @@
 		}
 	}
 
-	async function fetchUserById(id: string) {
-		const cache = users.get(id);
+	let fetchUserPromises: Map<string, Promise<OrgUser>> = new Map<string, Promise<OrgUser>>();
 
-		if (cache) {
-			return cache;
+	async function fetchUserById(id: string): Promise<OrgUser | undefined> {
```
ivyjeong13: Screenshot 2025-08-05 at 11 27 01 AM

Not seeing user being displayed for table, the pill, or the filters dropdown. Looking through it, I'm wondering why `fetchUserById` is needed? It doesn't look like the fetch for audit logs is an EventStream which I thought this be what it's used for. I still think `users` should be reactive and can't you just use that to display the information needed to display since it gets refetched after every `afterNavigate`?

Went through and quickly did:
* axed fetchUserById
* turned `users` reactive with `SvelteMap`
* passes `users` as a prop to `AuditLogs` and `AuditFilters`

And that worked for me. 

```
// page.svelte
let users = new SvelteMap<string, OrgUser>();

// ...

	function getFilterValue(label: keyof AuditLogURLFilters, value: string | number) {
		if (label === 'start_time' || label === 'end_time') {
			return new Date(value).toLocaleString(undefined, {
				year: 'numeric',
				month: 'short',
				day: 'numeric',
				hour: '2-digit',
				minute: '2-digit',
				second: '2-digit',
				hour12: true,
				timeZoneName: 'short'
			});


		}

		if (label === 'user_id' && value?.toString()) {
			const userMatch = users.get(value.toString());
			return userMatch?.email || userMatch?.username || value.toString();
		}

		return value;
	}

// ...

						{#each values as value (value)}
							{@const isMultiple = values.length > 1}
							{@const displayValue = getFilterValue(key, value)}

							{#if isMultiple}
								<span class="font-light">
									<span>{displayValue}</span>
								</span>


								<span class="mx-1 font-bold last:hidden">OR</span>
							{:else}
								<span class="font-light">{displayValue}</span>
							{/if}

						{/each}
```

```
// AuditLogs.svelte
	let {
		data = [],
		onSelectRow,
		emptyContent,

		currentFragmentIndex = 0,
		getFragmentIndex,
		getFragmentRowIndex,
		onLoadNextFragment,
		usersMap
	} = $props();

	let tableData = $derived(
		data.map((item) => ({
			...item,
			user: usersMap.get(item.userID)
		}))
	);

// ...
				<!-- Audit Data Rows -->
				{#each tableData as item, i (item.id)}
					{@const fragmentIndex = getFragmentIndex?.(i)}
					{@const fragmentRowIndex = getFragmentRowIndex?.(i)}
					<tr


@@ -133,11 +134,7 @@
								.replace(/,/g, '')}</td
						>
						<td class="px-6 py-4 text-sm whitespace-nowrap">
							{item.user?.email ?? item.user?.username ?? 'Unknown User'}
```

```
// AuditFilters.svelte
	interface Props {
		filters?: AuditLogURLFilters;
		onClose: () => void;
		usersMap: SvelteMap<string, OrgUser>;
		getFilterDisplayLabel?: (key: keyof AuditLogURLFilters) => string;
	}

	let { filters: externFilters, onClose, usersMap, getFilterDisplayLabel }: Props = $props();

// ...


// in processLog
			if (filterId === 'user_id') {
				return response.options.map((d) => {
					const user = usersMap.get(d);
					return { id: d, label: user?.email || user?.username || d };
				});
```

ryu-man: I am going to make users a reactive map, although it does work well on my end. I'm not sure what's going on.

image
image

ryu-man: @ivyjeong13 I made the `users` map reactive and got rid of the `getUserById` but I introduced a new function `getUserDisplayName` to keep one source of truth.

ui/user/src/routes/admin/audit-logs/+page.svelte
```diff
@@ -175,17 +176,9 @@
...
-
-		if (cache) {
-			return cache;
-		}
-
-		const remote = await getUser(id);
-		users.set(id, remote);
-
-		return remote;
+	function getUserDisplayName(id: string): string {
```
ivyjeong13, line 179: Screenshot 2025-08-05 at 2 35 54 PM


I'm seeing 'Unknown User' and I think it's because this function isn't reactive to when `users` changes. Logging inside the function:

```
// inside getUserDisplayName
SvelteMap
[[Entries]]
No properties

// outside checking with effect in +page.svelte
[[ Entries ]]
{
    "key": "2",
    "value": {
        "id": "2",
        "created": "2025-08-01T17:04:44.183324-04:00",
        "username": "107221547212253478536",
        "role": 1,
        "email": "ivy@acorn.io",
        "timezone": "America/New_York",
        "lastActiveDay": "2025-08-05T00:00:00Z"
    }
}
```

ryu-man: I just made a simulation of late initialization of the users map, and I found that the function is reactive and the values do update in all places.

ryu-man: `Unknown User` is displayed when there is a user that was deleted, so does that user still exist in the database?

ivyjeong13 (approved):

Issue #3721, opened 2025-08-02
Audit Logs - Not able to enter time as part of Time range


sangee2004:
**Describe the bug**
Audit Log - Not able to enter date as part of Time range

**To Reproduce**
1. Go to `Audit Log`
2. Click on `Filter By Date`
3. Try to clear the existing time to add an new date

Notice that `Invalid Date` error message gets presented and user is not allowed to enter time anymore

<img width="390" height="603" alt="Image" src="https://github.com/user-attachments/assets/d7157ae9-f366-4910-acf3-832999e55cb7" />

https://www.loom.com/share/f28bbb047bcd4b4f9f87d8757f83a926?sid=fddbe939-ef45-4dd7-881e-e0849465edd6

**Expected behavior**
Should be  able to enter date as part of Time range

sangee2004:
This issue is addressed when testing with latest builds.
Able to enter time range successfully when 1 day is chosen from the calendar view.

Filters get applied as expected and we are able to fetch the Audit entries relating to the applied filter.

PR #3741, opened 2025-08-04
fix: set value to zero when time input is cleared

ryu-man:
Addresses #3721

### DIFF
--- a/ui/user/src/lib/components/TimeInput.svelte
+++ b/ui/user/src/lib/components/TimeInput.svelte
@@ -27,6 +27,22 @@
 				if (['ArrowDown', 'ArrowUp'].includes(ev.key)) {
 					ev.preventDefault();
 				}
+
+				if (
+					(ev.key === 'Backspace' || ev.key === 'Delete') &&
+					ev.currentTarget.value.length === 1
+				) {
+					// Reset to 0 if the input is cleared
+
+					ev.preventDefault();
+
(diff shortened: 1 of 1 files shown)


ui/user/src/lib/components/TimeInput.svelte
```diff
@@ -27,6 +27,22 @@
 				if (['ArrowDown', 'ArrowUp'].includes(ev.key)) {
 					ev.preventDefault();
 				}
+
+				if (
```
ivyjeong13: https://www.loom.com/share/9e404690f5f24e3da07ea9e27f4b7159

Loom above summary:
nit: When I keep pressing 1 in the hour field, it cycles between 11 -> 3 -> 7, would be nice if it kept locked into a valid hour. Similarly would be nice for minutes.

Related to the input breaking: When I CMD+A and hit Delete, it clears out all the inputs and I can no longer supply a value into any of them until it re-renders (by closing out the time input and opening it up again).

ryu-man: >pressing 1 in the hour field, it cycles between 11 -> 3 -> 7

Now when a user enters a value over the max (hour/minute), it will set to the max.

>When I CMD+A and hit Delete, it clears out all the inputs and I can no longer supply

Now if the input is cleared, it will set to 00

ivyjeong13 (approved):

Issue #3579, opened 2025-07-28
Usage graphs


cjellick:
Details in this loom: 
https://www.loom.com/share/476b5ece984b4078ab79aef2ca7fa21e

## Graphs to Implement

Each graph below should support **time frame** and **filter controls**:

* **Time Frame**: User-selectable range (e.g., last 7 days, custom date range).
* **Filters**: At a minimum, by **server** and **user**.
* **Pagination**: Each graph should support pagination
* We want these both as a left-nav menu item (nested under MCP Servers, like audit logs) AND as a tab when you click on an individual MCP server (obviously scoped to just that server) 

### 1. **Most Frequent Tool Calls**

* X-axis: Tool name
* Y-axis: Number of calls
* Notes: Paginated (e.g., arrows to browse top N tools beyond top 10).

### 2. **Most Frequently Used Servers**

* X-axis: Server name
* Y-axis: Number of tool calls
* Note: Filtering by server will reduce this to a single bar, but that’s acceptable.

### 3. **Tool Call Response Times (Average)**

* X-axis: Tool name
* Y-axis: Average response time over the selected time frame

### 4. **Tool Call Response Times (Individual)**

* X-axis: Time (or sequential order of calls)
* Y-axis: Response time
* Notes: Intended for identifying outliers. One tool may dominate the graph.

### 5. **Errors by Tool Call**

* X-axis: Tool name
* Y-axis: Count of error responses

### 6. **Errors by Server**

* X-axis: Server name
* Y-axis: Count of error responses

### 7. **Server Creations**

* X-axis: Time
* Y-axis: Count of servers created over time

### 8. **Server Connections (by Server)**

* X-axis: Server name
* Y-axis: Count of connection initializations

### 9. **Most Active Users**

* X-axis: User name
* Y-axis: Number of tool calls

### 10. **Connections per User**

* X-axis: User name
* Y-axis: Number of connection events
* Notes: May surface “noisy” users who connect/disconnect frequently

### **Layout**

* Dedicated **Usage** tab or section, separated from **Audit Logs**
* Graphs grouped logically (e.g., Tool Metrics, Server Metrics, User Metrics)

cjellick:
@njhale once you clear your other times, let's see what you can come up with here ... i may kick it back to ivy if she frees up close to when you start this as i think there will be a lot of styling nitpicking going on with this one.

fyi - data and apis are already there though they may need some small tweaks.

njhale:
[...]

PR #3747, opened 2025-08-04
enhance: server side usage filtering and multiple values filter

njhale:
- Add server-side support for filtering by multiple user IDs and MCP
server display names to `/api/mcp-stats`
- User server side filtering to filter usage stats
- Enable filtering usage by multiple users and mcp server names

Follow up for https://github.com/obot-platform/obot/issues/3579

Loom: https://www.loom.com/share/031530f2016c4bc39bcf87f8b83dcdfe?sid=d7f448f7-1e98-4e19-8f64-da4cbd6c4f42

### DIFF
--- a/pkg/gateway/client/mcpauditlog.go
+++ b/pkg/gateway/client/mcpauditlog.go
@@ -154,11 +154,14 @@ func (c *Client) GetMCPUsageStats(ctx context.Context, opts MCPUsageStatsOptions
 		if opts.MCPID != "" {
 			tx = tx.Where("mcp_id = ?", opts.MCPID)
 		}
-		if opts.MCPServerDisplayName != "" {
-			tx = tx.Where("mcp_server_display_name = ?", opts.MCPServerDisplayName)
+		if len(opts.UserIDs) > 0 {
+			tx = tx.Where("user_id IN (?)", opts.UserIDs)
 		}
-		if opts.MCPServerCatalogEntryName != "" {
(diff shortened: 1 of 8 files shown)


pkg/gateway/client/mcpauditlog.go
```diff
@@ -154,11 +154,14 @@ func (c *Client) GetMCPUsageStats(ctx context.Context, opts MCPUsageStatsOptions
...
+		if len(opts.UserIDs) > 0 {
+			tx = tx.Where("user_id IN (?)", opts.UserIDs)
 		}
-		if opts.MCPServerCatalogEntryName != "" {
-			tx = tx.Where("mcp_server_catalog_entry_name = ?", opts.MCPServerCatalogEntryName)
+		if len(opts.MCPServerDisplayNames) > 0 {
+			tx = tx.Where("mcp_server_display_name IN (?)", opts.MCPServerDisplayNames)
+		}
+		if len(opts.MCPServerCatalogEntryNames) > 0 {
+			tx = tx.Where("mcp_server_catalog_entry_name IN (?)", opts.MCPServerCatalogEntryNames)
```
g-linville, line 164: Have you verified that this syntax works in both SQLite and Postgres?

njhale: Ran this with SQLite, but it should be compatible with Postgres. I'll spin up a Postgres instance and test it to be sure before this goes in

njhale: yup, looks like this works fine with Postgres as well

PR #3761, opened 2025-08-04
fix: stop leaking when client SSE is disconnected

thedadams:
Prior to this change, when the user's MCP client was disconnected, the message handler would leak goroutines because the msg.Send call would never complete. It was blocked waiting for someone to read the message.

A change was made to nanobot to detect when no one is listening to these messages. A "no reader" error is returned in this case, and Obot will handle this. When this error is returned, the audit log is not persisted, as if no message was sent at all (which would happen if the user's client wasn't connected to the downstream MCP server).

However, this led to quite a few changes to handle memory issues (ensuring that the server session is cached in memory) and client issues (ensuring each client has its own server session and message handler). The remaining code for loading MCP server tools in GPTScript was removed now the session manager is Obot is self-contained.

This change is bigger than it looks. A majority of the additions are pulling code from GPTScript into Obot so we can control it more easily. I've tried to call that out where appropriate here.


pkg/mcp/client.go
```diff
@@ -5,12 +5,22 @@ import (
...
-func (sm *SessionManager) ClientForMCPServerWithOptions(ctx context.Context, mcpServer v1.MCPServer, serverConfig ServerConfig, opts ...nmcp.ClientOption) (*mcp.Client, error) {
+type Client struct {
+	*nmcp.Client
+	ID     string
+	Config ServerConfig
+}
+
+func (c *Client) Capabilities() nmcp.ServerCapabilities {
+	return c.Session.InitializeResult.Capabilities
+}
```
thedadams, line 21: This is pulled from GPTScript.

pkg/mcp/client.go
```diff
@@ -41,16 +51,74 @@ func (sm *SessionManager) ClientForServer(ctx context.Context, userID, mcpServer
...
+
+func splitIntoMap(list []string) map[string]string {
+	result := make(map[string]string, len(list))
+	for _, s := range list {
+		k, v, ok := strings.Cut(s, "=")
+		if ok {
+			result[k] = v
+		}
+	}
+	return result
```
thedadams: This is pulled from GPTScript.

pkg/mcp/loader.go
```diff
@@ -110,16 +113,35 @@ func NewSessionManager(ctx context.Context, defaultLoader *gmcp.Local, tokenStor
...
+	}()
+
+	var errs []error
+	for id, session := range sm.sessions {
+		log.Infof("closing MCP session %s", id)
+		session.Session.Close()
+		session.Session.Wait()
+	}
+
+	return errors.Join(errs...)
```
thedadams: This is pulled from GPTScript.

pkg/mcp/loader.go
```diff
@@ -135,21 +157,44 @@ func (sm *SessionManager) CloseClient(ctx context.Context, server ServerConfig)
...
+
+	sm.lock.Unlock()
+
+	if session != nil && session.Client != nil {
+		session.Session.Close()
+		session.Session.Wait()
+	}
+
+	return nil
+}
```
thedadams: This is pulled from GPTScript.

pkg/mcp/runner.go
thedadams, line 1: This is pulled from GPTScript.

pkg/mcp/types.go
```diff
@@ -21,12 +20,25 @@ type Config struct {
...
+	Env                []string `json:"env"`
+	URL                string   `json:"url"`
+	Headers            []string `json:"headers"`
+	// Scope ensures separate deployments for each user and project.
+	Scope string `json:"scope"`
+	// ClientScope is used to scope clients to sessions.
+	// This value will not be used to determine if a new deployment of the server is needed, but is used to ensure the
+	// client is new for the session.
+	ClientScope  string   `json:"clientScope"`
+	AllowedTools []string `json:"allowedTools"`
```
thedadams: These fields, except for `ClientScope` are pulled from GPTScript. I also added `ClientScope` here.

g-linville (approved):

thedadams:
The `ClientScope` thing didn't feel right, but I couldn't put my finger on why. I just figured out why and I need to update this. I will add an extra commit for easier review.

g-linville (approved):

Issue #3725, opened 2025-08-03
Audit Logs - Accessing `Audit Logs` from MCP Server tab still serves old version of audit log view.


sangee2004:
**Describe the bug**
Audit Logs - Accessing `Audit Logs` from MCP Server tab still serves old version of audit logs.

**To Reproduce**
1. Select any MC Server entries that have some active users.
2. Go to Audit Logs`tab

Notice that old version of audit log view is presented with stats graphs and search being local UI search(not server side)

<img width="1222" height="833" alt="Image" src="https://github.com/user-attachments/assets/3eb76078-2eac-4b3c-ba48-81a8753dc077" />

**Expected behavior**
This Audit Log view should follow the same pattern as `Audit Log` presented when accessing `Audit Logs` option under `MCP Servers` .

sangee2004:
This issue is addressed when testing with latest build.

PR #3768, opened 2025-08-05
fix: make mcp server audit logs tab match audit logs page

ryu-man:
Addresses #3725 

- refactor audit logs page
- create audit log page content
- properly pass filters to audit logs slideover when declared in the MCP server tab


ui/user/src/lib/components/admin/McpServerEntryForm.svelte
```diff
@@ -362,41 +362,36 @@
...
-						This server has not had any active usage in the last 7 days.
-					</p>
-					{#if mcpId || mcpCatalogEntryId}
-						{@const param = mcpId ? 'mcpId=' + mcpId : 'entryId=' + mcpCatalogEntryId}
+		{@const name = 'manifest' in entry ? entry.manifest.name : null}
+		{@const mcpId = 'isCatalogEntry' in entry ? null : entry.id}
+		{@const mcpCatalogEntryId = 'isCatalogEntry' in entry ? entry.id : null}
+
+		<div class="mt-4 flex min-h-full flex-col gap-8 pb-8">
+			<AuditLogsPageContent {mcpId} {mcpCatalogEntryId} mcpServerDisplayName={name}>
```
ivyjeong13: Screenshot 2025-08-06 at 3 15 14 PM
Screenshot 2025-08-06 at 3 15 05 PM

Seeing 'Everything' logs in the /audit-logs page but when going to 'Everything' and clicking the Audit Logs tab, it's not showing anything.

Switched back to `main` to check and am seeing the logs in the Audit Logs tab there.

ryu-man: @ivyjeong13 Does this happen with every server?

ivyjeong13: > @ivyjeong13 Does this happen with every server?

I can't check right now cause I blew away my local to work on an issue, but this one was a multi-user server.

ryu-man: The issue is the `{@const mcpId = 'isCatalogEntry' in entry ? null : entry.id}` does not return the correct MCP ID, and I could not find the correct value. In my case the filters in the audit logs page show `msi1-2-ms1q94fl` and in the MCP servers logs tab it applies `ms1q94fl`

ryu-man: I skipped filtering by mcp id in the mcp server audit logs tab and data start to show

ivyjeong13: Maybe check with Donnie on that, ids that start with `ms` are the servers and ids starting with `msi` are server instances and I think either should be passable to mcpId filter?

mcpServerDisplayName should work fine though so will approve

Issue #3758, opened 2025-08-04
Audit Logs - Search for user name is case sensitive.


sangee2004:
**Describe the bug**
Audit Logs - search for user name is case sensitive.

**To Reproduce**
1. Go to 'Audit Logs`
2. Search for username in all lower case , when the username starts with upper case

Notice that no entries get returned
<img width="855" height="373" alt="Image" src="https://github.com/user-attachments/assets/0ff7bfeb-4c97-45ca-94ce-58115fbd1692" />

Searching with matched case works as exected
<img width="852" height="689" alt="Image" src="https://github.com/user-attachments/assets/a8a50664-9d0f-495e-9c47-f6a2622896e8" />

**Expected behavior**
Search for user name should be  case insensitive (similar to other fields like mcp server)

sangee2004:
This issue is addressed when testing with latest build.

PR #3771, opened 2025-08-05
fix: search audit logs by user case-insensitively

thedadams:
Issue: https://github.com/obot-platform/obot/issues/3758


pkg/gateway/client/mcpauditlog.go
```diff
@@ -43,8 +43,11 @@ func (c *Client) GetMCPAuditLogs(ctx context.Context, opts MCPAuditLogOptions) (
 		}
 
 		var userIDs []string
+		// Don't lowercase the query for the selecting from the database. Some databases aren't case-sensitive,
+		// and we want to preserve the case of the search term.
```
g-linville, line 47: This comment seems to contradict what we are doing? It says don't lowercase, but we are calling `strings.ToLower`

thedadams: It was confusing. I made a code-change to hopefully make it clearer.

njhale (approved):

PR #3791, opened 2025-08-06
add more quick actions

ryu-man:


ui/user/src/routes/admin/audit-logs/AuditLogsTimeline.svelte
```diff
@@ -234,12 +358,61 @@
...
+		};
+
+		let width = $state(getViewportWidth());
+
+		const onResize = debounce(() => {
+			width = getViewportWidth();
+		}, 1000 / 60);
+
+		$effect(() => {
+			window.addEventListener('resize', onResize);
```
ivyjeong13, line 388: Remove the addEventLIstener on destroy 

```
$effect(() => {
			window.addEventListener('resize', onResize);

			return () => {
				window.removeEventListener('resize', onResize);
			};
		});
```

ivyjeong13 (approved):
Just one comment but lgtm!

One thing I noticed while fiddling, not related to the timeline but the date time picker:

Screenshot 2025-08-07 at 4 24 04 PM

Would be a nice enhancement if the user can't pick an afterTime that is earlier than the beforeTime

ryu-man:
> One thing I noticed while fiddling, not related to the timeline but the date time picker:

Any guidance to reproduce the problem? 
The default behavior should be that when a user clicks on a date after the end date, the range expands forward, and if the user clicks on a date before the start date, the range should expand backwards.

ui/user/src/routes/admin/audit-logs/AuditLogsTimeline.svelte
ryu-man: done

PR #3821, opened 2025-08-07
chore: switch to gpt-5 by default

thedadams:
Just getting this ready so we can switch when we the time is right.


g-linville (approved):

tybalex (approved):

cjellick:
5 is to slow to be the default. we could do 5-instant.

thedadams:
@cjellick `gpt-5-instant` doesn't seem to be a thing in the API. Or, at least, my API key doesn't have access to it.

tybalex:
maybe we can do `gpt-5-mini`

Issue #3784, opened 2025-08-05
ability to control click the links in the profile menu to open in new tab


cjellick:
now that we have the links in the profile menu all opening in the same tab, i just want the ability to ctrl+click them like a normal link to open in a new tab. this will be impoartant when demoing

sangee2004:
Using Cmd+click on the profile menu for `chat` takes the user to a new tab as expected.

Using Cmd+click on the profile menu for `My Connectors`, `Admin Dashboard`, `My Account` opens a new tab but user is not taken to the new tab that was created.

@ivyjeong13 Reopening this issue for follow up.

ivyjeong13:
> Using Cmd+click on the profile menu for `chat` takes the user to a new tab as expected.
> 
> Using Cmd+click on the profile menu for `My Connectors`, `Admin Dashboard`, `My Account` opens a new tab but user is not taken to the new tab that was created.
> 
> [@ivyjeong13](https://github.com/ivyjeong13) Reopening this issue for follow up.

When clicking a link, the default behavior is 
cmd+click - new tab without focus
cmd+shift+click- new tab in focus

The chat link isnt actually a link, its a button and it does window.open which opens a new tab in focus when you click it with ctrl/cmd.

Trying to see if I can get the button to follow the same behavior as a link, otherwise I can change them all to be a button that does cmd+click with window.open

ivyjeong13:
@sangee2004 Just FYI for this, I did push through the change of making them all buttons, so the cmd+click action should be consistent with Chat, My Connectors, and Admin Dashboard (just realizing now I didn't do it for My Account 🤦‍♀️)

sangee2004:
@ivyjeong13 For `My Account` , I dont think  we want to open it in a new tab . We were not doing that even before https://github.com/obot-platform/obot/issues/3657#issuecomment-3138039712

ivyjeong13:
@cjellick Hey Craig, wondering if you have some time, can you check if cmd+clicking for you opens a new tab in focus for these links? Sangeetha is seeing it only opens the chat in focus. 🤔

cjellick:
In chrome, I see what sangeetha sees: chat is the only one that opens "in focus"
In firefox, they all consistently open in a new tab "out of focus"
In safar, chat doesnt open at all if i cmd+click


FWIW, i think "norma"l links open "out of focus", and i am fine with that behavior.

[...]

Issue #3271, opened 2025-06-26
Changes in catalog entries is not reflected after catalog refresh. Changes are reflected after a Manual refresh.


sangee2004:
**Describe the bug**
Changes in catalog entries is not reflected after catalog refresh. Changes are reflected after a Manual refresh.

**To Reproduce**
1. Create a MCP catalog .
2. Add MCP Servers using `Add servers from Git`.
3. This results in all existing entries from git location to get synced to catalog and list is presented to the user.
4. Add a new MCP Server entry to the github location.
5. Use "Refresh Catalog" option to get the new entry added in step4.

New entry does not get reflected in the UI .

At this time i see UI make `refresh` followed by `entries` api call. But `entries` api call does not fetch the new entry.

6. Refresh the page manually.

This results in the new entry being presented in the MCP Server list.

**Expected behavior**
New entries should be fetched without having to do a manual refresh of the page.

StrongMonkey:
@sangee2004 The refresh process is async and happens in the backend. The client is doing the right thing to call /entries after refresh but the entry might not be synced into the backend. I think the current behavior should be fine unless we want to do some aggressive await in the client/server side to wait for refresh to be done then refresh the entries.

cjellick:
@sangee2004 - i tend to agree with @StrongMonkey here, but please reply to his comment in case there is something we are missing about this issue.

sangee2004:
@cjellick @StrongMonkey Would it be worth putting a note to communicate to the user when `Refresh` is done that he would have to fetch the catalog entries manually to see the  synced changes?

If not , we can close this issue.

StrongMonkey:
Yes. I would suggest that the UI improvement we can do, is that when catalog is in refreshing state, show the status and when it is done, prompt user to refresh the entries.

g-linville:
Looks like this is a combo UX + backend change. I'll implement the backend to add an `isSyncing` status field to the catalog. The UI can do what it wants to display this.

g-linville:
Moved back to Triage for the UI portion of this.

sangee2004:
When testing with latest builds , I see the following behavior

When I use Refresh  in MCP Severs page, I am presented with `The catalog is currently syncing with the latest Git servers. Please check back in a few minutes or try refreshing` which never gets dismissed unless i reload the page. Do we expect for this message to get auto dismissed when syncing is actually done ? I waited long (like 1 minute)  and still dont see this message go away.

[...]

PR #3828, opened 2025-08-07
fix: change sync button text & info, add polling

ivyjeong13:
Addresses #3271 

Also addresses #3784 safari issue (verified safari,chrome,firefox)

### DIFF
--- a/ui/user/src/lib/components/navbar/Profile.svelte
+++ b/ui/user/src/lib/components/navbar/Profile.svelte
@@ -21,7 +21,6 @@
 	import { BOOTSTRAP_USER_ID } from '$lib/constants';
 	import { afterNavigate, goto } from '$app/navigation';
 	import PageLoading from '../PageLoading.svelte';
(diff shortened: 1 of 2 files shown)


ui/user/src/lib/components/navbar/Profile.svelte
```diff
@@ -142,7 +122,11 @@
 		<div class="flex flex-col gap-2 px-2 pb-4">
 			{#if profile.current.role === 1 && !inAdminRoute}
 				<button
-					onclick={(e) => handleNavigate('/admin/mcp-servers', e)}
+					onclick={(event) => {
+						if (!window) return;
```
ryu-man: I think checking for `window` using `typeof window === undefined` is better; I had server issues before using this check.

ivyjeong13: Went ahead and axed those checks since they're happening in the onclick!

Issue #3825, opened 2025-08-07
`/mcp-audit-logs` takes  2 secs to respond when there are lot of entries returned (~4500 entries)


sangee2004:
**Describe the bug**
`/mcp-audit-logs` takes  2 secs to respond when there are lot of entries returned (~4500 entries)

Setup with  ~4500 audit log entries.
**To Reproduce**
1. Go to `Audit Logs`

Notice that it takes 2 -3 secs for the the Audit entries to be presented after the default 3 months  time range filter is presented in the UI.

`/mcp-audit-logs` call made by UI takes  2 secs to respond 

<img width="1035" height="130" alt="Image" src="https://github.com/user-attachments/assets/6c682065-0be6-4a44-bc05-b019ccfb70f1" />

**Expected behavior**
Speed up the response time fro Audit log entries when there are lot of entries to be returned.

cjellick:
As mentioned by Ivy in the PR, I'd just prefer a loading screen.

cjellick:
I unassigned this one because I guess #3829 is the better one for tracking what I want done.

sangee2004:
When testing with latest build , we are presented with a loading screen.

There are 2 issues 
1. There is still the `No audit logs` page presented on which the `Loading Logs` message is presented.
2. Also  `Loading Logs` message continues to be presented for small time even after the audit entries are being presented.
https://www.loom.com/share/12781ec88ed74d528e286543b1550eeb
Logged this for tracking - https://github.com/obot-platform/obot/issues/3848

PR #3829, opened 2025-08-07
fix: reduce audit logs load time

ryu-man:
Addresses #3825 

- load data in `+page.ts` level.
- make small refactors.


ui/user/src/routes/admin/audit-logs/+page.ts
```diff
@@ -0,0 +1,30 @@
...
+			90
+		).toISOString();
+	}
+
+	if (!filters['end_time']) {
+		filters['end_time'] = set(new Date(), { milliseconds: 0, seconds: 0 }).toISOString();
+	}
+
+	return {
+		auditLogs: await fetchAuditLogs(filters)
```
ivyjeong13: If fetchAuditLogs is the one taking a long time to load, moving it to page.ts is not a great idea and won't make anything faster. Ideally, calls in page.ts should be super quick because it's the SSR part of svelte. (Tho on production, we have SSR turned off so this will happen clientside).

I'd think either we need to drop the time range to a reasonable range that the data won't take a long time to come back or need a graph database that can compile and give us the needed information much quicker?

ryu-man: > moving it to page.ts is not a great idea and won't make anything faster

Loading data in `page.ts` will start early as the user hovers over the sidebar audit logs link, and that will help reduce the waiting time; in the old approach, the data start loading after the UI is mounted.

>  it's the SSR part of svelte. (Tho on production, we have SSR turned off so this will happen clientside).

Load functions do work both on the server side and on the client side, even if SSR is turned off.

>I'd think either we need to drop the time range to a reasonable range that the data won't take a long time to come back

Indeed, this is a solution worth considering and could have a big impact on the loading time. But I will wait to see @cjellick's opinion.

ivyjeong13: I guess I'm wondering if we're worried about external navigation cases like if a user starts on a fresh chrome tab then puts `{url}/admin/audit-logs` in the url or has in their previous history and navigates directly to it. Or if they copy the url and send that to another admin via Slack or something. 😅 edit: These examples will not preload

ryu-man: >These examples will not preload

Yes, they will not preload, but the data still loads before the UI is ready. but I can't tell how effective it is going to be without testing it with pro data.

ivyjeong13: A Loom to kinda better explain:
https://www.loom.com/share/78af76d7b66d44509b6e3aff44d86b4f

And with internal navigation:
https://www.loom.com/share/bd2161d4caea481a89669d8a5192744c

Mostly just saying I prefer the long call happening as it was and we just add a spinner in that area where the data loads cause that looks better than any of the above happening in my opinion 😅 Then, in the meantime, we can figure out how to load the data faster.

cjellick: fwiw, im good with a spinner.

ryu-man:
Drop the load function; only add a page loading spinner.

ui/user/src/lib/components/admin/audit-logs/AuditLogsPageContent.svelte
```diff
@@ -66,6 +67,7 @@
 
 	const users = new SvelteMap<string, OrgUser>();
 
+	let showLoadingSpinner = $state(true);
```
ivyjeong13, line 70: Only thing, shouldn't showLoadingSpinner get set to true every time we need to make a call to fetch the audit logs?

ryu-man: just pushed the change, with an improved UI

Issue #3819, opened 2025-08-07
De configured auth provider shows up as configured until manual refresh.


sangee2004:
**Describe the bug**
De configured auth provider shows up as configured until manual refresh.

**To Reproduce**
1. De configured auth provider that is configured.

Notice that the de configured auth provider shows up as configured until manual refresh.

https://www.loom.com/share/bc4a820116234bf3bde1010b795d551f?sid=5a58af04-552b-4f88-95ca-f1fe00cc6ecf

**Expected behavior**
De configured auth provider should not show up as configured

sangee2004:
This issue is addressed when testing with latest build.

As soon as auth provider is de configured , user is brought back to this page as expected:

<img width="325" height="374" alt="Image" src="https://github.com/user-attachments/assets/d97fcd93-f23e-473f-ad57-04f0b7b23fec" />

Issue #3782, opened 2025-08-05
Clicking outside (or highlighting a text within the window) the configuration parameter window results in the window getting closed.


sangee2004:
**Describe the bug**
Clicking outside (or highlighting a text within the window) the configuration parameter window results in the window getting closed.

**To Reproduce**
1. Connect to any MCP server that has some configuration parameters to enter.
2. When entering  the configuration parameters , click outside this window. 
Notice that the window gets dismissed.
3. When entering  the configuration parameters , try to highlight any text . I was doing this to delete and correct what i had enetered.
This also results in configuration parameters window to be closed probably because as part of highlighting the text i might have just gotten out of the window's boundary.

https://www.loom.com/share/75873479dc8141f0a1f4d855e29599ea?sid=8c220598-0484-481b-a96f-d759974f7833

**Expected behavior**
When setting configuration parameters, users might have to go to other places and fetch the credentials they need to enter and might accidentally click out side of this window resulting in all eneterd parameters being lost with window getting closed.
It would be good to not auto close the window in such cases.

cjellick:
Tried this myself...it acutally isnt even quite the window boundary, clicking with the last 10ish pixels of the modal window seems to close it.

It is nice to click outside the modal to close, but I agree that it sucks if you do this accidentally and lose input. Could we add a "Are you sure you want to exit" if they have input some values into the modal? If the input fields are all blank, we should just close without asking.

sangee2004:
This issue is also seen when I try to edit the connector name from `Enabled Connector` list.

sangee2004:
When testing with latest build , we are not accidentally closing the window like we used to do before when highlighting the current text in the input field , or when attempting to delete the entire content (to start typing something)

@ivyjeong13 I dont see confirmation when closing the window ( explicitly closing the window or by clicking outside) when user has typed something.
```
Could we add a "Are you sure you want to exit" if they have input some values into the modal? If the input fields are all blank, we should just close without asking.
```
https://www.loom.com/share/3100c11d33b846a48a82b7affc86496f?sid=bff98fc7-994d-4c06-be4b-690b80f5d102

Reopening from followup

sangee2004:
User is now presented with the following confirmation message when user has something typed in the form and clicks outside of the window:


<img width="776" height="667" alt="Image" src="https://github.com/user-attachments/assets/68099c5c-37da-47d9-b23d-9c6c60a1fac3" />

Issue #3513, opened 2025-07-23
Add the ability to specify "badges" on a server


cjellick:
Will really likes the badges we put into the description of the tavily server (ie github starts and downloads):

<img width="926" height="275" alt="Image" src="https://github.com/user-attachments/assets/dbbaa34a-1d04-48b4-8898-25e3fe924477" />

He would like those to be exposed on the cards when you are browsing mcp servers, like here:

<img width="1548" height="525" alt="Image" src="https://github.com/user-attachments/assets/e9781668-947b-4861-b7c3-761ad28f24c6" />

So, we want the catalog object to have a "badges" or "badgeURLs" field that is a list of a badge URLs you can put in and we will render. Example URLs for tavily are:
`https://img.shields.io/github/stars/tavily-ai/tavily-mcp?style=social`
`https://img.shields.io/npm/dt/tavily-mcp`

We want the author of the mcp catalog to be able to drop those in and they get rendered on the card. I think the hardest part will be figuring out where to render them since the bottom is taken up by categories.

PR #3844, opened 2025-08-08
fix: prompt close on clickoutside of mcp config, auth deconfigured fix & mcp badges

ivyjeong13:
Addresses #3782 
Addresses #3513 
Addresses #3819


apiclient/types/mcpserver.go
g-linville, line 1: The fields added in this file would also need to be added to the storage types in order for them to persist. (We basically have a translation layer that sits between the types we store and the types here in `apiclient/types`.) This can be done by one of the backend engineers in a follow-up if that is easier.

ivyjeong13: Thanks Grant! Actually gonna drop the badge stuff here for now cause looks like we want to circle back to it after release!

PR #3910, opened 2025-08-12
fix: chat related bugfixes

ivyjeong13:


ui/user/src/lib/components/editor/MarkdownFile.svelte
```diff
@@ -0,0 +1,62 @@
...
+		onInvoke?: (invoke: InvokeInput) => void | Promise<void>;
+		items: EditorItem[];
+		mode: 'wysiwyg' | 'raw';
+	}
+
+	let { file: refFile, onFileChanged, mode = 'wysiwyg', onInvoke, items }: Props = $props();
+	// updating of the actual file.file.contents is handled during invoke of chat
+	// so have to keep a local state of the contents to share between the two variations
+	let contents = $state(refFile?.file?.contents ?? '');
+	let filename = $state(refFile?.name ?? '');
```
ryu-man: - I think it would be better to use a `$derived` rune instead of `$state + $effect`, $derived can hold temporary changes and recalculate when its deps change

ivyjeong13: Can you provide an example with temporary changes? I believe derived is meant to be used like a calculation, where it changes based on its dependencies. I can see what you mean for filename, cause that shouldn't change but $state(contents) needs to be writable and deviate from from refFile because refFile.file.contents is not going to change when onFileChanged is fired due to interaction between file editor changes and chat. Darren explained this better but basically file.file.contents shouldn't change until chat is invoked. 

edit: Ahh I see what you mean, the optimistic UI support for derived. Hmm, lemme think about that for a min!

ryu-man: Sure, here is an example of what I mean: [Writable Derived Rune](https://svelte.dev/playground/bf04fd8c59ee4068b5cd6937f44c124b?version=5.38.1).
If you click on the 'Override' button, the content will change, but in the next second, when a new date comes from props, the content will be recalculated.

What I mean is we can use $derived rune to achieve the same result as $state and $effect. because updating state inside an $effect rune is not recommended

ivyjeong13: :+1: Thanks! Made the change!

Issue #3731, opened 2025-08-03
Put deployment status on MCP server objects


thedadams:
We would like to add deployment/pod status to the MCP server objects so that the API returns this status. This involves two things:

1. The nanobot binary should fail when the underlying STDIO MCP server fails. Right now, the nanobot process will continue running when the STDIO server exits.
2. We need a controller handler to copy the deployment/pod information we need onto the status of the MCP server object.

thedadams:
The backend work for this has been merged. The changes add basic deployment status to the MCP server objects when they are deployed to Kubernetes. The intention is that this information indicates that there may be an issue and the user should "GetServerDetails" for more information.

Moving to Triage to start the conversation of how this will be reflected in the UI.

cjellick:
#4106 is closely tied to this.

cjellick:
Just to reiterate, the backend for this is more or less, in but im assigning to grant so that this and #4106 can be addressed holistically together.

cjellick:
i think the status we have at this point is sufficient

PR #3921, opened 2025-08-13
enhance: add deployment status information to MCP servers

thedadams:
Issue: https://github.com/obot-platform/obot/issues/3731


pkg/controller/controller.go
```diff
@@ -69,5 +82,45 @@ func (c *Controller) Start(ctx context.Context) error {
 	if err := c.router.Start(ctx); err != nil {
 		return fmt.Errorf("failed to start router: %w", err)
 	}
+
+	// Start local Kubernetes router if it exists
+	if c.localK8sRouter != nil {
+		if err := c.localK8sRouter.Start(ctx); err != nil {
+			fmt.Printf("Warning: Failed to start local Kubernetes router: %v\n", err)
```
StrongMonkey: Use logger here?

StrongMonkey (approved):

pkg/controller/controller.go
thedadams: I changed it so we return the error instead.

Issue #3765, opened 2025-08-05
Usage -Deleted user is shown as unknown in user filter drop down


sangee2004:
**Describe the bug**
Usage -Deleted user is shown as unknown in user filter drop down 

Pre Req -Have some usage generated for a user in last 7 days and delete the user.
**To Reproduce**
**Steps to reproduce the behavior:**
1. Go to `Usage`
2. Click on `Filters` and  access the user drop down presented for  `By User` filter
Notice that along with other user entries, the deleted user entries shows up as `Unknown`

<img width="405" height="145" alt="Image" src="https://github.com/user-attachments/assets/a40a08db-45f7-44d7-ad8d-2fd28dfe3b57" />

**Expected behavior**
Is there a way for us to show the deleted user's name in this case ? If not can we have some other text to indicate that this is a deleted user ?

This should be handled similar to how we will be handling this issue for Audit logs - https://github.com/obot-platform/obot/issues/3711

cjellick:
Yeah, we will want to soft delete users so that we can always report the user in the audit logs.

cjellick:
@g-linville an we soft delete uesrs so that we can always look them up for the audit logs?

g-linville:
certainly possible

g-linville:
I just merged the fix for this. Note that, for users that were deleted prior to this code change, there is no way for us to get that information back, so they will continue to show up as `Unknown`. For users that are deleted after this code change, in the audit logs, they should appear as `<Display Name> (Deleted)`.

sangee2004:
@g-linville I see the `<Display Name> (Deleted)` being present in `Audit Log` filters  for delete users as expected.

But in case of `Usage` filters , i dont see any entry relating to deleted user being presented in the `By Users` Filter drop down.

Also in the `Most Active Users` graph, the id of the deleted user is being shown (instead of the `<Display Name> (Deleted)`).

<img width="554" height="436" alt="Image" src="https://github.com/user-attachments/assets/cc41db50-f43d-40bb-92a2-2e58b3ee6994" />


Is there anything specific that needs to be done for `Usage` in this case ? we use `/mcp-stats` api call in this case

g-linville:
I'll work on getting things to display correctly for the Usage graphs as well.

g-linville:
Just merged the fix.

sangee2004:
This issue with presenting deleted users in `Usage` is addressed when testing with latest build.

[...]

PR #3927, opened 2025-08-13
fix: soft delete users so that we can show the right information in audit logs

g-linville:
for https://github.com/obot-platform/obot/issues/3765


pkg/gateway/client/apiactivity.go
```diff
@@ -41,7 +41,7 @@ func (c *Client) ActiveUsersByDate(ctx context.Context, start, end time.Time) ([
 			Pluck("user_id", &ids).Error; err != nil {
 			return err
 		}
-		return tx.Where("id IN (?) AND NOT internal", ids).Find(&users).Error
+		return tx.Where("id IN (?) AND NOT internal AND deleted_at IS NULL", ids).Find(&users).Error
```
g-linville, line 44: Do we want to include deleted users here? I am excluding them for now, but we can change it back in order to include them.

pkg/gateway/client/user.go
```diff
@@ -98,16 +120,40 @@ func (c *Client) DeleteUser(ctx context.Context, userID string) (*types.User, er
...
+		existingUser.OriginalEmail = existingUser.Email
+		existingUser.OriginalUsername = existingUser.Username
+
+		// Clear sensitive data and make fields unique to allow re-signup
+		// Append timestamp to prevent conflicts with new users using same email
+		deletedSuffix := fmt.Sprintf("_deleted_%d", now.Unix())
+		existingUser.Email = existingUser.Email + deletedSuffix
+		existingUser.Username = existingUser.Username + deletedSuffix
+		existingUser.HashedEmail = hash.String(existingUser.Email)
+		existingUser.HashedUsername = hash.String(existingUser.Username)
```
g-linville, line 139: Does this seem safe to do? We do have a constraint that all usernames in the database should be unique. I could try to find a way to drop this constraint, rather than adding this `deletedSuffix`, if we think that this is too dangerous or could somehow cause collisions.

pkg/gateway/client/user.go
```diff
@@ -98,16 +120,40 @@ func (c *Client) DeleteUser(ctx context.Context, userID string) (*types.User, er
 			}
 		}
 
-		if err := tx.Where("user_id = ?", existingUser.ID).Delete(new(types.Identity)).Error; err != nil {
```
g-linville, line 101: Deleting the identities here caused a bug, where we were unable to clean up the user's sessions in the UserDelete controller handler, meaning that the user's old `obot_access_token` cookie was still valid, and could be used to authenticate again as a new user, with the same username and email. I fixed this by waiting to delete identities until after we have deleted the user's sessions, and deleting them in the UserDelete controller handler, rather than here.

StrongMonkey (approved):

pkg/gateway/client/apiactivity.go
thedadams: I think this is correct.

apiclient/types/user.go
```diff
@@ -28,6 +28,9 @@ type User struct {
 	DailyPromptTokensLimit     int    `json:"dailyPromptTokensLimit,omitempty"`
 	DailyCompletionTokensLimit int    `json:"dailyCompletionTokensLimit,omitempty"`
 	DisplayName                string `json:"displayName,omitempty"`
+	DeletedAt                  Time   `json:"deletedAt,omitempty"`
```
thedadams: Should this be a pointer so that the `omitempty` works as expected? I think that without the pointer, the response will have a `"deletedAt": null` field for users that have not been deleted.
```suggestion
	DeletedAt                 *Time   `json:"deletedAt,omitempty"`
```

pkg/gateway/client/user.go
```diff
@@ -98,16 +120,40 @@ func (c *Client) DeleteUser(ctx context.Context, userID string) (*types.User, er
...
 	}); err != nil {
 		return nil, err
 	}
 
-	return existingUser, c.decryptUser(ctx, existingUser)
+	// Return the user with original values for the response (before encryption)
+	responseUser := *existingUser
+	responseUser.Email = responseUser.OriginalEmail
+	responseUser.Username = responseUser.OriginalUsername
+	return &responseUser, nil
```
thedadams: The intention is that the encrypted values for a user never leave this method. Since the `existingUser` is being encrypted on line 138 above, we should decrypt it here before returning. There have been a few bugs in this area caused by yours truly.

Instead of decrypting here, you could copy `existingUser` and copy over the values like you are doing here before encrypting `existingUser` above. Then here, just `return &responseUser, nil`.

thedadams (changes requested):

apiclient/types/user.go
g-linville: Fixed

pkg/gateway/client/user.go
g-linville: Thanks, fixed

PR #3946, opened 2025-08-15
feat: enhance MCP server management

thedadams:
Moves MCP server management logic into a backend interface. This allows for different implementations, such as Kubernetes, docker, or local execution.

Adds a new `ErrNotSupportedByBackend` error type to handle cases where a feature is not supported by the current backend.

Updates MCP handlers to use the backend interface for server details, logs, and restarts. Also handles remote servers.


thedadams:
Removing requests for review until I figure out why CI is failing.

thedadams:
OK. Ready for review.

A note about merging: I do want to be a little careful with merging this because I don't know if I'll have to do a terraform apply with the changes to the helm chart.

pkg/mcp/docker.go
```diff
@@ -0,0 +1,565 @@
...
+
+		// Set nanobot environment variables
+		env = []string{"NANOBOT_RUN_HEALTHZ_PATH=/healthz", "OBOT_KUBERNETES_MODE=true"}
+
+	case otypes.RuntimeContainerized:
+		// Use specified container image or base image
+		image = server.ContainerImage
+		if image == "" {
+			image = d.baseImage
+		}
```
g-linville: nit: I think we should return an error here, rather than falling back to the base image. Though, we should never be in this situation, as all the validation logic on MCPServerCatalogEntries and MCPServers should prevent this from happening.

thedadams: Good call. Fixed.

Issue #3939, opened 2025-08-14
Do not provide helper text for `Auth Provider` in auth enabled setups.


sangee2004:
**Describe the bug**
Do not provide helper text for `Auth Provider` in auth enabled setups.

**To Reproduce**
1. Have auth enabled obot setup
Notice that user is presented with following helper text for `Auth Provider` 
<img width="334" height="136" alt="Image" src="https://github.com/user-attachments/assets/90713f4d-0d69-465f-9a64-17b2078690fc" />

**Expected behavior**
Do not provide helper text for `Auth Provider` in auth enabled setups.

sangee2004:
This issue is addressed when testing with latest build.
There is no helper text presented for `Auth Providers` in auth enabled setups.

PR #3948, opened 2025-08-15
fix: remove tooltip from auth provider item

ryu-man:
Addresses #3939


ui/user/src/lib/components/Layout.svelte
```diff
@@ -181,9 +176,6 @@
 											{link.label}
 										</a>
 									{/if}
-									{#if tooltips[link.href as keyof typeof tooltips]}
```
ivyjeong13, line 184: Looking at the ticket, we don't want to completely remove the tooltip -- just only show it for setups with authentication disabled. Prolly can just add to the #if:

```
{#if tooltips[link.href as keyof typeof tooltips] && version.current.authEnabled} // version from $lib/stores
```

ryu-man: Fixed

PR #3951, opened 2025-08-15
feat: launch temp instance to generate tool previews for catalog entry

g-linville:


pkg/api/handlers/mcpcatalogs.go
```diff
@@ -313,6 +320,56 @@ func (h *MCPCatalogHandler) AdminListServersForEntryInCatalog(req api.Context) e
...
+	}
+
+	// Launch temporary instance and get tools
+	toolPreviews, err := h.sessionManager.LaunchTemporaryInstance(req.Context(), entry.Spec.Manifest, configRequest.URL, configRequest.Config)
+	if err != nil {
+		return fmt.Errorf("failed to launch temporary instance: %w", err)
+	}
+
+	// Update the catalog entry with the tool preview
+	entry.Spec.Manifest.ToolPreview = toolPreviews
```
njhale, line 382: Will this get wiped out when the catalog is refreshed?

pkg/mcp/loader.go
```diff
@@ -508,6 +509,89 @@ func (sm *SessionManager) updatedMCPPodName(ctx context.Context, url, id string,
...
+	// Create MCP client and list tools
+	client, err := sm.ClientForServer(ctx, "system", serverManifest.Name, tempName, serverConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create MCP client: %w", err)
+	}
+
+	// Ensure cleanup happens regardless of success or failure
+	defer func() {
+		if cleanupErr := sm.ShutdownServer(ctx, serverConfig); cleanupErr != nil {
+			fmt.Printf("failed to cleanup temporary instance %s: %v\n", tempName, cleanupErr) // TODO: use log package
```
njhale: reminder to switch over to the log package

pkg/mcp/loader.go
```diff
@@ -508,6 +509,89 @@ func (sm *SessionManager) updatedMCPPodName(ctx context.Context, url, id string,
 	}
 }
 
+// LaunchTemporaryInstance creates a temporary MCP server from a catalog entry, lists its tools,
+// then shuts it down and returns the tool preview data.
+func (sm *SessionManager) LaunchTemporaryInstance(ctx context.Context, catalogEntryManifest otypes.MCPServerCatalogEntryManifest, userURL string, configEnv map[string]string) ([]otypes.MCPServerTool, error) {
```
thedadams: This logic is very similar to the logic in `toolsForServer` in `pkg/api/handlers/mcp.go`, especially the part where we parse the input parameters. Can we find a way of not duplicating this admittedly complex logic?

Also, the name of this method is a bit misleading.

pkg/api/handlers/mcpcatalogs.go
g-linville: No, because this is only supported for manually created catalog entries, not ones from GitHub. If we did do this on ones from GitHub, yes it would just end up getting wiped.

pkg/mcp/loader.go
g-linville: Thanks, forgot to do this

pkg/mcp/loader.go
g-linville: Yeah I'll see if I can dry it up

pkg/mcp/loader.go

njhale (approved):

pkg/mcp/loader.go

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -20,12 +21,18 @@ import (
 type MCPCatalogHandler struct {
 	defaultCatalogPath string
 	serverURL          string
+	sessionManager     SessionManager
 }
 
-func NewMCPCatalogHandler(defaultCatalogPath string, serverURL string) *MCPCatalogHandler {
+type SessionManager interface {
+	GenerateToolPreviews(ctx context.Context, catalogEntryManifest types.MCPServerCatalogEntryManifest, userURL string, configEnv map[string]string) ([]types.MCPServerTool, error)
+}
```
thedadams: Do we need this interface? It looks superfluous to me.

g-linville:
@StrongMonkey It does work with OAuth; I just pushed the changes to make that happen. Thanks for asking about it, because I hadn't considered that.

This isn't something we want to do automatically, since the tools that a server provides can vary depending on the configuration (Zapier and GitMCP are good examples of this), so we don't want to do it without the admin explicitly asking for it.

StrongMonkey (approved):

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -313,6 +323,99 @@ func (h *MCPCatalogHandler) AdminListServersForEntryInCatalog(req api.Context) e
...
+// GenerateToolPreviews launches a temporary instance of an MCP server from a catalog entry
+// to generate tool preview data, then cleans up the instance.
+func (h *MCPCatalogHandler) GenerateToolPreviews(req api.Context) error {
+	catalogName := req.PathValue("catalog_id")
+	entryName := req.PathValue("entry_id")
+
+	// Check if user is admin
+	if !req.UserIsAdmin() {
+		return types.NewErrForbidden("only admins can launch temporary instances")
+	}
```
thedadams: This can be removed because the path for this handler is already not available to non-admin users.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -313,6 +323,99 @@ func (h *MCPCatalogHandler) AdminListServersForEntryInCatalog(req api.Context) e
...
+		return types.NewErrBadRequest("entry is not editable")
+	}
+
+	// Read configuration from request body
+	var configRequest struct {
+		Config map[string]string `json:"config"`
+		URL    string            `json:"url"`
+	}
+	if err := req.Read(&configRequest); err != nil {
+		return fmt.Errorf("failed to read configuration: %w", err)
```
thedadams: Does it make sense for this to be a `http.StatusBadRequest` error?

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -313,6 +323,99 @@ func (h *MCPCatalogHandler) AdminListServersForEntryInCatalog(req api.Context) e
...
+	}
+
+	// Convert catalog entry to server manifest
+	serverManifest, err := types.MapCatalogEntryToServer(entry.Spec.Manifest, configRequest.URL)
+	if err != nil {
+		return fmt.Errorf("failed to convert catalog entry to server config: %w", err)
+	}
+
+	// Create temporary MCPServer object to use existing conversion logic
+	tempName := "temp-preview-" + hash.Digest(serverManifest)[:8]
```
thedadams: nit: I would make this a little larger to avoid conflicts.
```suggestion
	tempName := "temp-preview-" + hash.Digest(serverManifest)[:32]
```

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -313,6 +323,99 @@ func (h *MCPCatalogHandler) AdminListServersForEntryInCatalog(req api.Context) e
...
+	}
+
+	if serverConfig.Runtime == types.RuntimeRemote {
+		oauthURL, err := h.oauthChecker.CheckForMCPAuth(req.Context(), tempMCPServer, serverConfig, "system", tempName, "")
+		if err != nil {
+			return fmt.Errorf("failed to check for MCP auth: %w", err)
+		}
+
+		if oauthURL != "" {
+			return req.Write(map[string]string{"oauthURL": oauthURL})
```
thedadams: It seems strange to me to have multiple types returned from the same API. The way this works in other cases is that there is a separate "OAuth URL" handler.

pkg/api/handlers/mcpcatalogs.go
g-linville: Yes

g-linville: There are a couple others I will change as well

pkg/api/handlers/mcpcatalogs.go
g-linville: Imo it would be a bit overkill to introduce a second route just for this. The UI can just check what it receives back and take action accordingly. I'm willing to change it if we really do want to, but my preference would be to keep it as it is.

pkg/api/handlers/mcpcatalogs.go

pkg/api/handlers/mcpcatalogs.go

pkg/api/handlers/mcpcatalogs.go

pkg/api/handlers/mcpcatalogs.go
thedadams: That was my nice way of asking you to change it. I don't think it's overkill to have good API design.

g-linville: Fair enough, lol. I couldn't tell if it was asking or just mentioning as an idea :)

g-linville: This is fixed now

thedadams: Yeah, bad communication by me. Thanks for addressing it.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -313,6 +323,149 @@ func (h *MCPCatalogHandler) AdminListServersForEntryInCatalog(req api.Context) e
...
+		return types.NewErrBadRequest("entry does not belong to catalog")
+	}
+
+	if !entry.Spec.Editable {
+		return types.NewErrBadRequest("entry is not editable")
+	}
+
+	if entry.Spec.Manifest.Runtime != types.RuntimeRemote {
+		return types.NewErrBadRequest("only catalog entries with a remote runtime can have their OAuth URL generated")
+	}
```
thedadams: nit: in other instances, we return an empty oauth URL in this case.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -313,6 +323,149 @@ func (h *MCPCatalogHandler) AdminListServersForEntryInCatalog(req api.Context) e
...
+	oauthURL, err := h.oauthChecker.CheckForMCPAuth(req.Context(), server, serverConfig, "system", server.Name, "")
+	if err != nil {
+		return types.NewErrBadRequest("failed to check for MCP auth: %v", err)
+	}
+
+	if oauthURL != "" {
+		return req.Write(map[string]string{"oauthURL": oauthURL})
+	}
+
+	return nil
```
thedadams: nit: in other instances, we always return an `oauthURL`, but it can be blank, indicating that OAut is not required.

pkg/api/handlers/mcpcatalogs.go

pkg/api/handlers/mcpcatalogs.go

Issue #3914, opened 2025-08-12
Audit Logs/Usage/Chat Threads - Make `Filter` page options consistent.


sangee2004:
**Describe the bug**
Audit Logs - Make `Clear All` button in `Filter` page behave the same way as in `Usage` and `Chat Threads`

1. Using `Clear All` button in `Filter` page in `Audit Logs` , results in values selected in the filters fields to get cleared. User needs to click on `Apply Filter` to actually get the changes applied.  Using  `Clear All` button in `Filter` page in `Usage` and `Chat Threads` will actually get the changes applied resulting in the `Filters` page getting auto closed and the new query being made

**Expected Behavior:**
Make  `Clear All` button in `Filter` page in `Audit Logs` behave the same way as `Usage` and `Chat Threads`

2. Have consistent name for clearing filters button - `Clear All` or `Clear All Filters`.
3. Have consistent order in which Apply filter and clear filter options appear in the `Filters` page.

<img width="1049" height="809" alt="Image" src="https://github.com/user-attachments/assets/85155a40-4721-44de-8a47-d9ad2ff99445" />
<img width="1063" height="809" alt="Image" src="https://github.com/user-attachments/assets/38b92a4e-ae85-4f92-bb9b-efcf226c5f09" />
<img width="1062" height="796" alt="Image" src="https://github.com/user-attachments/assets/e5d8d093-dcdc-47ad-b796-e8742911cd85" />

sangee2004:
When testing with latest build,  `Audit Logs` and `Chat Threads` are consistent with how `Clear All` option works.

`Usage` page still names this filed as `Clear All Filters` (and this option is present below `Apply Filters`)  and it will actually get the changes applied resulting in the `Filters` page getting auto closed and the new query being made. 

@ryu-man Reopening this issue to fix the inconsistencies with `Usage` .

ryu-man:
@cjellick I made some changes to use the same filters drawer across the app. but found out that chat threads are not server-side filtered. Should I implement a local filtering, or do we strictly need this to be server-side?

sangee2004:
Changes requested in https://github.com/obot-platform/obot/issues/3914#issuecomment-3192361860 are addressed when testing with latest build.
<img width="608" height="841" alt="Image" src="https://github.com/user-attachments/assets/e8bd85c4-2541-46cd-a1ac-33cdaf102660" />

There is an issue with selected filters not working which is tracked in https://github.com/obot-platform/obot/issues/3979

PR #3956, opened 2025-08-15
refactor usage filter to match audit logs filters

ryu-man:
Addresses #3914

### DIFF
--- a/ui/user/src/lib/components/admin/filters-drawer/FilterField.svelte
+++ b/ui/user/src/lib/components/admin/filters-drawer/FilterField.svelte
@@ -23,14 +23,21 @@
 
 	interface Props {
 		filter: FilterInput;
+		optionsSort?: (a: FilterOption, b: FilterOption) => number;
 		onSelect: SelectProps<{ id: string; label: string }>['onSelect'];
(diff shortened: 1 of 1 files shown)


ui/user/src/lib/components/admin/usage/UsageGraphs.svelte
```diff
@@ -348,19 +395,59 @@
...
+			user?.username ??
+			user?.email ??
+			'Unknown User';
+
+		if (user?.deletedAt) {
+			display += ' (Deleted)';
+		}
+
+		return display;
+	}
```
njhale: So there was an edge case where two users have the same name, that the original usage filter options would disambiguate by appending a `(user_email)` to the option. Is it possible to do that here as well?

ryu-man: I think that is a good idea. I can do that, but there will be a case where the secondary value might be undefined. And also I think using a short value is better. not sure which one it would be.

njhale: currently in main, when the secondary value is not available, nothing is appended to the display name.

PR #3966, opened 2025-08-18
enhance: allow setting ANTHROPIC_API_KEY for model setup

thedadams:
This adds the same functionality for an Anthropic API key as we support for an OpenAI API key.

Note that Anthropic doesn't have an image generation nor text-embedding model, so neither of these default model aliases will be set when using Anthropic API keys.

### DIFF
--- a/docs/docs/50-configuration/01-server-configuration.md
+++ b/docs/docs/50-configuration/01-server-configuration.md
@@ -5,6 +5,7 @@ The Obot server is configured via environment variables. The following configura
 | Environment Variable | Description |
 |---------------------|-------------|
 | `OPENAI_API_KEY` | The foundation of Obot is a large language model that supports function-calling. The default is OpenAI and specifying an OpenAI key here will ensure none of the users need to worry about specifying their own API key. |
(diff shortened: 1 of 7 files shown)


docs/docs/50-configuration/01-server-configuration.md
```diff
@@ -5,6 +5,7 @@ The Obot server is configured via environment variables. The following configura
 | Environment Variable | Description |
 |---------------------|-------------|
 | `OPENAI_API_KEY` | The foundation of Obot is a large language model that supports function-calling. The default is OpenAI and specifying an OpenAI key here will ensure none of the users need to worry about specifying their own API key. |
+| `ANTHROPIC_API_KEY` | You can also or instead set an Anthropic API key. |
```
njhale: femto-nit:


```suggestion
| `ANTHROPIC_API_KEY` | You can also provide an Anthropic API key in place of or in addition to an OpenAI API key. |
```

thedadams: Fixed.

Issue #3729, opened 2025-08-03
Provide ability to show `upgrade required` in chat view (similar to `My connectors` page).


sangee2004:
**Describe the bug**
Provide ability to show`upgrade required` in chat view (similar to `My connectors` page).

**To Reproduce**
1. As regular user, create a project and add any single user MCP server.
2. As admin user, update this MCP server and add a required parameter and update the server instance from step1.
3. Notice that `upgrade required` message is presented in the `My connectors` page as expected.
<img width="834" height="691" alt="Image" src="https://github.com/user-attachments/assets/35e29b83-7a57-4a3d-93bf-e189ab80a72f" />

This `upgrade required` message is not presented when in the chat view for the project.
<img width="794" height="558" alt="Image" src="https://github.com/user-attachments/assets/1782d5e0-801b-4880-b2e7-ea32951ab650" />

**Expected behavior**
We should provide the `upgrade required` message in chat view  for the  Connector entry in `Connectors` for the project.

ivyjeong13:
This may require some backend work, I don't see `configured`, `needsUrl` or `needsUpdate` passed in the /project/{id}/mcpservers endpoint

cjellick:
cc @g-linville for backend

g-linville:
Backend part is complete. Moving to Triage.

sangee2004:
This issue is addressed when testing with latest build.

There is an `Update Required` indication provided the user in project view. User is able to edit configurations by selecting this connector entry and using `Edit Configuration` option as expected.

<img width="1166" height="709" alt="Image" src="https://github.com/user-attachments/assets/b285d84f-58c6-4bd5-8b96-e42217cce7a8" />

PR #3968, opened 2025-08-18
enhance: return more information about project servers

g-linville:
backend portion of https://github.com/obot-platform/obot/issues/3729

### DIFF
--- a/pkg/controller/handlers/mcpserver/mcpserver_test.go
+++ b/pkg/controller/handlers/mcpserver/mcpserver_test.go
@@ -318,7 +318,7 @@ func TestConfigurationHasDrifted(t *testing.T) {
 				},
 			},
 			expectedDrift: true,
-			expectedError: true,
+			expectedError: false,
 		},
 		{
 			name: "no drift - different env order (order doesn't matter)",
@@ -391,7 +391,7 @@ func TestConfigurationHasDrifted(t *testing.T) {
 				},
 			},
 			expectedDrift: true,
-			expectedError: true,
+			expectedError: false,
 		},
 		{
 			name: "drift - missing runtime config",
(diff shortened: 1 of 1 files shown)


pkg/api/handlers/projectmcp.go
```diff
@@ -87,13 +97,22 @@ func (p *ProjectMCPHandler) ListServer(req api.Context) error {
 		return nil
 	}
 
-	credCtxs := make([]string, 0, len(servers.Items))
```
g-linville, line 90: All the credential stuff that was here before was unneeded and unused (the resulting `credMap` was never referenced anywhere). The new credentials stuff I have added is getting the credentials for the MCPServer that the ProjectMCPServer points to. We do not store any credentials scoped to ProjectMCPServers themselves, since they are basically just pointers.

pkg/api/handlers/projectmcp.go
```diff
@@ -87,13 +97,22 @@ func (p *ProjectMCPHandler) ListServer(req api.Context) error {
...
+		mcpServer, err := getMCPServerForProjectServer(req.Context(), req.Storage, server)
+		if err != nil {
+			return err
+		}
+
+		if mcpServer != nil {
+			mcpServers[server.Name] = *mcpServer
+
+			if mcpServer.Spec.SharedWithinMCPCatalogName == "" {
+				credCtxs = append(credCtxs, fmt.Sprintf("%s-%s", mcpServer.Spec.UserID, mcpServer.Name))
```
g-linville, line 119: We are using `mcpServer.Spec.UserID` here because the owner of this server is the one who configured the credentials. If I am not mistaken, I believe that this will always be the same as the ID of the user on the actual `req` object, unless perhaps that user is the admin.

pkg/controller/handlers/mcpserver/mcpserver.go
```diff
@@ -54,26 +54,16 @@ func configurationHasDrifted(needsURL bool, serverManifest types.MCPServerManife
...
 		drifted = npxConfigHasDrifted(serverManifest.NPXConfig, entryManifest.NPXConfig)
-
 	case types.RuntimeContainerized:
 		drifted = containerizedConfigHasDrifted(serverManifest.ContainerizedConfig, entryManifest.ContainerizedConfig)
-
 	case types.RuntimeRemote:
-		drifted, err = remoteConfigHasDrifted(needsURL, serverManifest.RemoteConfig, entryManifest.RemoteConfig)
-		if err != nil {
-			return drifted, err
-		}
```
g-linville, line 75: This was a bug before, where this error would get propagated all the way up this controller handler, resulting in the server never being marked as drifted. There is no need to do this, so I stopped propagating this error in this PR.

pkg/api/handlers/projectmcp.go
thedadams: That's what we use to create the MCP server credential, so I think this is the right thing to do.

pkg/api/handlers/projectmcp.go
```diff
@@ -171,11 +193,21 @@ func (p *ProjectMCPHandler) CreateServer(req api.Context) error {
...
 	}
 
+	var cred map[string]string
+	if mcpServer.Spec.SharedWithinMCPCatalogName == "" {
+		gptscriptCred, err := req.GPTClient.RevealCredential(req.Context(), []string{fmt.Sprintf("%s-%s", mcpServer.Spec.UserID, mcpServer.Name)}, mcpServer.Name)
+		if err != nil && !errors.As(err, &gptscript.ErrNotFound{}) {
+			return fmt.Errorf("failed to find credential: %w", err)
+		}
+
+		cred = gptscriptCred.Env
```
StrongMonkey, line 203: are you handling err not found? if it is not found, would gptscriptCred be nil and causing a panic?

pkg/api/handlers/projectmcp.go
```diff
@@ -189,7 +221,17 @@ func (p *ProjectMCPHandler) GetServer(req api.Context) error {
 		return err
 	}
 
-	return req.Write(convertProjectMCPServer(&projectServer, mcpServer))
+	var cred map[string]string
+	if mcpServer.Spec.SharedWithinMCPCatalogName == "" {
+		gptscriptCred, err := req.GPTClient.RevealCredential(req.Context(), []string{fmt.Sprintf("%s-%s", mcpServer.Spec.UserID, mcpServer.Name)}, mcpServer.Name)
```
StrongMonkey, line 226: same above

pkg/api/handlers/projectmcp.go
g-linville: We don't need to handle the err not found. `gptscriptCred` is not a pointer, so it will just have the zero values for everything if no cred was found, and `gptscriptCred.Env` will be a nil map that we assign to `cred`, so `cred` just remains nil.

pkg/api/handlers/projectmcp.go
g-linville: Same here

Issue #3941, opened 2025-08-14
When trying to connect to a misconfigured server , user is presented with partial error message which is actually a timeout error.


sangee2004:
**Describe the bug**
When trying to connect to a misconfigured server , user is presented with partial error message which is actually a timeout error.

**To Reproduce**
1. Create `Tavily` MCP server without proving the `TAVILY_API_KEY` which is a required parameter.
2. As regular user, try to connect to this MCP server

User is presented with following error message:

<img width="1012" height="792" alt="Image" src="https://github.com/user-attachments/assets/144b6080-0c50-4f82-86bc-0cbd86a46719" />

`/launch` call fails with `524: A timeout occurred` error.

**Expected behavior**
User should be presented with `503` errors like we used to see few day back - https://github.com/obot-platform/obot/issues/3767#issuecomment-3165342579

sangee2004:
When I test the same scenario in my local installs which does not have auth enabled , I am presented with the following error message

<img width="689" height="460" alt="Image" src="https://github.com/user-attachments/assets/eab88b62-d0cf-4aa5-8141-74a169e42045" />

thedadams:
The difference you are seeing in main versus your local setup is only related to the fact that main is behind Cloudflare.

sangee2004:
This issue is addressed when testing latest build.

user is presented with the following error message as expected.

<img width="518" height="334" alt="Image" src="https://github.com/user-attachments/assets/ca08ec37-e236-45c5-b3d9-36668b434aa4" />

PR #3973, opened 2025-08-18
fix: better detect MCP server misconfiguration

thedadams:
This change also contains a commit to replace `interface{}` with `any`.

Issue: https://github.com/obot-platform/obot/issues/3941

### DIFF
--- a/pkg/mcp/kubernetes.go
+++ b/pkg/mcp/kubernetes.go
@@ -613,10 +613,10 @@ func (k *kubernetesBackend) restartServer(ctx context.Context, id string) error 
 		return fmt.Errorf("failed to get deployment %s: %w", id, err)
 	}
 
-	patch := map[string]interface{}{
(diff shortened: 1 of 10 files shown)


pkg/mcp/kubernetes.go
```diff
@@ -568,61 +568,55 @@ func (k *kubernetesBackend) k8sObjectsForContainerized(id string, server ServerC
...
 	}
 
-	// Not get the pod name that is currently running, waiting for there to only be one pod.
+	// Now get the pod name that is currently running, waiting for there to only be one running pod.
 	var (
 		pods            corev1.PodList
 		runningPodCount int
 		podName         string
 	)
-	for {
```
thedadams, line 587: We don't need to loop here. Ensuring that the deployment has been updated and that the associated pods are ready is enough.

PR #3977, opened 2025-08-19
fix: stop redirect loop on fresh installs with auth enabled

thedadams:
When authentication is enabled on a fresh install, a "too many redirects" error is presented to the user after they enter the bootstrap token. Checking whether the current user is admin seems to stop this.

I found that this redirect and [this one](https://github.com/obot-platform/obot/blob/main/ui/user/src/routes/admin/+page.ts#L20) contend with each other and redirect until the error occurs.

### DIFF
--- a/ui/user/src/routes/admin/auth-providers/+page.ts
+++ b/ui/user/src/routes/admin/auth-providers/+page.ts
@@ -1,12 +1,13 @@
 import { handleRouteError } from '$lib/errors';
-import { AdminService } from '$lib/services';
+import { AdminService, ChatService } from '$lib/services';
 import type { AuthProvider } from '$lib/services/admin/types';
-import { profile, version } from '$lib/stores';
+import { profile } from '$lib/stores';
 import type { PageLoad } from './$types';
 import { redirect } from '@sveltejs/kit';
 
 export const load: PageLoad = async ({ fetch }) => {
-	if (!version.current.authEnabled) {
+	const version = await ChatService.getVersion({ fetch });
+	if (!version.authEnabled) {
 		throw redirect(302, '/admin');
 	}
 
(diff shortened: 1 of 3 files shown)


ui/user/src/routes/admin/auth-providers/+page.ts
```diff
@@ -6,7 +6,7 @@ import type { PageLoad } from './$types';
 import { redirect } from '@sveltejs/kit';
 
 export const load: PageLoad = async ({ fetch }) => {
-	if (!version.current.authEnabled) {
+	if (profile.current.isAdmin && !version.current.authEnabled) {
```
ivyjeong13: Ah okay, there's a case of the version store not being initialized. I'd change this to
```
// auth-providers/+page.ts
const version = await ChatService.getVersion({ fetch });
	if (!version.authEnabled) {
		throw redirect(302, '/admin');
	}
```

```
// operations.ts
export async function getVersion(opts?: { fetch?: Fetcher }): Promise<Version> {
	return (await doGet('/version', opts)) as Version;
}
```

And also in /admin/+page.ts, I'd update this case so if authEnabled isn't set, we don't do the redirect to /auth-providers. 
```
// admin/+page.ts
	let profile;
	let version;

	try {
		version = await ChatService.getVersion({ fetch });
		profile = await getProfile({ fetch });
	} catch (_err) {
		authProviders = await ChatService.listAuthProviders({ fetch });
	}

	if (profile?.role === Role.ADMIN) {
		throw redirect(
			307,
			profile.username === BOOTSTRAP_USER_ID && version?.authEnabled
				? '/admin/auth-providers'
				: '/admin/mcp-servers'
		);
	}
```

Sorry, totally forgot about this redirect case! Thanks for finding it!

thedadams: Thanks, Ivy! I've updated the PR with your suggested changes.

Issue #3982, opened 2025-08-19
Users - Sorting by lastActiveDay is inaccurate


sangee2004:
**Describe the bug**
Users - Default order in which users are presented is not sorted by lastActiveDay.

**To Reproduce**
1. Go to `Users`
2. Click on `Last Active` field.

Notice that users that are actively using the system today , show up in the top of the list.

But for users not active today , the list is not sorted correctly by `lastActiveDay`

Some snippets that show discrepancies 

<img width="998" height="286" alt="Image" src="https://github.com/user-attachments/assets/eb122a61-e430-4c59-87d0-95fad3d34529" />

Another instance:

<img width="998" height="286" alt="Image" src="https://github.com/user-attachments/assets/793523ed-408e-4ec9-8ca2-2be0ce3aa0c7" />

**Expected behavior**
Sorting by lastActiveDay should be accurate.

sangee2004:
This issue is addressed when testing with latest build.
Sorting by lastActiveDay works as expected.

Issue #3981, opened 2025-08-19
Users - Sorting by `Name` field is not working.


sangee2004:
**Describe the bug**
Users - Sorting by `Name` field is not working.

**To Reproduce**
1. Go to `Users`
2. Click on `Name` field to sort.

Notice that  nothing happens except for the sort icon flipping.

**Expected behavior**
Should be able to sort by `Name` field.

sangee2004:
Sorting `by Name` field works as expected now.

Issue #3967, opened 2025-08-18
Admin UI - Admin gets session expired message when deleting users.


sangee2004:
**Describe the bug**
Admin UI - Admin gets session expired message when deleting users.

**To Reproduce**
1. As admin, Go to `Users` page
2. Delete any existing user

Notice that user is presented with following session expired message and is forced to log in again.

<img width="564" height="350" alt="Image" src="https://github.com/user-attachments/assets/dfda6fef-871d-4171-a0e3-6957d14b0901" />

**Expected behavior**
Admin user should not get logged out in this case.

sangee2004:
This issue is fixed when testing with latest build.

Issue #3965, opened 2025-08-18
Admin UI - `Email` and `Last Active` is presented twice in `Users` page.


sangee2004:
**Describe the bug**
Admin UI - `Email` and `Last Active` is presented twice in `Users` page.

**To Reproduce**
1. As admin user, Go to `Users` page

Notice that  `Email` and `Last Active` is presented twice in `Users` page.

<img width="1017" height="84" alt="Image" src="https://github.com/user-attachments/assets/87c9aed1-b2c9-4f1a-8459-675c0fa9dd2e" />

sangee2004:
This issue is not seen anymore when testing with latest build.

PR #4002, opened 2025-08-21
fix: admin users table fixes and session expired fix

ivyjeong13:
* fixes admin `/users` page related bugs 

Addresses #3967 
Addresses #3981 
Addresses #3982
Addresses #3965

### DIFF
--- a/pkg/gateway/server/user.go
+++ b/pkg/gateway/server/user.go
@@ -164,15 +164,10 @@ func (s *Server) changeUserInternalStatus(apiContext api.Context, internal bool)
 
 func (s *Server) deleteUser(apiContext api.Context) (err error) {
 	userID := apiContext.PathValue("user_id")
-	if userID == "" {
+	isDeleteMe := userID == ""
+	if isDeleteMe {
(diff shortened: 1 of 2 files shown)


pkg/gateway/server/user.go
```diff
@@ -206,15 +201,18 @@ func (s *Server) deleteUser(apiContext api.Context) (err error) {
...
-	http.SetCookie(apiContext.ResponseWriter, &http.Cookie{
-		Name:     proxy.ObotAccessTokenCookie,
-		Value:    "",
-		Path:     "/",
-		MaxAge:   -1,
-		HttpOnly: true,
-		Secure:   strings.HasPrefix(s.uiURL, "https://"),
-	})
+	// Only clear the cookie if this is a "delete me" operation
+	if isDeleteMe {
```
g-linville, line 205: Sorry about this! It was my fault that we were deleting the token for everyone, and not just "delete me" requests. I didn't consider that at first when I added this code.

g-linville (approved):

Issue #3992, opened 2025-08-20
Provide dropdown of all existing categories when adding categories to MCP servers.


sangee2004:
**Describe the bug**
Provide dropdown of all existing categories when adding categories to MCP servers.

**To Reproduce**
1. Create MCP server and add categories.

Notice that there is no guidance to the user around being able to pick existing categories. 

**Expected behavior**
Provide dropdown of all existing categories when adding categories to MCP servers , in addition to allowing users to add new categories .

sangee2004:
This issue is addressed when testing with latest build.

PR #4005, opened 2025-08-22
feat: allow re-utilization of existing categories

ryu-man:
Addresses #3992 

- convert category handling to a multi-select input
- fetch categories from API
- If the user types a value and hits the enter key: 
  - it will select the highlighted option. 
  - will create a new category if no option is available.

### DIFF
--- a/ui/user/src/lib/components/admin/CatalogServerForm.svelte
+++ b/ui/user/src/lib/components/admin/CatalogServerForm.svelte
@@ -622,13 +622,12 @@
 				onUpdate={async (categories) => {
 					formData.categories = [
 						// Avoid duplicates
-						...new Set([
-							...formData.categories,
-							...categories
+						...new Set(
+							categories
 								.split(',')
 								.map((c) => c.trim())
 								.filter(Boolean)
-						])
+						)
 					];
 				}}
 			/>


ivyjeong13:
@ryu-man Can this PR be rebased, seeing a lot of other merges

ryu-man:
@ivyjeong13 PR is cleaned

ui/user/src/lib/components/Select.svelte
```diff
@@ -63,17 +69,17 @@
...
 	let availableOptions = $derived(
-		options.filter((option) => option.label.toLowerCase().includes(search.toLowerCase()))
+		options.filter((option) => option.label.toLowerCase().includes(query?.toLowerCase() ?? ''))
 	);
 
 	let selectedOptions = $derived(
 		selectedValues
+			.filter(Boolean)
 			.map((selectedValue) => options.find((option) => option.id === selectedValue))
 			.filter(Boolean) as T[]
```
ivyjeong13, line 84: Do you need both `.filter`? Thinking can just need it once

ryu-man: Yes, I encountered a scenario having undefined options; the `selectedValues` array could contain values that are not set in the options. That's why I added the last filter.

PR #4017, opened 2025-08-23
fix: refresh auth groups for requests with mcp auth tokens

njhale:
Addresses https://github.com/obot-platform/obot/issues/4022

### DIFF
--- a/pkg/gateway/client/user.go
+++ b/pkg/gateway/client/user.go
@@ -26,13 +26,14 @@ var (
 	}
 )
 
-func (c *Client) UserFromToken(ctx context.Context, token string) (*types.User, string, string, []string, error) {
+func (c *Client) UserFromToken(ctx context.Context, token string) (*types.User, string, string, string, []string, error) {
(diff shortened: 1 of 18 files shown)


pkg/gateway/client/user.go
```diff
@@ -26,13 +26,14 @@ var (
 	}
 )
 
-func (c *Client) UserFromToken(ctx context.Context, token string) (*types.User, string, string, []string, error) {
+func (c *Client) UserFromToken(ctx context.Context, token string) (*types.User, string, string, string, []string, error) {
```
g-linville, line 29: This might be the record for the most return values I have seen on one function, lol.

njhale: Yeah, I should probably refactor this a bit instead of just letting keep growing

njhale: Yeah, I should probably refactor this a bit instead of just letting keep growing

drpebcak (approved):

Issue #3319, opened 2025-07-03
Pre-populate and Persist MCP Server Tools and Capabilities via Temporary Instance Launch


cjellick:
**Problem**

The MCP server detail screen in the admin catalog currently does not display toosl and capability information as desired. Also, the user-facing version of the screen suffers from long loading times due to live data fetching.
---

**Proposed Solution**

We need to implement a system to pre-populate and persist MCP tools and capability data for each server (or more accurately each catalog entry).

### Requirements

1. **Static Population Strategy**

   * Introduce a section on the MCP server detail screen (maybe overlaid on top of the current tools section if it's unpopulated) with a button/link:
     **“Launch a temporary instance to populate capabilities and tools.”**
   * Secondary text should indicate:
     *“If not launched, tools will populate when the first user launches this server.”*

2. **On Click Behavior**

   * Clicking the button spins up a temporary MCP instance.
   * This instance gathers tool and capability data.
   * Collected data is stored in a persistent structure (e.g. API resource) associated with the MCP server.
   * Once complete, the instance is torn down automatically.

3. **Display Behavior**

   * When visiting the MCP detail screen, tool and capability data is read from this stored structure.
   * Descriptions are hidden by default to declutter the UI.

4. **Handling Upgrades**

   * Admins must be able to explicitly trigger a “Regenerate Tools & Capabilities” action to refresh data after MCP upgrades.
   * Do **not** auto-refresh tools every time a user connects

5. **Future Enhancement (Optional)**

   * Consider a TTL (e.g. 5 days). If expired, the next user connection can trigger a refresh.

This issue was AI generated mostly from this loom:
https://www.loom.com/share/40a56c38e77a4f71a252af52ac5978ff

njhale:
I spoke with @cjellick a bit last Friday and we came to the conclusion that we also want the ability to pre-populate the tools and capabilities section with static "hints" provided on the catalog entries themselves. This lets us show a tool/capability "preview" without having to start up a server. I'm going to implement this bit first because it gets us to a nice demoable MVP for select MCP servers more quickly.

[Edit] Here's the PR for that: https://github.com/obot-platform/obot/pull/3348

njhale:
Pausing work on this until I get my really-high's done.

cjellick:
@njhale - i am going to assign this to @g-linville to pickup since I expect you'll be on the group stuff for a bit once things settle. if you have a wip pr, let him know.

g-linville:
[...]

PR #4038, opened 2025-08-25
enhance: add a lastUpdated field to catalog entries

g-linville:
supports https://github.com/obot-platform/obot/issues/3319

I implemented this in the controller, because there was no clean way to do it in the API, due to the fact that we use `apply` for stuff coming from GitHub.


pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
```diff
@@ -48,3 +52,37 @@ func DeleteEntriesWithoutRuntime(req router.Request, _ router.Response) error {
...
+	hash := sha256.Sum256(data)
+	return hex.EncodeToString(hash[:]), nil
+}
+
+// UpdateManifestHashAndLastUpdated updates the manifest hash and last updated timestamp when configuration changes
+func UpdateManifestHashAndLastUpdated(req router.Request, _ router.Response) error {
+	entry := req.Object.(*v1.MCPServerCatalogEntry)
+
+	// Compute current config hash
+	currentHash, err := computeManifestHash(entry)
```
thedadams: nit:
```suggestion
	currentHash, err := hash.Digest(entry.Spec.Manifest)
```

g-linville: Fixed

Issue #4039, opened 2025-08-25
Having leading or trailing spaces in allowed github user list, results in unauthorized log in errors for the user.


sangee2004:
**Describe the bug**
Having leading or trailing spaces in allowed github user list, results in unauthorized log in errors for the user.

**To Reproduce**
1. Have github auth enabled setup
2. Restrict access to certain set of github users by setting - `Allowed GitHub Users` in Github auth provider with  leading or trailing spaces for users like `sangee2004, sangee-girish ,sangeetest1` where `sangee-girish` has leading and trailing space.
3. Attempt to log in as user - `sangee-girish`

Notice that user will be denied access.

**Expected behavior**
User should not be denied access in this case.
leading or trailing spaces for the user should be stripped before it can be used for validating authorization.

cjellick:
@njhale should this be moved to stalled out?

njhale:
@cjellick I suspect the current UX provided by the UI for entering array fields in the provider dialogs will trim leading/trailing space, so this may not be a problem anymore.

I'll check it out and flip this over to testing if I find it mitigates the issue.

cjellick:
im just going to kick it to qa to test that.

sangee2004:
This issue is not seen anymore.
When adding an `Allowed GitHub Users` in Github auth provider page with leading or trailing spaces and hitting enter, leading and trailing spaces get stripped before being saved as expected.

PR #4044, opened 2025-08-26
enhance: validate auth provider options before (re)configuring

njhale:
- Validate auth provider options before (re)configuration using the target auth provider's `validate` tool (when present)
- Capture and display field-specific validation errors in auth provider config dialog

Loom: https://www.loom.com/share/09e93ef0c92f4113b6bb5cedb6c72641?sid=8e1455eb-7b5a-4d06-9ccf-f1104a07199e

Addresses part of https://github.com/obot-platform/obot/issues/4039

### DIFF
--- a/pkg/api/handlers/authprovider.go
+++ b/pkg/api/handlers/authprovider.go
@@ -5,20 +5,25 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"net/http"
+	"strings"
 
 	"github.com/gptscript-ai/go-gptscript"
 	"github.com/obot-platform/obot/apiclient/types"
 	"github.com/obot-platform/obot/pkg/api"
 	"github.com/obot-platform/obot/pkg/api/handlers/providers"
 	"github.com/obot-platform/obot/pkg/gateway/server/dispatcher"
+	"github.com/obot-platform/obot/pkg/invoke"
 	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
 	"github.com/obot-platform/obot/pkg/system"
 	"github.com/tidwall/gjson"
 	"gorm.io/driver/postgres"
 	"gorm.io/gorm"
 	"gorm.io/gorm/logger"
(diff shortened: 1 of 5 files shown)


pkg/gateway/server/dispatcher/dispatcher.go
```diff
@@ -24,7 +24,7 @@ import (
 )
 
 type Dispatcher struct {
-	invoker                     *invoke.Invoker
```
thedadams, line 27: This should remain unexported. If you need the invoker somewhere else, then you can use the `services.Invoker`. There is nothing special about this one.

thedadams (changes requested):
It seems easier to have the auth provider strip leading and trailing spaces, but I'm not entirely opposed to this idea.

njhale:
> It seems easier to have the auth provider strip leading and trailing spaces, but I'm not entirely opposed to this idea.

It's not just leading/trailing spaces. There's domain specific validation for certain fields as well; e.g. organization must be a valid github login handle and can only be a list of size 1, allowed users needs to be valid github user names, etc. 

That being said, it looks like not much is exposed for other auth providers ATM. I'd rather this validation logic _not_ bleed up into Obot and for it to be in-place when/if we need it.

pkg/api/handlers/authprovider.go
```diff
@@ -126,6 +131,60 @@ func (ap *AuthProviderHandler) listAuthProviders(req api.Context) ([]types.AuthP
...
+			log.Debugf("Auth provider %q does not provide a validate tool. Looking for 'validate from %s'", ref.Name, ref.Spec.Reference)
+			return nil
+		}
+
+		return types.NewErrHTTP(http.StatusUnprocessableEntity, strings.Trim(err.Error(), "\"'"))
+	}
+
+	var validationErr authProviderValidationErr
+	if json.Unmarshal([]byte(res.Output), &validationErr) == nil && validationErr.Err != nil {
+		return types.NewErrHTTP(http.StatusUnprocessableEntity, res.Output)
```
StrongMonkey, line 182: I think it should return `validationErr.Err` so that it gives back clean error message

StrongMonkey (approved):

njhale:
Got stalled out on this and it looks like the initial issue may have been mitigated with some more robust UI. Closing this out.

PR #4054, opened 2025-08-26
enhance: use a friendlier error when listing tools for chat fails

thedadams:
This will change the message from something like:
```
500 /assistants/a1-obot/projects/p1sfq8x/threads/t1vjptc/invoke: failed to create MCP client for server Single User Everything 1: failed to create MCP client: failed to create MCP client: error from server: failed to send initialize message to server ms14h5xq: timed out waiting for MCP server to be ready
```

to something like
```
500 /assistants/a1-obot/projects/p1sfq8x/threads/t1vjptc/invoke: failed to populate tools for MCP server "Single User Everything 1": failed to create MCP client: error from server: failed to send initialize message to server ms14h5xq: timed out waiting for MCP server to be ready
```


pkg/mcp/docker.go
```diff
@@ -246,7 +246,7 @@ func (d *dockerBackend) getContainer(ctx context.Context, name string) (*contain
 
 	for _, c := range containers {
 		for _, containerName := range c.Names {
-			if strings.TrimPrefix(containerName, "/") == name {
+			if strings.TrimPrefix(containerName, "/")+"dope" == name {
```
thedadams: This was mistakenly committed. I was using this as a way of consistently getting an error. I'll remove it.

njhale: don't know how I missed this. sorry!

Issue #4037, opened 2025-08-25
`rate limit exceeded, please try again later` error presented to user when refresh was done after user was revoked access.


sangee2004:
**Describe the bug**
`rate limit exceeded, please try again later` error presented to user when refresh was done after user was revoked access.

Pre Req - Have github auth enabled in the obot setup and provide access to few users using `Allowed GitHub Users` in github auth provider.

**To Reproduce**
1. As regular user who has access to obot setup , log in and chat.
2. As admin user , revoke access to this user by removing this user from `Allowed GitHub Users`list
3. As regular user , manually refresh the chat page you are currently in (from step 2)

Notice that user is presented with the following error message:

<img width="1183" height="151" alt="Image" src="https://github.com/user-attachments/assets/77cda08e-debe-4337-a7c5-6eb983b75e2d" />

On reload of this page , user presented with `Log in` page
I see these error in obot logs
```
ERROR: github-auth-provider: failed to get state: failed to load cookied session: session ticket cookie failed validation: <nil>
ERROR: github-auth-provider: failed to get state: failed to load cookied session: session ticket cookie failed validation: <nil>
ERROR: github-auth-provider: failed to get state: failed to load cookied session: session ticket cookie failed validation: <nil>
ERROR: github-auth-provider: failed to get state: failed to load cookied session: session ticket cookie failed validation: <nil>
ERROR: github-auth-provider: failed to get state: failed to load cookied session: session ticket cookie failed validation: <nil>
ERROR: github-auth-provider: failed to get state: failed to load cookied session: session ticket cookie failed validation: <nil>
ERROR: github-auth-provider: failed to get state: failed to load cookied session: session ticket cookie failed validation: <nil>
ERROR: github-auth-provider: failed to get state: failed to load cookied session: session ticket cookie failed validation: <nil>
ERROR: github-auth-provider: failed to get state: failed to load cookied session: session ticket cookie failed validation: <nil>
ERROR: github-auth-provider: failed to get state: failed to load cookied session: session ticket cookie failed validation: <nil>
ERROR: github-auth-provider: failed to get state: failed to load cookied session: session ticket cookie failed validation: <nil>
ERROR: github-auth-provider: failed to get state: failed to load cookied session: session ticket cookie failed validation: <nil>
ERROR: github-auth-provider: failed to get state: failed to load cookied session: session ticket cookie failed validation: <nil>
[...]

PR #4065, opened 2025-08-27
fix: tell browser to delete cookie when session is invalid

g-linville:
for https://github.com/obot-platform/obot/issues/4037

### DIFF
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -34,11 +34,7 @@ const (
 	ObotAuthProviderQueryParam = "obot-auth-provider"
 )
 
-type InvalidSessionErr struct{}
-
-func (InvalidSessionErr) Error() string {
-	return "invalid session"
-}
+var ErrInvalidSession = errors.New("invalid session")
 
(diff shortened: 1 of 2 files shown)


njhale (approved):

pkg/proxy/proxy.go
```diff
@@ -34,6 +34,12 @@ const (
 	ObotAuthProviderQueryParam = "obot-auth-provider"
 )
 
+type InvalidSessionErr struct{}
+
+func (InvalidSessionErr) Error() string {
+	return "invalid session"
+}
```
thedadams: I recently learned that you could also do `var InvalidSessionErr = errors.New("invalid session")` here and use `errors.Is(err, proxy.InvalidSessionErr)` in the other spot. Not sure if you think that would be cleaner.

thedadams (approved):

pkg/proxy/proxy.go
g-linville: I like this a lot more. Fixed

PR #4067, opened 2025-08-27
chore: refactor AccessControlRules to live in the default catalog

g-linville:


pkg/api/handlers/accesscontrolrules.go
```diff
@@ -63,6 +108,11 @@ func (*AccessControlRuleHandler) Create(req api.Context) error {
 		return types.NewErrBadRequest("invalid access control rule manifest: %v", err)
 	}
 
+	// Validate that referenced resources exist in the same catalog
+	if err := (*AccessControlRuleHandler)(nil).validateResourcesInCatalog(req, manifest.Resources, catalogID); err != nil {
```
g-linville: fix this

pkg/api/handlers/accesscontrolrules.go
```diff
@@ -72,22 +122,46 @@ func (*AccessControlRuleHandler) Create(req api.Context) error {
...
+	}
+
 	existing.Spec.Manifest = manifest
 
 	if err := manifest.Validate(); err != nil {
 		return types.NewErrBadRequest("invalid access control rule manifest: %v", err)
 	}
 
+	// Validate that referenced resources exist in the same catalog
+	if err := (*AccessControlRuleHandler)(nil).validateResourcesInCatalog(req, manifest.Resources, catalogID); err != nil {
```
g-linville: and this

pkg/api/handlers/accesscontrolrules.go
```diff
@@ -97,14 +171,76 @@ func (*AccessControlRuleHandler) Update(req api.Context) error {
...
+				} else if server.Spec.MCPServerCatalogEntryName != "" {
+					// Verify the catalog entry belongs to the catalog
+					var entry v1.MCPServerCatalogEntry
+					if err := req.Get(&entry, server.Spec.MCPServerCatalogEntryName); err != nil {
+						return types.NewErrBadRequest("MCPServerCatalogEntry %s not found for server %s: %v", server.Spec.MCPServerCatalogEntryName, resource.ID, err)
+					}
+					if entry.Spec.MCPCatalogName != catalogID {
+						return types.NewErrBadRequest("MCPServer %s belongs to catalog entry in catalog %s, not %s", resource.ID, entry.Spec.MCPCatalogName, catalogID)
+					}
+				}
```
g-linville: fix this

pkg/controller/handlers/accesscontrolrule/accesscontrolrule.go
```diff
@@ -23,27 +23,47 @@ func New(acrHelper *accesscontrolrule.Helper) *Handler {
...
+	// Default to default catalog for legacy rules without catalog ID
+	if catalogID == "" {
+		catalogID = "default"
+	}
+
 	for _, resource := range acr.Spec.Manifest.Resources {
 		switch resource.Type {
 		case types.ResourceTypeMCPServerCatalogEntry:
-			if err := req.Get(&mcpservercatalogentry, req.Namespace, resource.ID); err == nil {
+			if resource.ID == "*" {
```
g-linville: fix this

pkg/api/handlers/accesscontrolrules.go
```diff
@@ -97,14 +171,76 @@ func (*AccessControlRuleHandler) Update(req api.Context) error {
...
 		},
 	})
 }
 
+// validateResourcesInCatalog validates that referenced resources exist in the specified catalog
+func (*AccessControlRuleHandler) validateResourcesInCatalog(req api.Context, resources []types.Resource, catalogID string) error {
+	for _, resource := range resources {
+		switch resource.Type {
+		case types.ResourceTypeMCPServerCatalogEntry:
+			if resource.ID != "*" {
```
g-linville: fix this and similar

pkg/controller/handlers/accesscontrolrule/accesscontrolrule.go
```diff
@@ -23,27 +23,47 @@ func New(acrHelper *accesscontrolrule.Helper) *Handler {
...
+					newResources = append(newResources, resource)
+				}
+				// If entry belongs to different catalog, remove it from the rule
 			} else if !errors.IsNotFound(err) {
 				return fmt.Errorf("failed to get MCPServerCatalogEntry %s: %w", resource.ID, err)
 			}
+			// If entry not found, remove it from the rule
 		case types.ResourceTypeMCPServer:
-			if err := req.Get(&mcpserver, req.Namespace, resource.ID); err == nil {
+			if resource.ID == "*" {
```
g-linville: and this

pkg/controller/handlers/accesscontrolrule/accesscontrolrule.go
```diff
@@ -56,3 +76,22 @@ func (h *Handler) PruneDeletedResources(req router.Request, _ router.Response) e
...
+	// Check if server came from a catalog entry
+	if server.Spec.MCPServerCatalogEntryName != "" {
+		var entry v1.MCPServerCatalogEntry
+		if err := req.Get(&entry, req.Namespace, server.Spec.MCPServerCatalogEntryName); err == nil {
+			return entry.Spec.MCPCatalogName == catalogID
+		}
+	}
+	
+	// For servers without catalog association, they belong to default catalog only
+	return catalogID == "default"
```
g-linville: fix this and remove this helper function

pkg/controller/handlers/accesscontrolrule/migration.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+const (
+	// Migration annotation to track completion
+	migrationAnnotation = "obot.ai/acr-catalog-migration"
+	migrationVersion    = "v1"
+)
+
+// MigrateToDefaultCatalog migrates existing AccessControlRules to the default catalog
+func (h *Handler) MigrateToDefaultCatalog(req router.Request, _ router.Response) error {
```
g-linville, line 10: Fix this whole thing

pkg/controller/handlers/accesscontrolrule/migration.go
```diff
@@ -0,0 +1,83 @@
...
+	// Update the access control rule
+	if err := req.Client.Update(req.Ctx, acr); err != nil {
+		return fmt.Errorf("failed to migrate access control rule to default catalog: %w", err)
+	}
+
+	return nil
+}
+
+// CheckMigrationStatus checks if all AccessControlRules have been migrated
+func CheckMigrationStatus(ctx context.Context, c client.Client) (bool, error) {
```
g-linville: remove this crap

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -411,7 +421,36 @@ func (h *Handler) DeleteUnauthorizedMCPServerInstances(req router.Request, _ rou
...
+				return fmt.Errorf("failed to delete MCPServerInstance %s: %w", instance.Name, err)
+			}
+			continue
+		}
+		
+		// Determine the catalog ID for the server
+		var catalogID string
+		if mcpServer.Spec.SharedWithinMCPCatalogName != "" {
+			catalogID = mcpServer.Spec.SharedWithinMCPCatalogName
+		} else if mcpServer.Spec.MCPServerCatalogEntryName != "" {
```
g-linville: fix this

pkg/api/handlers/accesscontrolrules.go
```diff
@@ -16,16 +16,30 @@ func NewAccessControlRuleHandler() *AccessControlRuleHandler {
...
 		return fmt.Errorf("failed to list access control rules: %w", err)
 	}
 
 	items := make([]types.AccessControlRule, 0, len(list.Items))
 	for _, item := range list.Items {
-		items = append(items, convertAccessControlRule(item))
+		// Filter by catalog ID
+		if item.Spec.MCPCatalogID == catalogID {
+			items = append(items, convertAccessControlRule(item))
+		}
```
thedadams, line 42: Does it make sense to use one of the helper functions you created here?

pkg/api/handlers/accesscontrolrules.go
```diff
@@ -71,23 +119,42 @@ func (*AccessControlRuleHandler) Create(req api.Context) error {
...
-	existing.Spec.Manifest = manifest
+	// Verify rule belongs to the requested catalog
+	if existing.Spec.MCPCatalogID != catalogID {
+		return types.NewErrBadRequest("access control rule does not belong to catalog %s", catalogID)
+	}
 
-	if err := manifest.Validate(); err != nil {
-		return types.NewErrBadRequest("invalid access control rule manifest: %v", err)
+	if err := h.validateResourcesInCatalog(req, manifest.Resources, catalogID); err != nil {
+		return types.NewErrBadRequest("%s", err.Error())
```
thedadams: Two things here:
1. What are your thoughts on adding the above check (`existing.Spec.MCPCatalogID != catalogID`) into this validation function? My thought is that all the validation logic would be self-contained. I realize that that specific validation check is superfluous for creations, but I think I like the idea of having a single place to validate.
2. `h.valiateResourcesInCatalog` already returns an `ErrBadRequest`. Since it is unexported and we control it, we can just return the error here instead of creating another `ErrBadReqest`.

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -361,7 +361,17 @@ func (h *Handler) DeleteUnauthorizedMCPServers(req router.Request, _ router.Resp
 			continue
 		}
 
-		hasAccess, err := h.accessControlRuleHelper.UserHasAccessToMCPServerCatalogEntry(user, server.Spec.MCPServerCatalogEntryName)
+		// Get the catalog ID for the server's catalog entry to check access properly
+		var catalogEntry v1.MCPServerCatalogEntry
+		if err := req.Get(&catalogEntry, system.DefaultNamespace, server.Spec.MCPServerCatalogEntryName); err != nil {
```
thedadams, line 367: We should check that the error here is actually a not-found error. Same for the change below.

pkg/api/handlers/accesscontrolrules.go
g-linville: I don't think so. Most of those helper functions are just getting the rules for a specific server/catalog entry/user/etc.

pkg/api/handlers/accesscontrolrules.go
g-linville: 1. Imo I would rather leave it as is. I would have to pass both the `existing.Spec.MCPCatalogID` and the catalog ID that we pulled from the path value into the function, which as you said is superfluous for creates. That function is specifically for validating that the resources are in the catalog. I think it is better separation of concerns to keep it that way.
2. Thanks, I didn't catch this

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
g-linville: Fixed

pkg/storage/apis/obot.obot.ai/v1/accesscontrolrule.go
```diff
@@ -15,13 +15,15 @@ type AccessControlRule struct {
 }
 
 type AccessControlRuleSpec struct {
-	Manifest types.AccessControlRuleManifest `json:"manifest"`
+	MCPCatalogID string                          `json:"mcpCatalogId,omitempty"`
```
StrongMonkey: json field should be `mcpCatalogID` since we use `ID` everywhere

g-linville: Good catch! I'm blaming Claude. I'll get the fix up for this soon.

PR #4074, opened 2025-08-28
chore: add tool previews last generated field to catalog entries

g-linville:
supports https://github.com/obot-platform/obot/issues/3319


pkg/api/handlers/mcpcatalogs.go
```diff
@@ -28,15 +29,17 @@ type MCPCatalogHandler struct {
...
 
-func NewMCPCatalogHandler(defaultCatalogPath string, serverURL string, sessionManager *mcp.SessionManager, oauthChecker MCPOAuthChecker, gatewayClient *gclient.Client) *MCPCatalogHandler {
+func NewMCPCatalogHandler(defaultCatalogPath string, serverURL string, sessionManager *mcp.SessionManager, oauthChecker MCPOAuthChecker, gatewayClient *gclient.Client, storageClient kclient.Client) *MCPCatalogHandler {
 	return &MCPCatalogHandler{
 		defaultCatalogPath: defaultCatalogPath,
 		serverURL:          serverURL,
 		sessionManager:     sessionManager,
 		oauthChecker:       oauthChecker,
 		gatewayClient:      gatewayClient,
+		storageClient:      storageClient,
```
njhale: I believe you can use the `Storage` field on the `api.Context` passed to your handler methods instead of adding a new field to the handler type.

njhale (changes requested):

pkg/api/handlers/mcpcatalogs.go
g-linville: Thanks, just pushed the fix

thedadams (approved):

PR #4075, opened 2025-08-28
chore: doc updates for MCP catalogs

cloudnautique:
This goes over the different types of MCP servers, and how to add them.

### DIFF
--- a/docs/docs/20-concepts/03-admin/05-mcp-servers.md
+++ b/docs/docs/20-concepts/03-admin/05-mcp-servers.md
@@ -43,13 +43,15 @@ Multi-User servers will still require the user to authenticate to the Obot Gatew
 
 ### Remote server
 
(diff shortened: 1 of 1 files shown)


docs/docs/20-concepts/03-admin/05-mcp-servers.md
```diff
@@ -0,0 +1,131 @@
...
+- The organization owns shared service accounts or API keys
+- You want to simplify user onboarding by eliminating individual setup
+- The service supports organizational or tenant-based access
+- Usage monitoring and control at the organizational level is important
+
+Multi-User servers will still require the user to authenticate to the Obot Gateway's configured identity provider.
+
+### Remote server
+
+MCP Servers that are HTTP Streaming compatible should be configured this way. These servers can be provided by trusted 3rd party vendors and must conform to the OAUTH specs. Remote servers also work for MCP servers deployed through existing CI/CD pipeline within the organization.
```
drpebcak: is conforming to the oauth spec a requirement? can we not do bearer or other header-based auth?

cloudnautique: I think it is fair to say we will work with those that conform to the SPEC, and you MAY be able to use other forms but not guaranteed.

PR #4096, opened 2025-08-29
chore: update MCP audit log batching defaults to be less conservative

thedadams:
This change also adds this configuration to the helm chart.

### DIFF
--- a/pkg/services/config.go
+++ b/pkg/services/config.go
@@ -100,8 +100,8 @@ type Config struct {
 	// Sendgrid webhook
 	SendgridWebhookUsername           string `usage:"The username for the sendgrid webhook to authenticate with"`
 	SendgridWebhookPassword           string `usage:"The password for the sendgrid webhook to authenticate with"`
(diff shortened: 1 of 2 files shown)


pkg/services/config.go
```diff
@@ -100,8 +100,8 @@ type Config struct {
 	// Sendgrid webhook
 	SendgridWebhookUsername           string `usage:"The username for the sendgrid webhook to authenticate with"`
 	SendgridWebhookPassword           string `usage:"The password for the sendgrid webhook to authenticate with"`
-	MCPAuditLogPersistIntervalMinutes int    `usage:"The interval in minutes to persist MCP audit logs to the database" default:"1"`
-	MCPAuditLogsPersistBatchSize      int    `usage:"The number of MCP audit logs to persist in a single batch" default:"10000"`
+	MCPAuditLogPersistIntervalSeconds int    `usage:"The interval in minutes to persist MCP audit logs to the database" default:"5"`
```
njhale: The usage tag says minutes, but the field name and code are in seconds

njhale (changes requested):

pkg/services/config.go
thedadams: Thanks. Fixed.

Issue #4049, opened 2025-08-26
Power User and Power User Plus roles


g-linville:
We want to add two new user roles to Obot. Here will be the new role lineup:

- User: current end-user functionality
- Power User: can create MCP servers for himself only
- Power User Plus: can create MCP servers and manage AccessControlRules for them
- Admin: current admin functionality

Names are subject to change.

For Power Users and Power User Plusses, they should have a personal workspace where they manage their MCPServerCatalogEntries and multi-user MCPServers. For now, I will call this a PowerUserWorkspace in the API. This will be the logical container for their MCPServerCatalogEntries and multi-user MCPServers that they create.

Various details:

- Existing AccessControlRules are scoped to a global level. This needs to be migrated to live under the default MCPCatalog.
- Admins will also have their own PowerUserWorkspace.
- Power Users will not be allowed to create multi-user servers (single and remote only), but Power User Plusses can create all three.
- When a Power User or Power User Plus is demoted to a regular User, their PowerUserWorkspace and everything in it will be destroyed.

Additionally, we want some kind of setting that Admins can change that determines what the default role for new users is (User, Power User, or Power User Plus).

cjellick:
> Should admins also have a PowerUserWorkspace that they can use?

Yes, but they would access it as a normal user - through a separate app link in the profile dropdown

> The UI will need to find a way to give admins visibility into all existing PowerUserWorkspaces.

What do you mean? is this different from the new list page I show in the figma?

> Should Power Users (not Power User Plusses) be allowed to create multi-user MCP servers? Or single-user and remote only?
>   - There is seemingly not much use for a multi-user server for a Power User, since they can't share it with others anyway.

Single and remote only. but this is one area where i feel like the story isnt really complete. like it going to feel disjointed to create an mcp server entry and then hop back over to Chat or My Connectors to instantiate it. i havent figured out this detail yet

> When a Power User [Plus] is demoted to a regular User, should we automatically destroy their PowerUserWorkspace and all MCPServerCatalogEntries/MCPServers contained therein?

Yes, with a warning in the UI that thats what will hapen

g-linville:
> What do you mean? is this different from the new list page I show in the figma?

Oh you're right; we do already have a wireframe for that. I forgot.

ivyjeong13:
[...]

PR #4097, opened 2025-08-29
feat: add power user and power user plus roles

g-linville:
for https://github.com/obot-platform/obot/issues/4049

^ Be sure to read up on this issue before reviewing this.

### DIFF
--- a/pkg/controller/handlers/mcpserver/mcpserver.go
+++ b/pkg/controller/handlers/mcpserver/mcpserver.go
@@ -203,6 +203,7 @@ func (h *Handler) MigrateSharedWithinMCPCatalogName(req router.Request, _ router
 
 	if server.Spec.SharedWithinMCPCatalogName != "" && server.Spec.MCPCatalogID == "" {
 		server.Spec.MCPCatalogID = server.Spec.SharedWithinMCPCatalogName
+		server.Spec.SharedWithinMCPCatalogName = ""
 		return req.Client.Update(req.Ctx, server)
 	}
 


apiclient/types/user.go
```diff
@@ -3,6 +3,8 @@ package types
 const (
 	RoleUnknown Role = iota
 	RoleAdmin
+	RolePowerUserPlus
+	RolePowerUser
```
g-linville, line 7: Do we want these to be roles number 2 and 3? I know we probably want to migrate away from this iota setup eventually anyway, but I wasn't totally sure if these were the best numbers for us to use for this.

apiclient/types/poweruserworkspace.go
```diff
@@ -0,0 +1,9 @@
+package types
+
+type PowerUserWorkspace struct {
+	Metadata
+	UserID string `json:"userID,omitempty"`
+	Role   Role   `json:"role,omitempty"`
+}
+
+type PowerUserWorkspaceList List[PowerUserWorkspace]
```
thedadams: I thought `go fmt` always put an empty line at the end of a file.

apiclient/types/userrolechange.go
```diff
@@ -0,0 +1,10 @@
+package types
+
+type UserRoleChange struct {
```
thedadams: A quick search indicates that this is not used anywhere and can be removed.

pkg/api/authz/poweruserworkspace.go
```diff
@@ -0,0 +1,64 @@
...
+	// Validate role-based access to workspace endpoints
+	if !a.validateWorkspaceRoleAccess(req.URL.Path, isPowerUserPlus, user.GetUID()) {
+		return false, nil
+	}
+
+	var workspace v1.PowerUserWorkspace
+	if err := a.cache.Get(req.Context(), kclient.ObjectKey{
+		Namespace: system.DefaultNamespace,
+		Name:      resources.WorkspaceID,
+	}, &workspace); err != nil {
```
thedadams, line 36: Only do an uncached Get if the error is a not-found error.

pkg/api/authz/poweruserworkspace.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	// Only PowerUserPlus can access MCP servers and access control rules
+	if strings.Contains(path, "/workspaces/") &&
+		(strings.Contains(path, "/servers") || strings.Contains(path, "/access-control-rules")) {
+		return isPowerUserPlus
+	}
+
+	return false
+}
```
thedadams: I think these path-based rules belong here: https://github.com/obot-platform/obot/blob/main/pkg/api/authz/authz.go#L25

pkg/gateway/server/user.go
```diff
@@ -31,6 +67,13 @@ func (s *Server) getCurrentUser(apiContext api.Context) error {
 		return err
 	}
 
+	// Check if this is an admin user who needs a PowerUserWorkspace created
+	if user.Role.HasRole(types2.RoleAdmin) {
+		if err := s.ensureAdminWorkspace(apiContext, user); err != nil {
+			pkgLog.Warnf("failed to ensure admin workspace for user %s: %v", user.Username, err)
+		}
+	}
```
thedadams: Something about a `GET` possible creating resources rubs me the wrong way. However, the only other way I can think of to do this is to, on startup, create a "User role changed" object for all admins and check it there.

I'm not sure if I like that better, but at least a `GET` won't create objects.

What are your thoughts?

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -339,10 +339,27 @@ func (h *Handler) DeleteUnauthorizedMCPServers(req router.Request, _ router.Resp
 	// Iterate through each MCPServer and make sure it is still allowed to exist.
 	for _, server := range mcpServers.Items {
 		if server.Spec.ThreadName != "" || server.Spec.SharedWithinMCPCatalogName != "" {
-			// For project-scoped servers and multi-user servers created by the admin, we don't need to check them.
+			// For legacy project-scoped servers and multi-user servers created by the admin, we don't need to check them.
```
thedadams, line 342: nit: I disagree with using the term legacy here. We still have project-scoped MCP servers and I don't think they are going anywhere.

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -339,10 +339,27 @@ func (h *Handler) DeleteUnauthorizedMCPServers(req router.Request, _ router.Resp
...
 		}
 
+		if server.Spec.PowerUserWorkspaceID != "" {
+			// For multi-user servers in a PowerUserWorkspace, make sure that the user on that workspace is a PowerUserPlus, and not a normal PowerUser
+			var workspace v1.PowerUserWorkspace
+			if err := req.Get(&workspace, req.Namespace, server.Spec.PowerUserWorkspaceID); err != nil {
+				return fmt.Errorf("failed to get power user workspace %s: %w", server.Spec.PowerUserWorkspaceID, err)
+			}
+
+			if workspace.Spec.Role != types.RoleAdmin && workspace.Spec.Role != types.RolePowerUserPlus {
```
thedadams: nit:
```suggestion
			if !workspace.Spec.Role.HasRole(types.RolePowerUserPlus) {
```

pkg/storage/apis/obot.obot.ai/v1/mcpserver.go
```diff
@@ -79,6 +83,8 @@ type MCPServerSpec struct {
 	NeedsURL bool `json:"needsURL,omitempty"`
 	// PreviousURL contains the URL of the server before it was updated to match the catalog entry.
 	PreviousURL string `json:"previousURL,omitempty"`
+	// PowerUserWorkspaceID contains the name of the PowerUserWorkspace that owns this MCP server, if there is one.
+	PowerUserWorkspaceID string `json:"powerUserWorkspaceID,omitempty"`
```
thedadams, line 89: nit: it seems that this is analogous to `SharedWithinMCPCatalogName`, so could it be changed to `SharedWithinMCPWorkspaceName` or something similar?

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
```diff
@@ -0,0 +1,261 @@
...
+	} else if oldPrivileged && newPrivileged {
+		if err := h.updateWorkspaceRole(req.Ctx, req.Client, req.Namespace, userIDStr, roleChange.Spec.NewRole); err != nil {
+			return err
+		}
+
+		// If demoting to PowerUser from PowerUserPlus or Admin, clean up workspace resources.
+		// PowerUsers are not allowed to manage Access Control Rules or multi-user MCPServers.
+		if roleChange.Spec.NewRole == types.RolePowerUser &&
+			(roleChange.Spec.OldRole == types.RolePowerUserPlus ||
+				roleChange.Spec.OldRole == types.RoleAdmin) {
```
thedadams: nit:
```suggestion
			roleChange.Spec.OldRole.HasRole(types.RolePowerUserPlus) {
```

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
```diff
@@ -0,0 +1,261 @@
...
+	// Create the default access control rule for this workspace
+	if err := h.createDefaultAccessControlRule(req.Ctx, req.Client, req.Namespace, workspace); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (h *Handler) isPrivilegedRole(role types.Role) bool {
+	return role.HasRole(types.RolePowerUser) || role.HasRole(types.RolePowerUserPlus) || role.HasRole(types.RoleAdmin)
```
thedadams:
```suggestion
	return role.HasRole(types.RolePowerUser)
```

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
```diff
@@ -0,0 +1,261 @@
...
+		FieldSelector: fields.SelectorFromSet(map[string]string{
+			"spec.userID": userIDStr,
+		}),
+	}); err != nil {
+		return err
+	}
+
+	if len(existingWorkspaces.Items) > 0 {
+		return nil
+	}
```
thedadams: nano-nit:
```suggestion
	}); err != nil || len(existingWorkspaces.Items) > 0 {
		return err
	}
```

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
```diff
@@ -0,0 +1,261 @@
...
+		Namespace: namespace,
+		FieldSelector: fields.SelectorFromSet(map[string]string{
+			"spec.userID": userID,
+		}),
+	}); err != nil {
+		return err
+	}
+
+	for _, workspace := range workspaces.Items {
+		if err := client.Delete(ctx, &workspace); err != nil {
```
thedadams: Should we ignore not found just in case?

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
```diff
@@ -0,0 +1,261 @@
...
+		}),
+	}); err != nil {
+		return err
+	}
+
+	for _, server := range servers.Items {
+		if err := client.Delete(ctx, &server); err != nil {
+			return err
+		}
+	}
```
thedadams: I understand that a user can only have one workspace right now, but we get support for multiple here with very minimal change.
```suggestion
	for _, workspace := range workspaces.Items {
		// Delete non-generated AccessControlRules in this workspace
		var acrs v1.AccessControlRuleList
		if err := client.List(ctx, &acrs, &kclient.ListOptions{
			Namespace: namespace,
			FieldSelector: fields.SelectorFromSet(map[string]string{
				"spec.powerUserWorkspaceID": workspace.Name,
			}),
		}); err != nil {
			return err
		}

		for _, acr := range acrs.Items {
			if !acr.Spec.Generated {
				if err := client.Delete(ctx, &acr); err != nil {
					return err
				}
			}
		}

		// Delete all MCPServers in this workspace
		var servers v1.MCPServerList
		if err := client.List(ctx, &servers, &kclient.ListOptions{
			Namespace: namespace,
			FieldSelector: fields.SelectorFromSet(map[string]string{
				"spec.powerUserWorkspaceID": workspace.Name,
			}),
		}); err != nil {
			return err
		}

		for _, server := range servers.Items {
			if err := client.Delete(ctx, &server); err != nil {
				return err
			}
		}
	}
```

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
```diff
@@ -0,0 +1,261 @@
...
+					{
+						Type: types.ResourceTypeSelector,
+						ID:   "*",
+					},
+				},
+			},
+		},
+	}
+
+	return create.OrGet(ctx, client, defaultACR)
```
thedadams: I know this works, but using `create.OrGet` with something that has a generated name seems weird.

pkg/api/handlers/accesscontrolrules.go
```diff
@@ -97,18 +126,30 @@ func (h *AccessControlRuleHandler) Create(req api.Context) error {
 			Namespace:    req.Namespace(),
 		},
 		Spec: v1.AccessControlRuleSpec{
-			MCPCatalogID: catalogID,
-			Manifest:     manifest,
+			Manifest: manifest,
```
thedadams: nit, and then you can get rid of the added code below:
```suggestion
			Manifest:             manifest,
			MCPCatalogID:         catalogID,
			PowerUserWorkspaceID: workspaceID,
```

pkg/api/handlers/accesscontrolrules.go
```diff
@@ -162,23 +226,33 @@ func (h *AccessControlRuleHandler) Update(req api.Context) error {
 	return req.Write(convertAccessControlRule(existing))
 }
 
-// Delete deletes an access control rule (admin only).
+// Delete deletes an access control rule for a catalog or workspace.
 func (*AccessControlRuleHandler) Delete(req api.Context) error {
 	catalogID := req.PathValue("catalog_id")
-	if catalogID == "" {
-		return types.NewErrBadRequest("catalog_id is required")
+	workspaceID := req.PathValue("workspace_id")
```
thedadams, line 227: This is the only handler where we are not "verifying that the scope exists." I don't think it's a big deal since we are deleting, but, personally, I would add it for consistency.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -153,30 +161,45 @@ func (h *MCPCatalogHandler) ListEntriesForCatalog(req api.Context) error {
...
+		var catalog v1.MCPCatalog
+		if err := req.Get(&catalog, catalogName); err != nil {
+			return fmt.Errorf("failed to get catalog: %w", err)
+		}
+	} else if workspaceID != "" {
+		var workspace v1.PowerUserWorkspace
+		if err := req.Get(&workspace, workspaceID); err != nil {
+			return fmt.Errorf("failed to get workspace: %w", err)
+		}
+	} else {
```
thedadams: nit: like you did in other areas since we don't need the object.
```suggestion
	if catalogName != "" {
		if err := req.Get(&v1.MCPCatalog{}, catalogName); err != nil {
			return fmt.Errorf("failed to get catalog: %w", err)
		}
	} else if workspaceID != "" {
		if err := req.Get(&PowerUserWorkspace{}, workspaceID); err != nil {
			return fmt.Errorf("failed to get workspace: %w", err)
		}
	} else {
```

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -212,20 +242,34 @@ func (h *MCPCatalogHandler) CreateEntry(req api.Context) error {
...
```
thedadams: Same here, and for the rest of the file.

Also pointing out that we are not doing this check for list, get, and tool previews. Personally, I would add it for consistency.

pkg/api/handlers/mcp.go
```diff
@@ -59,13 +59,24 @@ func (m *MCPHandler) GetCatalogEntryFromDefaultCatalog(req api.Context) error {
...
+			hasAccess bool
+			err       error
+		)
+
+		if entry.Spec.MCPCatalogName != "" {
+			hasAccess, err = m.acrHelper.UserHasAccessToMCPServerCatalogEntryInCatalog(req.User, entry.Name, entry.Spec.MCPCatalogName)
+		} else if entry.Spec.PowerUserWorkspaceID != "" {
+			hasAccess, err = m.acrHelper.UserHasAccessToMCPServerCatalogEntryInWorkspace(req.User, entry.Name, entry.Spec.PowerUserWorkspaceID)
+		}
+
```
thedadams: Check the error as soon as possible, but no sooner.
```suggestion
```

pkg/api/handlers/mcp.go
```diff
@@ -1572,13 +1676,24 @@ func (m *MCPHandler) GetServerFromDefaultCatalog(req api.Context) error {
...
+			hasAccess bool
+			err       error
+		)
+
+		if server.Spec.SharedWithinMCPCatalogName != "" {
+			hasAccess, err = m.acrHelper.UserHasAccessToMCPServerInCatalog(req.User, server.Name, server.Spec.SharedWithinMCPCatalogName)
+		} else if server.Spec.PowerUserWorkspaceID != "" {
+			hasAccess, err = m.acrHelper.UserHasAccessToMCPServerInWorkspace(req.User, server.Name, server.Spec.PowerUserWorkspaceID)
+		}
+
```
```suggestion
```

pkg/api/handlers/mcp.go
```diff
@@ -1485,32 +1568,53 @@ func convertMCPServer(server v1.MCPServer, credEnv map[string]string, serverURL
...
+				// Check workspace-scoped servers
+				hasAccess, err := m.acrHelper.UserHasAccessToMCPServerInWorkspace(req.User, server.Name, server.Spec.PowerUserWorkspaceID)
+				if err != nil {
+					return err
+				}
+				if hasAccess {
+					allowedServers = append(allowedServers, server)
+				}
+			}
+		}
```
thedadams: nit: this check is handled slightly differently in (at least) 2 other places. For consistency, I would suggest that this be updated to match the other places.

thedadams (changes requested):
Lots of comments, but my main concern here is that the path-based authorization be moved to the correct place.

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
g-linville: Well we have ProjectServers, which are different from what I was calling "project-scoped servers", which are MCP servers with a `Spec.ThreadName` field on them. I think this is something that has been deprecated/unsupported for months now, and is not possible to create with our current APIs.

apiclient/types/poweruserworkspace.go
g-linville: It might. Idk if we lint this folder or not? This file was AI-written so it didn't necessarily get `go fmt`'d.

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
thedadams: OK. My mistake.

pkg/api/authz/poweruserworkspace.go
g-linville: Discussed in Slack. The static rules won't work for this, but I will work on reimplementing this in a clearer/more extensible way.

pkg/gateway/server/user.go
g-linville: Yeah I agree with you. I didn't think about the weirdness of a `GET` causing something to be created. I'll go with your suggestion. :+1:

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
g-linville: So I implemented it this way because the `HasRole` function relies on the numerical values of the roles, which iirc is something we intend to do away with in the future.

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
g-linville: Yeah good point, this is worth doing

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
thedadams: I brought up the idea of changing this and was shot down. Also, I think that it is better to use the functionality as-is, and change it all at once.

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
g-linville: Agreed

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
g-linville: Sounds good. I didn't realize that your idea for that was shot down.

apiclient/types/poweruserworkspace.go
g-linville: Fixed

apiclient/types/userrolechange.go
g-linville: Deleted

pkg/api/authz/poweruserworkspace.go
g-linville: Fixed

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
g-linville: Fixed

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
g-linville: Fixed

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
g-linville: Fixed

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
g-linville: Fixed

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
g-linville: Fixed

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
g-linville: Fixed

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
g-linville: Fixed

pkg/api/handlers/accesscontrolrules.go

pkg/api/handlers/accesscontrolrules.go
g-linville: Fixed

pkg/api/handlers/mcpcatalogs.go

pkg/api/handlers/mcpcatalogs.go
g-linville: Fixed, and improved the consistency in this file

pkg/api/handlers/mcp.go

pkg/api/handlers/mcp.go

pkg/api/handlers/mcp.go
g-linville: Fixed, and improved the consistency in this file

pkg/api/authz/poweruserworkspace.go
g-linville: I refactored this to use regex patterns to match routes that should be PowerUserPlus only. Hopefully the logic is a little clearer now.

pkg/gateway/server/user.go
g-linville: Discussing this more in Slack

pkg/api/router/router.go
```diff
@@ -383,24 +384,24 @@ func Router(services *services.Services) (http.Handler, error) {
 	// Slack event receiver
 	mux.HandleFunc("POST /api/slack/events", slackEventHandler.HandleEvent)
 
-	// MCP Catalog Entries (user routes to access single-user and remote MCP servers configured by the admin)
-	mux.HandleFunc("GET /api/all-mcp-catalogs/entries", mcp.ListEntriesInDefaultCatalog)
-	mux.HandleFunc("GET /api/all-mcp-catalogs/entries/{entry_id}", mcp.GetCatalogEntryFromDefaultCatalog)
+	// MCP Catalog Entries (user routes to access single-user and remote MCP servers from all sources)
+	mux.HandleFunc("GET /api/all-mcps/entries", mcp.ListEntriesFromAllSources)
```
StrongMonkey, line 388: Will UI need to accommadate this route change?

g-linville: No, I made the necessary frontend changes for this in this PR

StrongMonkey: Nvm, I saw you changed the path in ui

pkg/api/handlers/poweruserworkspace.go
```diff
@@ -0,0 +1,65 @@
...
+
+	return req.Write(types.PowerUserWorkspaceList{
+		Items: items,
+	})
+}
+
+// Get returns a specific power user workspace by ID.
+func (*PowerUserWorkspaceHandler) Get(req api.Context) error {
+	var workspace v1.PowerUserWorkspace
+	if err := req.Get(&workspace, req.PathValue("workspace_id")); err != nil {
```
StrongMonkey, line 52: Nit: should we prevent user from accessing workspace when you don't have access? looks like if you know the id then you can get the workspace

pkg/api/handlers/accesscontrolrules.go
```diff
@@ -47,43 +57,62 @@ func (*AccessControlRuleHandler) List(req api.Context) error {
...
+	} else if workspaceID != "" && rule.Spec.PowerUserWorkspaceID != workspaceID {
+		return types.NewErrBadRequest("access control rule does not belong to workspace %s", workspaceID)
 	}
 
 	return req.Write(convertAccessControlRule(rule))
 }
 
-// Create creates a new access control rule (admin only).
+// Create creates a new access control rule for a catalog or workspace.
 func (h *AccessControlRuleHandler) Create(req api.Context) error {
```
StrongMonkey, line 98: Are we grant access control to a whole workspace if power user plus want to share their mcpserver? would there be a way to share invidiually?

pkg/api/handlers/poweruserworkspace.go
g-linville: This is prevented in the authz logic

pkg/api/handlers/accesscontrolrules.go
g-linville: The access control rules work the same way for workspaces as for catalogs. You can share all things, or individual servers.

pkg/gateway/server/user.go
g-linville: Fixed. We are handling it on startup for existing admins, and also at the time of user creation using a callback

pkg/storage/apis/obot.obot.ai/v1/mcpserver.go
g-linville: Fixed, by renaming `SharedWithinMCPCatalogName` to `MCPCatalogID`.

pkg/api/handlers/accesscontrolrules.go
StrongMonkey: Got it.

pkg/api/authz/poweruserworkspace.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+	}
+
+	return workspace.Spec.UserID == user.GetUID(), nil
+}
+
+// Workspace access patterns that require PowerUserPlus privileges, and not PowerUser
+var powerUserPlusRequiredPatterns = []*regexp.Regexp{
+	regexp.MustCompile(`/workspaces/[^/]+/servers`),              // MCP servers management
+	regexp.MustCompile(`/workspaces/[^/]+/access-control-rules`), // Access control rules management
```
thedadams, line 55: I'm good with this for now, but we should extend our authorization framework if we find ourselves needing this again.

pkg/controller/handlers/mcpserver/mcpserver.go
```diff
@@ -197,3 +197,14 @@ func (h *Handler) DeleteServersWithoutRuntime(req router.Request, _ router.Respo
 
 	return nil
 }
+
+func (h *Handler) MigrateSharedWithinMCPCatalogName(req router.Request, _ router.Response) error {
+	server := req.Object.(*v1.MCPServer)
+
+	if server.Spec.SharedWithinMCPCatalogName != "" && server.Spec.MCPCatalogID == "" {
+		server.Spec.MCPCatalogID = server.Spec.SharedWithinMCPCatalogName
+		return req.Client.Update(req.Ctx, server)
```
thedadams: nit: so it doesn't ever need to be used again.
```suggestion
	if server.Spec.SharedWithinMCPCatalogName != "" && server.Spec.MCPCatalogID == "" {
		server.Spec.MCPCatalogID = server.Spec.SharedWithinMCPCatalogName
		server.Spec.SharedWithinMCPCatalogName = ""
		return req.Client.Update(req.Ctx, server)
```

Issue #4108, opened 2025-08-29
drop (or reduce) the fade out at the top of the chat well


cjellick:
<img width="1136" height="213" alt="Image" src="https://github.com/user-attachments/assets/d541bc18-a73d-4a57-aabe-29115eb2c5db" />

I dont like this....it takes up a lot of veritical space and makes me feel claustrophobic. Can we eliminate or reduce it?

My vote is for eliminate and just have a hard cut-off like chatgpt, but if theres a good reason for having some amount of fade out, im open.

sangee2004:
This issue is addressed when testing with latest build.
I dont see the fade out at the top of the chat well anyone when there is a very large chat message.


<img width="1316" height="765" alt="Image" src="https://github.com/user-attachments/assets/405f4174-d39f-42ff-aa43-e5f823a3290d" />

Issue #4014, opened 2025-08-22
If there's an everything rule, skip the access control rule prompt at the end of creating a server


cjellick:
We get this prompt at the end of creating a server

<img width="808" height="248" alt="Image" src="https://github.com/user-attachments/assets/b3acbbc1-7ab9-43bd-b0fa-ef9c61175461" />

I think we should just skip it if there's an everyone everything rule in place.

sangee2004:
We are not presented with the following page when creating MCP severs when we have the default  `Everything` rule is already present.


<img width="788" height="274" alt="Image" src="https://github.com/user-attachments/assets/4a800d14-9287-4691-ba84-3af8c546aa06" />

Issue #4008, opened 2025-08-22
Chat Threads - By default chats should be sorted by Created time with most recent entries shown at the top of the list.


sangee2004:
**Describe the bug**
Chat Threads - By default chats should be sorted by Created time with most recent entries shown at the top of the list.

**To Reproduce**
1. Go to 'Chat Threads'
Notice that the chats presented are sorted by `Name` field with chats with no name appearing in the top of the list followed by names which start with numbers and then alphabetically sorted list.

**Expected behavior**
By default chats should be sorted by `Created time` with most recent entries shown at the top of the list.

This is regression from the fix provided to address this issue -  https://github.com/obot-platform/obot/issues/3742

sangee2004:
This issue is addressed when testing with latest builds.

PR #4114, opened 2025-09-02
fix: minor UI updates & skip acr step if everything rule in place

ivyjeong13:
Addresses #4014
Addresses #4008
Addresses #4108

### DIFF
--- a/ui/user/src/lib/components/admin/CatalogServerForm.svelte
+++ b/ui/user/src/lib/components/admin/CatalogServerForm.svelte
@@ -511,7 +511,10 @@
 			const entryResponse = await handleFns[type]?.(catalogId);
 			savedEntry = entryResponse;
 
-			if (!entry) {
+			const existingRules = await AdminService.listAccessControlRules();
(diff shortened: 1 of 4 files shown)


ui/user/src/lib/components/admin/CatalogServerForm.svelte
```diff
@@ -511,7 +511,10 @@
 			const entryResponse = await handleFns[type]?.(catalogId);
 			savedEntry = entryResponse;
 
-			if (!entry) {
+			const existingRules = await AdminService.listAccessControlRules();
+			const hasEverythingRule = existingRules.some((rule) => rule.id === 'acr1-everything');
```
g-linville: The only downside to this approach is that, if the pre-created `Everything` rule gets removed, but then the admin manually creates the equivalent of it, we won't detect that here, because we are doing it based on name rather than checking the subjects and resources on the rule. Not sure whether this matters, but I wanted to mention it.

ivyjeong13: :+1: Will change to check for the subjects and resources instead

Issue #4084, opened 2025-08-28
Copy and pasting large text into chat well results in jankiness


cjellick:
if you copy and paste a large block of text into our chat area and then try to type, the ui gets super janky and you cant easily finish.

Can see it here
https://www.loom.com/share/370c132d8c45440ba3fa4cd4a1efd8a0

shwetapsdet:
I've tested the functionality on main, it's working as expected

https://www.loom.com/share/aa7ee8e9ee0e4f5b939ab6fe0309151f?sid=a673c8bb-5c22-470a-9931-8f7bc9700d3e

PR #4119, opened 2025-09-02
feat: implement Milkdown plaintext editor to be used in chat input

ryu-man:
Addresses #4084

### DIFF
--- a/ui/user/src/lib/components/messages/Input.svelte
+++ b/ui/user/src/lib/components/messages/Input.svelte
@@ -1,10 +1,11 @@
 <script lang="ts">
-	import { type InvokeInput } from '$lib/services';
-	import { autoHeight } from '$lib/actions/textarea.js';
-	import { ArrowUp, LoaderCircle } from 'lucide-svelte';
 	import { onMount, type Snippet, tick } from 'svelte';
+	import { ArrowUp, LoaderCircle } from 'lucide-svelte';
+
+	import { type InvokeInput } from '$lib/services';
 	import type { EditorItem } from '$lib/services/editor/index.svelte';
-	import { twMerge } from 'tailwind-merge';
+
+	import PlaintextEditor from './PlaintextEditor.svelte';
 
 	interface Props {
 		id?: string;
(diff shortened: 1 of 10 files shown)


ui/user/src/lib/components/messages/Input.svelte
```diff
@@ -142,38 +146,55 @@
...
-				{placeholder}
-			></textarea>
-			{#if !children}
-				{@render submitButton()}
+<div
+	class=" focus-within:ring-blue bg-surface1 mt-4 flex h-fit max-h-[80svh] overflow-hidden rounded-2xl focus-within:shadow-md focus-within:ring-1"
+>
+	<!-- <div class="max-h-full min-h-full w-full flex-1 overflow-y-auto"></div> -->
+
+	<div class="flex min-h-full w-full flex-col" {id}>
```
ivyjeong13: Can you check if the `/` command to show prompts work with this input change? Wasn't popping up for me on this branch but did see it pop up in main: 

Screenshot 2025-09-04 at 11 35 44 AM

Would also check that it works and supplies the prompt into the input after selecting one of them and filling out any fields if it requires it

ryu-man: Wow, I did not know this feature existed. Going to fix any issues

ryu-man: Fixed

ivyjeong13: https://www.loom.com/share/3ccc536345154077b43e0293f892675f

I see prompts popping up! But it doesn't look to supply the filled out prompt after filling out the information? Switched to main to give an example in the Loom if it helps!

ryu-man: Thanks for the loom; I will address that.

ryu-man: The issue is fixed 
https://www.loom.com/share/9f7bd5321e2641eb8642ddea6f5f62ef?sid=2e603d1e-2976-48ae-ae25-b2db6df4279b

PR #4122, opened 2025-09-03
fix: migrate access control rules before controller start

thedadams:
Migrating these in the controller causes a race condition where MCP servers could be erroneously deleted.


njhale (approved):
lgtm

g-linville (approved):

PR #4126, opened 2025-09-03
fix: retain tool previews when updating catalog entries

g-linville:


thedadams (approved):

sangee2004:
I see the following use case work now
1. Create a single entry MCP server
2. Use `Launch Temporary Instance` option in `Tools` tab to get the tools to be displayed.
3. Update MCP server in step 1 (added a new configuration parameter)
4. Revisit `Tools` tab
Existing tools from step 2 are still displayed.
User is additionally presented with `Regenerate tools and Capabilities` option.
Screenshot 2025-09-03 at 10 45 51 AM
Able to use this option to refetch the tools

Issue #4016, opened 2025-08-23
Bring back URL parameterization for remote MCP servers


njhale:
At some point we dropped the ability to parameterize remote MCP server URLs for user configuration.

e.g. 

If an admin defines a remote MCP server catalog entry with a parameterized URL like `https://gitmcp.io/${ORG}/${REPO}`, `ORG` and `URL` become required "arguments" for users connecting to the remote server. The provided arguments then used to build that target URL when Obot connects to the remote MCP server.

Note: We need to carefully consider how this will interact with the advanced config options for locking down remote MCP server URLs as well as catalog entry config drift.

cjellick:
@njhale @g-linville can you give me explicit reasons why this will be tricky or we should be cautious about doing it? Just want to understand before I decided what to do with it.

g-linville:
@cjellick main thing is that I recently tightened up the way that we validate URLs. There's no way the current system would let you include anything like `${}` in it. So that logic would need to change. But we would still need to validate it somewhere. So that might involve kind of using a fake value to fill in for the template and then validating. That's the main thing that I think would be annoying about this.

I also think it's just not as good from a security perspective. But I guess it's a "use at your own risk" kind of thing for the admin who decides to set up a catalog entry like this, so not the biggest deal.

g-linville:
Oh also, drift detection, as Nick mentioned in the issue.

tybalex:
This feature is required for configuring Databricks Remote MCP Servers, I am looking for something like:
```
env:
  - name: Databricks workspace hostname
    key: DATABRICKS_WORKSPACE_URL
    description: Databricks workspace hostname, (e.g., https://your-workspace.cloud.databricks.com)
    required: true
    sensitive: false
  - name: Genie space ID
    key: DATABRICKS_GENIE_SPACE_ID
    description: Your Databricks Genie space ID
    required: true
    sensitive: false

runtime: remote
remoteConfig:
  fixedUrl: https://${DATABRICKS_WORKSPACE_URL}/api/2.0/mcp/genie/${DATABRICKS_GENIE_SPACE_ID}
```

cjellick:
@StrongMonkey - as mentioned, I need you to pick this up.

StrongMonkey:
This should be done as @tybalex verified that it works. There is one remaining issue which i will address

PR #4129, opened 2025-09-03
Enhance: Add URL templating

StrongMonkey:
Add URL template substitution. For now, it just does simpler substitution with ${ENV_NAME} where env_name is the env variable you defined when supplying to the mcp server. 

The catalog entry will have to define urlTemplate field instead of fixedUrl, as we introduced separate logic for validating url Template. 

The final connection url will be updated to mcp server once user configures the env variables that are defined in catalog.

Disclosure: Tests file are added by AI.

https://github.com/obot-platform/obot/issues/4016

### DIFF
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -688,7 +688,6 @@ func (m *MCPHandler) GetPrompt(req api.Context) error {
 		if nse := (*mcp.ErrNotSupportedByBackend)(nil); errors.As(err, &nse) {
 			return types.NewErrHTTP(http.StatusBadRequest, nse.Error())
 		}
-
 		var are nmcp.AuthRequiredErr
 		if errors.As(err, &are) {
 			return types.NewErrHTTP(http.StatusPreconditionFailed, "MCP server requires authentication")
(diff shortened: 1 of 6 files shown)


pkg/api/handlers/mcp.go
```diff
@@ -633,18 +633,10 @@ func (m *MCPHandler) GetPrompts(req api.Context) error {
...
-		}
 		if nse := (*mcp.ErrNotSupportedByBackend)(nil); errors.As(err, &nse) {
 			return types.NewErrHTTP(http.StatusBadRequest, nse.Error())
 		}
-
-		var are nmcp.AuthRequiredErr
-		if errors.As(err, &are) {
-			return types.NewErrHTTP(http.StatusPreconditionFailed, "MCP server requires authentication")
-		}
-		return fmt.Errorf("failed to list prompts: %w", err)
```
g-linville: What is the reason for this change?

pkg/api/handlers/mcp.go
```diff
@@ -682,18 +682,10 @@ func (m *MCPHandler) GetPrompt(req api.Context) error {
...
-		}
 		if nse := (*mcp.ErrNotSupportedByBackend)(nil); errors.As(err, &nse) {
 			return types.NewErrHTTP(http.StatusBadRequest, nse.Error())
 		}
-
-		var are nmcp.AuthRequiredErr
-		if errors.As(err, &are) {
-			return types.NewErrHTTP(http.StatusPreconditionFailed, "MCP server requires authentication")
-		}
-		return fmt.Errorf("failed to get prompt: %w", err)
```
g-linville: Same here.

pkg/api/handlers/mcp_test.go
```diff
@@ -0,0 +1,289 @@
...
+			if err != nil {
+				t.Errorf("unexpected error: %v", err)
+				return
+			}
+
+			// For edge cases, we just verify the function doesn't crash
+			// and returns a result
+			if result == "" && tt.template != "" {
+				t.Errorf("expected non-empty result for template %q", tt.template)
+			}
```
g-linville: This seems strange to me. Why don't we have an expected result for each of these tests?

pkg/api/handlers/mcp.go
StrongMonkey: Oh shoot, I though AI added these so i removed it. Will add it back

pkg/api/handlers/mcp_test.go
StrongMonkey: yeah, will remove. AI generated tests

pkg/api/handlers/mcp.go
StrongMonkey: reverted back.

Issue #4090, opened 2025-08-29
Show in admin ui that you’re running enterprise


cjellick:
In the admin, add a "enterprise" pill to the top right, similar to the "chat" pill. 

@thedadams - can you add whatever backend bit would be necessary for ivy to determine this? I am thinking a new field on the /api/version endpoint.

thedadams:
Backend change has merged. The `/api/version` API will now return a `"enterprise": true` when appropriate.

sangee2004:
This issue is addressed when testing with latest build

<img width="1306" height="848" alt="Image" src="https://github.com/user-attachments/assets/acb0e790-7de8-4222-86e1-4a6048c761e1" />

PR #4132, opened 2025-09-03
enhance: include enterprise info in version API

thedadams:
Whether the container is running the enterprise version is controlled via an environment variable that can only be set when building the image.

Issue: https://github.com/obot-platform/obot/issues/4090

### DIFF
--- a/pkg/api/handlers/version.go
+++ b/pkg/api/handlers/version.go
@@ -30,6 +30,7 @@ type VersionHandler struct {
 	supportDocker    bool
 	authEnabled      bool
 	sessionStore     SessionStore
+	enterprise       bool
 }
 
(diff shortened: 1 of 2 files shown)


pkg/api/handlers/version.go
```diff
@@ -70,6 +71,8 @@ func (v *VersionHandler) getVersionResponse() map[string]any {
 	values["dockerSupported"] = v.supportDocker
 	values["authEnabled"] = v.authEnabled
 	values["sessionStore"] = v.sessionStore
+	values["enterpriseVersion"] = v.enterprise
```
StrongMonkey: should the name be `enterprise` since v.enterprise is just a boolean?

thedadams: Yeah, I'll change it.

StrongMonkey (approved):

Issue #4093, opened 2025-08-29
Audit Logs - Not all audit log entries are shown when there are lot of rows(~5000) to be displayed.


sangee2004:
**Describe the bug**
Audit Logs - Not all audit log entries are shown when there are lot of rows(~5000) to be displayed.

Pre Req - Have a setup were there are many audit entries (~5000) 
**To Reproduce**
1. Go to 'Audit Logs`
2. Choose 90 day window for listing audit entries

Notice that only few entries (out of the ~5200 entries returned by `/mcp-audit-logs` api call ) are presented in the Audit logs . This is after waiting for 10 minutes in the `Audit Log` page. Initially i saw entries fetched until `Aug 8` . After waiting for 10 more minutes , I see entries fetched up to `Aug 5` now.

<img width="1309" height="233" alt="Image" src="https://github.com/user-attachments/assets/bbdcb2c9-df79-44d3-b5a6-3ad07d656315" />

Time Line Graph is able account for all the entries that are available

<img width="1281" height="419" alt="Image" src="https://github.com/user-attachments/assets/5e83f0d4-dba1-414f-a2a9-f3efb85a1a52" />

**Expected behavior**
We should be able to show all audit entries in a reasonably good time.
If we are not able to load 5000 entries in a reasonably good time , we should not have pagination set to 10,000. It should be set to a lower value that allows for UI to display entries without long delays.

sangee2004:
This issue is addressed when testing with latest build.

When trying to get the audit logs for last 90 days which in my case is ~6790 entries ,  I am able to scroll down the entire list of all the 6790 entries very smoothly with no delays.

PR #4140, opened 2025-09-04
fix: improve audit logs table performace when working with large datasets

ryu-man:
Addresses #4093 

- Create a virtual table component to render only visible rows.
- Dynamically calculate the number of rows to render in the viewport.

### DIFF
--- a/ui/user/src/lib/components/admin/audit-logs/AuditLogs.svelte
+++ b/ui/user/src/lib/components/admin/audit-logs/AuditLogs.svelte
@@ -1,119 +1,105 @@
 <script lang="ts">
 	import { twMerge } from 'tailwind-merge';
 	import { tooltip } from '$lib/actions/tooltip.svelte';
+	import { VirtualTable } from '$lib/components/ui/virtual-table';
 
-	let {
-		data = [],
-		onSelectRow,
-		emptyContent,
-		getUserDisplayName,
-		currentFragmentIndex = 0,
-		getFragmentIndex,
-		getFragmentRowIndex,
-		onLoadNextFragment
-	} = $props();
+	let { data = [], onSelectRow, emptyContent, getUserDisplayName } = $props();
+
+	let element: HTMLElement | undefined = $state();
+	let navElementHeight = $state(64);
+
+	const itemHeight = 56;
+
+	$effect(() => {
+		if (!element) return;
+
+		const navElement = element.closest('main')?.querySelector('nav');
+
+		navElementHeight = navElement?.clientHeight || 64;
+	});
 </script>
 
 <!-- Data Table -->
 <div
-	class="dark:bg-surface2 w-full overflow-hidden overflow-x-auto rounded-lg border border-transparent bg-white shadow-sm"
+	bind:this={element}
+	class="dark:bg-surface2 relative flex h-screen w-full min-w-full divide-y divide-gray-200 overflow-hidden overflow-x-auto rounded-lg border border-transparent bg-white shadow-sm"
+	style="height: calc(100vh - {navElementHeight}px - 54px);"
 >
 	{#if data.length}
-		<table class="min-w-full divide-y divide-gray-200">
-			<thead class="dark:bg-surface1 bg-surface2">
-				<tr class="sticky top-0">
-					<th
-						scope="col"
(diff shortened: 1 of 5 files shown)


ui/user/src/lib/components/ui/virtual-table/virtual-table.svelte
```diff
@@ -0,0 +1,263 @@
...
+	</div>
+</div>
+
+<style>
+	.virtual-list-viewport {
+		-webkit-overflow-scrolling: touch;
+	}
+
+	.virtual-list-row {
+		/* Remove display: block to maintain table-row behavior */
```
ivyjeong13: Just wondering if this comment is correct? Don't see `display` being changed here, maybe can be removed if the `display:block` line was removed?

ivyjeong13 (dismissed):
Personally, not a fan of now there's two scrollbars, one to scroll down to the table and another to scroll through the table. But if Craig's good with it :+1: 

Otherwise was wondering if like a virtualized infinite scroll experience was possible or local pagination

ui/user/src/lib/components/ui/virtual-table/virtual-table.svelte
ryu-man: Old code. Not needed any more

ryu-man:
@cjellick  @ivyjeong13  The virtual table can't work with the old scroll experience (page scroll) because the table has to have a `max-height` to be able to calculate how many rows to render in the viewport; changing the height will increase/decrease how many rows to render in the viewport.

Showing both scrolls so the users have the ability to scroll down the page to see the data table in full height, the table dynamically calculates how many rows need to be shown within the viewport; smaller screens fewer rows, larger screens more rows.

I am not sure if the whole page can be virtualized based on the table rows!

ui/user/src/lib/components/ui/virtual-page/virtual-page-viewport.svelte
```diff
@@ -0,0 +1,318 @@
...
+
+		viewportElement.scrollTo({
+			top: targetScrollTop,
+			behavior: 'smooth'
+		});
+	}
+</script>
+
+<Render
+	class="virtual-page-root flex h-[100svh] max-h-[100svh] w-full overflow-hidden"
```
ivyjeong13: Just wondering what's the `virtual-page-root` and `virtual-page-viewport` classes used for?

ryu-man: I used them to quickly identify the elements in the DevTools element inspector.

ui/user/src/lib/components/admin/audit-logs/AuditLogs.svelte
```diff
@@ -1,119 +1,90 @@
...
-						>Client</th
-					>
-					<th
-						scope="col"
-						class="sticky top-0 px-6 py-3 text-left text-xs font-medium tracking-wider text-gray-500 uppercase"
-						>IP Address</th
-					>
-				</tr>
-			</thead>
+		<VirtualPageTable class={twMerge('w-full flex-1')}>
```
ivyjeong13: Just quick loom of a couple things I noticed! 
https://www.loom.com/share/3dd27629587040feb5859fa3441ebc93

ryu-man: Thanks for the loom

ryu-man: To fix the second issue, I had to convert the table layout to be `fixed` and manually set each column width.

ui/user/src/lib/components/ui/virtual-page/virtual-page-viewport.svelte
ivyjeong13: Ah to debug? I'd just axe classes that aren't actually being used.

ui/user/src/lib/components/admin/audit-logs/AuditLogs.svelte
ivyjeong13: :+1: Prolly need main's dataset to see how these fixed widths play out

ui/user/src/lib/components/ui/virtual-page/virtual-page-viewport.svelte
ryu-man: Will axe them

Issue #4107, opened 2025-08-29
Project Templates & Shareable Project Links


cjellick:
## Overview

Enable users to create shareable links for their projects, allowing others to generate a point-in-time copy (template) of a project. The copied project includes all relevant configuration and knowledge, but excludes user-specific data and history. This feature supports easy onboarding, sharing, and demonstration of project setups.

## Goals

- Allow project owners to generate a shareable link for any project.
- Recipients of the link can create a new project based on a snapshot/template of the original.
- Ensure the copied project is a static snapshot (not live or synced with the original).
- Exclude sensitive or user-specific data from the template.
- Support use cases like onboarding, demos, and blog-driven adoption.

### 1. As a project owner:
- I want to generate a shareable link for my project.
- I want the link to create a copy of my project for others, without exposing my data or history.

### 2. As a recipient:
- I want to click a shared link and instantly create a new project based on the template.
- I want to configure any required mcp servers  myself.
- I want to see all relevant knowledge and tasks included in the template.

### 3. As a content creator (e.g., for a blog):
- I want to embed a link that lets readers try a pre-configured project setup.

### Functional

- **Link Generation:**  
  - Project owners can generate a unique, shareable link for any project.
  - The link represents a snapshot of the project at the time of link creation.

- **Project Copying:**  
  - Clicking the link creates a new project for the recipient, based on the snapshot.
  - The new project is not connected to the original (no live sync).

- **Included in Template:**  
  - Project-level configuration (excluding user-specific secrets).
  - Knowledge files attached at the project level.
  - Task definitions.

- **Excluded from Template:**  
  - Chat threads.
  - Task runs/history.
  - User-specific configuration (e.g., MCP server credentials).

- **MCP Server Handling:**  
  - If the template references MCP servers, recipients must configure these upon first use.
  - Provide clear prompts/guidance for required configuration.

### Non-Functional

  - No sensitive data or user history is included in the template.
  - Recipients should have a clear, guided experience when setting up required configurations.

## Open Questions

- Should project owners be able to update or revoke links after creation?
  - Answer: yes
- Should there be an expiration or access control on shared links?
  - Answer: no, not needed at this point
- Are there additional project elements (besides knowledge and tasks) that should be included or excluded?

sangee2004:
Project Sharing feature is now available in latest build.

Following basic workflows relating  Project Sharing was validated.

[...]

PR #4154, opened 2025-09-05
feat: project templates and shareable links

njhale:
Addresses the bulk of https://github.com/obot-platform/obot/issues/4107

Loom: https://www.loom.com/share/964fd03595814f92a8072df77b95634d?sid=45918fb2-f0be-4c76-bfa9-1e7a59308a41

To be addressed in a follow-up:
- Preventing users from copying projects containing MCP servers they
  don't have access to
- Preventing copied project upgrades when the user doesn't have access
  to MCP servers in the latest snapshot

### DIFF
--- a/pkg/api/handlers/projects.go
+++ b/pkg/api/handlers/projects.go
@@ -529,7 +529,7 @@ func convertProject(thread *v1.Thread, parentThread *v1.Thread) types.Project {
 		TemplatePublicID:             thread.Status.UpgradePublicID,
 	}
 
-	if !thread.Status.LastUpgraded.Time.IsZero() {
+	if !thread.Status.LastUpgraded.IsZero() {
 		p.TemplateLastUpgraded = types.NewTime(thread.Status.LastUpgraded.Time)
 	}
 
--- a/pkg/api/handlers/template.go
+++ b/pkg/api/handlers/template.go
@@ -101,10 +101,6 @@ func (h *TemplateHandler) DeleteProjectTemplate(req api.Context) error {
 		return err
 	}
 
-	if len(templateThreadList.Items) < 1 {
-		return types.NewErrNotFound("template not found for project %s", projectID)
-	}
-
 	var templateThread *v1.Thread
 	for _, thread := range templateThreadList.Items {
 		if thread.Spec.Template {
@@ -174,10 +170,6 @@ func (h *TemplateHandler) GetProjectTemplate(req api.Context) error {
 		"spec.sourceThreadName": projectThreadName,
 	}); err != nil {
 		return err
-	}
-
-	if len(templateThreadList.Items) < 1 {
-		return types.NewErrNotFound("template not found for project %s", projectID)
 	}
 
 	var templateThread *v1.Thread
--- a/pkg/api/handlers/threads.go
+++ b/pkg/api/handlers/threads.go
@@ -44,7 +44,7 @@ func convertTemplateThread(thread v1.Thread, share *v1.ThreadShare) types.Projec
 
 	// Populate LastUpdated from status field
 	if !thread.Status.LastUpgraded.IsZero() {
-		template.ProjectSnapshotLastUpgraded = *types.NewTime(thread.Status.LastUpgraded.Time)
+		template.ProjectSnapshotLastUpgraded = types.NewTime(thread.Status.LastUpgraded.Time)
 	}
 
 	if share != nil {
--- a/pkg/controller/handlers/threads/template.go
+++ b/pkg/controller/handlers/threads/template.go
@@ -292,7 +292,6 @@ func newProjectThreadConfig(
 		// Normalize MCP ID to catalog entry name
 		mcpID := projectMCPID
 		if spec, ok := mcpServerSpecs[mcpID]; ok {
-			// TOOD(njhale): Do we need to worry about the catalog this belongs to?
 			mcpID = strings.Join([]string{spec.MCPCatalogID, spec.MCPServerCatalogEntryName}, "/")
 		}
 
--- a/pkg/storage/apis/obot.obot.ai/v1/thread.go
+++ b/pkg/storage/apis/obot.obot.ai/v1/thread.go
@@ -48,17 +48,13 @@ func (in *Thread) Has(field string) (exists bool) {
 
 // SetLatestConfigRevision sets the latest config revision for the thread and returns true if the
 // operation has changed the revision history.
-func (in *Thread) SetLatestConfigRevision(revision string) (changed bool) {
+func (in *Thread) SetLatestConfigRevision(revision string) bool {
 	if len(in.Status.ConfigRevisions) == 0 {
 		// No revision history, initialize the revision history
 		in.Status.ConfigRevisions = []string{revision}
 		return true
 	}
 
-	var (
-		revisionCount = len(in.Status.ConfigRevisions)
-		latest        = in.Status.ConfigRevisions[revisionCount-1]
-	)
 	if in.Spec.Template {
 		// Add the new revision to the end of the history (marking it as the latest revision)
 		// Then deduplicate the history so that the latest revisions are kept and order is preserved
@@ -79,7 +75,11 @@ func (in *Thread) SetLatestConfigRevision(revision string) (changed bool) {
 	// Non-template thread, if the revision differs from the latest revision or there's more than one
 	// revision in the history, replace it with the latest revision.
 	// For non-template threads, we only want to keep the latest revision in the history.
-	if latest != revision || len(in.Status.ConfigRevisions) > 1 {
+	var (
+		revisionCount = len(in.Status.ConfigRevisions)
+		latest        = in.Status.ConfigRevisions[revisionCount-1]
+	)
+	if latest != revision || revisionCount > 1 {
 		in.Status.ConfigRevisions = []string{revision}
 		return true
 	}
@@ -91,18 +91,19 @@ func (in *Thread) SetLatestConfigRevision(revision string) (changed bool) {
 // The revisions are deduplicated by keeping the latest revision and removing duplicates.
 // The order of the revisions is preserved.
 func dedupeRevisions(revisions []string) []string {
-	var (
-		visited = make(map[string]struct{}, len(revisions))
-		deduped = make([]string, 0, len(revisions))
-	)
-	for i := len(revisions) - 1; i >= 0; i-- {
-		revision := revisions[i]
-		if _, ok := visited[revision]; !ok {
+	// First pass: find the last index of each revision
+	lastIndex := make(map[string]int, len(revisions))
+	for i, revision := range revisions {
+		lastIndex[revision] = i
+	}
+
+	// Second pass: build result keeping only items at their last index
+	deduped := make([]string, 0, len(revisions))
+	for i, revision := range revisions {
+		if lastIndex[revision] == i {
 			deduped = append(deduped, revision)
-			visited[revision] = struct{}{}
 		}
 	}
-	slices.Reverse(deduped)
 
 	return deduped
 }
@@ -157,7 +158,7 @@ func (in *Thread) Get(field string) string {
 }
 
 func (in *Thread) FieldNames() []string {
-	return []string{"spec.userUID", "spec.project", "spec.template", "spec.parentThreadName", "spec.sourceThreadName"}
+	return []string{"spec.userUID", "spec.project", "spec.template", "spec.agentName", "spec.parentThreadName", "spec.sourceThreadName"}
 }
 
 func (in *Thread) GetColumns() [][]string {
--- a/apiclient/types/template.go
+++ b/apiclient/types/template.go
@@ -3,7 +3,7 @@ package types
 type ProjectTemplate struct {
 	Metadata
 	ProjectSnapshot                  ThreadManifest `json:"projectSnapshot,omitempty"`
-	ProjectSnapshotLastUpgraded      Time           `json:"projectSnapshotLastUpgraded,omitempty"`
+	ProjectSnapshotLastUpgraded      *Time          `json:"projectSnapshotLastUpgraded,omitempty"`
 	ProjectSnapshotStale             bool           `json:"projectSnapshotStale,omitempty"`
 	ProjectSnapshotUpgradeInProgress bool           `json:"projectSnapshotUpgradeInProgress,omitempty"`
 	MCPServers                       []string       `json:"mcpServers,omitempty"`
--- a/apiclient/types/zz_generated.deepcopy.go
+++ b/apiclient/types/zz_generated.deepcopy.go
@@ -2799,7 +2799,10 @@ func (in *ProjectTemplate) DeepCopyInto(out *ProjectTemplate) {
 	*out = *in
 	in.Metadata.DeepCopyInto(&out.Metadata)
 	in.ProjectSnapshot.DeepCopyInto(&out.ProjectSnapshot)
-	in.ProjectSnapshotLastUpgraded.DeepCopyInto(&out.ProjectSnapshotLastUpgraded)
+	if in.ProjectSnapshotLastUpgraded != nil {
+		in, out := &in.ProjectSnapshotLastUpgraded, &out.ProjectSnapshotLastUpgraded
+		*out = (*in).DeepCopy()
+	}
 	if in.MCPServers != nil {
 		in, out := &in.MCPServers, &out.MCPServers
 		*out = make([]string, len(*in))
--- a/pkg/storage/apis/obot.obot.ai/v1/thread_test.go
+++ b/pkg/storage/apis/obot.obot.ai/v1/thread_test.go
@@ -1,0 +1,113 @@
+package v1
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDedupeRevisions(t *testing.T) {
+	for _, tt := range []struct {
+		name      string
+		revisions []string
+		expected  []string
+	}{
+		{name: "no duplicates", revisions: []string{"a", "b", "c"}, expected: []string{"a", "b", "c"}},
+		{name: "one duplicate out of order", revisions: []string{"a", "b", "c", "a"}, expected: []string{"b", "c", "a"}},
+		{name: "multiple consecutive duplicates", revisions: []string{"a", "a", "a", "b", "b", "b"}, expected: []string{"a", "b"}},
+		{name: "multiple duplicates out of order", revisions: []string{"a", "d", "a", "c", "b", "c", "b"}, expected: []string{"d", "a", "c", "b"}},
+		{name: "multiple duplicates out of order with consecutives", revisions: []string{"a", "d", "a", "c", "b", "c", "b", "b"}, expected: []string{"d", "a", "c", "b"}},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.expected, dedupeRevisions(tt.revisions))
+		})
+	}
+}
+
+func TestSetLatestConfigRevision(t *testing.T) {
+	type expected struct {
+		changed   bool
+		revisions []string
+	}
+	for _, tt := range []struct {
+		name     string
+		thread   *Thread
+		revision string
+		expected expected
+	}{
+		{
+			name:     "empty revision history",
+			thread:   &Thread{},
+			revision: "a",
+			expected: expected{changed: true, revisions: []string{"a"}},
+		},
+		{
+			name: "template thread with new revision",
+			thread: &Thread{
+				Spec:   ThreadSpec{Template: true},
+				Status: ThreadStatus{ConfigRevisions: []string{"c", "b"}},
+			},
+			revision: "a",
+			expected: expected{changed: true, revisions: []string{"c", "b", "a"}},
+		},
+		{
+			name: "template thread with old latest revision",
+			thread: &Thread{
+				Spec:   ThreadSpec{Template: true},
+				Status: ThreadStatus{ConfigRevisions: []string{"c", "b", "a"}},
+			},
+			revision: "a",
+			expected: expected{changed: false, revisions: []string{"c", "b", "a"}},
+		},
+		{
+			name: "template thread with duplicate revisions",
+			thread: &Thread{
+				Spec:   ThreadSpec{Template: true},
+				Status: ThreadStatus{ConfigRevisions: []string{"b", "c", "b", "c", "a", "a", "b"}},
+			},
+			revision: "a",
+			expected: expected{changed: true, revisions: []string{"c", "b", "a"}},
+		},
+		{
+			name: "thread with no revisions",
+			thread: &Thread{
+				Spec: ThreadSpec{Project: true},
+			},
+			revision: "a",
+			expected: expected{changed: true, revisions: []string{"a"}},
+		},
+		{
+			name: "thread with old latest revision",
+			thread: &Thread{
+				Spec:   ThreadSpec{Project: true},
+				Status: ThreadStatus{ConfigRevisions: []string{"b"}},
+			},
+			revision: "a",
+			expected: expected{changed: true, revisions: []string{"a"}},
+		},
+		{
+			name: "thread with multiple revisions keeps old latest revision",
+			thread: &Thread{
+				Spec:   ThreadSpec{Project: true},
+				Status: ThreadStatus{ConfigRevisions: []string{"b", "c", "a"}},
+			},
+			revision: "a",
+			expected: expected{changed: true, revisions: []string{"a"}},
+		},
+		{
+			name: "thread with multiple revisions replaces old latest revision",
+			thread: &Thread{
+				Spec:   ThreadSpec{Project: true},
+				Status: ThreadStatus{ConfigRevisions: []string{"b", "c", "a"}},
+			},
+			revision: "d",
+			expected: expected{changed: true, revisions: []string{"d"}},
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			changed := tt.thread.SetLatestConfigRevision(tt.revision)
+			assert.Equal(t, tt.expected.changed, changed)
+			assert.Equal(t, tt.expected.revisions, tt.thread.Status.ConfigRevisions)
+		})
+	}
+}


pkg/api/handlers/projectmcp.go
```diff
@@ -56,7 +56,7 @@ func convertProjectMCPServer(projectServer *v1.ProjectMCPServer, mcpServer *v1.M
 	}
 	pmcp.Alias = mcpServer.Spec.Alias
 
-	if cred != nil && mcpServer.Spec.SharedWithinMCPCatalogName == "" {
+	if mcpServer.Spec.SharedWithinMCPCatalogName == "" {
```
njhale: Note: I removed the `cred != nil` check because it was preventing the fields below from being set for "copied" mcp servers

pkg/api/handlers/mcp.go
```diff
@@ -210,6 +210,10 @@ func (m *MCPHandler) ListServer(req api.Context) error {
 	items := make([]types.MCPServer, 0, len(servers.Items))
 
 	for _, server := range servers.Items {
+		if server.Annotations[v1.HideMCPServerAnnotation] == "true" {
```
njhale: This prevents users from seeing the snapshot project's (potentially unconfigured/unauthenticated) MCP servers in their set of enabled connectors. 

These MCP servers are part of the snapshot resources that get copied by users when they hit a share link or upgrade their project.

pkg/api/handlers/threads.go
```diff
@@ -33,19 +34,23 @@ func NewThreadHandler(dispatcher *dispatcher.Dispatcher, events *events.Emitter)
...
+		ProjectSnapshot:                  thread.Spec.Manifest,
+		ProjectSnapshotStale:             thread.Status.SourceThreadUpgradeAvailable,
+		ProjectSnapshotUpgradeInProgress: thread.Status.UpgradeInProgress || thread.Annotations[v1.ThreadUpgradeApprovedAnnotation] == "true",
+		AssistantID:                      thread.Spec.AgentName,
+		ProjectID:                        strings.Replace(thread.Spec.SourceThreadName, system.ThreadPrefix, system.ProjectPrefix, 1),
+		Ready:                            thread.Status.Created,
+	}
+
+	// Populate ProjectSnapshotRevision from annotation if present
+	if ts := thread.Annotations[v1.ProjectSnapshotRevisionAnnotation]; ts != "" {
```
njhale: `v1.ProjectSnapshotRevisionAnnotation` is just an informational timestamp so publishers and consumers can reason about when a snapshot was last updated. It could also be a status field on the thread that gets set after an upgrade has finished.

pkg/controller/handlers/threads/threads.go
```diff
@@ -380,51 +415,48 @@ func (t *Handler) CopyTasksFromSource(req router.Request, _ router.Response) err
...
-		err          error
-	)
-	for _, taskName := range thread.Spec.Manifest.SharedTasks {
-		var task v1.Workflow
-		if err := req.Get(&task, thread.Namespace, taskName); apierrors.IsNotFound(err) {
-			modified = true
-			continue
-		} else if err != nil {
+	// Delete all existing tasks for this thread
+	var existing v1.WorkflowList
```
njhale, line 419: I changed this to copy all tasks

pkg/storage/apis/obot.obot.ai/v1/thread.go
```diff
@@ -190,6 +190,29 @@ type ThreadStatus struct {
...
+	// - introduction message
+	// - starter messages
+	// - tool set
+	// - tasks
+	// - knowledge files
+	// - model provider
+	// - model
+	// - prompt
+	// - allowed MCP tools
+	// - project MCP servers
```
njhale: I've since dropped some of these from the hash since they didn't feel like they would be destructive (e.g. allowed MCP tools).

ui/user/src/lib/components/agents/AgentCatalog.svelte
```diff
@@ -1,262 +0,0 @@
-<script lang="ts">
```
njhale, line 1: Dropped this since it wasn't being used and simplified things a bit

ui/user/src/lib/components/agents/AgentCopy.svelte
```diff
@@ -1,6 +1,5 @@
 <script lang="ts">
```
njhale, line 1: Planning on moving this to a more aptly named component/directory down the road.

ui/user/src/lib/components/chat/McpServerRequirements.svelte
```diff
@@ -0,0 +1,223 @@
+<script lang="ts">
```
njhale, line 1: This got a bit messy, but I wanted to force users to walk through configuring MCP servers when they copy a project. I'm going to refactor this a bit.

pkg/api/authz/template.go
```diff
@@ -16,6 +16,7 @@ func (a *Authorizer) checkTemplate(req *http.Request, resources *Resources) (boo
 	var templateShareList v1.ThreadShareList
 	err := a.cache.List(req.Context(), &templateShareList, kclient.InNamespace(system.DefaultNamespace), kclient.MatchingFields{
 		"spec.publicID": resources.TemplateID,
+		"spec.template": "true",
```
thedadams: Typically, cached clients don't support more than one field selector. I would be surprised if this was an exception.

pkg/api/handlers/projectmcp.go
```diff
@@ -56,7 +56,7 @@ func convertProjectMCPServer(projectServer *v1.ProjectMCPServer, mcpServer *v1.M
 	}
 	pmcp.Alias = mcpServer.Spec.Alias
 
-	if cred != nil && mcpServer.Spec.MCPCatalogID == "" {
+	if mcpServer.Spec.MCPCatalogID == "" {
```
thedadams, line 59: Is this expected? The comment below seems to indicate that this was intentional behavior.

pkg/api/handlers/projects.go
```diff
@@ -247,6 +247,51 @@ func (h *ProjectsHandler) GetProject(req api.Context) error {
...
+	}
+
+	// Verify the source is actually a template
+	if !source.Spec.Template {
+		return types.NewErrBadRequest("source project is not a template")
+	}
+
+	// Project has diverged from the snapsbot, upgrade to the latest snapshot by setting the upgrade annotation
+	if thread.Annotations == nil {
+		thread.Annotations = map[string]string{}
```
thedadams: What is the reason for doing this with annotations instead of fields on the object itself? Personally, I would prefer fields.

If there is some reason to use annotations that I am not aware of, then nit:
```suggestion
		thread.Annotations = make(map[string]string, 2)
```

pkg/api/handlers/projects.go
```diff
@@ -469,6 +514,8 @@ func convertProject(thread *v1.Thread, parentThread *v1.Thread) types.Project {
 		Editor:                       thread.IsEditor(),
 		UserID:                       thread.Spec.UserID,
 		WorkflowNamesFromIntegration: thread.Status.WorkflowNamesFromIntegration,
+		TemplateUpgradeAvailable:     thread.Status.SourceThreadUpgradeAvailable || thread.Annotations[v1.ThreadUpgradeApprovedAnnotation] == "true",
+		TemplateUpgradeInProgress:    thread.Status.UpgradeInProgress,
```
thedadams: Yeah, I think that if `SourceThreadUpgradeAvailable` and `UpgradeInProgress` are fields then we can use a field for the "upgrade approved" part.

I think it makes sense to put this field on the spec of the `Thread`, but not in the manifest.

pkg/api/handlers/threads.go
thedadams: +1 for status field

pkg/storage/apis/obot.obot.ai/v1/thread.go
thedadams: I think these all make sense.

For allowed tools for example, if I change them myself, then I won't ever be able to upgrade from the template again. What if the next upgrade of the template has the allowed tools list that I need?

pkg/api/handlers/mcp.go
thedadams: Does it make sense for these to be separate object types instead of having a tracking annotation?

pkg/api/handlers/template.go
```diff
@@ -42,121 +40,83 @@ func (h *TemplateHandler) CreateProjectTemplate(req api.Context) error {
...
+					v1.ProjectSnapshotRevisionAnnotation: now,
+				},
+			},
+			Spec: v1.ThreadSpec{
+				Manifest:         projectThread.Spec.Manifest,
+				AgentName:        projectThread.Spec.AgentName,
+				SourceThreadName: projectThread.Name,
+				UserID:           projectThread.Spec.UserID,
+				Project:          true,
+				Template:         true,
```
thedadams, line 63: Same question: does it make sense for these to be separate objects?

pkg/api/handlers/template.go
```diff
@@ -42,121 +40,83 @@ func (h *TemplateHandler) CreateProjectTemplate(req api.Context) error {
...
-	var err error
-	switch {
-	case updated.CreationTimestamp.IsZero():
-		err = req.Create(updated)
-	case !equality.Semantic.DeepEqual(templateThreadShare.Spec, updated.Spec):
-		err = req.Update(updated)
+	// The project has diverged from the snapshot.
+	// Set the annotation to trigger an upgrade and return the latest snapshot.
+	if snapshotThread.Annotations == nil {
+		snapshotThread.Annotations = map[string]string{}
```
thedadams: nit:
```suggestion
		snapshotThread.Annotations = make(map[string]string, 2)
```

pkg/controller/handlers/threads/template.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	if thread.Status.UpgradeInProgress && thread.Status.ThreadConfigHash == source.Status.ThreadConfigHash {
+		thread.Status.UpgradeInProgress = false
+		return req.Client.Status().Update(req.Ctx, thread)
+	}
+
+	if thread.Annotations[v1.ThreadUpgradeApprovedAnnotation] != "true" || thread.Status.UpgradeInProgress {
+		return nil
+	}
```
thedadams: nit: put this higher in the function.

pkg/controller/handlers/threads/template.go
```diff
@@ -0,0 +1,285 @@
...
+	// Clear derived statuses to trigger downstream copy controllers
+	desired.Status.CopiedTasks = false
+	desired.Status.CopiedTools = false
+	desired.Status.ThreadConfigHash = ""
+	desired.Status.SourceThreadUpgradeAvailable = false
+	desired.Status.SharedKnowledgeSetName = ""
+	desired.Status.KnowledgeSetNames = nil
+	desired.Status.UpgradeInProgress = true
+
+	if err := req.Client.Update(req.Ctx, desired); err != nil {
```
thedadams: There is a bug in `kinm` where this will update the status, too. Not sure if we want to rely on that.

pkg/controller/handlers/threads/template.go
```diff
@@ -0,0 +1,285 @@
...
+	return hex.EncodeToString(sum[:])
+}
+
+func hashJSON(v any) (string, error) {
+	b, err := json.Marshal(v)
+	if err != nil {
+		return "", err
+	}
+	return hashBytes(b), nil
+}
```
thedadams: There is a `hash.Digest` you can use. It's in gptscript, but we've used it in Obot.

pkg/controller/handlers/threads/threads.go
```diff
@@ -503,3 +515,141 @@ func (t *Handler) RemoveOldFinalizers(req router.Request, _ router.Response) err
...
+				Spec: v1.MCPServerSpec{
+					Manifest:                  sourceMCPServer.Spec.Manifest,
+					Alias:                     sourceMCPServer.Spec.Alias,
+					UserID:                    thread.Spec.UserID,
+					MCPServerCatalogEntryName: sourceMCPServer.Spec.MCPServerCatalogEntryName,
+				},
+			}
+			if thread.Spec.Template {
+				// Hide the MCP user from the user view
+				copiedMCPServer.Annotations = map[string]string{v1.HideMCPServerAnnotation: "true"}
```
thedadams: If we are going to use the same objects here, my preference would be to make this a field on the spec rather than an annotation.

pkg/controller/handlers/threads/threads.go
```diff
@@ -503,3 +515,141 @@ func (t *Handler) RemoveOldFinalizers(req router.Request, _ router.Response) err
 	}
 	return nil
 }
+
+// copyMCPServersFromSource copies MCP servers from the source thread to the destination thread,
+// creating appropriate per-user instances and project mappings.
+func (t *Handler) copyMCPServersFromSource(req router.Request, thread *v1.Thread) error {
+	var sourceProjectMcpList v1.ProjectMCPServerList
```
thedadams: nano-nit:
```suggestion
	var sourceProjectMCPList v1.ProjectMCPServerList
```

pkg/controller/handlers/threads/threads.go
```diff
@@ -503,3 +515,141 @@ func (t *Handler) RemoveOldFinalizers(req router.Request, _ router.Response) err
...
+	if err := req.Client.List(req.Ctx, &sourceProjectMcpList, kclient.InNamespace(thread.Namespace), kclient.MatchingFields{
+		"spec.threadName": thread.Spec.SourceThreadName,
+	}); err != nil {
+		return err
+	}
+
+	desiredPMSNames := map[string]struct{}{}
```
thedadams: nit:
```suggestion
	desiredPMSNames := make(map[string]struct{}, len(sourceProjectMcpList))
```

ui/user/src/lib/components/chat/McpServerRequirements.svelte
```diff
@@ -0,0 +1,223 @@
...
+		} finally {
+			configuring = false;
+		}
+	}
+</script>
+
+{#each oauthQueue as mcpServer, i (mcpServer.id)}
+	<dialog bind:this={oauthDialogs[i]} class="flex w-full flex-col gap-4 p-4 md:w-sm">
+		<div class="absolute top-2 right-2">
+			<button class="icon-button" onclick={nextOauth}>
```
ivyjeong13, line 185: If you want to force users, wondering if we should remove the ability to close out of the set up process so like the close button here and if in the configure form? But I think for oauths, we did allow the option for them to exit and oauth at a later time (they'd see the alert on the connector in the sidebar and click the connector to see the oauth banner)

pkg/api/handlers/mcp.go
njhale: So something like `TemplateMCPServer` for example?

pkg/storage/apis/obot.obot.ai/v1/thread.go
njhale: That's a fair point, Ill add it back.

ui/user/src/lib/components/chat/McpServerRequirements.svelte
njhale: Yeah we show the unconfigurec status in the sidebar now as well for project mcp servers

pkg/api/authz/template.go
njhale: since we have a unique public ID, this should work without `spec.template` too. Not sure how this snuck in. will drop.

njhale:
@thedadams yeah, I had the same thought. WDYT about a single `Template` resource that contains each of the objects that needs to be copied, instead of having separate `Template*` resources for each one?

Also, FWIW, the reason it isn't this way in the PR currently is to reuse most of the same logic between:
- copying a project thread and its resource to a template thread
- copying a template thread and its resources to a consumer project

pkg/controller/handlers/threads/template.go
njhale: yeah, this was out of order here. this update should come before the status fields are changed and `req.Client.Status().Update(...)` is called.

pkg/api/handlers/projectmcp.go
njhale: So, not sure if this is correct, but it was preventing the fields below from being updated for copied project mcpservers in certain cases. I'll double check to make sure that's the case.

pkg/api/handlers/projects.go
njhale: So the controller needs to "unapprove" the upgrade when a project/snapshot starts upgrading so that it doesn't just upgrade automatically after it finishes when the next update is available.

IIRC, historically, we didn't like controllers changing spec fields that the user set on a resource, which is why I did this as an annotation.

pkg/api/handlers/mcp.go
njhale: For reference, https://github.com/obot-platform/obot/pull/4154#issuecomment-3281351756 explains why I did this with the same objects (to reuse copying logic).

pkg/controller/handlers/threads/threads.go
njhale: Just noticed that there's a side-effect here that causes task threads to get nuked when you upgrade from a template. I would like to follow up with a fix for this if possible.

njhale:
So I stuck with copying existing resource types since it was more straightforward.

@thedadams @ivyjeong13 requesting another round of review to try to get this in

njhale:
hmm, go lint isn't failing locally. Checking it out.

pkg/api/handlers/threads.go
njhale: Done

pkg/storage/apis/obot.obot.ai/v1/thread.go
njhale: Done

pkg/api/authz/template.go
njhale: Removed

pkg/api/handlers/projects.go
njhale: I dropped all the annotations and switched to spec and status fields.

pkg/api/handlers/projects.go
njhale: Agreed. Done.

pkg/api/handlers/template.go
njhale: Switched from annotations to spec/status fields for everything

pkg/controller/handlers/threads/template.go
njhale: Removed

pkg/controller/handlers/threads/template.go
njhale: Code was refactored.

pkg/controller/handlers/threads/template.go
njhale: Refactored all of the hashing logic. It now takes advantage of `hash.Digest` and performs more robust "revision numbers" (project state hashes)

pkg/controller/handlers/threads/threads.go
njhale: Done

pkg/controller/handlers/threads/threads.go
njhale: Fixed

njhale:
> hmm, go lint isn't failing locally. Checking it out.

had a stale local work tree. fixed

pkg/api/handlers/projects.go
```diff
@@ -469,6 +524,13 @@ func convertProject(thread *v1.Thread, parentThread *v1.Thread) types.Project {
 		Editor:                       thread.IsEditor(),
 		UserID:                       thread.Spec.UserID,
 		WorkflowNamesFromIntegration: thread.Status.WorkflowNamesFromIntegration,
+		TemplateUpgradeAvailable:     (thread.Status.UpgradeAvailable && !thread.Spec.UpgradeApproved),
+		TemplateUpgradeInProgress:    thread.Status.UpgradeInProgress,
+		TemplatePublicID:             thread.Status.UpgradePublicID,
+	}
+
+	if !thread.Status.LastUpgraded.Time.IsZero() {
```
thedadams: nit:
```suggestion
	if !thread.Status.LastUpgraded.IsZero() {
```

ivyjeong13 (approved):
https://www.loom.com/share/0f48e1839e0c4940a94c2e3596088266

Think there's one more location that could use the `dontLogErrors` for the reveal endpoint! (Loom above for reference) Also thinking the init configure form data could be turned into a common function since some form of it is being used in a few locations now. But that's more of a note to myself, not PR blocking! : ) Frontend changes LGTM

Frontend changes LGTM

pkg/api/handlers/template.go
```diff
@@ -42,122 +39,85 @@ func (h *TemplateHandler) CreateProjectTemplate(req api.Context) error {
...
+		"spec.sourceThreadName": projectThreadName,
+	}); err != nil {
 		return err
 	}
 
-	return req.Delete(&templateThread)
+	if len(templateThreadList.Items) < 1 {
+		return types.NewErrNotFound("template not found for project %s", projectID)
+	}
+
```
thedadams: nano-nit:
```suggestion
```

pkg/api/handlers/template.go
```diff
@@ -205,118 +164,49 @@ func (h *TemplateHandler) CopyTemplate(req api.Context) error {
...
-	if err := req.List(&templateShareList, kclient.MatchingFields{
-		"spec.template": "true",
-	}); err != nil {
-		return err
-	}
-
-	templateShares := make(map[string]v1.ThreadShare, len(templateShareList.Items))
-	for _, templateShare := range templateShareList.Items {
-		templateShares[templateShare.Spec.ProjectThreadName] = templateShare
+		return types.NewErrNotFound("template not found for project %s", projectID)
```
thedadams: same nano-nit.

pkg/api/handlers/template.go
```diff
@@ -205,118 +164,49 @@ func (h *TemplateHandler) CopyTemplate(req api.Context) error {
...
+		if share.Spec.Template {
+			templateShare = &share
+			break
 		}
-
-		templateList.Items = append(templateList.Items, convertTemplateThread(templateThread, threadShare))
 	}
 
-	return req.Write(templateList)
+	return req.Write(convertTemplateThread(*templateThread, templateShare))
```
thedadams, line 201: Do we need to catch `threadShare == nil` here?

pkg/api/handlers/threads.go
```diff
@@ -33,19 +33,21 @@ func NewThreadHandler(dispatcher *dispatcher.Dispatcher, events *events.Emitter)
...
+		ProjectSnapshotStale:             thread.Status.UpgradeAvailable,
+		ProjectSnapshotUpgradeInProgress: thread.Status.UpgradeInProgress || thread.Spec.UpgradeApproved,
+		AssistantID:                      thread.Spec.AgentName,
+		ProjectID:                        strings.Replace(thread.Spec.SourceThreadName, system.ThreadPrefix, system.ProjectPrefix, 1),
+		Ready:                            thread.Status.Created,
+	}
+
+	// Populate LastUpdated from status field
+	if !thread.Status.LastUpgraded.IsZero() {
+		template.ProjectSnapshotLastUpgraded = *types.NewTime(thread.Status.LastUpgraded.Time)
```
thedadams: Based on the way this is being set, should it be a pointer so that this is omitted when empty?

pkg/storage/apis/obot.obot.ai/v1/thread.go
```diff
@@ -46,6 +46,96 @@ func (in *Thread) Has(field string) (exists bool) {
...
+		deduped = make([]string, 0, len(revisions))
+	)
+	for i := len(revisions) - 1; i >= 0; i-- {
+		revision := revisions[i]
+		if _, ok := visited[revision]; !ok {
+			deduped = append(deduped, revision)
+			visited[revision] = struct{}{}
+		}
+	}
+	slices.Reverse(deduped)
```
thedadams: nit: I think if you go through the revisions in order above, then you don't have to reverse.

pkg/storage/apis/obot.obot.ai/v1/thread.go
```diff
@@ -67,7 +157,7 @@ func (in *Thread) Get(field string) string {
 }
 
 func (in *Thread) FieldNames() []string {
-	return []string{"spec.userUID", "spec.project", "spec.template", "spec.agentName", "spec.parentThreadName", "spec.sourceThreadName"}
+	return []string{"spec.userUID", "spec.project", "spec.template", "spec.parentThreadName", "spec.sourceThreadName"}
```
thedadams: Let's not remove `spec.agentName` here. It is used in some places, even if those places aren't exercised by our current usage.

pkg/storage/apis/obot.obot.ai/v1/thread.go
```diff
@@ -46,6 +46,96 @@ func (in *Thread) Has(field string) (exists bool) {
 	return slices.Contains(in.FieldNames(), field)
 }
 
+// SetLatestConfigRevision sets the latest config revision for the thread and returns true if the
+// operation has changed the revision history.
+func (in *Thread) SetLatestConfigRevision(revision string) (changed bool) {
```
thedadams: nit:
```suggestion
func (in *Thread) SetLatestConfigRevision(revision string) bool {
```

njhale:
@ivyjeong13 Thanks for the review and actually taking the time to run this!

> Think there's one more location that could use the dontLogErrors for the reveal endpoint! 

nice catch. I'll fix that ASAP

pkg/controller/handlers/threads/template.go
```diff
@@ -0,0 +1,383 @@
...
+		mcpServerInstancesSpecs[mcpServerInstance.Name] = spec
+	}
+
+	// Build allowed MCP tools data with normalized MCP IDs
+	config.AllowedMCPTools = make(map[string][]string, len(manifest.AllowedMCPTools))
+	for projectMCPID, toolNames := range manifest.AllowedMCPTools {
+		// Normalize MCP ID to catalog entry name
+		mcpID := projectMCPID
+		if spec, ok := mcpServerSpecs[mcpID]; ok {
+			// TOOD(njhale): Do we need to worry about the catalog this belongs to?
```
thedadams:
```suggestion
```

njhale:
@ivyjeong13 done

pkg/api/handlers/projects.go
njhale: fixed

pkg/api/handlers/template.go
njhale: removed

pkg/api/handlers/template.go
njhale: just think of the extra operations that would have been saved!!! (or not)

removed

pkg/api/handlers/template.go
njhale: The thread shares for templates are created in a controller in the backend, so are not guaranteed to exist immediately. I think this should be fine. Also there are conditional statements in `convertTemplateThread` to protect against nil pointer exceptions.

pkg/api/handlers/threads.go
njhale: Yeah, I actually already did this for the `TemplateLastUpgraded` field on the public API's `Project` type. Changed `ProjectSnapshotLastUpgraded` to match. Thanks.

pkg/storage/apis/obot.obot.ai/v1/thread.go
njhale: Good call. Rewrote this function and added a few table tests.

pkg/storage/apis/obot.obot.ai/v1/thread.go
njhale: un-removed

pkg/storage/apis/obot.obot.ai/v1/thread.go
njhale: Dropped

pkg/controller/handlers/threads/template.go
njhale: fuzzy find failed me.

removed

thedadams (approved):

Issue #4155, opened 2025-09-05
When configuring catalog entry with file param, textarea isn't used


ivyjeong13:
**Describe the bug**
When configuring an mcp catalog entry with a file param, it doesn't change to a textarea like the Value field does for an admin when they configure a multi-user server. 

Ex: 
<img width="1247" height="731" alt="Image" src="https://github.com/user-attachments/assets/c240343b-373e-4594-8a86-95f5f2845715" />

**To Reproduce**
1. Configure single-user catalog entry with a File param
2. See the field is an input instead of a textarea

**Expected behavior**
Should be a similar experience as the admin setting Value for a File type in a multi-user server

**Screenshots**

**Additional context**

sangee2004:
When configuring MCP servers which have parameters of type  - `file` they are presented as text area as expeted

BigQuery Toolbox uses file type for `Google Application Credentials` and `Tools YAML` parameters both odf which show up as text area as expected.

```
env:
  - key: GOOGLE_CLOUD_PROJECT
    name: Google Cloud Project ID
    required: true
    sensitive: false
    description: Your Google Cloud project ID
  - key: GOOGLE_APPLICATION_CREDENTIALS
    name: Google Application Credentials
    required: true
    sensitive: true
    description: Your service account JSON credentials file
    file: true
  - key: BIGQUERY_DATASET_NAME
    name: BigQuery Dataset Name
    required: false
    sensitive: false
    description: BigQuery dataset name to use for the tools to replace ${BIGQUERY_DATASET_NAME} placeholder
  - key: TOOLS_YAML
    name: Tools YAML
    required: false
    sensitive: false
    description: Optional custom tools configuration YAML file to extend available tools
    file: true
```
<img width="728" height="626" alt="Image" src="https://github.com/user-attachments/assets/41435be2-f7c3-494a-88b4-6ac056eec3d5" />

There is issue with contents shown in clear text when marked as sensitive which will be tracked in separate issue.

PR #4164, opened 2025-09-05
fix: catalog entry with a sensitive file configuration textarea

ivyjeong13:
Addresses #4155 

* accounts for if file & sensitive are selected for a catalog entry configuration


ryu-man:
This is what I am getting when setting a file parameter for a single user catalog entry.

image

ivyjeong13:
> This is what I am getting when setting a file parameter for a single user catalog entry.
> 
> image

Yes, that is correct! For a single user catalog entry, the file textarea happens when a user configures the file from connecting in My Connectors. That is what this ticket addresses!

StrongMonkey (approved):

Issue #4173, opened 2025-09-06
Microsoft Entra - Not able to connect to MCP severs from VS code/Claude when authenticating as Microsoft Entra user.


sangee2004:
**Describe the bug**
Microsoft Entra - Not able to connect to MCP severs from VS code/Claude when authenticating as Microsoft Entra user.

Pre Req - obot server that is auth enabled with `Microsoft Entra` auth provider configured.
**To Reproduce**
1. As admin user, Create an ACL rule that allows access to any mcp server say `server1` for a group say `group1`
2. Log in to obot server as a regular user who is a member of  `group1`
3. Connect with `server1` (In my case it was a `DuckduckGo Search` server)
4. Get the connection string for VS code/Claude client
5. Use this connection string to configure MCP server in VS code/Claude client
6. Try to start the MCP server

Authentication work flow succeeds.
After this MCP server gets to Error state

Following errors seen in VS code logs:

``` 
2025-09-05 16:50:10.486 [info] Connection state: Starting
2025-09-05 16:50:10.488 [info] Starting server from LocalProcess extension host
2025-09-05 16:50:10.491 [info] Connection state: Running
2025-09-05 16:50:10.520 [info] Connection state: Error 401 status sending message to http://localhost:8080/mcp-connect/ms12w5rf: unauthorized

2025-09-05 16:50:10.520 [error] Server exited before responding to `initialize` request.
```

I see the following errors in obot logs:

```
2025-09-06 00:05:04.631 UTC [8924] ERROR:  duplicate key value violates unique constraint "uni_users_hashed_username"
2025-09-06 00:05:04.631 UTC [8924] DETAIL:  Key (hashed_username)=(6ef4319419f294baa17c296c08793f807a52f4acc592ca22e3994d07a42445b9) already exists.
2025-09-06 00:05:04.631 UTC [8924] STATEMENT:  INSERT INTO "users" ("created_at","display_name","username","hashed_username","email","hashed_email","verified_email","role","icon_url","timezone","last_active_day","internal","daily_prompt_tokens_limit","daily_completion_tokens_limit","encrypted","deleted_at","original_email","original_username") VALUES ($1,$2,$3,$4,$5,$6,$7,$8,$9,$10,$11,$12,$13,$14,$15,$16,$17,$18) RETURNING "id"
[...]

PR #4174, opened 2025-09-06
fix: add entra to verified auth providers

njhale:
Add `entra-auth-provider` to the set of verified providers.
This fixes MCP token auth when Entra is enabled.

Addresses https://github.com/obot-platform/obot/issues/4173

### DIFF
--- a/pkg/gateway/client/identity.go
+++ b/pkg/gateway/client/identity.go
@@ -20,6 +20,7 @@ var (
 	verifiedAuthProviders = []string{
 		"default/google-auth-provider",
 		"default/github-auth-provider",
+		"default/entra-auth-provider",
 	}
 
 	identityGroupResource = schema.GroupResource{


g-linville:
I'm not sure why this would fix anything. The "verified" auth providers are the auth providers that are known to always validate user email addresses. Last I knew, Entra did not fall into this category, as there is no way to prove that the admin set it up such that each user's email address is an email address they actually own. This is an important distinction, because it determines whether or not we can map multiple Identities to the same User on the basis of a matching email address. Why are you making this change?

njhale:
@g-linville 

> I'm not sure why this would fix anything. 

I should have explained my reasoning more in the PR description, sorry about that.

The `Identity` produced by MCP token auth has a `ProviderUserID` field that is set to the actual Obot user ID [at this point](https://github.com/obot-platform/obot/blob/82488e2aca7c1f700184db91b04de604afc920bd/pkg/gateway/client/identity.go#L115). This means it doesn't match the existing `Identity` in the database (since that has the real ID from the provider) on the first use and creates a new one for the token. The catch is that this doesn't have a `UserID` field set yet, which causes the user creation [logic here](https://github.com/obot-platform/obot/blob/82488e2aca7c1f700184db91b04de604afc920bd/pkg/gateway/client/identity.go#L258) to trigger and fail with `"ERROR: duplicate key value violates unique constraint \"uni_users_hashed_username\"` because a user already exists with the same hashed username.

Adding `entra-auth-provider` to the set of verified providers fixes this by triggering [this logic](https://github.com/obot-platform/obot/blob/82488e2aca7c1f700184db91b04de604afc920bd/pkg/gateway/client/identity.go#L176), causing the user being looked up by email (which is on the passed in identity) instead; avoiding the error and passing auth

> Last I knew, Entra did not fall into this category, as there is no way to prove that the admin set it up such that each user's email address is an email address they actually own.

Ah, yup! You're right. Thank you for catching that.

This isn't the _right_ way to fix the issue. I'm going to close this out and come up with another fix.

PR #4175, opened 2025-09-08
fix: pass obot user id from mcp token auth in user info extra

njhale:
The `Identity` produced by MCP token auth has a `ProviderUserID` field that is set to the actual Obot user ID [at this point](https://github.com/obot-platform/obot/blob/82488e2aca7c1f700184db91b04de604afc920bd/pkg/gateway/client/identity.go#L115). This means it doesn't match the existing `Identity` in the database (since that has the real ID from the provider) on the first use and creates a new one for the token. The catch is that this doesn't have a `UserID` field set yet, which causes the user creation [logic here](https://github.com/obot-platform/obot/blob/82488e2aca7c1f700184db91b04de604afc920bd/pkg/gateway/client/identity.go#L258) to trigger and fail with `"ERROR: duplicate key value violates unique constraint \"uni_users_hashed_username\"` because a user already exists with the same hashed username.

To fix this, pass the Obot user ID from the mcp token authenticators to the user decorator authenticator in the user info's extra map. This makes the user ID available when ensuring the initial token identity.

Addresses https://github.com/obot-platform/obot/issues/4173

Note: I'm still not sure this is the _correct_ solution, but at first glance it does appear to fix the problem.

### DIFF
--- a/pkg/gateway/server/tokenreview.go
+++ b/pkg/gateway/server/tokenreview.go
@@ -58,6 +58,7 @@ func (g *gatewayTokenReview) AuthenticateRequest(req *http.Request) (*authentica
 				"auth_provider_namespace": {namespace},
 				"auth_provider_name":      {name},
 				"auth_provider_groups":    groupIDs,
+				"obot:userID":             {strconv.FormatUint(uint64(u.ID), 10)},
 			},
 		},
 	}, true, nil
--- a/pkg/gateway/client/auth.go
+++ b/pkg/gateway/client/auth.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/http"
 	"slices"
+	"strconv"
 
 	types2 "github.com/obot-platform/obot/apiclient/types"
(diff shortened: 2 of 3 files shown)


pkg/gateway/server/tokenreview.go
```diff
@@ -58,6 +58,7 @@ func (g *gatewayTokenReview) AuthenticateRequest(req *http.Request) (*authentica
 				"auth_provider_namespace": {namespace},
 				"auth_provider_name":      {name},
 				"auth_provider_groups":    groupIDs,
+				"obot:userID":             {strconv.FormatUint(uint64(u.ID), 10)},
```
njhale, line 61: I used `obot:userID` here to match the existing convention for passing the user ID in extra, but I get the feeling the `obot:` prefix may be reserved for "internal usage". Is it OK to use this, or should I change the name?

njhale:
Superseded by https://github.com/obot-platform/obot/pull/4180

Issue #4146, opened 2025-09-04
User changes get applied to the end of the text when trying to edit using WYSIWG editor.


sangee2004:
**Describe the bug**
Edits done to workspace files using WYSIWYG editor mode is not saved.

**To Reproduce**
1. Chat in any project and get a response in markdown mode 
2. Edit the message response 
3. When message is presented in `WYSIWYG` editor mode  and make some changes to the text in the middle of the response.

Notice that the user is taken to the end of the content and changes get applied to the end of the file.

See this happen multiple times after 40 secs into the video

https://www.loom.com/share/dfc7662c0c0943329aadc8c9d73778da?sid=abeb4194-ac8c-4712-85c5-59157ebe4624

sangee2004:
This issue is addressed when testing with latest build.
Able to edit contents in WYSIWG editor successfully.

PR #4177, opened 2025-09-08
fix: when contents change, keep track of user focused cursor position

ivyjeong13:
Addresses #4146 

* root issue: when `contents` changes from `setValue`, it updates the milkdown content but doesn't keep track where the user's original position was so focus gets moved to the bottom -- keep track and reapply cursor position after update

### DIFF
--- a/ui/user/src/lib/components/editor/Milkdown.svelte
+++ b/ui/user/src/lib/components/editor/Milkdown.svelte
@@ -7,6 +7,7 @@
 	import { listener, listenerCtx } from '@milkdown/kit/plugin/listener';
 	import type { InvokeInput } from '$lib/services';
 	import type { EditorState } from '@milkdown/prose/state';
+	import { TextSelection } from '@milkdown/prose/state';
 	import type { EditorView } from '@milkdown/prose/view';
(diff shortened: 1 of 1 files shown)


ui/user/src/lib/components/editor/Milkdown.svelte
```diff
@@ -79,17 +80,37 @@
 			setValue(overrideContent, true);
 		}
 
-		if (!isEditing && !overrideContent && contents) {
+		if (!isEditing && !overrideContent && contents && !focused) {
```
ryu-man, line 83: If I am editing or I am focused in, all the changes from the outside will be discarded?

ivyjeong13: The changes that come through are previous changes made by the user! It's causing issues where if a user is typing sporadically, the changes that come in are previous changes and they lose what they had typed or removed. The only outside changes is if the user is making changes by talking to the AI agent but at that point, they won't be focused or editing directly in the editor.

ryu-man: make sense

PR #4192, opened 2025-09-09
enhance: make MCP URLs global when possible

thedadams:
Instead of each user that connects to an MCP server getting a different URL, this change introduces the concept of a global URL for an MCP server that any user can use, so long as they have configured the MCP server.

If a user has configured an MCP server more than once, then the oldest MCP server will have the "universal" URL, and the remaining servers will have a unique URL as was before this change. If the user deletes the oldest MCP server, then the oldest remaining server will have the universal URL. The unique URL will always work.

If the MCP server has not been configured for a user when they try to connect to it with this global URL, they will get a 404 error and message explaining that the server needs to be configured for them first.


njhale (approved):
lgtm

pkg/api/handlers/mcp.go
```diff
@@ -742,15 +716,137 @@ func (m *MCPHandler) GetPrompt(req api.Context) error {
...
+			var instances v1.MCPServerInstanceList
+			if err := req.List(&instances, &kclient.ListOptions{
+				FieldSelector: fields.SelectorFromSet(map[string]string{
+					"spec.mcpServerName": id,
+					"spec.userID":        req.User.GetUID(),
+				}),
+			}); err != nil {
+				return "", server, mcp.ServerConfig{}, err
+			}
+			if len(instances.Items) == 0 {
```
tybalex, line 742: is it possible for multiple users to trigger this at the same time?

thedadams: Yes, but this is scoped to one user by the field selector. Even if this is triggered multiple times simultaneously by the same user, the result would change. The URL would refer to the first created server.

tybalex: sounds good

tybalex (approved):

sangee2004:
All users get the same connection url for any MCP server like `http://localhost:8080/mcp-connect/default-tavily-search-e8fa7b88` .

When this connection url is used by a user who has not configured the MCP server , a new mcp server will be deployed for this user if the MCP server has no required parameters like DuckDuckGo Search /Ant v MCP server.

When this connection url is used by a user who has not configured the MCP server and the MCP server has some required parameters like Tavily Search/Gmail , user will be presented with following message 

Screenshot 2025-09-10 at 2 00 02 PM

When  this connection url is used by a user who does not have access to the MCP server (not granted access to this MCP server using ACL) , then user is presented with following error message

Screenshot 2025-09-10 at 2 15 09 PM


When there are multiple instances of the same MCP server , then the oldest instance gets the "universal" URL. Other MCP sever instances get connection url with mcpid - `http://localhost:8080/mcp-connect/ms1vqprp` as expected.

Issue #4196, opened 2025-09-09
Need better logging when connecting to remote DBs


cloudnautique:
We need to add logging when the Obot server is attempting to connect to the database in all scenarios. When Obot starts with a local Postgres we log that it's getting configured in the shell script.

When connecting to a remote Postgres DB and the connection is blocked by something like a security group, there are no logs, it the process just appears to hang. We should log when connecting during startup from the Go binary.

sangee2004:
This issue is addressed when testing with latest build.

When obot server is able to connect successfully to an remote DB  configured using `OBOT_SERVER_DSN` env variable , following message is seen in obot logs

```
2025/10/09 22:20:23 INFO Connecting to database dsn=postgres://[REDACTED]@host.docker.internal/obot
time="2025-10-09T22:20:23Z" level=info msg="Generated self-signed cert (apiserver.local.config/certificates/apiserver.crt, apiserver.local.config/certificates/apiserver.key)"
time="2025-10-09T22:20:23Z" level=info msg="setting componentGlobalsRegistry in SetFallback. We recommend calling componentGlobalsRegistry.Set() right after parsing flags to avoid using feature gates before their final values are set by the flags."
```

When obot server fails to connect with a remote DB (due to misconfiguration) , we see the following error messages being logged in obot server:

```
2025/10/09 22:25:07 INFO Connecting to database dsn=postgres://[REDACTED]@localhost/obot
time="2025-10-09T22:25:07Z" level=error msg="failed to initialize database, got error failed to connect to `user=sangee database=obot`:\n\t127.0.0.1:5432 (localhost): dial error: dial tcp 127.0.0.1:5432: connect: connection refused\n\t[::1]:5432 (localhost): dial error: dial tcp [::1]:5432: connect: connection refused\n\t127.0.0.1:5432 (localhost): dial error: dial tcp 127.0.0.1:5432: connect: connection refused\n\t[::1]:5432 (localhost): dial error: dial tcp [::1]:5432: connect: connection refused"
[...]

PR #4198, opened 2025-09-10
Add DB logging and clarify DB requirements

cjellick:
- Add more logging around DB connections
- Clarify DB requirements in the docs


cjellick:
https://github.com/obot-platform/obot/issues/4196

PR #4199, opened 2025-09-10
feat: MCP Publisher app for power users

ivyjeong13:
Addresses #4049 

* sets up routes and app for MCP Publisher for power users & power user plus

Adding reviews to hopefully get some eyes with these changes 👀 

~~TODO: hook up default user role settings~~ done

### DIFF
--- a/ui/user/src/lib/components/Layout.svelte
+++ b/ui/user/src/lib/components/Layout.svelte
@@ -127,7 +127,7 @@
 					{
 						id: 'obot-chat',
 						icon: MessageCircle,
-						label: 'Obot Chat',
+						label: 'Chat Management',
 						disabled: isBootStrapUser,
 						collapsible: true,
 						items: [


pkg/services/config.go
```diff
@@ -386,14 +388,39 @@ func New(ctx context.Context, config Config) (*Services, error) {
...
+			w, err := storageClient.Watch(ctx, &v1.UserDefaultRoleSettingList{}, &kclient.ListOptions{
+				Namespace: system.DefaultNamespace,
+			})
+			if err != nil {
+				slog.Error("failed to watch default role setting", "error", err)
+				time.Sleep(time.Second * 1)
+				continue
+			}
+			defer func() {
+				w.Stop()
```
ivyjeong13: @StrongMonkey Hey Daishan, seeing this lint error and wondering if this is an appropriate alternative?

```
	go func() {
		for {
			w, err := storageClient.Watch(ctx, &v1.UserDefaultRoleSettingList{}, &kclient.ListOptions{
				Namespace: system.DefaultNamespace,
			})
			if err != nil {
				slog.Error("failed to watch default role setting", "error", err)
				time.Sleep(time.Second * 1)
				continue
			}

			for event := range w.ResultChan() {
				if roleSetting, ok := event.Object.(*v1.UserDefaultRoleSetting); ok {
					*defaultRole = roleSetting.Spec.Role
				}
			}

			// moving the defer func() here
			w.Stop()
			//nolint:revive
			for range w.ResultChan() { 
			}
		}
	}()
```

ryu-man:
@ivyjeong13 Here is what I have noticed:

- Categories in `/mcp-publisher?new=*` are not populated.

image

- If I go to `/mcp-publisher` and select new MCP server, clicking on Save does not do anything; no error is seen.

- Accessing `/mcp-publisher/access-control` shows a `404` page

ivyjeong13:
> @ivyjeong13 Here is what I have noticed:
> 
> * Categories in `/mcp-publisher?new=*` are not populated.
> 
> image
> * If I go to `/mcp-publisher` and select new MCP server, clicking on Save does not do anything; no error is seen.
> * Accessing `/mcp-publisher/access-control` shows a `404` page

@ryu-man 
Asked a question to Craig about what categories should show, if we should use the catalog API to get categories or if a workspace specific api endpoint needs to be created.

> * If I go to `/mcp-publisher` and select new MCP server, clicking on Save does not do anything; no error is seen.

For this one, could you be more specific about what steps you took? Just tried creating and saving all three MCP servers types, so wondering if there's an odd edge case.

> * Accessing `/mcp-publisher/access-control` shows a `404` page

I'm not seeing this, if it's a 404, I'm wondering if by any chance, can you refetch and see if you get new commits? I did see a forbidden when I directly accessed the route (as a Power User Plus) from the browser URL so looking into that.

ryu-man:
> I'm not seeing this, if it's a 404, I'm wondering if by any chance, can you refetch and see if you get new commits? I did see a forbidden when I directly accessed the route (as a Power User Plus) from the browser URL so looking into that.

https://www.loom.com/share/b86dbde853074634a6236cdcbb9edf73?sid=da9f0eed-b9ce-4bd6-a6f1-0ca1e69f00c3

ryu-man:
> For this one, could you be more specific about what steps you took? Just tried creating and saving all three MCP servers types, so wondering if there's an odd edge case.

https://www.loom.com/share/1a99ce2fd582403db79cbe344b93776f?sid=e3cd9111-0fda-4a4d-891d-73a733c40ca1

pkg/services/config.go
```diff
@@ -355,9 +357,29 @@ func New(ctx context.Context, config Config) (*Services, error) {
...
+	} else if err != nil {
+		return nil, err
+	}
+
+	defaultRole := &defaultRoleSetting.Spec.Role
+
 	// Create callback for new privileged user workspace creation
 	onNewPrivilegedUser := func(ctx context.Context, user *types.User) {
-		// Create a UserRoleChange event to trigger PowerUserWorkspace creation
+		fmt.Println("onNewPrivilegedUser", *defaultRole)
```
njhale: Should drop this print statement or add more info to the log and switch over to `log.Debug/Info` if we're actually interested with this showing up in the logs.

pkg/services/config.go
```diff
@@ -366,14 +388,39 @@ func New(ctx context.Context, config Config) (*Services, error) {
 			Spec: v1.UserRoleChangeSpec{
 				UserID:  user.ID,
 				OldRole: apiclienttypes.RoleBasic, // New users start as basic
-				NewRole: user.Role,
+				NewRole: *defaultRole,
```
njhale, line 394: As it stands, since the `defaultRole` pointer can be written to asynchronously by the goroutine below, we'll need to synchronize access to it. You can do this by creating an [`RWMutex`](https://pkg.go.dev/sync#RWMutex) shared by this anonymous function and the goroutine.

e.g.

```go
var defaultRoleLock sync.RWMutex
defaultRole := &defaultRoleSetting.Spec.Role

onNewPrivilegedUser := func(ctx context.Context, user *types.User) {
		fmt.Println("onNewPrivilegedUser", *defaultRole)
        defaultRoleLock.RLock()
        defer defaultRoleLock.RUnlock()
        // Dereference `defaultRole` and create UserRoleChange...
        // Note: You should also check if `defaultRole` is `nil` before attempting to dereference it.
}

// ...
go func() {
		for {
			// ...

			for event := range w.ResultChan() {
				if roleSetting, ok := event.Object.(*v1.UserDefaultRoleSetting); 
ok {
                    defaultRoleLock.Lock() // Write lock
					*defaultRole = roleSetting.Spec.Role
                    defaultRoleLock.Unlock() // Unlock after operation instead of deferring because defer is only called when the function the call is scoped to returns, which won't happen since this runs forever
				}
			}
		}
	}()
```

Of course, you'll also need to pass the same lock down to `gatewayClient` if it's being read there as well, so it may make more sense to just encapsulate this all into a single type that manages thread-safe access to the `defaultRole`.

**It may also just be cleaner and more robust to make a get call for the `UserDefaultRoleSetting` when it's needed instead of managing state across multiple components like this.**

If we're dead-set on not making the extra API calls, I think it would be better to write a controller handler to update the cached `defaultRole` instead of trying to instantiate and manage a watch here.

pkg/services/config.go
```diff
@@ -355,9 +357,29 @@ func New(ctx context.Context, config Config) (*Services, error) {
...
+			Spec: v1.UserDefaultRoleSettingSpec{
+				Role: apiclienttypes.RoleBasic,
+			},
+		}
+		if err := storageClient.Create(ctx, &defaultRoleSetting); err != nil {
+			return nil, err
+		}
+	} else if err != nil {
+		return nil, err
+	}
```
njhale, line 377: I think it may make more sense to do this as part of the controller's `PreStart` hook in `pkg/controller/controller.go` if possible.

pkg/services/config.go
```diff
@@ -366,14 +388,39 @@ func New(ctx context.Context, config Config) (*Services, error) {
...
+				}
+			}()
+
+			for event := range w.ResultChan() {
+				if roleSetting, ok := event.Object.(*v1.UserDefaultRoleSetting); ok {
+					*defaultRole = roleSetting.Spec.Role
+				}
+			}
+		}
+	}()
```
njhale, line 425: See https://github.com/obot-platform/obot/pull/4199/files#r2349884650 and https://github.com/obot-platform/obot/pull/4199/files#r2349887662

pkg/api/handlers/mcp.go
```diff
@@ -145,6 +149,8 @@ func convertMCPServerCatalogEntry(entry v1.MCPServerCatalogEntry) types.MCPServe
 		UserCount:                 entry.Status.UserCount,
 		LastUpdated:               v1.NewTime(entry.Status.LastUpdated),
 		ToolPreviewsLastGenerated: v1.NewTime(entry.Status.ToolPreviewsLastGenerated),
+		PowerUserWorkspaceID:      powerUserWorkspaceID,
```
StrongMonkey, line 152: @ivyjeong13 This look good to me

ivyjeong13: Thanks Daishan!

ivyjeong13:
> @ivyjeong13 Here is what I have noticed:
> 
> * Categories in `/mcp-publisher?new=*` are not populated.
> 
> image
> * If I go to `/mcp-publisher` and select new MCP server, clicking on Save does not do anything; no error is seen.
> * Accessing `/mcp-publisher/access-control` shows a `404` page

Synced with Abdel and these issues no longer happening after re-running make dev after switching to branch.

pkg/services/config.go
StrongMonkey: yeah added for debugging, fixed.

pkg/services/config.go
StrongMonkey: Good call, I missed that. Added lock.

pkg/services/config.go
StrongMonkey: @ivyjeong13 yeah this is my bad, should be fixed

pkg/gateway/client/identity.go
```diff
@@ -213,12 +213,18 @@ func (c *Client) ensureIdentity(ctx context.Context, tx *gorm.DB, id *types.Iden
...
+			if user.Role == types2.RoleUnknown {
 				user.Role = role
 				userChanged = true
 			}
 
+			if c.IsExplicitAdmin(user.Email) && user.Role != types2.RoleAdmin {
+				user.Role = types2.RoleAdmin
+				userChanged = true
+			}
+
```
thedadams, line 229: This is probably fine, but I think it is already handled on line 50 above.

StrongMonkey: It does ensure an edge case where if an explicit admin is changed to standard user, this will ensure that it will be reverted back to admin.

pkg/api/authz/authz.go
```diff
@@ -104,6 +104,44 @@ var staticRules = map[string][]string{
 		"GET /api/all-mcps/servers",
 		"GET /api/all-mcps/servers/{mcp_server_id}",
 	},
+	PowerUserGroup: {
+		// Workspace-scoped MCP Server Catalog Entries
+		"GET /api/workspaces/{workspace_id}/entries",
```
njhale, line 109: IIUC, this will allow _ANY_ user with the `PowerUserGroup` to access resources in _ALL_ workspaces, not just the ones they own.

I had thought authz for this was handled by https://github.com/obot-platform/obot/blob/05ae9b4e7c556c74555e0c98fd74ba6f0bf312c5/pkg/api/authz/resources.go#L304

pkg/api/handlers/accesscontrolrules.go
```diff
@@ -337,6 +327,7 @@ func convertAccessControlRule(rule v1.AccessControlRule) types.AccessControlRule
 		Metadata:                  MetadataFrom(&rule),
 		MCPCatalogID:              rule.Spec.MCPCatalogID,
 		PowerUserWorkspaceID:      rule.Spec.PowerUserWorkspaceID,
+		PowerUserID:               "", // Not available in this context
```
StrongMonkey: you can just remove it if it is empty

pkg/api/handlers/poweruserworkspace.go
```diff
@@ -56,6 +62,119 @@ func (*PowerUserWorkspaceHandler) Get(req api.Context) error {
...
+			servers = append(servers, convertMCPServer(server, credMap[server.Name], p.serverURL, slug))
+		}
+	}
+
+	return req.Write(types.MCPServerList{
+		Items: servers,
+	})
+}
+
+func (p *PowerUserWorkspaceHandler) ListAllAccessControlRules(req api.Context) error {
```
StrongMonkey, line 143: @ivyjeong13 👍 looks good!

cjellick (approved):

pkg/api/authz/authz.go
```diff
@@ -104,6 +104,44 @@ var staticRules = map[string][]string{
...
+		"GET /api/workspaces/{workspace_id}/servers/{mcp_server_id}/instances",
+	},
+
+	PowerUserPlusGroup: {
+		// Access Control Rules
+		"GET /api/workspaces/{workspace_id}/access-control-rules",
+		"GET /api/workspaces/{workspace_id}/access-control-rules/{access_control_rule_id}",
+		"POST /api/workspaces/{workspace_id}/access-control-rules",
+		"PUT /api/workspaces/{workspace_id}/access-control-rules/{access_control_rule_id}",
+		"DELETE /api/workspaces/{workspace_id}/access-control-rules/{access_control_rule_id}",
```
thedadams, line 139: We should discuss what's necessary here.

pkg/api/handlers/poweruserworkspace.go
```diff
@@ -56,6 +62,119 @@ func (*PowerUserWorkspaceHandler) Get(req api.Context) error {
 	return req.Write(convertPowerUserWorkspace(workspace))
 }
 
+func (*PowerUserWorkspaceHandler) ListAllEntries(req api.Context) error {
+	var list v1.PowerUserWorkspaceList
+	if err := req.List(&list); err != nil {
+		return fmt.Errorf("failed to list power user workspaces: %w", err)
+	}
+
+	catalogEntries := make([]types.MCPServerCatalogEntry, 0)
```
thedadams, line 71: nit:
```suggestion
	catalogEntries := make([]types.MCPServerCatalogEntry, 0, len(list.Items))
```

pkg/api/handlers/poweruserworkspace.go
```diff
@@ -56,6 +62,119 @@ func (*PowerUserWorkspaceHandler) Get(req api.Context) error {
...
+	})
+}
+
+func (p *PowerUserWorkspaceHandler) ListAllServers(req api.Context) error {
+	var list v1.PowerUserWorkspaceList
+	if err := req.List(&list); err != nil {
+		return fmt.Errorf("failed to list power user workspaces: %w", err)
+	}
+
+	servers := make([]types.MCPServer, 0)
```
thedadams, line 94: nit:
```suggestion
	servers := make([]types.MCPServer, 0, len(list.Items))
```

pkg/api/handlers/poweruserworkspace.go
```diff
@@ -56,6 +62,119 @@ func (*PowerUserWorkspaceHandler) Get(req api.Context) error {
...
+	})
+}
+
+func (p *PowerUserWorkspaceHandler) ListAllAccessControlRules(req api.Context) error {
+	var list v1.PowerUserWorkspaceList
+	if err := req.List(&list); err != nil {
+		return fmt.Errorf("failed to list power user workspaces: %w", err)
+	}
+
+	accessControlRules := make([]types.AccessControlRule, 0)
```
thedadams, line 149: nit:
```suggestion
	accessControlRules := make([]types.AccessControlRule, 0, len(list.Items))
```

pkg/api/router/router.go
```diff
@@ -488,7 +493,7 @@ func Router(services *services.Services) (http.Handler, error) {
 	mux.HandleFunc("POST /api/workspaces/{workspace_id}/entries/{entry_id}/generate-tool-previews", mcpCatalogs.GenerateToolPreviews)
 	mux.HandleFunc("POST /api/workspaces/{workspace_id}/entries/{entry_id}/generate-tool-previews/oauth-url", mcpCatalogs.GenerateToolPreviewsOAuthURL)
 
-	// Workspace-scoped MCP Servers (PowerUserPlus and higher only)
+	// Workspace-scoped MCP Servers (PowerUser and higher only)
```
thedadams, line 496: Power users should not be able to create servers here.

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
```diff
@@ -234,8 +234,8 @@ func (h *Handler) createDefaultAccessControlRule(ctx context.Context, client kcl
 				DisplayName: "Default Access Rule",
 				Subjects: []types.Subject{
 					{
-						Type: types.SubjectTypeUser,
-						ID:   workspace.Spec.UserID,
+						Type: types.SubjectTypeSelector,
+						ID:   "*",
```
thedadams, line 238: The original intent was that this gives a user access to their workspace. This change gives all users access to every workspace that is created.

pkg/gateway/client/client.go
```diff
@@ -21,9 +22,11 @@ type Client struct {
 	kickAuditPersist chan struct{}
 	// Callback function called when new privileged users are created
 	onNewPrivilegedUser func(ctx context.Context, user *types.User)
+	defaultRole         *types2.Role
+	lock                *sync.RWMutex
 }
 
-func New(ctx context.Context, db *db.DB, encryptionConfig *encryptionconfig.EncryptionConfiguration, adminEmails []string, auditLogPersistenceInterval time.Duration, auditLogBatchSize int, onNewPrivilegedUser func(ctx context.Context, user *types.User)) *Client {
+func New(ctx context.Context, db *db.DB, encryptionConfig *encryptionconfig.EncryptionConfiguration, adminEmails []string, auditLogPersistenceInterval time.Duration, auditLogBatchSize int, onNewPrivilegedUser func(ctx context.Context, user *types.User), defaultRole *types2.Role, lock *sync.RWMutex) *Client {
```
thedadams, line 29: We should use a storage client instead of trying to manage this in memory.

pkg/gateway/client/identity.go
```diff
@@ -210,12 +215,18 @@ func (c *Client) ensureIdentity(ctx context.Context, tx *gorm.DB, id *types.Iden
...
-			if role != types2.RoleUnknown && user.Role != role {
+			if user.Role == types2.RoleUnknown {
 				user.Role = role
 				userChanged = true
 			}
 
+			if c.IsExplicitAdmin(user.Email) && user.Role != types2.RoleAdmin {
+				user.Role = types2.RoleAdmin
+				userChanged = true
+			}
```
thedadams, line 228: I still think this should be removed.

pkg/services/config.go
```diff
@@ -366,14 +391,40 @@ func New(ctx context.Context, config Config) (*Services, error) {
...
+					lock.Unlock()
+				}
+			}
+
+			w.Stop()
+			//nolint:revive
+			for range w.ResultChan() {
+			}
+		}
+	}()
```
thedadams, line 425: If this was something that needed to happen, then we should be using a controller. However, I think that should be reworked.

thedadams:
Did a post-merge review. I have submitted PRs to handle everything except for the permissions stuff. Just need to figure out what the requirements are and go from there.

Issue #4149, opened 2025-09-04
`Error: MPC -32603: failed to send logging/setLevel message to server ` error seen when connecting to MCP server from VS code.


sangee-girish:
**Describe the bug**
`Error: MPC -32603: failed to send logging/setLevel message to server ms1bgqmr:` error seen when connecting to MCP server from VS code.

**To Reproduce**
1. Connect to any MCP server (Tavily Search in my case)
2. Get connection string for VS code
3. From VS code, add this MCP server and start it.

Notice that  errors relating to `Error: MPC -32603: failed to send logging/setLevel message to server ` is seen in logs

```
2025-09-04 15:21:13.527 [info] Connection state: Starting
2025-09-04 15:21:13.531 [info] Starting server from LocalProcess extension host
2025-09-04 15:21:13.538 [info] Connection state: Running
2025-09-04 15:21:17.412 [error] Failed to set MCP server log level: Error: MPC -32603: failed to send logging/setLevel message to server ms1bgqmr: error from server: JSON RPC method not found: logging/setLevel
2025-09-04 15:21:17.837 [info] Discovered 4 tools
2025-09-04 15:26:19.874 [info] Error reading from async stream, we will reconnect: TypeError: terminated
```

Audit logs also show this error

<img width="1286" height="288" alt="Image" src="https://github.com/user-attachments/assets/30296861-6585-43d0-8071-b7e457ef2c08" />

**Note** -  This error does not prevent me from making successful tool calls.

**Expected behavior**
There should be no errors when connecting to MCP server

cjellick:
some findings: it is nanobot declaring the logging capability (not the tavily server) but then nanobot doesnt implement the setLevel call.

now, the MCP spec doesnt say the server has to implement that call, but it does say a client MAY call setLevel. To me that implies that If the server says it supports a capability, then it SHOULD (and maybe even must?) implement the setLevel call.

cjellick:
discussion - https://acorn-io.slack.com/archives/C03FU91U4SX/p1757101069959619

takeaway is that nanobot is doing the wrong thing. it should support this method if its going to decalre it, but donnie also noted that the latest version of nanobot commented this out.

sangee2004:
@thedadams I am still able to reproduce this issue when testing with latest build - https://github.com/obot-platform/obot/commit/7ccaa2dc

MCP server logs (from VS code) when connecting from VS code:
```
25-09-10 16:32:27.034 [info] Starting server Tavily Search
[...]

PR #4204, opened 2025-09-10
chore: bump nanobot to pickup request ID changes

thedadams:
There were some interface changes in nanobot, so I had to adjust to those.


cjellick (approved):

tybalex (approved):

thedadams:
This change also inadvertently fixes https://github.com/obot-platform/obot/issues/4149 because we are updating nanobot.

Issue #4102, opened 2025-08-29
Stop failing in chat if unable to list tools from an MCP server


thedadams:
When we are not able to list tools from an MCP server when doing chat, the message fails without going to the LLM. Instead, we should return some text to the caller indicating that we weren't able to list tools from specific MCP servers.

thedadams:
There are two ways of testing this:
- Chat with an MCP server that needs OAuth, but don't go through the OAuth process
- Purposely misconfigure an MCP server (I always use Tavily without providing an API key) and try to chat with it.

For any message that you send, you should see a little warning symbol. There is a message about the error when you hover over it.

Note that these symbols/messages are temporary. They will go away anytime the chat is reloaded (like refreshing the page or navigating to a different thread).

sangee2004:
Tested with latest build.

Able to chat successfully in projects that have MCP servers that are failing to list tools. User is also notified about these MCP servers that having problems listing tools.

 
Have a project configured with following MCP servers
1. DuckDuckGo Search
2. Outlook with no credentials
3. Tavily Search MCP server which was created without the required TAVILY_API_KEY configuration.

Chat in this project and ask to retrieve current information which results in tool calls made to DuckDuckGo Search. 

Tool calls to DuckDuckGo Search succeed as expected. 
User is presented with warning symbol in the chat message. When hovering over it , we are presented with error messages relating to tool call failures for Outlook and Tavily Search MCP server.



<img width="1065" height="665" alt="Image" src="https://github.com/user-attachments/assets/9c6e2937-e9c4-41ae-ae4b-8c14343d5a49" />

PR #4221, opened 2025-09-11
enhance: continue chat if listing tools from MCP servers fail

thedadams:
If listing tools fails for some reason, then the invoke API will return a message explaining which servers had issues and why. The chat and events streams will continue as normal.

An example when OAuth is not setup for an MCP server:
Screenshot 2025-09-11 at 18 18 34

Issue: https://github.com/obot-platform/obot/issues/4102


thedadams:
The last commit changes the way this is displayed.
Screenshot 2025-09-12 at 09 51 33

ivyjeong13:
> The last commit changes the way this is displayed. Screenshot 2025-09-12 at 09 51 33

Oo yeah I like that better!

PR #4250, opened 2025-09-15
Fix: Update apk before adding chromium

cjellick:
Ran into an issue building locally. This resolved it.


tools/package-chrome.sh
```diff
@@ -1,4 +1,5 @@
 set -e
 mkdir -p /opt/google/chrome
+apk update
 apk add chromium
```
drpebcak: To be optimal, I think this should just be 
```suggestion
apk add --no-cache chromium
```

Issue #4183, opened 2025-09-08
Audit logs - When user name is long , User name field overlaps with Server field .


sangee2004:
**Describe the bug**
Audit logs - When user name is long , User name field overlaps with Server field .

**To Reproduce**
1. Log in to obot server as a regular user with very long display name > 30 character (first and last name for a gmail account in my case)
2. Connect with any MCP server and chat with it to make some tool calls.
3. As admin user , go to `Audit Logs`

Notice that the  User name field overlaps with Server field .

<img width="1043" height="326" alt="Image" src="https://github.com/user-attachments/assets/93ee802d-7d9b-4c6f-9740-f26bd182ef10" />

**Expected behavior**
There should be no overlap of user name and server field values.

sangee2004:
This issue is addressed when testing with latest build.

All columns in the `Audit Logs` view are resizable and there is no overlapping of fields anymore.

<img width="939" height="204" alt="Image" src="https://github.com/user-attachments/assets/f77f042a-64ff-4908-b275-e547d8cc5957" />

PR #4256, opened 2025-09-16
feat: make table cells resizable

ryu-man:
Addresses #4183

- Make table cells resizable for better UX.
- Cell content will be truncated by default.
- Each table cell has a min width.

image

### DIFF
--- a/ui/user/src/lib/components/admin/audit-logs/AuditLogs.svelte
+++ b/ui/user/src/lib/components/admin/audit-logs/AuditLogs.svelte
@@ -146,11 +146,11 @@
 
 						{@render th('Identifier', { class: 'w-[24ch]', minWidth: '24ch' })}
 
-						{@render th('Response Code', { class: 'w-[20ch]', minWidth: '14ch' })}
+						{@render th('Response Code', { class: 'w-[22ch]', minWidth: '22ch' })}
 
-						{@render th('Response Time (ms)', { class: 'w-[24ch]', minWidth: '24ch' })}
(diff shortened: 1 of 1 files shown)


ui/user/src/lib/components/admin/audit-logs/AuditLogs.svelte
```diff
@@ -1,70 +1,147 @@
...
+			return () => {
+				node.removeEventListener('pointerdown', pointerDownHandler);
+			};
+		}}
+	>
+		<GripVertical class="w-3" />
+	</button>
+{/snippet}
+
+{#snippet tdResizeHandler()}
```
ivyjeong13, line 64: https://www.loom.com/share/4e79a8e0bacb45a5944d3d99f65cb428

Just one lil bug I noticed with resizing columns near the end of the table!

ryu-man: Hey Ivy, Does the horizontal scrollbar stop working when the resize handler goes off-table, or are you suggesting to keep the resize handle visible during resize?

ryu-man: Added a small change
https://www.loom.com/share/1f991e8d39ca4814a668c64d5aa27683?sid=0f98d6ba-4e03-43e6-b5bc-04a8a3aae430

ivyjeong13: Sorry didn't see your last comment till just now but that change looks good to me! :+1:

PR #4266, opened 2025-09-17
fix: adjust step calculation for minute frame based on duration

ryu-man:


ui/user/src/lib/components/admin/audit-logs/AuditLogsTimeline.svelte
```diff
@@ -297,11 +297,17 @@
...
+			if (duration < 30) {
+				step = 1 * ticksRatio;
+			} else if (duration < 60) {
+				step = 2 * ticksRatio;
+			} else {
 				step = 5 * ticksRatio;
 			}
 		}
 
+		console.log(frame);
```
ivyjeong13: nit: remove debug console.log

Issue #4286, opened 2025-09-17
Project Sharing  - Do not show mcp server id when prompting for configuration parameters during project copy.


sangee2004:
**Describe the bug**
Project Sharing  - Do not show mcp server id when promoting for configuration parameters during project copy.

**To Reproduce**
1. As user1, Create a project with MCP servers `Tavily Search` (in my case)
2. Enable project sharing
3. As user2 , use project sharing link to create project.

Notice that  mcp server id is presented  when prompting for configuration parameters during project copy.
<img width="739" height="360" alt="Image" src="https://github.com/user-attachments/assets/8fc3eba8-77ad-49f7-94d3-2e4b46075563" />

**Expected behavior**
Do not show mcp server id when prompting for configuration parameters during project copy.

njhale:
From what I can tell, this UX predated project sharing. Even so, I have a PR up that removes the server ID from dialog title.

sangee2004:
This issue is addressed when testing with latest build

User is only presented with MCP server name as expected
<img width="845" height="321" alt="Image" src="https://github.com/user-attachments/assets/0a0e65c1-bba8-42cd-b71f-cb52926621e5" />

Issue #4275, opened 2025-09-17
Project Sharing - Do not show `Shared Project Info` details when parent project gets deleted.


sangee2004:
**Describe the bug**
Project Sharing - Do not show `Shared Project Info` details when parent project gets deleted.

**To Reproduce**
1. As user1, Create a project with MCP servers.
2. Enable project sharing
3. As user2 , use project sharing link to create project.
4. As user1,  delete the project created in step1
5. As user2 , access the project configuration page for the project created in step2

<img width="1098" height="276" alt="Image" src="https://github.com/user-attachments/assets/fa71be21-0b52-424c-b7d5-c58daf7e3207" />

**Expected behavior**
Should we not want to show `Shared Project Info` details all together at all or the better thing to do here may be to indicate that the parent project is more available ?

cjellick:
We should probably still have this indicator but change some of the text to clearly state the upstream project was deleted.

sangee2004:
Following message is presented to user when the parent project template is deleted.

<img width="1022" height="378" alt="Image" src="https://github.com/user-attachments/assets/c7af0e2f-d995-428a-a722-a7f378882c45" />

Issue #4273, opened 2025-09-17
Project Sharing - Provide Upgrade available indication for project in project list view (or in `project configuration` ) when new template version is available.


sangee2004:
**Describe the bug**
Project Sharing - Provide Upgrade available indication for project in project list view (or in `project configuration` ) when new template version is available.

**To Reproduce**
**To Reproduce**
1. As user1, Create a project with few MCP servers.
2. Enable project sharing
3. As user2 , use project sharing link to create project.
4. As user1, add another mcp server to this project,
5. As user2 , access project created in step 3.
Notice that there is no indication in project panel relating to Upgrade being available for this project.

<img width="1356" height="843" alt="Image" src="https://github.com/user-attachments/assets/f8c36611-600c-4d58-a536-20354bbd13df" />

User need to go to `Project configuration` page to see the `Upgrade` option 

<img width="1357" height="844" alt="Image" src="https://github.com/user-attachments/assets/77c82f7d-9a83-4eab-8283-2496f6539b39" />

**Expected behavior**
Provide Upgrade available indication for project in project list view (or in `project configuration` ) when new template version is available.

sangee2004:
`Upgrade Available` indication is now presented in the `Project Configuration` option in the project panel.

<img width="1284" height="775" alt="Image" src="https://github.com/user-attachments/assets/cdbb16b7-1fce-42b1-a215-ed278e46582b" />

njhale:
@ivyjeong13 is going to make some changes to this soon.

@cjellick should we just reopen this, or create a new issue

cjellick:
Just leave this issue closed. Ivy will handle what she needs to handle.

cjellick:
Just leave this issue closed. Ivy will handle what she needs to handle.

Issue #4272, opened 2025-09-17
Project Sharing - After user upgrades project with latest template version, project view does not reflect the changes made to the project unless refreshed.


sangee2004:
**Describe the bug**
Project Sharing - After user upgrades project with latest template version, project view does not reflect the changes made to the project unless refreshed manually.

**To Reproduce**
1. As user1, Create a project with few MCP servers.
2. Enable project sharing
3. As user2 , use project sharing link to create project.
4. As user1, add another mcp server to this project,
5. As user2 , access project created in step 3.
6. Use Upgrade option to from `Project configuration` page.

<img width="893" height="493" alt="Image" src="https://github.com/user-attachments/assets/35bb18af-6d73-42cb-9c3d-4681006fb8c6" />

7. After this notice that the new MCP server that got added to the project  is not being presented in the `Connectors` list

Manual refresh of the project page results in new MCP sever being shown in `Connectors` list

**Expected behavior**
Should be able to show all the changes fetched from the upgrade (including new tasks in Project panel and Knowledge files in `Project Configuration` page right after update is done ) without having to reload the page.

sangee2004:
This issue is addressed when testing with latest builld.

PR #4292, opened 2025-09-18
Fix project share UI reactiveness

njhale:
Contains several fixes for project share UI components:

### fix: remove edit mode component

The `EditMode` component periodically updates the focused project if it's changed.
Upgrading a project from a share will also result in the project being
updated out-of-band in the backend. This can sometimes cause the update
issued by `EditMode` to fail with a conflict and result in an error
message being presented to the user.

Since we've already switched over to a manual `Update` button for
changing project configuration, we can drop this component and prevent
the failure from occurring.

### fix: show changes after successful copied project upgrade

After a successful upgrade of a copied project, refresh its config form data,
MCP servers, tasks, and knowledge files so that changes are immediately
reflected in the chat sidebar and project config.

Addresses https://github.com/obot-platform/obot/issues/4272

### enhance: add upgrade available badge to project settings gear

Display an icon above the project settings gear for copied projects when
it's eligible for upgrade to a new version of the project it was copied
from.

Additionally, to detect changes to eligibility without user interaction,
poll the project in the `ChatSidebar` component at a 5 minute
interval.

Loom: https://www.loom.com/share/4c27a47eef7f432d837545262a4e8363?sid=4f466202-4aeb-4a54-8051-22f63e14e0b6

Addresses https://github.com/obot-platform/obot/issues/4273

### fix: notify user when upstream shared project is deleted

Also fix styling of "Project Share Info" text and copy button.

e.g. New styling/upgrade/copy button

Screenshot 2025-09-18 at 5 23 05 AM

e.g. New UX when the upstream project is deleted

Screenshot 2025-09-18 at 5 23 38 AM

Addresses https://github.com/obot-platform/obot/issues/4275

### fix: prevent requeue on not found

Prevent controller requeues when a source project is not found.

### fix: remove server id from mcp config dialog

Addresses https://github.com/obot-platform/obot/issues/4286


ui/user/src/lib/components/chat/ChatSidebar.svelte
```diff
@@ -26,6 +28,47 @@
...
+		}
+
+		// Restart polling when the component is mounted or the project is updated elsewhere.
+		// The main use case for polling here is to keep the upgradeAvailable state fresh
+		// and inform the user of available upgrades when they haven't done anything that
+		// caused the project to be fetched for a while.
+		pollProject();
+	});
+
+	onDestroy(() => {
```
ryu-man, line 68: Looks good, but you can directly clear the interval in the `$effect` rune:
```
	$effect(() => {
		// ... code
		
		// return a clearnup function; runs on dep change and when component is destroyed
		return ()=> {
		  clearInterval(projectPollInterval);
		}
	});
```

ryu-man (approved):
UI looks good.

ui/user/src/lib/components/chat/ChatSidebar.svelte
njhale: Oh that's a neat feature. TIL.

PR #4312, opened 2025-09-19
Default user role

thedadams:


cjellick (approved):

pkg/gateway/client/identity.go
```diff
@@ -164,9 +168,11 @@ func (c *Client) ensureIdentity(ctx context.Context, tx *gorm.DB, id *types.Iden
 	}
 
 	if user.Role == types2.RoleUnknown {
-		c.lock.RLock()
-		role = *c.defaultRole
-		c.lock.RUnlock()
+		var err error
+		user.Role, err = c.getDefaultRole(ctx)
```
StrongMonkey, line 172: The reason I did pointer and watch thing is that I don't want to call the api to get role every time ensureEntity is called.

thedadams: Move this so that it is only called when a new user is created.

PR #4313, opened 2025-09-19
Fix: fix permission related issue for power users

StrongMonkey:
This PR reverts the changes to allow power user and plus to have more route access, as it should be granted based on workspace and permission in another handler. Also change the UI to use a differente endpoint to list servers for the current users.

### DIFF
--- a/ui/user/src/lib/context/poweruserWorkspace.svelte.ts
+++ b/ui/user/src/lib/context/poweruserWorkspace.svelte.ts
@@ -36,7 +36,7 @@ export async function fetchMcpServerAndEntries(
 	const context = mcpServerAndEntries || getPoweruserWorkspace();
 	context.loading = true;
 	const entries = await ChatService.listWorkspaceMCPCatalogEntries(workspaceID);
-	const servers = await ChatService.listWorkspaceMCPCatalogServers(workspaceID);
+	const servers = await ChatService.listMCPCatalogServers();
 	context.entries = entries;
 	context.servers = servers;
 	context.loading = false;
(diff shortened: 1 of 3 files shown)


ui/user/src/lib/context/poweruserWorkspace.svelte.ts
```diff
@@ -36,7 +36,7 @@ export async function fetchMcpServerAndEntries(
 	const context = mcpServerAndEntries || getPoweruserWorkspace();
 	context.loading = true;
 	const entries = await ChatService.listWorkspaceMCPCatalogEntries(workspaceID);
-	const servers = await ChatService.listWorkspaceMCPCatalogServers(workspaceID);
+	const servers = await ChatService.listMCPCatalogServers();
```
ivyjeong13: This will make it so that the power user will receive multi-user servers created from the global catalog (and may look like as if they created and can modify it without further UI changes). Just wanted to make sure this is intended? edit: I've been major derping and saying remote but I meant multi-user!

ivyjeong13: Screenshot 2025-09-19 at 2 13 40 PM

Was able to confirm with:
1. Logged in as just a Power User. Went to MCP Publisher and created a Multi-User server called Power User Everything.
2. Logged out and logged in as a Admin.
3. Went to Admin Dashboard and created a Multi-User server called Global Everything.
4. Went to MCP Publisher and created a Multi-User server called Admin Power User Everything.
5. Logged out and logged back in as the regular Power User. Went to MCP Publisher. Expected to just see Power User Everything but I also see the other two (one from someone else's workspace and one created in the mcp-catalog).

StrongMonkey: discussed offline, ivy will address the changes

ivyjeong13: :+1: pushed up the change!

ivyjeong13 (approved):

Issue #4342, opened 2025-09-21
Connectors list becomes  empty after upgrading project.


sangee2004:
**Describe the bug**
Connector list becomes  empty after upgrading project.

**To Reproduce**
1. As user1, Create a project with `Tavily Search`. Enable Project sharing.
2. As another user , create project using this share link.
3. As user1, Add `DuckDuckgo Search`. Update project share.
4. As same user in step1 , update project from `Project Configuration` page.

Notice that after upgrade , `Connectors list` becomes empty.

<img width="1397" height="847" alt="Image" src="https://github.com/user-attachments/assets/aaa62f4d-2be6-49d5-88cb-e4950274b9ac" />

Manual refresh of the page , brings back both the connectors.

<img width="1320" height="755" alt="Image" src="https://github.com/user-attachments/assets/d82db677-1494-4030-99c0-490f1b8cefbf" />

**Expected behavior**
Connectors lists should reflect the correct list of connectors after project upgrade.

njhale:
This is a recent regression. Fixing now.

sangee2004:
This issue is not seen anymore when testing with latest build.
Able to see connector list reflect all the MCP servers as expected after a project upgrade.

Issue #4337, opened 2025-09-20
Remote MCP server - User is prompted to  authenticate even after successful oauth flow.


sangee2004:
**Describe the bug**
Remote MCP server - User is prompted to  authenticate even after successful oauth flow.

**To Reproduce**
1. Connect to any Remote MCP server
2. When user is presented with Authetication workflow , complete the oauth workflow successfuly.

<img width="1302" height="975" alt="Image" src="https://github.com/user-attachments/assets/330657fc-fec6-4a45-816c-a9a979751b58" />

4. Once we are back in project view , notice that user is presented with `Authenticate` option again. When following this , we are presented with error message

<img width="1397" height="837" alt="Image" src="https://github.com/user-attachments/assets/49c28a51-4844-4b53-a913-58c036ba37ea" />

<img width="1429" height="535" alt="Image" src="https://github.com/user-attachments/assets/4fb7d03d-0333-46bf-afd3-87d967bc6100" />

Note - we are already authenticated successfully after the first oauth flow and tool calls succeed.

https://www.loom.com/share/381340825f0648f6b4bd27ed1a88b17a?sid=c9dea5a2-7015-46c1-82b1-911181836715

**Expected behavior**
User should not be prompted with oauth workflow once it is completed successfully.

sangee2004:
This issue is not seen anymore. Once user  completes aouth workflow and closes the oauth workflow , he is brought back to project view were the `Autheticate` pop up flashes for a minute but then autocloses after that.

I will log a different issue to track this.

Issue #4332, opened 2025-09-20
Project Sharing - `Update` button used when prompting users to enter credentials for the MCP servers that are part of project created using `Copy Project`


sangee2004:
**Describe the bug**
Project Sharing - `Update` button used when prompting users to enter credentials for the MCP servers that are part of project created using `Copy Project`

**To Reproduce**
**Steps to reproduce the behavior:**
1. As user1, Create a project with MCP that has  required configuration parameters like `Tavily Search` 
2. Enable project sharing
3. As user2 , use project sharing link to create project.

Notice that when user is prompted to enter credentials , we are presented with `Update` button.

<img width="845" height="321" alt="Image" src="https://github.com/user-attachments/assets/ca8d8e01-fc49-4572-9fda-7c9d04063be0" />

**Expected behavior**
User should not be presented with message suggesting update is required because of recent server updates.

Instead of having `Update` button when entering configuration details, can this be changed to some other text like `Save` (we use `Launch` when connecting to new MCP servers but it may be misleading to use  this text here)?

sangee2004:
User is now presented with `Save` option when being prompted to enter credentials for projects created from share link.

<img width="872" height="319" alt="Image" src="https://github.com/user-attachments/assets/c35107fa-9a50-4e17-90d9-d39eb1bd26fd" />

PR #4354, opened 2025-09-21
fix: don't reopen oauth prompts after successful auth and close

njhale:
Addresses:
- https://github.com/obot-platform/obot/issues/4337
- https://github.com/obot-platform/obot/issues/4342
- https://github.com/obot-platform/obot/issues/4332


njhale:
> 👍 Pulled down and tested.

Thanks Ivy!

> One thing I'm wondering if is they're should be something like Retry Authentication if the user closes out of the oauth tab prematurely?

The unauthed status isn't cleared, so users will be able to click on the MCP server (which still has a warning icon), and hit "authenticate" there.

Issue #4341, opened 2025-09-20
Admin user is allowed to deploy MCP servers from user registry without having been given access to it.


sangee2004:
**Describe the bug**
Admin user is allowed to deploy MCP servers from user registry without having been given access to it.

**To Reproduce**
 **Scenario 1:**
1. As PUP user , create MCP server.
2. Delete the default ACL rule that is allows access to all MCP servers for all users.
3. As admin , notice that the MCP server created in step1 is available in `Available Connectors` list
Admin is able to connect and chat with it.

 **Scenario 2:**
1. As PU user , create MCP server.
2. As admin , notice that the MCP server created in step1 is available in `Available Connectors` list
Admin is able to connect and chat with it.

**Expected behavior**
Admin users should not be allowed to  deploy MCP servers from user registry without having been given access to it like in both the scenarios with PUP and PU users.

StrongMonkey:
After discussion, we are going to make the following change:

1. Owner of the mcp server bypass ACR, so they should always see their entries even when ACR is deleted.
2. Admin see all the servers from all users in admin page, but should not see the mcp server in my connector page.

sangee2004:
@StrongMonkey Assigning this issue back to you for making the changes that we discussed over huddle.

ACRs should only control which user can connect to MCP servers.

Owner (and admins) of  MCP server should always have access to MCP servers in their workspace  for listing , editing and seeing server details , audit log and usage.

sangee2004:
This issue is addressed when testing with latest build.

Admin user does not get access to MCP servers deployed by PU user for connecting to them.They do not appear in the `Available Connectors` list.
Admin user does not get access to MCP servers deployed by PUP user , unless there is a ACR that grants access to admin user for connecting to them.They do not appear in the `Available Connectors` list unless there is a ACR that grants access to admin user

PR #4368, opened 2025-09-22
Fix: don't bypass admin user to view user-owned mcp server

StrongMonkey:
Before, we are bypassing admin user to view user owned mcp servers and entries. This change will make admin also subject to ACR to be able to view mcp server and entries from power user and power user plus.

https://github.com/obot-platform/obot/issues/4341


pkg/api/handlers/mcp.go
```diff
@@ -94,18 +92,6 @@ func (m *MCPHandler) ListEntriesFromAllSources(req api.Context) error {
...
-		for _, entry := range list.Items {
-			if entry.Spec.MCPCatalogName == system.DefaultCatalog || entry.Spec.PowerUserWorkspaceID != "" {
-				items = append(items, convertMCPServerCatalogEntry(entry))
-			}
-		}
-
-		return req.Write(types.MCPServerCatalogEntryList{Items: items})
-	}
-
 	// For non-admin users, check access via AccessControlRules across all sources
```
njhale: nit: comment should be dropped

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -169,12 +172,29 @@ func (h *MCPCatalogHandler) ListEntries(req api.Context) error {
 		return fmt.Errorf("failed to list entries: %w", err)
 	}
 
-	var items []types.MCPServerCatalogEntry
+	var entries []types.MCPServerCatalogEntry
```
njhale: nit: you can initialize with a max capacity here:

e.g.

```go
entries := make([]types.MCPServerCatalogEntry, 0, len(list.Items))
```

pkg/api/handlers/mcp.go
njhale: @StrongMonkey The commit I was referring to was

> // For non-admin users, check access via AccessControlRules across all sources

Looks like you dropped different comments

Issue #4366, opened 2025-09-22
`ERROR: : exec: environment variable contains NUL` error seen when chatting with `AntV Chart` MCP server.


sangee2004:
**Describe the bug**

**To Reproduce**
1. Connect with `AntV Chart` MCP server.
2. Chat with it to make some tool calls like
```
draw graph for following 
1:00 2
2:00 3
```
Notice that user is presented with errors:

<img width="1297" height="630" alt="Image" src="https://github.com/user-attachments/assets/b90818d6-640a-4a82-bfe8-efb1a761721d" />

Note - I am able to chat in the project when there are no tool calls involved.

Following errors seen in obot logs:

```
time="2025-09-22T19:26:09Z" level=info msg="Handling request: method POST, path /credentials/reveal" id=37f4bf9f-bef2-454a-bff7-4136ee948c26
time="2025-09-22T19:26:09Z" level=info msg="Handled request: method POST, path /credentials/reveal" id=37f4bf9f-bef2-454a-bff7-4136ee948c26
time="2025-09-22T19:26:09Z" level=info msg="Handling request: method POST, path /evaluate" id=65cb417f-cf0c-4b31-bc21-b5e4348962d7
time="2025-09-22T19:26:09Z" level=error msg="failed to run tool [input_parse] cmd [/bin/sh -c exec ${GPTSCRIPT_TOOL_DIR}/bin/gptscript-go-tool input]: exec: environment variable contains NUL" logger=-ai/gptscript@v0.9.6-0.20250826061404-f962ce4ca884/pkg/engine
time="2025-09-22T19:26:09Z" level=info msg="Handled request: method POST, path /evaluate" id=65cb417f-cf0c-4b31-bc21-b5e4348962d7
time="2025-09-22T19:26:10Z" level=error msg="failed to save state: ERROR: : exec: environment variable contains NUL" logger=/app/pkg/invoke/invoker.go
time="2025-09-22T19:26:10Z" level=error msg="error syncing 'default/r1chat729x9': handler obot-controller obot.obot.ai/v1, Kind=Run: [routes.go:85] failed to stream: ERROR: : exec: environment variable contains NUL, requeuing"

```

**Expected behavior**

No errors should be seen when tool calls are made.

cjellick:
@sangee2004 this issue doesnt have those headers we talked about

cjellick:
The description of this issue is no longer accurate. This reproduces any time you have a newline in chat.

sangee2004:
[...]

PR #4371, opened 2025-09-22
Fix: Dont replace newlines with null character

cjellick:
The fourth parameter of this textBetween function call is causing us to
replaced newlines with the unicode null character. Here's the docs for
that function:

> Get all text between positions from and to. When blockSeparator is given,
> it will be inserted to separate text from different block nodes.
> If leafText is given, it'll be inserted for every non-text leaf node
> encountered, otherwise leafText will be used.

LeafText is the 4th param. The last sentence of that is confusing, but
regardless I don't think we need to set this 4th parameter.


cjellick:
Addresses https://github.com/obot-platform/obot/issues/4366

cjellick:
This was introduced in https://github.com/obot-platform/obot/pull/4119

ryu-man (approved):

ivyjeong13 (approved):
LGTM

PR #4381, opened 2025-09-23
enhance: always pull MCP images in docker backend

thedadams:


pkg/mcp/docker.go
```diff
@@ -696,12 +694,14 @@ func (d *dockerBackend) createVolumeWithFiles(ctx context.Context, files []File,
...
 }
 
-func (d *dockerBackend) ensureImageExists(ctx context.Context, imageName string) error {
-	// Check if image exists locally
-	_, err := d.client.ImageInspect(ctx, imageName)
-	if err == nil {
-		// Image exists locally
-		return nil
+func (d *dockerBackend) pullImage(ctx context.Context, imageName string, ifNotExists bool) error {
+	if ifNotExists {
```
StrongMonkey, line 698: If the same tag exists, do you still pull it? like do we always pull obot:latest to check latest sha?

thedadams: We always pull it. Docker does the right thing though and only pulls new layers.

We discussed doing it this way during standup to ensure that the latest of the `ghcr.io/obot-platform/mcp-images/phat:main` is always used.

Issue #4382, opened 2025-09-23
PUP user does not get access to the MCP servers in his own registry unless an ACL rule is created for allowing access to the mcp server.


sangee2004:
**Describe the bug**
PUP user does not get access to the MCP servers in his own registry unless an ACL rule is created for allowing access to this user.

**To Reproduce**
1. As a PUP user , create an MCP server.
2. Delete the default ACL rule that grants access to all server to everyone
3. Go to `My Connectors` page.
Notice that MCP server created in step1  is not listed in `Available Connectors` list.

**Note** If I add an ACL rule allowing PUP user access to this MCP server , then he is able to connect to it as expected.

**Expected Behavior:**
PUP user should be able to connect to their own MCP servers (that they have created) without having to need an ACL rule to access them.

sangee2004:
This issue is addressed when testing with latest build.
PUP user gets access to the MCP servers in his own registry without having to need an ACL rule that grants access to it.

PR #4389, opened 2025-09-24
Fix: fix admin ACR to view entries and servers

StrongMonkey:
This PR fixes a couple of issue regarding to admin and ACR

- Admin by default still see all entries and servers from all users, be able to view logs/server details
- Admin should not see mcp server from connector page from PU/PUP users unless a ACR is granted to admin
- PU/PUP user should see their mcp servers and use them without having a ACR if they are the owner.

https://github.com/obot-platform/obot/issues/4382
https://github.com/obot-platform/obot/issues/4341

### DIFF
--- a/pkg/accesscontrolrule/helper.go
+++ b/pkg/accesscontrolrule/helper.go
@@ -1,6 +1,7 @@
 package accesscontrolrule
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/obot-platform/obot/apiclient/types"
@@ -8,15 +9,18 @@ import (
 	"github.com/obot-platform/obot/pkg/system"
 	kuser "k8s.io/apiserver/pkg/authentication/user"
 	gocache "k8s.io/client-go/tools/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type Helper struct {
 	acrIndexer gocache.Indexer
+	client     client.Client
 }
 
-func NewAccessControlRuleHelper(acrIndexer gocache.Indexer) *Helper {
+func NewAccessControlRuleHelper(acrIndexer gocache.Indexer, client client.Client) *Helper {
 	return &Helper{
 		acrIndexer: acrIndexer,
+		client:     client,
 	}
 }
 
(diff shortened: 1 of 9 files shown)


pkg/accesscontrolrule/helper.go
```diff
@@ -381,11 +385,16 @@ func (h *Helper) UserHasAccessToMCPServerInWorkspace(user kuser.Info, serverName
 		}
 	}
 
+	// If the server is owned by the current user, they have access to it to ignore the AccessControlRules
+	if serverUserID == userID {
+		return true, nil
+	}
+
```
thedadams: I think we should do this check before making any of the "get access control rules" call because this is easier.

pkg/api/handlers/mcp.go
```diff
@@ -203,22 +204,28 @@ func (m *MCPHandler) ListServer(req api.Context) error {
 			continue
 		}
 
+		var hasAccess bool
+		var err error
+
```
```suggestion
		var (
			hasAccess bool
			err error
		)
```

pkg/api/handlers/mcp.go
```diff
@@ -203,22 +204,28 @@ func (m *MCPHandler) ListServer(req api.Context) error {
...
 			}
-			if !hasAccess {
-				continue
-			}
+		}
+
+		// if the server is owned by the current user, they have access to it
+		if server.Spec.UserID == req.User.GetUID() {
+			hasAccess = true
+		}
```
thedadams, line 227: Same, do this first since it is much easier than the others.

pkg/accesscontrolrule/helper.go
StrongMonkey: Good point 👍

pkg/accesscontrolrule/helper.go
```diff
@@ -337,6 +341,12 @@ func (h *Helper) UserHasAccessToMCPServerInWorkspace(user kuser.Info, serverName
 		userID = user.GetUID()
 		groups = authGroupSet(user)
 	)
+
+	// If the server is owned by the current user, they have access to it to ignore the AccessControlRules
+	if serverUserID == userID {
+		return true, nil
+	}
```
thedadams, line 342: Why not do this as the very first thing in the function?

thedadams (approved):

pkg/accesscontrolrule/helper.go
StrongMonkey: no reason than being dumb, moved :)

cjellick (approved):

PR #4390, opened 2025-09-24
fix: shorten copied project share resource names

njhale:
Shorten the names of API objects copied when:
- creating project share snapshots
- copying project share snapshots

Note:
- Resources with the old long names should be deleted on upgrade
- Not sure how collision resistant this is; may want to switch to hashing/truncating the current name (just to normalize the length to the standard type prefix + 5 characters), then appending the project thread name as a hyphenated suffix. That should give us MCP server names like `ms1mcrprp-t1pw77s`.


cjellick:
closing for now

Issue #4372, opened 2025-09-22
When there are multiple MCP servers that has required configurations in template , user is prompted to enter credentials for only one of them.


sangee2004:
**Describe the bug**
When there are multiple MCP servers that has required configurations in template , user is prompted to enter credentials for only one of them.

**Environment**: local
**Consistently Reproduces**: Yes .
**Additional environments attempted in**: main

**To Reproduce**
1. Create a template with 2 MCP servers that have required configuration like `Tavily Search` and `Exa Search`
2. Enable project sharing
3. Use the share link to create a project
Notice that user is prompted to enter credentials only for 1 of the MCP servers.
Other MCP server is also put in configuration required state. User needs to select this server from Connectors list and then use `Edit Configuration` to enter credentials.

https://www.loom.com/share/f6da596a48a64743871398e1778ff68a?sid=d8e720ff-f806-4389-843c-07a67ea2c202

**Expected behavior**
User should be prompted to enter credentials for both the MCP servers.

sangee2004:
This issue is addressed when testing with  latest builds.
User is prompted to enter credentials for both the MCP servers now.

PR #4391, opened 2025-09-24
fix: multiple config prompt sequence

njhale:
- Add keys for config prompt elements
- Hard refresh page on project upgrade to ensure new config/auth dialogs are
  triggered

Addresses https://github.com/obot-platform/obot/issues/4372

Loom: https://www.loom.com/share/1c2ceed88d6643f0a39032232950930d?sid=6fa39560-5b32-4813-b352-65c1e62fc3a8

### DIFF
--- a/ui/user/src/lib/components/chat/McpServerRequirements.svelte
+++ b/ui/user/src/lib/components/chat/McpServerRequirements.svelte
@@ -30,6 +30,9 @@
 	type OauthRequirement = Extract<Requirement, { type: 'oauth' }>;
 	let requirements = $derived([
 		...projectMcps.items
+			.filter((m) => (m.configured === false || m.needsURL) && !closed.has(m.id!))
+			.map((m) => ({ type: 'config', id: m.id!, mcpID: m.mcpID! }) as Requirement),
+		...projectMcps.items
 			.filter((m) => !m.authenticated && m.oauthURL && !closed.has(m.id!))
 			.map(
 				(m) =>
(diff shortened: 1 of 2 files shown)


ui/user/src/lib/components/chat/McpServerRequirements.svelte
```diff
@@ -280,23 +281,24 @@
 				</a>
 			</dialog>
 		{/if}
+		{#if requirements[0].type === 'config'}
```
thedadams: nit: use `:else if` and remove the `{/if}` above.

njhale: done

ui/user/src/lib/components/edit/ProjectConfiguration.svelte
```diff
@@ -59,15 +58,11 @@
...
 			// The upgrade was successful
-			// Refresh the sidebar mcp servers, tasks, and config form data
-			// Knowledge files are refreshed when the project changes because we're using
-			// bind:project={modifiedProject} in the ProjectConfigurationKnowledge component below
-			modifiedProject = project;
-			layout.tasks = (await ChatService.listTasks(project.assistantID, project.id)).items;
-			projectMCPs.items = await ChatService.listProjectMCPs(project.assistantID, project.id);
+			// Hard refresh the page to ensure all components are updated and auth/config prompts
+			// are triggered
+			window.location.href = window.location.pathname + '?edit=true';
```
ivyjeong13, line 63: If this is already the path w/ params you're on, you can also use `window.location.reload()`!

ivyjeong13 (approved):

ui/user/src/lib/components/edit/ProjectConfiguration.svelte
njhale: so `?edit=true` is actually not in the path when you navigate to the project settings (it gets stripped out), but adding it to the path will ensure we always return to the settings.

Issue #3911, opened 2025-08-12
Renaming thread is not working.


sangee2004:
**Describe the bug**
Renaming thread is not working.

**To Reproduce**
1. Chat with any MCP server
2. Using `Rename` option , rename chat thread.
3. Start a new chat or chat in the same thread.

Notice that the thread names reverts back to its original name(that was assigned to it automatically after the 1 st conversation in the chat thread)
https://www.loom.com/share/992136eddd5f46c5b7a743bb72fc86b3?sid=e38afc51-862c-4262-9756-84636127bdc3

**Expected behavior**
Changes made to the thread name should persist.

sangee2004:
@ivyjeong13 I am able to save the thread name (and it persists) when I hit enter after renaming the thread. But after this , i dont get to the see the option "Delete"/"Rename" being presented for this chat thread - https://www.loom.com/share/61ce021806f04b11bcc7dc79aa44d604?sid=06fa540c-5642-411b-bf85-f5c169bc7163

ivyjeong13:
Having a hard time reproducing this locally even though I see it on main, moving back for now to work on other issues

sangee2004:
This issue is addressed when testing with latest build.

Able to rename threads by changing the name and hitting enter as expected.

PR #4392, opened 2025-09-24
fix: use one if-else block to render the chat layout

ryu-man:
Addresses #3911

- This issue does not appear in the local environment.
- This is a blind patch that may or may not work.

### DIFF
--- a/ui/user/src/lib/components/chat/sidebar/Threads.svelte
+++ b/ui/user/src/lib/components/chat/sidebar/Threads.svelte
@@ -265,21 +265,11 @@
 									break;
 							}
 						}}
-						class="mx-2 w-0 grow border-none bg-transparent ring-0 outline-hidden dark:text-white"
+						class="mx-2 w-0 flex-1 grow border-none bg-transparent ring-0 outline-hidden dark:text-white"
 						placeholder="Enter name"
 						type="text"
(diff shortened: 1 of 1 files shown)


ivyjeong13:
> Addresses #3911
> 
> * This issue does not appear in the local environment.
> * This is a blind patch that may or may not work.

@ryu-man After working with a production only issue bug, realized I could probably do the same for this issue. Was able to reproduce locally by running production build.

```
pnpm run build 
make build
./bin/obot server
```

Can potentially put some console.logs and figure out what's not causing the proper rendering with this!

ryu-man:
> > Addresses #3911
> > 
> > * This issue does not appear in the local environment.
> > * This is a blind patch that may or may not work.
> 
> @ryu-man After working with a production only issue bug, realized I could probably do the same for this issue. Was able to reproduce locally by running production build.
> 
> ```
> pnpm run build 
> make build
> ./bin/obot server
> ```
> 
> Can potentially put some console.logs and figure out what's not causing the proper rendering with this!

@ivyjeong13 this screenshot is from the local build, which seems to be working; the clear & save buttons are shown with this change.

image

ivyjeong13 (approved):
Ran production ui build and checked!

Issue #4220, opened 2025-09-11
File parameter marked as sensitive shows text in plain text with Reveal option.


sangee2004:
**Describe the bug**
File parameter marked as sensitive shows text in plain text .

**To Reproduce**
1. Connect to `BigQuery Toolbox` MCP server
2. When prompted to configure parameters, try to enter text in `Google Application Credentials` which of type - `file` and and marked as sensitive

Notice that the text appears in clear text when there is a Reveal option.

<img width="750" height="711" alt="Image" src="https://github.com/user-attachments/assets/c1c1dcc7-6bd3-44fc-a826-aabb98acc3ec" />

Toggling to `Reveal` and then `Hide` results in text being hidden as expected.

https://www.loom.com/share/ab520b1c917b4090be5bc627876b9ef1?sid=f7308e75-07ba-4a8b-821e-a0cbeaaa20f0

**Expected behavior**
When typing text in `File` parameters that are marked as sensitive we expect  the text to be hidden.

sangee2004:
When testing with latest build , after i enter the `Google Application credential` (which I cut form some place and copy it over in this field) it is in hidden state as expected. The reveal options starts blinking 

https://www.loom.com/share/67ce33e9fef048cbaa62a4b4572797ad?sid=9ea326a3-5110-4d47-8a7f-e0032437afd6

I am not able to delete the text that was eneterd.
If just start to type in this field, no text shows up ( i expect it to be shown as hidden (*) as i type which is not happening)

@ryu-man Moving this issue to you for follow up.

ryu-man:
@sangee2004 The user should click on the eye icon to reveal the password so he can edit it. At the moment it is not possible to edit a hidden multiline section.

sangee2004:
When testing with latest build , I donot see the problem with reveal option starting to blink like I did before.

I am also able to type in this text area and delete as expected.

There is an issue with text not growing along with the text being entered resulting in incorrect behavior which i will track in a different issue.

PR #4403, opened 2025-09-24
fix: do not show sensitive info automatically

ryu-man:
Addresses #4220

- The user has to click on the eye icon to show/edit sensitive info.
- If a user tries to edit hidden sensitive info, the eye icon will start pulsing to grab the user's attention.

### DIFF
--- a/ui/user/src/lib/components/SensitiveInput.svelte
+++ b/ui/user/src/lib/components/SensitiveInput.svelte
@@ -15,28 +15,10 @@
 	let { name, value = $bindable(''), error, oninput, textarea, disabled }: Props = $props();
 	let showSensitive = $state(false);
 	let textareaElement = $state<HTMLTextAreaElement>();
-	let isEditing = $state(false);
+	let isPulsing = $state(false);
 
 	function getMaskedValue(text: string): string {
 		return text.replace(/[^\n\r]/g, '•');
-	}
-
-	function handleTextareaFocus() {
-		if (!showSensitive) {
-			isEditing = true;
-		}
-	}
-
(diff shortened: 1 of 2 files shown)


ui/user/src/lib/components/SensitiveInput.svelte
```diff
@@ -98,3 +100,21 @@
 		{/if}
 	</button>
 </div>
+
+<style>
+	@keyframes pulse {
```
ivyjeong13, line 112: Curious if tailwind's `animate-pulse` couldn't be used?

ryu-man: I am gonna give it a shot

ryu-man: @ivyjeong13 The TW `animate-pulse` does not seem to support animating two colors, which I think is necessary to attract attention.

ivyjeong13: Got it! One thing on my mind is if we should try to get this behavior to match an `<input type="password">` where a user can type while it's in sensitive mode. I'm thinking it'll be more effort to implement than forcing a user to click the eye to start typing so will go ahead and approve this.

ryu-man: Yes, I agree with you on this point, and I thought of this too. I just implemented a quick fix for the issue because implementing a textarea-like password input needs more work to sync between the real value, the masked value, the keyboard event, and the cursor position so the changes that happen on the masked value get applied to the real value too. But I will implement the behavior you describe in a separate PR in the future.

PR #4424, opened 2025-09-25
fix: improve performance when listing all servers in workspaces

g-linville:


pkg/api/handlers/poweruserworkspace.go
```diff
@@ -90,32 +90,35 @@ func (p *PowerUserWorkspaceHandler) ListAllEntries(req api.Context) error {
...
 	}
 
-	servers := make([]types.MCPServer, 0)
-	for _, item := range list.Items {
-		fieldSelector := kclient.MatchingFields{"spec.powerUserWorkspaceID": item.Name}
-		var serverList v1.MCPServerList
-		if err := req.List(&serverList, fieldSelector); err != nil {
-			return fmt.Errorf("failed to list servers: %w", err)
+	// Filter servers that have a non-empty PowerUserWorkspaceID
+	filteredServers := make([]v1.MCPServer, 0)
```
thedadams: nano-nit:
```suggestion
	var filteredServers []v1.MCPServer
```

pkg/api/handlers/poweruserworkspace.go
```diff
@@ -125,18 +128,21 @@ func (p *PowerUserWorkspaceHandler) ListAllServers(req api.Context) error {
 				credMap[cred.ToolName] = c.Env
 			}
 		}
+	} else {
+		credMap = make(map[string]map[string]string)
```
thedadams: Reading from a nil map is fine, so you don't have to do that.

pkg/api/handlers/poweruserworkspace.go
g-linville: fixed

pkg/api/handlers/poweruserworkspace.go
g-linville: fixed

thedadams (approved):

Issue #4327, opened 2025-09-19
Power User Plus - Provide user access to `Audit logs` and `Usage` for MCP servers.


sangee2004:
As a `Power User Plus` user , I  do not have have access to `Audit logs` and `Usage` for MCP servers that I have created.

It would be good to provide `Power User Plus` users access to `Audit logs` and `Usage` for MCP servers that they have created.

cjellick:
Note taht they should have limited access to audit logs. @thedadams is logging an "auditor" feature wherein you must explicitly have the auditor role to see the actual payloads of tool calls. If and admin doesnt have that role, they can only see the metadata.

I want PU/PUP users to only be able to see the metadata and not the payloads.

...If possible, it would be fine to allow them to see payloads for THEIR OWN tool calls (meanings ones made by them).

g-linville:
Backend for this is merged. I will check and see if any UI work is needed for this. Since there are no new components necessary, I can probably just get Claude to implement this pretty easily, if needed.

g-linville:
Moving to Triage for UI work.

cjellick:
@ryu-man - sending this to you for UI. What we want to do is just have a basic version of the usage and audit log pages for power user and power user plus type users in the "MCP Publisher" app.

Note that they will only be able to see audit log "metadata". Meaning they wont be able to see the payloads of tool request. This is the same for admin users in the admin app now. However, in the admin app, we give the admin a note that is like "You need the auditor permission to see this." We dont want to show that note for this PU/PUP usecase because there will be no permission they could be granted to be able to see this info. So, I just simply want the data to not be displayed, no warning or note.

You can ask @g-linville any follow up questions.

sangee2004:
This issue is addressed when testing with latest build.
Both PUP and PU user are given access to `Audit Logs` and `Usage` tab to monitor MCP servers in their own workspace.

<img width="1309" height="743" alt="Image" src="https://github.com/user-attachments/assets/5160a205-aff5-48e4-aeae-0c153c67611e" />

PR #4476, opened 2025-09-29
feat: give PUs and PUPs metadata audit access

g-linville:
Backend for https://github.com/obot-platform/obot/issues/4327

### DIFF
--- a/pkg/api/authz/authz.go
+++ b/pkg/api/authz/authz.go
@@ -204,6 +204,11 @@ var (
 		types.GroupPowerUser: {
 			"GET /api/users",
 			"GET /api/users/{user_id}",
+			"GET /api/mcp-audit-logs",
+			"GET /api/mcp-audit-logs/filter-options/{filter}",
+			"GET /api/mcp-audit-logs/{mcp_id}",
+			"GET /api/mcp-stats",
+			"GET /api/mcp-stats/{mcp_id}",
 		},
 
 		types.GroupAuthenticated: {
--- a/pkg/api/handlers/mcpgateway/auditlog.go
+++ b/pkg/api/handlers/mcpgateway/auditlog.go
@@ -10,6 +10,7 @@ import (
 	"github.com/obot-platform/obot/pkg/api"
 	gateway "github.com/obot-platform/obot/pkg/gateway/client"
 	gatewaytypes "github.com/obot-platform/obot/pkg/gateway/types"
+	"github.com/obot-platform/obot/pkg/system"
 )
 
 type AuditLogHandler struct{}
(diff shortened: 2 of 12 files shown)


pkg/api/authz/authz.go
```diff
@@ -204,6 +204,11 @@ var (
 		types.GroupPowerUser: {
 			"GET /api/users",
 			"GET /api/users/{user_id}",
+			"GET /api/mcp-audit-logs",
+			"GET /api/mcp-audit-logs/filter-options/{filter}",
+			"GET /api/mcp-audit-logs/{mcp_id}",
+			"GET /api/mcp-stats",
+			"GET /api/mcp-stats/{mcp_id}",
```
g-linville, line 211: TODO: make sure that the PUs and PUPs can't use the `mcp_id` parameter to get info for servers outside of their workspace.

g-linville: This is working as expected.

pkg/api/handlers/mcpgateway/auditlog.go
```diff
@@ -71,6 +72,11 @@ func (h *AuditLogHandler) ListAuditLogs(req api.Context) error {
 		Query:                     strings.TrimSpace(query.Get("query")),
 	}
 
+	// Apply workspace filtering for Power Users
+	if req.UserIsPowerUser() && !req.UserIsAdmin() {
```
StrongMonkey, line 76: Nit: would it be better to move `!req.UserIsAdmin()` into `UserIsPowerUser`?

pkg/api/request.go
```diff
@@ -279,6 +279,12 @@ func (r *Context) UserIsAuditor() bool {
 	return slices.Contains(r.User.GetGroups(), types.GroupAuditor)
 }
 
+func (r *Context) UserIsPowerUser() bool {
+	groups := r.User.GetGroups()
+	return slices.Contains(groups, types.GroupPowerUser) ||
+		slices.Contains(groups, types.GroupPowerUserPlus)
```
thedadams: A power user plus is a power user, so just need to check for this group.
```suggestion
	return slices.Contains(groups, types.GroupPowerUser)
```

pkg/api/handlers/mcpgateway/handler.go
```diff
@@ -142,10 +142,22 @@ func (h *Handler) OnMessage(ctx context.Context, msg nmcp.Message) {
 		return
 	}
 
+	// Determine PowerUserWorkspaceID: use server's workspace ID for multi-user servers,
+	// or look up catalog entry's workspace ID for single-user servers
+	powerUserWorkspaceID := m.mcpServer.Spec.PowerUserWorkspaceID
+	if powerUserWorkspaceID == "" && m.mcpServer.Spec.MCPServerCatalogEntryName != "" {
+		// This is a single-user server created from a catalog entry, look up the entry
+		var entry v1.MCPServerCatalogEntry
+		if err := h.storageClient.Get(ctx, kclient.ObjectKey{Namespace: m.mcpServer.Namespace, Name: m.mcpServer.Spec.MCPServerCatalogEntryName}, &entry); err == nil {
```
StrongMonkey, line 151: should It handle the error when failed to get?

thedadams (changes requested):

pkg/api/handlers/mcpgateway/handler.go
g-linville: probably not necessary

pkg/api/request.go
g-linville: fixed

pkg/api/handlers/mcpgateway/auditlog.go
g-linville: probably not I think

thedadams (approved):

Issue #4456, opened 2025-09-26
Updating a PUP user to PU user does not delete existing access control rules created for the user. Accessing Access control results in 403 errors.


sangee2004:
**Describe the bug**
Updating a PUP user to PU user does not delete existing access control rules created for the user. Accessing Access control results in 403 errors.

**To Reproduce**
1. Create a PUP user ( I have PUP as default user role and logged in as this user)
2. As admin, we will be the default ACR that allows access to all MCP servers for all users
3. Update the user role for this user to PU.
4. Notice that the ACR created in step2 continues to be present.

<img width="1454" height="827" alt="Image" src="https://github.com/user-attachments/assets/c3776229-e7c5-43df-b4dd-b4088a51dcb1" />

<img width="1142" height="459" alt="Image" src="https://github.com/user-attachments/assets/9bea4152-9516-43df-91cf-0f03b3c9aba4" />

5. Log in as the same user (as in step1)
6. Access `MCP Publisher` page.
Notice that we have `Access Control` option that is presented in the UI
Accessing it , results in 403 errors.

<img width="1159" height="658" alt="Image" src="https://github.com/user-attachments/assets/3c55c92e-7b15-4eea-89ff-77279a56e764" />
**Expected behavior**
When PUP user downgrades to PU user , all existing ACRs should get deleted.

cjellick:
@StrongMonkey did you ask me about this functionality?

StrongMonkey:
@cjellick No, I think we are aligned with the behavior. In grant's initial design, when a PUP user is downgraded to PU, we should delete all existing ACR so that PU will not accidentally inherit the ACR from (PUP) that share all its entries in the ongoing future.(since they can't update the ACR anymore). 

But @sangee2004 I think you said that you can't reliably reproduce this? we should re-test it. From issue description this is more like a front refresh issue as you shouldn't see the access control page, the ACR should be already deleted in backend.

sangee2004:
@StrongMonkey This is not a front issues. I see the default ACR of PUP role (giving access to everyone) continues to be present (even reported in admin view when listing all ACRs) . With the latest build as of Friday , I was able to reproduce this issue consistently.

g-linville:
I'm thinking this should already be fixed. I will go test it.

g-linville:
Yep, this is fixed.

sangee2004:
[...]

PR #4483, opened 2025-09-29
Fix: delete all ACR when depromoted to power user

StrongMonkey:
When downgrading from power user plus to power user, we should delete every ACR rules from PUP so that power user won't be able to share their entries with existing ACR rules in place. This include default generated rules to share everything from power user plus.

https://github.com/obot-platform/obot/issues/4456

### DIFF
--- a/pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
+++ b/pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
@@ -172,10 +172,8 @@ func (h *Handler) cleanupWorkspaceForDemotionToPowerUser(ctx context.Context, cl
 		}
 
 		for _, acr := range acrs.Items {
-			if !acr.Spec.Generated {
-				if err := client.Delete(ctx, &acr); err != nil {
-					return err
-				}
+			if err := client.Delete(ctx, &acr); err != nil {
+				return err
 			}
 		}
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,11 +20,11 @@ RUN --mount=type=cache,id=pnpm,target=/root/.local/share/pnpm/store \
 
 FROM cgr.dev/chainguard/postgres:latest-dev AS build-pgvector
 RUN apk add build-base git postgresql-dev clang-19
-RUN git clone --branch v0.8.0 https://github.com/pgvector/pgvector.git && \
+RUN git clone --branch v0.8.1 https://github.com/pgvector/pgvector.git && \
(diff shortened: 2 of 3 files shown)


pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
```diff
@@ -172,10 +173,8 @@ func (h *Handler) cleanupWorkspaceForDemotionToPowerUser(ctx context.Context, cl
 		}
 
 		for _, acr := range acrs.Items {
-			if !acr.Spec.Generated {
-				if err := client.Delete(ctx, &acr); err != nil {
-					return err
-				}
+			if err := client.Delete(ctx, &acr); err != nil {
```
thedadams, line 175: I don't think we want to delete the generated access control rule completely. This will mean that all of the user's custom MCP servers will also get deleted because the access control rule is gone.

The handler that creates the generated access control rule makes this distinction between power user and power user plus: https://github.com/obot-platform/obot/blob/main/pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go#L226-L233

I think we should just update the generated access control rule. Similarly, I think if the user it upgraded from power user to power user plus, then we should also make the corresponding change.

StrongMonkey: Yeah 

> This will mean that all of the user's custom MCP servers will also get deleted because the access control rule is gone.

We made a change to ensure that the owner of the workspace will see their entries regardless of ACR for a better UX. So when PU user deletes all its ACR they should still see their entries.

>  Similarly, I think if the user it upgraded from power user to power user plus, then we should also make the corresponding change.

Yeah, we dicussed this a bit. I specifically didn't make it this case, because there could be a case where PU is upgraded to PUP and immediately their mcp server are shared by the world. So rather making the change automatically, when they got promoted to power user plus, they should be able to modify ACR to decide who to share.

thedadams: We talked about this "offline." The reason that the MCP servers aren't getting deleted when we delete all the access control rules is because of another bug in our cleanup logic. Both of these bugs should be addressed so that users don't have access to MCP servers they shouldn't have access to.

StrongMonkey: Pushed a fix it. It should just follow regular access pattern to determine access rules based on catalogId and poweruserWorkspaceId.

Issue #4409, opened 2025-09-25
Default ACR that gets created for PU user is not necessary.


sangee2004:
**Describe the bug**
Default ACR that gets created for PU user is not necessary.

**To Reproduce**
1. Log in to obot as any user.
2. User gets created with `Basic User` role (`Basic User` is the default use role on fresh installs)
3. As admin user , update the user to `Power User` role.
4. Go to `Access Control`

Notice that following default ACR that allows user access to all MCP entry in the user registry gets created.

<img width="945" height="614" alt="Image" src="https://github.com/user-attachments/assets/8fa1b8d3-8881-4bc6-a098-ff3d98b4ec81" />

PU users will have the ability to deploy MCP servers  in their own registry and be able to connect to them even with out having this ACR.

**Expected behavior**
There is no need to be creating this Default ACR  for PU users.

g-linville:
Merged the fix for this. Note that these existing ACRs won't get cleaned up automatically. I didn't bother to add a migration for that, since they are harmless anyway. But new ones will not be created.

sangee2004:
There are no Default ACRs created for PU users anymore when testing with latest builds.

PR #4519, opened 2025-09-30
fix: do not create needless ACR for power users

g-linville:
for https://github.com/obot-platform/obot/issues/4409

There is separate logic that already gives PUs access to things in their own PUW, so we don't need to create this default ACR for them anymore.

### DIFF
--- a/pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
+++ b/pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
@@ -160,7 +160,7 @@ func (h *Handler) cleanupWorkspaceForDemotionToPowerUser(ctx context.Context, cl
 	}
 
 	for _, workspace := range workspaces.Items {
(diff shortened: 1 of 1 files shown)


pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
```diff
@@ -177,6 +177,11 @@ func (h *Handler) cleanupWorkspaceForDemotionToPowerUser(ctx context.Context, cl
 			}
 		}
 
+		workspace.Status.DefaultAccessControlRuleGenerated = false
```
thedadams: nit: only update if we know the value is changing.

thedadams (approved):

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
g-linville: fixed

StrongMonkey (approved):

PR #4520, opened 2025-09-30
chore: separate catalog and workspace cleanup

thedadams:
The clean up of unauthorized MCP servers and instances was a universal handler. This could cause slowdown as we list the world every time an MCP server or instance is changed.

This change separates cleanup for catalogs and workspaces in an effort to be more efficient.

### DIFF
--- a/pkg/controller/routes.go
+++ b/pkg/controller/routes.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"github.com/obot-platform/nah/pkg/handlers"
+	"github.com/obot-platform/nah/pkg/router"
 	"github.com/obot-platform/obot/pkg/controller/generationed"
 	"github.com/obot-platform/obot/pkg/controller/handlers/accesscontrolrule"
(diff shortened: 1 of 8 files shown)


pkg/controller/routes.go
```diff
@@ -261,6 +262,12 @@ func (c *Controller) setupRoutes() {
 	// AccessControlRule
 	root.Type(&v1.AccessControlRule{}).HandlerFunc(cleanup.Cleanup)
 	root.Type(&v1.AccessControlRule{}).HandlerFunc(accesscontrolrule.PruneDeletedResources)
+	// This is a hack. We use field selectors to trigger other resources. However, when an access control rule is deleted,
+	// we don't trigger because we don't have the object to match the field selectors against.
+	// Having a finalizer that does nothing will ensure that the other resources are triggered.
+	root.Type(&v1.AccessControlRule{}).FinalizeFunc(v1.AccessControlRuleFinalizer, func(router.Request, router.Response) error {
+		return nil
+	})
```
g-linville, line 270: nice hack

Issue #4261, opened 2025-09-17
When tool listing results in MCP server instance to be redeployed because of availability of phat image version , tool listing fails.


sangee2004:
**Describe the bug**
When tool listing results in MCP server instance to be redeployed because of availability of phat image version , tool listing fails.

**Pre Req** - Local obot install with helm install. Have few stdio mcpservers already running  (like Tavily Search, Duckduckgo search)

**To Reproduce**
1. Upgrade obot install with latest chart and obot version (that requires a new version of phat image - `ghcr.io/obot-platform/mcp-images/phat:main`).
2. Launch chat view with any of the existing connectors from the `Enabled Connectors` list  and go to `Tools` tab of the MCP server from connect list.

Notice that this results in MCP server pod being redeployed with new phat image .

`\tools` call itself fails with `500 error` and no tools are fetched.

<img width="831" height="298" alt="Image" src="https://github.com/user-attachments/assets/d339b08b-26d6-47e2-88a5-1a15e74c940b" />

https://www.loom.com/share/075438c981174dc4b86a3743123af18a?sid=68b5890d-a3ea-4f6b-9ae0-baffaf45d06e

Next retry to list tools succeeds as expected.

**Expected behavior**
First `\tools` call made after a new `phat` image is made avaialable should not report failure.

cjellick:
Any thoughts on this one @thedadams ? Can we do better as Sangeetha is suggesting?

cjellick:
@sangee2004 @thedadams does this issue even make sense? I dont think we redeploy when a new phat-image is availble

thedadams:
I think the issue is more concerned with why the pod redeployed. She guessed at why the pod was redeployed, and the new image was not the reason. You're right that we don't do anything with existing pods when there is a new image version.

cjellick:
SO two action items here:
1. build a 4xx 5xx http error dashboard to get a handle on how often these occur in our hosted envs
2. donnie has thoughts on how to more intelligently handle these scenarios and he'll comment on that.

thedadams:
Right now, we are naively waiting for the deployment to progress here: https://github.com/obot-platform/obot/blob/main/pkg/mcp/kubernetes.go#L460-L464

It would be better if we did this in a smarter way. For example, watch the pods and don't timeout if the pod is continuing to progress.

cjellick:
@drpebcak - the first item above:

> build a 4xx 5xx http error dashboard to get a handle on how often these occur in our hosted envs

[...]

PR #4530, opened 2025-10-01
enhance: be smarter about timeouts when waiting for deployments

g-linville:
for https://github.com/obot-platform/obot/issues/4261

### DIFF
--- a/pkg/mcp/kubernetes.go
+++ b/pkg/mcp/kubernetes.go
@@ -458,27 +458,20 @@ func (k *kubernetesBackend) k8sObjects(id string, server ServerConfig, serverDis
 	return objs, nil
 }
 
-// analyzePodStatuses combines the results of analyzing each individual pod status.
-func analyzePodStatuses(pods []corev1.Pod) (bool, error) {
-	var (
-		shouldRetry bool
-		errs        []error
-	)
-	for i, pod := range pods {
-		podShouldRetry, podErr := analyzePodStatus(&pod)
-		shouldRetry = shouldRetry || podShouldRetry
-		if podErr != nil {
-			errs = append(errs, fmt.Errorf("pod %d: %w", i+1, podErr))
+// getNewestPod finds and returns the most recently created pod from the list.
+func getNewestPod(pods []corev1.Pod) (*corev1.Pod, error) {
+	if len(pods) == 0 {
+		return nil, fmt.Errorf("no pods provided")
+	}
+
+	newest := &pods[0]
+	for i := range pods {
+		if pods[i].CreationTimestamp.After(newest.CreationTimestamp.Time) {
+			newest = &pods[i]
 		}
 	}
 
-	if len(errs) == 0 {
-		return shouldRetry, nil
-	}
-	if len(errs) == 1 {
-		return shouldRetry, errs[0]
-	}
-	return shouldRetry, fmt.Errorf("multiple pod issues: %v", errs)
+	return newest, nil
 }
 
 // analyzePodStatus examines a pod's status to determine if we should retry waiting for it
@@ -622,7 +615,18 @@ func (k *kubernetesBackend) updatedMCPPodName(ctx context.Context, url, id strin
 			return "", fmt.Errorf("%w: %v", ErrHealthCheckTimeout, lastErr)
 		}
 
-		shouldRetry, podErr := analyzePodStatuses(pods.Items)
+		// Get the newest pod and analyze its status
+		newestPod, err := getNewestPod(pods.Items)
+		if err != nil {
+			olog.Debugf("failed to get newest pod: id=%s error=%v attempt=%d", id, err, attempt+1)
+			lastErr = err
+			if attempt < maxRetries {
+				continue
+			}
+			return "", fmt.Errorf("%w: %v", ErrHealthCheckTimeout, lastErr)
+		}
+
+		shouldRetry, podErr := analyzePodStatus(newestPod)
 		lastErr = podErr
 
 		if !shouldRetry {


g-linville:
Pausing my work on this to pick up something higher priority. I think this will work; I just haven't actually tested it in k8s yet.

g-linville:
Tested, this works as expected!a

pkg/mcp/kubernetes.go
```diff
@@ -455,45 +458,177 @@ func (k *kubernetesBackend) k8sObjects(id string, server ServerConfig, serverDis
...
 
-func (k *kubernetesBackend) updatedMCPPodName(ctx context.Context, url, id string, server ServerConfig) (string, error) {
-	// Wait for the deployment to be updated.
-	_, err := wait.For(ctx, k.client, &appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: id, Namespace: k.mcpNamespace}}, func(dep *appsv1.Deployment) (bool, error) {
-		return dep.Generation == dep.Status.ObservedGeneration && dep.Status.Replicas == 1 && dep.Status.UpdatedReplicas == 1 && dep.Status.ReadyReplicas == 1 && dep.Status.AvailableReplicas == 1, nil
-	}, wait.Option{Timeout: time.Minute})
-	if err != nil {
-		return "", ErrHealthCheckTimeout
+// analyzePodStatus examines a pod's status to determine if we should retry waiting for it
+// or if we should fail immediately. Returns (shouldRetry, reason).
```
njhale: This is looking very familiar

pkg/mcp/kubernetes.go
```diff
@@ -27,6 +28,8 @@ import (
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+var klog = logger.Package()
```
thedadams: nit: if I saw `klog`, I would think this was kubernetes related. Could we change it to `olog` or `plog`?

pkg/mcp/kubernetes.go
```diff
@@ -455,45 +458,177 @@ func (k *kubernetesBackend) k8sObjects(id string, server ServerConfig, serverDis
...
+				// Unschedulable could be transient (e.g., waiting for autoscaler)
+				return true, fmt.Sprintf("pod unschedulable: %s", cond.Message)
+			}
+		}
 	}
 
-	if err = ensureServerReady(ctx, url, server); err != nil {
-		return "", fmt.Errorf("failed to ensure MCP server is ready: %w", err)
+	// Check container statuses
+	allContainerStatuses := append([]corev1.ContainerStatus{}, pod.Status.ContainerStatuses...)
```
thedadams: Is this necessary? This intent seems to be to copy the array so that modifications don't affect the original array, but we don't seem to be modifying the array or elements anywhere.

If it is necessary, could we add a comment explaining why?

pkg/mcp/kubernetes.go
```diff
@@ -455,45 +458,177 @@ func (k *kubernetesBackend) k8sObjects(id string, server ServerConfig, serverDis
...
+	// Default: pod is in Pending or Running but not ready yet - should retry
+	return true, fmt.Sprintf("pod in phase %s, waiting for containers to be ready", pod.Status.Phase)
+}
+
+func (k *kubernetesBackend) updatedMCPPodName(ctx context.Context, url, id string, server ServerConfig) (string, error) {
+	const maxRetries = 5
+	var lastReason string
+
+	// Retry loop with smart pod status checking
+	for attempt := 0; attempt <= maxRetries; attempt++ {
```
thedadams: nit:
```suggestion
	for attempt := range maxRetries {
```

pkg/mcp/kubernetes.go
```diff
@@ -455,45 +458,177 @@ func (k *kubernetesBackend) k8sObjects(id string, server ServerConfig, serverDis
...
+		// Wait for the deployment to be updated.
+		_, err := wait.For(ctx, k.client, &appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: id, Namespace: k.mcpNamespace}}, func(dep *appsv1.Deployment) (bool, error) {
+			return dep.Generation == dep.Status.ObservedGeneration && dep.Status.Replicas == 1 && dep.Status.UpdatedReplicas == 1 && dep.Status.ReadyReplicas == 1 && dep.Status.AvailableReplicas == 1, nil
+		}, wait.Option{Timeout: time.Minute})
+
```
thedadams:
```suggestion
```

pkg/mcp/kubernetes.go
```diff
@@ -455,45 +458,177 @@ func (k *kubernetesBackend) k8sObjects(id string, server ServerConfig, serverDis
...
+				if p.Status.Phase == corev1.PodRunning {
+					podName = p.Name
+					runningPodCount++
+				}
+			}
+			if runningPodCount == 1 {
+				return podName, nil
+			}
+
+			return "", ErrHealthCheckTimeout
```
thedadams: I would expect us to retry here if we haven't hit the `maxRetries` yet or determine why there is not exactly one running pod.

pkg/mcp/kubernetes.go
```diff
@@ -455,45 +458,177 @@ func (k *kubernetesBackend) k8sObjects(id string, server ServerConfig, serverDis
...
+		if len(pods.Items) == 0 {
+			klog.Debugf("no pods found for MCP server: id=%s attempt=%d", id, attempt+1)
+			lastReason = "no pods found"
+			if attempt < maxRetries {
+				continue
+			}
+			return "", fmt.Errorf("%w: %s", ErrHealthCheckTimeout, lastReason)
+		}
+
+		// Analyze the first pod (there should only be one for a deployment with replicas=1)
```
thedadams: I don't think this comment is correct.

For example, if the deployment has been updated, there could be 1 old running pod and 1 new pod that is crashing. I think we would want to handle this case.

pkg/mcp/kubernetes.go
```diff
@@ -455,45 +458,177 @@ func (k *kubernetesBackend) k8sObjects(id string, server ServerConfig, serverDis
...
+			if strings.Contains(reason, "CrashLoopBackOff") {
+				return "", fmt.Errorf("%w: %s", ErrPodCrashLoopBackOff, reason)
+			} else if strings.Contains(reason, "ImagePull") || strings.Contains(reason, "InvalidImageName") {
+				return "", fmt.Errorf("%w: %s", ErrImagePullFailed, reason)
+			} else if strings.Contains(reason, "unschedulable") || strings.Contains(reason, "Evicted") {
+				return "", fmt.Errorf("%w: %s", ErrPodSchedulingFailed, reason)
+			} else if strings.Contains(reason, "CreateContainer") || strings.Contains(reason, "RunContainer") {
+				return "", fmt.Errorf("%w: %s", ErrPodConfigurationFailed, reason)
+			}
+			return "", fmt.Errorf("%w: %s", ErrHealthCheckTimeout, reason)
```
thedadams: Instead of doing string contains here, can `analyzePodStatus` return the appropriate error and we return it here?

pkg/mcp/kubernetes.go
g-linville: Nope, I think this was left behind after I made some changes. Thanks for catching it

pkg/mcp/kubernetes.go
g-linville: LLM didn't know about this I guess :)

pkg/mcp/kubernetes.go
g-linville: Fixed (`olog`)

pkg/mcp/kubernetes.go
g-linville: Fixed

pkg/mcp/kubernetes.go
g-linville: Fixed

pkg/mcp/kubernetes.go
g-linville: Fixed

pkg/mcp/kubernetes.go
g-linville: Yeah I think by the time we reach here there will always be one running pod. I added a comment and added retries in case somehow there isn't.

pkg/mcp/kubernetes.go
g-linville: Yes, thank you! I added a function that will loop over the pod statuses instead.

pkg/mcp/kubernetes.go
g-linville: Yes, fixed

pkg/mcp/kubernetes.go
```diff
@@ -455,45 +458,191 @@ func (k *kubernetesBackend) k8sObjects(id string, server ServerConfig, serverDis
...
-		return "", ErrHealthCheckTimeout
+// analyzePodStatuses combines the results of analyzing each individual pod status.
+func analyzePodStatuses(pods []corev1.Pod) (bool, error) {
+	var (
+		shouldRetry bool
+		errs        []error
+	)
+	for i, pod := range pods {
+		podShouldRetry, podErr := analyzePodStatus(&pod)
+		shouldRetry = shouldRetry || podShouldRetry
```
thedadams: I'm not sure the logic here is correct. For example, if there is an old pod running and a new pod failing to pull an image, then my understanding is that this logic will always retry because of the old running pod.

It will eventually fail after all the retries, but I'm not sure if that is the desired outcome.

pkg/mcp/kubernetes.go
```diff
@@ -455,45 +458,191 @@ func (k *kubernetesBackend) k8sObjects(id string, server ServerConfig, serverDis
...
+			// Permanent failure - return the error with the appropriate type already wrapped
+			olog.Debugf("pod in non-retryable state: id=%s error=%v attempt=%d", id, podErr, attempt+1)
+			return "", podErr
+		}
+
+		// Transient issue - retry if we haven't exceeded max retries
+		if attempt < maxRetries {
+			olog.Debugf("pod not ready, will retry: id=%s error=%v attempt=%d maxRetries=%d", id, podErr, attempt+1, maxRetries)
+			continue
+		}
```
thedadams: This block will always get executed in this loop.

pkg/mcp/kubernetes.go
```diff
@@ -455,45 +458,191 @@ func (k *kubernetesBackend) k8sObjects(id string, server ServerConfig, serverDis
...
+
+		// Exceeded max retries
+		olog.Debugf("exceeded max retries waiting for pod: id=%s lastError=%v attempts=%d", id, lastErr, attempt+1)
+		return "", fmt.Errorf("%w after %d retries: %v", ErrHealthCheckTimeout, maxRetries+1, lastErr)
```
thedadams: `maxRetries+1` seems strange here. If `maxRetries` is 5, I would expect at most 5 retries, not 6.

pkg/mcp/kubernetes.go
g-linville: Hmm, that is a good point. Do you think it makes the most sense to check the more recently created pod, if there are multiple, and just use that as the source of whether to retry?

pkg/mcp/kubernetes.go
g-linville: Yep, just something dumb the LLM did that I didn't catch

g-linville: Fixed

pkg/mcp/kubernetes.go
g-linville: Fixed

pkg/mcp/kubernetes.go
thedadams: Yeah, I think that is a good approach.

g-linville: Done

Issue #4434, opened 2025-09-25
Usage - When there is no usage details to be reported , MCP server tabs are lost .


sangee2004:
**Describe the bug**
A clear and concise description of what the bug is.Usage - When there is no usage details to be reported , MCP server tabs are lost .

**To Reproduce**
1. As admin/owner , access any MCP server entry from Dashboard that no user has connected with.
2. Go to Usage tab

Notice that all the MCP Server tabs like Overview , Server Details , Audit logs etc get lost.

<img width="1013" height="626" alt="Image" src="https://github.com/user-attachments/assets/808f3307-1a18-42c9-83c9-a70f64cb07c4" />

**Expected behavior**
All MCP Server tabs  should be present when `No Usage Status` page is presented.

sangee2004:
This issue is addressed when testing with latest builds.

PR #4531, opened 2025-10-01
feat: add Audit Logs page with layout and content components

ryu-man:
Addresses #4327, #4434

- Add audit logs & usage pages for PUP | PU users
- hide payload data that requires special permissions

image

The log drawer shows only public data.

image

### DIFF
--- a/ui/user/src/lib/components/admin/McpServerEntryForm.svelte
+++ b/ui/user/src/lib/components/admin/McpServerEntryForm.svelte
@@ -350,7 +350,7 @@
 	{/if}
 	<div class="flex grow flex-col gap-2">
 		<OverflowContainer
-			class="scrollbar-none flex w-full items-center gap-2 overflow-x-auto"
+			class="scrollbar-none flex min-h-fit w-full items-center gap-2 overflow-x-auto"
 			style="scroll-behavior: smooth;"
 			{@attach (node: HTMLDivElement) => (scrollContainer = node)}
 		>


ui/user/src/lib/components/admin/audit-logs/AuditLogDetails.svelte
```diff
@@ -107,41 +110,45 @@
...
-				>
-					<div class="absolute top-0 left-0 h-full w-1 bg-blue-800"></div>
-					<div class="flex flex-col gap-1">
-						{#each Object.entries(auditLog.requestHeaders ?? {}) as [key, value] (key)}
-							<p>
-								<span class="font-medium">{key}</span>: {value}
-							</p>
-						{/each}
+			{#if shouldShowPayload}
+				{#if auditLog.requestHeaders}
```
ivyjeong13, line 114: Just checking, are these coming back from the backend now even if a user doesn't have the access to view them? I remembered talking to Donnie before about these wouldn't return if they didn't have the auditor role and wondering if that changed. Same for requestBody, responseHeader, responseBody.

ryu-man: The real data is not shown if that condition is omitted; only a message stating the user does not have enough permission, and I am here just hiding that message because Craig has required that to be done. However, I need to double-check if the backend API returns the data or not.

ryu-man: After checking that info does not exist in the response data

ivyjeong13: Okay cool! Just a nit of wondering if those extra `if shouldShowPayload` checks are necessary if those values don't come back from the backend if they don't have access.

ryu-man: Yes, because Craig said that we don't want to show those permission messages for PU/P users.

Issue #4544, opened 2025-10-06
PU and PUP user roles are interchanged when upgrading from v0.11.0 to latest main.


sangee2004:
**Describe the bug**
PU and PUP user roles are interchanged when upgrading from v0.11.0 to latest main.

**Pre Req:**
1. Fresh install of v0.11.0 that is auth enabled.
2. Have an admin ,  PUP, PU and regular users 
3. As PUP user,  create MCP server and have some Access control rules created.
4. As PU user ,create MCP server and connect to it.

<img width="1147" height="494" alt="Image" src="https://github.com/user-attachments/assets/5ba0ac99-e1e1-4fc4-aacc-a5dcba5ddc4b" />

**To Reproduce**
1. Upgrade to latest version from `main` - [v0.0.0-dev+bdca12ef](https://github.com/obot-platform/obot/commit/bdca12ef)

After upgrade , notice that PUP user gets reported as PU user and does not have access to any ACRs that he had created and is not provided with ability to create any new ACRs.

PU user  gets reported as PUP user and has access to ACRs and is allowed to create new ACRs.

<img width="1080" height="500" alt="Image" src="https://github.com/user-attachments/assets/63051c0b-801e-4c0a-bef5-2c7ddd41ad88" />

**Expected behavior**
User roles should remain the same for PU and PUP users after upgrade

sangee2004:
This issue is addressed when testing with latest build - [v0.0.0-dev+4549b8f3](https://github.com/obot-platform/obot/commit/4549b8f3)

PR #4548, opened 2025-10-07
fix: correct user role migration

thedadams:
The power user and power user plus roles were migrated incorrectly.


StrongMonkey (approved):

StrongMonkey:
https://github.com/obot-platform/obot/issues/4544

Issue #4324, opened 2025-09-19
Streamline bulk-updating MCP server instances


drpebcak:
Currently, you need to click though each individual MCP Server in the catalog, click a bunch of times to get to the 'Server Details' page, and then select the instances and hit 'update'. This needs to be done for every entry in the catalog that has pending updates (and you won't know if updates are pending until you click into that page).

We should provide a page that lists all available pending updates and also allows kicking off the update process.

cjellick:
@ivyjeong13 - I have been thinking about this and what I'd like to see is a "Deployed Servers" view that is much more "tactical" around operating these servers. Think, like:
- a table of the individual deployed server instances with the most relevant info
- ability to filter down that table
- ability to perform bulk updates on them
- ability to see "status" of each server at a glance
- server and audiit logs are one click away

sangee2004:
We now have `Deployment and Connection` tab presented in MCP Serve page with all individual deployed server instances listed here.

Issues found when testing this feature will be tracked in separate issues.

Following issue have been logged so far relating to these changes:

https://github.com/obot-platform/obot/issues/4575
https://github.com/obot-platform/obot/issues/4576
https://github.com/obot-platform/obot/issues/4577
https://github.com/obot-platform/obot/issues/4579
https://github.com/obot-platform/obot/issues/4581
https://github.com/obot-platform/obot/issues/4582
https://github.com/obot-platform/obot/issues/4584
https://github.com/obot-platform/obot/issues/4586

PR #4553, opened 2025-10-07
feat: deployed servers admin tab for MCP Servers

ivyjeong13:
Addresses #4324

### DIFF
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -2417,9 +2417,9 @@ func (m *MCPHandler) ListServerInstances(req api.Context) error {
 	}
 
 	// Filter instances that belong to servers in this catalog
-	var catalogServerNames = make(map[string]bool)
+	var catalogServerNames = make(map[string]struct{})
 	for _, server := range catalogServers {
-		catalogServerNames[server.Name] = true
+		catalogServerNames[server.Name] = struct{}{}
 	}
 
 	var filteredInstances []v1.MCPServerInstance
@@ -2428,7 +2428,7 @@ func (m *MCPHandler) ListServerInstances(req api.Context) error {
 			// Hide template instances
 			continue
 		}
-		if catalogServerNames[instance.Spec.MCPServerName] {
+		if _, exists := catalogServerNames[instance.Spec.MCPServerName]; exists {
 			filteredInstances = append(filteredInstances, instance)
 		}
 	}


pkg/api/router/router.go
```diff
@@ -455,6 +455,7 @@ func Router(services *services.Services) (http.Handler, error) {
 	mux.HandleFunc("PUT /api/mcp-catalogs/{catalog_id}/entries/{entry_id}", mcpCatalogs.UpdateEntry)
 	mux.HandleFunc("DELETE /api/mcp-catalogs/{catalog_id}/entries/{entry_id}", mcpCatalogs.DeleteEntry)
 	mux.HandleFunc("GET /api/mcp-catalogs/{catalog_id}/entries/{entry_id}/servers", mcpCatalogs.AdminListServersForEntryInCatalog)
+	mux.HandleFunc("GET /api/mcp-catalogs/{catalog_id}/entries/all-servers", mcpCatalogs.AdminListServersForAllEntriesInCatalog)
```
ivyjeong13, line 458: Needed an endpoint/endpoints to get all deployed servers for all catalog entries in a catalog as well as all deployed servers for all entries in all workspaces. Vibe-coded my way through to get these, would like a backend check on these just in case! 😅

pkg/api/handlers/poweruserworkspace.go
```diff
@@ -148,6 +148,104 @@ func (p *PowerUserWorkspaceHandler) ListAllServers(req api.Context) error {
...
+func (p *PowerUserWorkspaceHandler) ListAllServersForAllEntries(req api.Context) error {
+	// Get all workspaces
+	var workspaceList v1.PowerUserWorkspaceList
+	if err := req.List(&workspaceList); err != nil {
+		return fmt.Errorf("failed to list power user workspaces: %w", err)
+	}
+
+	// Get all entries from all workspaces
+	var allEntries []v1.MCPServerCatalogEntry
+	var entryList v1.MCPServerCatalogEntryList
```
njhale:
```suggestion
	var (
	    allEntries []v1.MCPServerCatalogEntry
	    entryList v1.MCPServerCatalogEntryList
	)
```

pkg/api/handlers/poweruserworkspace.go
```diff
@@ -148,6 +148,104 @@ func (p *PowerUserWorkspaceHandler) ListAllServers(req api.Context) error {
...
+	if err := req.List(&entryList); err != nil {
+		return fmt.Errorf("failed to list entries: %w", err)
+	}
+
+	// Create a map of workspace names for efficient lookup
+	workspaceNames := make(map[string]bool)
+	for _, workspace := range workspaceList.Items {
+		workspaceNames[workspace.Name] = true
+	}
```
njhale: This can be initialized with a size and it's idiomatic to use the empty struct instead of bool for sets in go:



```suggestion
	workspaceNames := make(map[string]struct{}, len(workspaceList.Items))
	for _, workspace := range workspaceList.Items {
		workspaceNames[workspace.Name] = struct{}{}
	}
```

pkg/api/handlers/poweruserworkspace.go
```diff
@@ -148,6 +148,104 @@ func (p *PowerUserWorkspaceHandler) ListAllServers(req api.Context) error {
...
+	for _, server := range allServers {
+		if server.Spec.Template {
+			continue
+		}
+		if seenServers[server.Name] {
+			continue
+		}
+		seenServers[server.Name] = true
+		filteredServers = append(filteredServers, server)
+	}
```
njhale:
```suggestion
	var (
	    filteredServers []v1.MCPServer
	    seenServers = make(map[string]bool)
	)
	for _, server := range allServers {
		if _, seen := seenServers[server.Name]; seen || server.Spec.Template {
			continue
		}
		
		seenServers[server.Name] = true
		filteredServers = append(filteredServers, server)
	}
```

pkg/api/authz/authz.go
```diff
@@ -50,6 +50,7 @@ var (
 		"/api/cronjobs/",
 		"/api/mcp-catalogs",
 		"/api/mcp-catalogs/",
+		"GET /api/mcp-catalogs/{catalog_id}/entries/all-servers",
```
njhale: I believe this is already handled by the `/api/mcp-catalogs/` (trailing slash) line above.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -449,6 +449,81 @@ func (h *MCPCatalogHandler) AdminListServersForEntryInCatalog(req api.Context) e
...
+		return fmt.Errorf("failed to list entries: %w", err)
+	}
+
+	// Filter entries that belong to this catalog
+	var catalogEntries []v1.MCPServerCatalogEntry
+	for _, entry := range entriesList.Items {
+		if entry.Spec.MCPCatalogName == catalogName {
+			catalogEntries = append(catalogEntries, entry)
+		}
+	}
```
g-linville: nit: it would be easier (and possibly more efficient? not sure) to do this with a field selector on `"spec.mcpCatalogName"`

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -449,6 +449,81 @@ func (h *MCPCatalogHandler) AdminListServersForEntryInCatalog(req api.Context) e
...
+		if server.Spec.Template {
+			// Hide template servers
+			continue
+		}
+		if seenServers[server.Name] {
+			continue
+		}
+		seenServers[server.Name] = true
+		filteredServers = append(filteredServers, server)
+	}
```
g-linville: Duplicates are not possible, as each server has a uniquely generated ID, and each will only show up once in the list.

Also, for safety, we should filter out workspace servers here too. Unless something strange has happened, there shouldn't be any in the list, but it would make me feel better, lol.

```suggestion
	// Filter out template servers and servers in workspaces
	var filteredServers []v1.MCPServer
	for _, server := range allServers {
		if server.Spec.Template || server.Spec.PowerUserWorkspaceID != "" {
			// Hide template servers and servers in workspaces.
			// Servers in workspaces should not be possible,
			// unless somehow someone (like an admin) created one from
			// an entry in the default catalog.
			// Though the UI does not expose the ability to do this,
			// nor would ordinary users have the authz rules to allow them to.
			continue
		}

		filteredServers = append(filteredServers, server)
	}
```

pkg/api/handlers/poweruserworkspace.go
```diff
@@ -148,6 +148,104 @@ func (p *PowerUserWorkspaceHandler) ListAllServers(req api.Context) error {
...
+
+	// Create a map of workspace names for efficient lookup
+	workspaceNames := make(map[string]bool)
+	for _, workspace := range workspaceList.Items {
+		workspaceNames[workspace.Name] = true
+	}
+
+	// Filter entries that belong to any workspace
+	for _, entry := range entryList.Items {
+		if entry.Spec.PowerUserWorkspaceID != "" && workspaceNames[entry.Spec.PowerUserWorkspaceID] {
```
g-linville: We don't need the `workspaceNames` lookup here. If an entry has a workspace ID set on it, it is guaranteed that that workspace exists. There are automatic cleanup handlers in the code to delete entries if their workspace gets deleted.

pkg/api/handlers/poweruserworkspace.go
g-linville: No need to filter for duplicates here as well

pkg/api/handlers/poweruserworkspace.go
```diff
@@ -148,6 +148,104 @@ func (p *PowerUserWorkspaceHandler) ListAllServers(req api.Context) error {
...
+			continue
+		}
+		seenServers[server.Name] = true
+		filteredServers = append(filteredServers, server)
+	}
+
+	// Build credential contexts for all filtered servers
+	credCtxs := make([]string, 0, len(filteredServers))
+	for _, server := range filteredServers {
+		credCtxs = append(credCtxs, fmt.Sprintf("%s-%s", server.Spec.PowerUserWorkspaceID, server.Name))
```
g-linville: The credential context to be used here is the one for the user, not for the workspace. The workspace context is to be used for multi-user MCP servers created in the workspace by the PUP. The user context is for servers created from catalog entries (single user or remote).

```suggestion
		credCtxs = append(credCtxs, fmt.Sprintf("%s-%s", server.Spec.UserID, server.Name))
```

pkg/api/handlers/poweruserworkspace.go
```diff
@@ -148,6 +148,104 @@ func (p *PowerUserWorkspaceHandler) ListAllServers(req api.Context) error {
...
+			}
+		}
+	}
+
+	servers := make([]types.MCPServer, 0, len(filteredServers))
+	for _, server := range filteredServers {
+		// Add extracted env vars to the server definition
+		addExtractedEnvVars(&server)
+
+		slug, err := slugForMCPServer(req.Context(), req.Storage, server, req.User.GetUID(), "", server.Spec.PowerUserWorkspaceID)
```
g-linville: Similarly, the slug here should only include the user ID and not the workspace ID, since the "container" that holds the server is just the user, not the workspace, which contained the catalog entry from which the server was created.

```suggestion
		slug, err := slugForMCPServer(req.Context(), req.Storage, server, req.User.GetUID(), "", "")
```

pkg/api/handlers/poweruserworkspace.go
```diff
@@ -183,6 +281,60 @@ func convertAccessControlRuleWithWorkspace(rule v1.AccessControlRule, powerUserI
 	}
 }
 
+func (p *PowerUserWorkspaceHandler) ListAllServersInstances(req api.Context) error {
```
g-linville: micro nit:

```suggestion
func (p *PowerUserWorkspaceHandler) ListAllServerInstances(req api.Context) error {
```

g-linville (changes requested):

ryu-man (commented):
nit: I found it a little bit hard to find selectable servers. I think it would be better to sort servers by selectable first and then all unselectable.

Screenshot 2025-10-08 210653

ui/user/src/routes/admin/mcp-servers/+page.svelte
```diff
@@ -213,179 +168,78 @@
...
-							>
-								<Trash2 class="size-4" />
-							</button>
-						{/if}
-						<button class="icon-button hover:text-blue-500" use:tooltip={'View Entry'}>
-							<Eye class="size-4" />
-						</button>
-					{/snippet}
-				</Table>
+				</RegistriesView>
```
ryu-man, line 225: nit: I think we should show the total of the selectable servers. rather than all servers

Screenshot 2025-10-08 210832

ui/user/src/routes/admin/mcp-servers/+page.svelte
```diff
@@ -213,179 +168,78 @@
...
+			{:else if view === 'urls'}
+				<SourceUrlsView
```
ryu-man, line 227: nit: Tabs are overcrowded on small devices; maybe it would be better to set text to no wrap and make them horizontally scrollable.

Screenshot 2025-10-08 210922

ui/user/src/routes/admin/mcp-servers/+page.svelte
```diff
@@ -213,179 +168,78 @@
...
+					catalog={defaultCatalog}
```
ryu-man, line 228: Table overflow is visible

Screenshot 2025-10-08 211023

ui/user/src/routes/admin/mcp-servers/s/[id]/details/+page.svelte
```diff
@@ -0,0 +1,68 @@
...
+	import { fly } from 'svelte/transition';
+
+	let { data } = $props();
+	let { mcpServer } = data;
+	let loading = $state(false);
+	let users = $state<OrgUser[]>([]);
+	let instances = $state<MCPServerInstance[]>([]);
+	let usersMap = $derived(new Map(users.map((u) => [u.id, u])));
+
+	onMount(async () => {
```
Comment (ryu-man) [inline ui/user/src/routes/admin/mcp-servers/s/[id]/details/+page.svelte:19]: nit: data is better to be loaded before the component is mounted, just to make sure the response comes asap

ryu-man (approved):
Other than the comments above, the front looks good.

ui/user/src/routes/admin/mcp-servers/+page.svelte
ivyjeong13: Yup, kind of a conflicting behavior when trying to use sticky and the parent has an overflow css but will address!

pkg/api/handlers/mcp.go
```diff
@@ -2383,3 +2383,68 @@ func (m *MCPHandler) TriggerUpdate(req api.Context) error {
...
+	}
+
+	// Get all instances for these catalog servers
+	var allInstances v1.MCPServerInstanceList
+	if err := req.List(&allInstances); err != nil {
+		return fmt.Errorf("failed to list server instances: %w", err)
+	}
+
+	// Filter instances that belong to servers in this catalog
+	var catalogServerNames = make(map[string]bool)
```
njhale: nit: this can be a `map[string]struct{}`

PR #4560, opened 2025-10-08
Support composite MCP servers

njhale:


cjellick:
I changed the issue to #4635

pkg/mcp/client.go
```diff
@@ -50,21 +50,21 @@ func (sm *SessionManager) ClientForServer(ctx context.Context, userID, mcpServer
 func (sm *SessionManager) clientForServerWithScope(ctx context.Context, clientScope, userID, mcpServerDisplayName, mcpServerName string, serverConfig ServerConfig) (*Client, error) {
 	clientName := "Obot MCP Gateway"
 	var tokenStorage nmcp.TokenStorage
-	if serverConfig.Runtime == types.RuntimeRemote && strings.HasPrefix(serverConfig.URL, fmt.Sprintf("%s/mcp-connect/", sm.baseURL)) {
-		// If the URL points back to us, then this is Obot chat. Ensure the client name reflects that.
+	if (serverConfig.Runtime == types.RuntimeRemote || serverConfig.Runtime == types.RuntimeComposite) && strings.HasPrefix(serverConfig.URL, fmt.Sprintf("%s/mcp-connect/", sm.baseURL)) {
```
thedadams: We don't want to do this.

pkg/mcp/types.go
```diff
@@ -236,17 +217,13 @@ func ServerToServerConfig(mcpServer v1.MCPServer, scope string, credEnv map[stri
...
-					val = applyPrefix(val, header.Prefix)
-				}
-
 				serverConfig.Headers = append(serverConfig.Headers, fmt.Sprintf("%s=%s", header.Key, val))
 			}
 		} else {
 			return serverConfig, missingRequiredNames, fmt.Errorf("runtime %s requires remote config", mcpServer.Spec.Manifest.Runtime)
 		}
+	case types.RuntimeComposite:
+		return serverConfig, nil, nil
```
thedadams, line 251: Do we want to check that the component MCP servers are all configured?

pkg/validation/mcpvalidators.go
```diff
@@ -416,13 +416,204 @@ func (v RemoteValidator) validateRemoteCatalogConfig(config types.RemoteCatalogC
...
+	if len(manifest.CompositeConfig.ComponentServers) == 0 {
+		return types.RuntimeValidationError{
+			Runtime: types.RuntimeComposite,
+			Field:   "compositeConfig.componentServers",
+			Message: "at least one component server is required",
+		}
+	}
+
+	// Check for duplicate component servers
+	seen := make(map[string]bool)
```
thedadams: nit:
```suggestion
	seen := make(map[string]bool, len(manifest.CompositeConfig.ComponentServers))
```

pkg/validation/mcpvalidators.go
```diff
@@ -416,13 +416,204 @@ func (v RemoteValidator) validateRemoteCatalogConfig(config types.RemoteCatalogC
...
```
thedadams: Same here and anywhere else in this file (I think there are two more below).

pkg/api/handlers/projectmcp.go
```diff
@@ -344,7 +344,7 @@ func (p *ProjectMCPHandler) GetOAuthURL(req api.Context) error {
 		return err
 	}
 
-	u, err := p.mcpOAuthChecker.CheckForMCPAuth(req.Context(), server, serverConfig, req.User.GetUID(), server.Name, "")
+	u, err := p.mcpOAuthChecker.CheckForMCPAuth(req, server, serverConfig, req.User.GetUID(), server.Name, "")
```
thedadams, line 347: Can we also do this?
```suggestion
	u, err := p.mcpOAuthChecker.CheckForMCPAuth(req, server, serverConfig, server.Name, "")
```

pkg/api/authz/resources.go
```diff
@@ -17,6 +17,10 @@ var apiResources = map[string][]string{
 		"GET    /api/all-mcps/servers/{mcpserver_id}/prompts/{prompt_name}",
 		"GET    /oauth/callback/{oauth_request_id}/{mcp_id}",
 		"GET    /oauth/mcp/callback",
+		// Composite MCP OAuth UI and API endpoints
+		"GET    /auth/mcp/composite/{mcp_id}",
+		// TODO(cmcp): Switch this back to /oauth/composite/{mcp_id}
```
thedadams: Do we need to do something here?

pkg/api/handlers/mcpgateway/oauth/handler.go
```diff
@@ -44,4 +44,7 @@ func SetupHandlers(gatewayClient *client.Client, oauthChecker *MCPOAuthHandlerFa
 
 	mux.HandleFunc("GET /oauth/jwks.json", h.tokenService.ServeJWKS)
 	mux.HandleFunc("POST /oauth/replace-jwks", h.tokenService.ReplaceJWK)
+
+	// TODO(cmcp): Switch this to /oauth/composite/{mcp_id}
```
thedadams: Same question here.

pkg/api/handlers/mcpgateway/clientmessagehandler.go
```diff
@@ -185,6 +185,7 @@ func (c *clientMessageHandler) handleMessage(ctx context.Context, msg nmcp.Messa
 
 	var ch <-chan nmcp.Message
 	if msg.ID != nil {
+		msg.ID = uuid.NewString()
```
thedadams, line 190: I know we talked about this, but a comment explaining why we are doing this would be nice here.

pkg/api/handlers/mcpgateway/oauth/mcpoauthhandler.go
```diff
@@ -38,18 +40,61 @@ func NewMCPOAuthHandlerFactory(baseURL string, sessionManager *mcp.SessionManage
 		tokenStore:        globalTokenStore,
 	}
 }
-func (f *MCPOAuthHandlerFactory) CheckForMCPAuth(ctx context.Context, mcpServer v1.MCPServer, mcpServerConfig mcp.ServerConfig, userID, mcpID, oauthAppAuthRequestID string) (string, error) {
-	if mcpServerConfig.Runtime != types.RuntimeRemote {
-		// OAuth is only support for remote MCP servers.
+func (f *MCPOAuthHandlerFactory) CheckForMCPAuth(req api.Context, mcpServer v1.MCPServer, mcpServerConfig mcp.ServerConfig, userID, mcpID, oauthAppAuthRequestID string) (string, error) {
+	switch mcpServerConfig.Runtime {
+	case types.RuntimeRemote:
+	case types.RuntimeComposite:
```
thedadams: I think this is a time where the switch statement is more confusing. I thought this was a bug at first.

pkg/api/handlers/mcpgateway/oauth/mcpoauthhandler.go
```diff
@@ -38,18 +40,61 @@ func NewMCPOAuthHandlerFactory(baseURL string, sessionManager *mcp.SessionManage
...
+			if err != nil {
+				continue
+			}
+
+			u, err := f.CheckForMCPAuth(req, componentServer, componentConfig, userID, componentServer.Name, oauthAppAuthRequestID)
+			if err != nil {
+				if req.Context().Err() != nil {
+					return "", fmt.Errorf("failed to check component server OAuth: %w", req.Context().Err())
+				}
+				continue
```
thedadams: Should we error here no matter what?

apiclient/types/mcpserver.go
```diff
@@ -54,6 +55,34 @@ type RemoteCatalogConfig struct {
...
+
+type CompositeRuntimeConfig struct {
+	ComponentServers []ComponentServer `json:"componentServers"`
+}
+
+type ComponentServer struct {
+	// CatalogEntryID is used when the component server is a single-user or remote server.
+	CatalogEntryID string `json:"catalogEntryID"`
+	// MCPServerID is used when the component server is a multi-user server.
+	MCPServerID     string            `json:"mcpServerID,omitempty"`
```
thedadams: nit:
```suggestion
	MultiUserMCPServerID     string            `json:"multiUserMCPServerID,omitempty"`
```

pkg/api/handlers/mcpgateway/oauth/authorize.go
```diff
@@ -369,6 +369,26 @@ func (h *handler) oauthCallback(req api.Context) error {
 		return nil
 	}
 
+	// Check if the MCP server is a component of a composite; only finalize if it's not
+	_, server, _, err := handlers.ServerForActionWithConnectID(req, mcpID)
```
thedadams: Based on my understanding of the code, the `mcpID` here will always be an MCPServer name. I think we can get the MCP server object directly here instead of using this special function.

pkg/api/handlers/mcpgateway/oauth/composite.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// checkCompositeAuth checks if the composite OAuth flow is complete.
+// If it is not complete, it returns the list of component OAuth URLs still needed (respecting session-scoped skips).
+func (h *handler) checkCompositeAuth(req api.Context) error {
+	var (
+		compositeMCPID     = req.PathValue("mcp_id")
+		oauthAuthRequestID = req.URL.Query().Get("oauth_auth_request")
+	)
+	_, compositeServer, _, err := handlers.ServerForActionWithConnectID(req, compositeMCPID)
```
thedadams: Same comment here.

pkg/api/handlers/mcpgateway/handler.go
```diff
@@ -103,26 +104,69 @@ func (h *Handler) StreamableHTTP(req api.Context) error {
...
+
+	req.Request = req.WithContext(withMessageContext(req.Context(), messageCtx))
 
 	nmcp.NewHTTPServer(nil, h, nmcp.HTTPServerOptions{SessionStore: h}).ServeHTTP(req.ResponseWriter, req.Request)
 
 	return nil
 }
 
-type messageContext struct {
+type serverContext struct {
```
thedadams: Is this necessary?

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -311,6 +315,7 @@ func (h *MCPCatalogHandler) CreateEntry(req api.Context) error {
 }
 
 func (h *MCPCatalogHandler) UpdateEntry(req api.Context) error {
+	// TODO(cmcp): Update this handler to handle updating composite entries
```
thedadams: Is this still a TODO?

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -664,9 +676,12 @@ func (h *MCPCatalogHandler) GetServerFromEntry(req api.Context) error {
...
 // to generate tool preview data, then cleans up the instance.
 func (h *MCPCatalogHandler) GenerateToolPreviews(req api.Context) error {
-	catalogName := req.PathValue("catalog_id")
-	workspaceID := req.PathValue("workspace_id")
-	entryName := req.PathValue("entry_id")
+	var (
+		catalogName = req.PathValue("catalog_id")
+		workspaceID = req.PathValue("workspace_id")
+		entryName   = req.PathValue("entry_id")
+		dryRun      = req.Request.URL.Query().Get("dryRun") == "true"
```
thedadams: What is a dry run?

pkg/api/handlers/mcp.go
```diff
@@ -965,11 +1005,11 @@ func ServerForAction(req api.Context, id string) (v1.MCPServer, mcp.ServerConfig
 		return server, mcp.ServerConfig{}, err
 	}
 
-	serverConfig, err := serverConfigForAction(req, server)
+	serverConfig, err := ServerConfigForAction(req, server)
 	return server, serverConfig, err
 }
 
-func serverConfigForAction(req api.Context, server v1.MCPServer) (mcp.ServerConfig, error) {
+func ServerConfigForAction(req api.Context, server v1.MCPServer) (mcp.ServerConfig, error) {
```
thedadams: I would prefer that this stay unexported.

pkg/api/handlers/mcpgateway/oauth/mcpoauthhandler.go
njhale: I'll switch to an if-else

pkg/api/handlers/mcpcatalogs.go
njhale: "dryRun" lets us get the previews for an MCP server without updating its CatalogEntry. This is used when we populate the tools for individual MCP servers when creating a composite CatalogEntry (configuring tool overrides). Without this, `GenerateToolPreviews` would attempt to update the entry we're generating previews for, and fail if that entry was sourced from git.

pkg/api/handlers/projectmcp.go
njhale: Get the user ID from the `req.User` in the method itself? I don't see why not.

njhale: Took a second look here and it seems like we pass `system` for this field when generating tool previews, so I don't think we should change this for now.

pkg/mcp/types.go
njhale: Not all of them need to be configured for composite MCP servers to work though. They should work with at least one configured.

pkg/api/handlers/mcpgateway/oauth/authorize.go
```diff
@@ -341,7 +341,7 @@ func (h *handler) callback(req api.Context) error {
 
 // oauthCallback handles the second-level third-party OAuth for MCP servers.
 func (h *handler) oauthCallback(req api.Context) error {
-	oauthAuthRequestID, err := h.oauthChecker.stateCache.createToken(req.Context(), req.URL.Query().Get("state"), req.URL.Query().Get("code"), req.URL.Query().Get("error"), req.URL.Query().Get("error_description"))
+	oauthAuthRequestID, mcpID, err := h.oauthChecker.stateCache.createToken(req.Context(), req.URL.Query().Get("state"), req.URL.Query().Get("code"), req.URL.Query().Get("error"), req.URL.Query().Get("error_description"))
```
thedadams: nano-nit: I know you didn't name this variable, but it is really an `mcpServerID`. Using `mcpID` implicitly means that it could be an MCP server ID, an MCP server instance ID, or an MCP catalog entry ID.

If it is straightforward, we should rename this variable to `mcpServerID` here and where it is referenced in the session storage.

pkg/api/handlers/mcpgateway/oauth/mcpoauthhandler.go
```diff
@@ -38,18 +40,70 @@ func NewMCPOAuthHandlerFactory(baseURL string, sessionManager *mcp.SessionManage
...
+			if err != nil {
+				continue
+			}
+
+			u, err := f.CheckForMCPAuth(req, componentServer, componentConfig, userID, componentServer.Name, oauthAppAuthRequestID)
+			if err != nil {
+				if req.Context().Err() != nil {
+					return "", fmt.Errorf("failed to check component server OAuth: %w", req.Context().Err())
+				}
+			}
```
thedadams, line 81: Do we want to continue if `err != nil` here?

pkg/api/handlers/mcpgateway/handler.go
```diff
@@ -121,8 +176,10 @@ type messageContext struct {
 	userID, mcpID string
 	mcpServer     v1.MCPServer
 	serverConfig  mcp.ServerConfig
-	req           *http.Request
-	resp          http.ResponseWriter
+	enabled       bool
+	compositeContext
```
thedadams: nit: move embedded fields to the top of the struct.

pkg/api/handlers/mcpgateway/handler.go
```diff
@@ -121,8 +176,10 @@ type messageContext struct {
 	userID, mcpID string
 	mcpServer     v1.MCPServer
 	serverConfig  mcp.ServerConfig
-	req           *http.Request
-	resp          http.ResponseWriter
+	enabled       bool
```
thedadams: I think it is worth adding a comment here to indicate that this field is only used for composite MCP servers.

pkg/api/handlers/mcpgateway/composite.go
```diff
@@ -0,0 +1,610 @@
...
+	if catalogName == "" && m.mcpServer.Spec.MCPServerCatalogEntryName != "" {
+		var entry v1.MCPServerCatalogEntry
+		if err := h.storageClient.Get(ctx, kclient.ObjectKey{Namespace: m.mcpServer.Namespace, Name: m.mcpServer.Spec.MCPServerCatalogEntryName}, &entry); err != nil {
+			log.Errorf("Failed to get catalog for server %s: %v", m.mcpServer.Name, err)
+			return
+		}
+		catalogName = entry.Spec.MCPCatalogName
+	}
+
+	for _, componentServer := range m.componentServers {
```
thedadams, line 128: Need to fire webhooks before getting here. Also need to firewebhooks after getting/compiling the responses.

pkg/api/handlers/mcpgateway/composite.go
```diff
@@ -0,0 +1,610 @@
...
+		return
+
+	case methodNotificationsProgress, methodNotificationsRootsListChanged, methodNotificationsCancelled, methodLoggingSetLevel:
+		// These methods don't require a result.
+		result = nmcp.Notification{}
+		for _, client := range clients {
+			if err = client.Session.Exchange(ctx, msg.Method, &msg, &result); err != nil {
+				log.Errorf("Failed to send %s message to server %s: %v", msg.Method, client.mcpID, err)
+				return
+			}
```
thedadams, line 448: We don't want to return on the first error.

- progress is really only meant for one of the component servers. The others are likely to error.
- Not every component server will support logging. Those that don't are likely to error.

pkg/api/handlers/mcpgateway/composite.go
```diff
@@ -0,0 +1,610 @@
...
+	}
+
+	// Merge Prompts capability
+	if compositePrompts, componentPrompts := composite.Capabilities.Prompts, component.Capabilities.Prompts; componentPrompts != nil && (compositePrompts == nil || !compositePrompts.ListChanged) {
+		composite.Capabilities.Prompts = &nmcp.PromptsServerCapability{
+			ListChanged: componentPrompts.ListChanged,
+		}
+	}
+
+	return composite
```
thedadams, line 509: Should we include logging and resources in this?

pkg/api/handlers/mcp.go
```diff
@@ -969,6 +1027,35 @@ func ServerForAction(req api.Context, id string) (v1.MCPServer, mcp.ServerConfig
 	return server, serverConfig, err
 }
 
+func (m *MCPHandler) serverConfigForAction(req api.Context, server v1.MCPServer) (mcp.ServerConfig, error) {
```
thedadams: Having a function and a method with the same name here confused me.

How do you feel about add ing the `tokenService` as a parameter to the function version of `serverConfigForAction` and moving this logic there?

pkg/api/handlers/mcp.go
```diff
@@ -1028,18 +1115,66 @@ func serverForAction(req api.Context) (v1.MCPServer, mcp.ServerConfig, error) {
 	return ServerForAction(req, req.PathValue("mcp_server_id"))
 }
 
-func serverForActionWithCapabilities(req api.Context, mcpSessionManager *mcp.SessionManager) (v1.MCPServer, mcp.ServerConfig, nmcp.ServerCapabilities, error) {
-	server, serverConfig, err := serverForAction(req)
+func (m *MCPHandler) serverForActionWithCapabilities(req api.Context) (v1.MCPServer, mcp.ServerConfig, nmcp.ServerCapabilities, error) {
```
thedadams: My opinion: either all of these function should be methods on `*MCPHandler` or they should all be functions. Prior to this change, they were all functions so they could be easily used in other areas of the code if necessary.

pkg/api/handlers/mcp.go
```diff
@@ -1115,6 +1250,16 @@ func (m *MCPHandler) CreateServer(req api.Context) error {
 		input.MCPServerManifest.RemoteConfig.URL = "https://" + input.MCPServerManifest.RemoteConfig.URL
 	}
 
+	// Normalize component URLs for composite input
+	if input.MCPServerManifest.CompositeConfig != nil {
+		for i := range input.MCPServerManifest.CompositeConfig.ComponentServers {
+			rc := input.MCPServerManifest.CompositeConfig.ComponentServers[i].Manifest.RemoteConfig
+			if rc != nil && rc.URL != "" && !strings.HasPrefix(rc.URL, "http") {
+				rc.URL = "https://" + rc.URL
+			}
```
thedadams: Do we do this for remote MCP servers? If not, then I think we should be consistent here.

pkg/api/handlers/mcp.go
```diff
@@ -1428,6 +1655,149 @@ func (m *MCPHandler) ConfigureServer(req api.Context) error {
...
+
+		// Persist disabled state to parent composite manifest if present
+		if idx, ok := parentComps[component.Spec.MCPServerCatalogEntryName]; ok && compositeServer.Spec.Manifest.CompositeConfig != nil {
+			compositeServer.Spec.Manifest.CompositeConfig.ComponentServers[idx].Disabled = componentConfig.Disabled
+		}
+
+		if componentConfig.Disabled {
+			// On disable, clear remote URL (if any) and persist; credentials were already cleared above
+			if component.Spec.Manifest.RemoteConfig != nil {
+				component.Spec.Manifest.RemoteConfig.URL = ""
```
thedadams: If this URL is long and complicated, then the user would have to provide it again to enable right? I think it would be easier on our users if we had a different way of disabling this.

pkg/api/handlers/mcp.go
```diff
@@ -1428,6 +1655,149 @@ func (m *MCPHandler) ConfigureServer(req api.Context) error {
...
+			}
+			if err := req.Update(&component); err != nil {
+				return fmt.Errorf("failed to update disabled component server %s: %w", component.Name, err)
+			}
+			if compositeServer.Spec.Manifest.CompositeConfig != nil {
+				if err := req.Update(&compositeServer); err != nil {
```
thedadams: We want to avoid calling update on this composite server multiple times in a single API request. Can we save it for the end?

pkg/api/handlers/mcp.go
```diff
@@ -1520,6 +1938,71 @@ func (m *MCPHandler) Reveal(req api.Context) error {
...
+		}
+
+		cfg := map[string]string{}
+		if err == nil {
+			for k, v := range compCred.Env {
+				if v != "" {
+					cfg[k] = v
+				}
+			}
+		}
```
```suggestion
		for k, v := range compCred.Env {
			if v != "" {
				cfg[k] = v
			}
		}
```

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -664,9 +678,12 @@ func (h *MCPCatalogHandler) GetServerFromEntry(req api.Context) error {
...
```
thedadams: Please add a comment explaining what `dryRun` is.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -750,9 +851,12 @@ func (h *MCPCatalogHandler) GenerateToolPreviews(req api.Context) error {
...
 
 func (h *MCPCatalogHandler) GenerateToolPreviewsOAuthURL(req api.Context) error {
-	catalogName := req.PathValue("catalog_id")
-	workspaceID := req.PathValue("workspace_id")
-	entryName := req.PathValue("entry_id")
+	var (
+		catalogName = req.PathValue("catalog_id")
+		workspaceID = req.PathValue("workspace_id")
+		entryName   = req.PathValue("entry_id")
+		dryRun      = req.Request.URL.Query().Get("dryRun") == "true"
```
thedadams: Same comment about adding a comment.

pkg/mcp/loader.go
```diff
@@ -297,7 +298,10 @@ func (sm *SessionManager) ensureDeployment(ctx context.Context, server ServerCon
 				}
 			}
 		}
-		// This is a remote MCP server, so there is nothing to deploy.
+
+		return server, nil
+	case otypes.RuntimeComposite:
+		server.URL = fmt.Sprintf("%s/mcp-connect/%s", sm.baseURL, mcpServerName)
```
thedadams: Ideally, this transformation would completely transform the server config to "remote".

pkg/mcp/types.go
```diff
@@ -294,3 +296,21 @@ func ProjectServerToConfig(tokenService *ephemeral.TokenService, projectMCPServe
...
+	})
+	if err != nil {
+		return ServerConfig{}, fmt.Errorf("failed to create token: %w", err)
+	}
+
+	return ServerConfig{
+		URL:          fmt.Sprintf("%s/mcp-connect/%s", baseURL, mcpServer.Name),
+		Headers:      []string{fmt.Sprintf("Authorization=Bearer %s", token)},
+		Scope:        fmt.Sprintf("%s-%s", mcpServer.Name, scope),
+		AllowedTools: []string{},
```
thedadams:
```suggestion
```

pkg/api/handlers/mcpgateway/oauth/mcpoauthhandler.go
njhale: Disabled components are already skipped in the code above, so I'm on the fence about whether to continue here. If we return the error here, at least it can bubble up to the user.

pkg/api/handlers/mcpgateway/oauth/authorize.go
njhale: I took a look and the naming goes all the way down to the `MCPOAuthToken`, so it would be a larger update all references IIUC. I renamed the variable here and can follow up with renaming after I get composites squared away if that works for you.

pkg/api/handlers/mcpgateway/handler.go
njhale: So we're actually dropping disabled components from the `messageContext` before we even attempt to create the clients for them, so this field is no longer necessary. I'm going to remove it entirely.

pkg/api/handlers/mcpgateway/composite.go
njhale: I switched this to continue. LMK if you feel we need something more complex.

pkg/api/handlers/mcpgateway/composite.go
njhale: So resources are explicitly out of scope, which is why they are purposefully excluded here; I added logging to this though.

pkg/api/handlers/mcpgateway/composite.go
njhale: Id like to follow up with a fix for this so that Ivy can start iterating on the frontend if possible.

pkg/api/handlers/mcp.go
njhale: we do this a few lines above when creating regular remote mcp servers.

pkg/api/handlers/mcp.go
njhale: Done

pkg/api/handlers/mcp.go
njhale: Done.

pkg/api/handlers/mcpgateway/handler.go
njhale: Done

pkg/api/handlers/mcp.go
njhale: do you have any suggestions there? I was thinking we could just leave the URL on the remote config when it's disabled.

pkg/api/handlers/mcpcatalogs.go
njhale: Done.

pkg/mcp/types.go
njhale: Done

pkg/mcp/loader.go
njhale: Done

pkg/api/handlers/mcp.go
njhale: I changed all of these back to functions and plumbed the dependencies in as parameters.

pkg/mcp/types.go
```diff
@@ -296,3 +296,25 @@ func ProjectServerToConfig(tokenService *ephemeral.TokenService, projectMCPServe
...
+		UserGroups: []string{types.GroupBasic},
+	})
+	if err != nil {
+		return ServerConfig{}, fmt.Errorf("failed to create token: %w", err)
+	}
+
+	finalScope := scope
+	if !strings.HasPrefix(scope, mcpServerName+"-") {
+		finalScope = fmt.Sprintf("%s-%s", mcpServerName, scope)
+	}
```
thedadams: I think the scope should be what the scope is at this point. We don't need to modify it.

njhale: Fixed

PR #4562, opened 2025-10-08
chore: require a db dsn to be set unless a special dev value is enabled

drpebcak:


chart/templates/secret.yaml
```diff
@@ -1,4 +1,7 @@
 {{- if not .Values.config.existingSecret -}}
+{{- if not .Values.dev.useEmbeddedDb }}
+# {{- required "You must set .Values.config.OBOT_SERVER_DSN for production use. Otherwise, you can set .Values.dev.useEmbeddedDb to true for testing purposes." .Values.config.OBOT_SERVER_DSN | b64enc }}
```
StrongMonkey, line 3: This should be uncommented?

drpebcak: No, its commented on purpose so that it doesnt effect the final yaml

PR #4583, opened 2025-10-09
feat: support for pod annotations

nlamirault:


cjellick:
@nlamirault - purely out of curiosity - what do you plan to use this for? I assume it is allow the pod to get picked up by some controller you have running in the cluster but just wanted to hear the details if youre able to share.

chart/templates/deployment.yaml
```diff
@@ -13,9 +13,14 @@ spec:
...
   template:
     metadata:
-      {{- if not .Values.config.existingSecret }}
+      {{- if or .Values.podAnnotations (not .Values.config.existingSecret)  }}
       annotations:
+        {{- if not .Values.config.existingSecret }}
         checksum/config-secret: {{ include (print $.Template.BasePath "/secret.yaml") . | sha256sum }}
+        {{- end }}
+        {{- with .Values.podAnnotations }}
+        {{- toYaml . | nindent 4 }}
```
drpebcak: This needs to be indented 8 spaces to be valid

```suggestion
        {{- toYaml . | nindent 8 }}
```

chart/templates/deployment.yaml
```diff
@@ -13,9 +13,14 @@ spec:
       {{- include "obot.selectorLabels" . | nindent 6 }}
   template:
     metadata:
-      {{- if not .Values.config.existingSecret }}
+      {{- if or .Values.podAnnotations (not .Values.config.existingSecret)  }}
```
drpebcak: This is a little complicated for my taste. Might be better to wrap each item in conditionals and just leave `annotations` blank if there are none in the final template


```suggestion
        {{- if not .Values.config.existingSecret }}
        checksum/config-secret: {{ include (print $.Template.BasePath "/secret.yaml") . | sha256sum }}
        {{- end }}
        {{- with .Values.podAnnotations }}
        {{- toYaml . | nindent 8 }}
        {{- end }}
```

Issue #4160, opened 2025-09-05
There is unnecessary tool  calls being made to fetch groups and users when creating ACL rules and when accessing existing ACL rules.


sangee2004:
**Describe the bug**
There is unnecessary tool calls being made to fetch groups and users when creating ACL rules and when accessing existing ACL rules

Pre Req - Have an `Microsoft Entra` auth enabled setup
**To Reproduce**

**Scenario 1:**

1. Log in as admin user
2. Go to `Access Control` page and. Click on `Add New rule`  
Notice that a call to `/groups` and /users` is made when there is no need to make since group/users are not presented anywhere at this time

4. Then add users/group using "+ Add Users/Groups` 
Notice that another call to `/groups` and `/users` is made

<img width="1026" height="164" alt="Image" src="https://github.com/user-attachments/assets/080416f3-9255-4aad-9a5c-7fc5951e90fb" />

https://www.loom.com/share/5d8fbf1aa06249b69efa7b3c0c653028?sid=3ccd5d56-8ed2-4c10-a587-8e8e7733517d

**Scenario 2:**
1. Log in as admin user
2. Go to `Access Control` page and. 
3. Open any existing ACL rule
Notice that a call to `/groups` and `/users` is made when there is no need to make these calls.

https://www.loom.com/share/62ceb7c28b7f4d95b6bca923fa396325?sid=9fb5833f-be4e-496c-b1f6-1c9d6f7935a9

**Expected behavior**
UI should not be making tool calls o fetch groups and users when creating ACL rules and when accessing existing ACL rules.

cjellick:
i dont know if this is valid or not. @ivyjeong13 going to throw it to you for assessment.

cjellick:
sorry maybe better for @njhale to look at this

njhale:
The page pre-populates the users and groups at page load to map names and photos to users and groups, even if none are present; we can improve on this, but not doing that would likely be a minor improvement.

When you click `Add Users/Groups`, it makes the calls again to get fresh data, especially for the live group search. I think this part can be improved, especially if we don't want to perform a live group search. We could just get all the cached groups (from users that have logged in, up-front), and filter that client-side.

sangee2004:
@ryu-man Could you please let me know what fixes have been made to address this issue?

For Scenario 1 , I see that there is no calls to `/groups` and `/users` when `Add New rule` is clicked.

[...]

PR #4588, opened 2025-10-09
feat: intialize a KV store solution built on `IndexedDB` to cache frequent used data

ryu-man:
Addresses #4160

- This PR helps reduce the number of times to call the API by caching data locally for 5 mn.
- Currently it is used to cache `users` & `groups`

### DIFF
--- a/ui/user/src/lib/kv/kv.ts
+++ b/ui/user/src/lib/kv/kv.ts
@@ -2,11 +2,9 @@ import { getContext, setContext } from 'svelte';
 
 const CONTEXT_KEY = '@obot/context/kv';
 
-type Callback = (data?: unknown) => void;
-
 /**
  * Production-ready IndexedDB Key-Value Store with expiration support
- * Features: TTL, batch operations, automatic cleanup, listeners, and error handling
+ * Features: TTL,  cleanup, and error handling
  */
 export class KV {
 	dbName: string;
@@ -14,7 +12,6 @@ export class KV {
 	version: number;
 	db: IDBDatabase | null;
 	cleanupInterval: number | null;
-	listeners: Map<string, Callback[]>;
 	initPromise: Promise<IDBDatabase>;
 
 	constructor(dbName = 'obot', storeName = 'main', version = 1) {
@@ -23,7 +20,6 @@ export class KV {
 		this.version = version;
 		this.db = null;
 		this.cleanupInterval = null;
-		this.listeners = new Map();
 		this.initPromise = this.init();
 	}
 
@@ -97,7 +93,6 @@ export class KV {
 			const request = store.put(item);
 
 			request.onsuccess = () => {
-				this.emit('set', { key, value, ttlMs: ttl });
 				resolve(value);
 			};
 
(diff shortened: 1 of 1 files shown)


ui/user/src/lib/kv/kv.ts
```diff
@@ -0,0 +1,548 @@
...
+
+const CONTEXT_KEY = '@obot/context/kv';
+
+type Callback = (data?: unknown) => void;
+
+/**
+ * Production-ready IndexedDB Key-Value Store with expiration support
+ * Features: TTL, batch operations, automatic cleanup, listeners, and error handling
+ */
+export class KV {
```
ivyjeong13, line 9: Maybe just me, but this seems like overkill for a simple cache to fetch users and groups every 5 minutes?

ryu-man: Oh, I think I can make it more simple.

ryu-man: Remove batch operation and listeners logic. kept only the need methods

ryu-man: If you meant going with IndexedDB was an overkill, then the reason I chose it is because it does handle large data better than `localstorage`, and because users and groups will grow in the future.

ivyjeong13: I guess just feels like a svelte store to hold/pass around the user/groups and do a refetch every 5 mins would be enough currently? I'd assume we'd switch over to backend pagination as we scale.

I'm actually thinking wouldn't the audit logs instead be a better use case for indexedDB?

ryu-man: > I guess just feels like a svelte store to hold/pass around the user/groups and do a refetch every 5 mins would be enough currently?

The downside of a store is that it is not persistent. If the user reloads the page before timeout, then it will fetch data again.

> I'm actually thinking wouldn't the audit logs instead be a better use case for indexedDB?

Yes, I do agree. But the case of audit logs is that data is dynamic due to filtering. So it looks to me that using the store would not be efficient; the only good reason to use it is to load data ASAP before fetching from the API, which is possibly outdated.

ivyjeong13: Yeah, that sounds okay to me, make one call on page load then cache for 5mins?

IndexedDB still feels overkill to me over a store but not a blocker as it not's negatively impacting!

Since we are caching the fetching of user/groups, we should go ahead and utilize it everywhere else and not just the access-control page. Loom for what I mean:
https://www.loom.com/share/125858277f78402691e6178ed32e1cfe

ryu-man: I do totally agree on using caching with other entities across the application. The current implementation addresses the issues on the access control. I will go ahead and implement an expanded solution to be used  in the whole app

cjellick:
I'm not comfortable with this change. Let's not pursue it for now. Right now my priority will just be to make that groups call faster.

cjellick:
dont think we're going to do this

Issue #4603, opened 2025-10-10
Invalid characters in MCP server names cause issues in nanobot that prevent npx servers from running


g-linville:
If you set up an MCP server with invalid characters like `@` and `/`, then Obot will allow the configuration, but Nanobot will fail to launch the npx command for some reason and never appear ready. This is likely a bug in Nanobot itself, or we could be missing some input validation in Obot. This likely affects uvx servers as well, but I haven't tried it there.

The nanobot version where I saw this is `v0.0.0-dev+122be441`, which is the version in the newest `ghcr.io/obot-platform/mcp-images/phat:main` image at the time of writing.

g-linville:
After using Claude to analyze the code, I am pretty sure that the `/` is an issue, and nanobot might be parsing the MCP server name and dropping everything after the `/` at some point.

cjellick:
@g-linville can you give an example that reproduces the problem?

g-linville:
Yes, `@modelcontextprotocol/server-sequential-thinking` as the actual value of the Name field for the server will cause this problem. The nanobot config looks like this:

```yaml
mcpServers:
  '@modelcontextprotocol/server-sequential-thinking':
    args:
    - '@modelcontextprotocol/server-sequential-thinking'
    command: npx
publish:
  mcpServers:
  - '@modelcontextprotocol/server-sequential-thinking'
```

g-linville:
Also worth noting, this is an older version of Nanobot, but the one contained in the newest version of the "phat" image.

cjellick:
SO just to be clear, this is if those characters are in the *Name* field?

sangee2004:
@g-linville Wanted to confirm that this is the same issue tracked in https://github.com/obot-platform/obot/issues/4248 ?

sangee2004:
This issue is addressed when testing with latest build - `[v0.0.0-dev+75b14879](https://github.com/obot-platform/obot/commit/75b14879)`

Create an MCP server with name  having "/" and "@" like `@modelcontextprotocol/server-sequential-thinking`

Able to connect and chat with this MCP server successfully.


<img width="1342" height="914" alt="Image" src="https://github.com/user-attachments/assets/5755477c-7099-40fb-bbd3-a53bea3b06de" />

g-linville:
@sangee2004 yes that is the same. Sorry, didn't realize we already had an issue for that when I created this one.

PR #4610, opened 2025-10-11
fix: remove / from MCP server names for nanobot

thedadams:
Nanobot expects the server name to not have any forward-slashes in it because it uses forward-slashes for tool mapping. So, if the server name is something like @modelcontextprotocol/server-sequential-thinking, then nanobot will parse this as a server with name @modelcontextprotocol and a tool with name server-sequential-thinking.

This change will replace / with - and nanobot will correctly parse the config.

Issue: https://github.com/obot-platform/obot/issues/4603

### DIFF
--- a/pkg/mcp/loader.go
+++ b/pkg/mcp/loader.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"strings"
 	"sync"
 
 	"github.com/gptscript-ai/gptscript/pkg/hash"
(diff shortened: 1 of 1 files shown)


pkg/mcp/loader.go
```diff
@@ -325,7 +327,7 @@ func constructNanobotYAML(name, command string, args []string, env map[string]st
 }
 
 type nanobotConfig struct {
-	Publish    nanobotConfigPublish              `json:"publish,omitempty"`
+	Publish    nanobotConfigPublish              `json:"publish,omitzero"`
```
njhale, line 330: First time I've seen this used

Issue #4182, opened 2025-09-08
Provide a way to get logs for MCP server instances that fails to start successfully and  get auto deleted.


sangee2004:
**Describe the bug**
Provide a way to get logs for MCP server instances that fails to start successfully and  get auto deleted.

**To Reproduce**
1. Create an MCP server entry like Tavily Search MCP server (`npx tavily-mcp@0.2.9`) without the required parameter - `TAVILY_API_KEY`
2. As regular user try to connect to this MCP server

MCP Server fails to start which is as expected

User is presented with following error message:

<img width="580" height="420" alt="Image" src="https://github.com/user-attachments/assets/ddda0b96-ae8a-4156-9d40-916cc42f7952" />

At this time, mcp server instance that failed to start successfully gets deleted already.

As admin or regular user, there is no way to get access to the MCP server logs to understand what might be causing the MCP server to launch successfully.

**Expected behavior**
Provide a way for admin ( and regular user) to get access to the MCP server logs when MCP server fails to start successfully.
This may require us to not delete the MCP server instances when they fail to get started successfully.

cjellick:
I dont want to auto-delete anymore. But we do have to figure out the "next steps" when you get into this situation.

sangee2004:
@ivyjeong13 When I tried to connect to an MCP server that is misconfigured (Tavily Search MCP server- `npx tavily-mcp@0.2.9` without the required parameter - TAVILY_API_KEY)  , I am presented with this page:

<img width="1311" height="411" alt="Image" src="https://github.com/user-attachments/assets/109dc76b-c1f2-4ef2-bed1-e179d350f7ed" />

If I close the window or use `Cancel and Delete Server` , mcp server instance is already deleted. 

How would I be able to not delete the instance and get access to. the logs of this instance ?

ivyjeong13:
> [@ivyjeong13](https://github.com/ivyjeong13) When I tried to connect to an MCP server that is misconfigured (Tavily Search MCP server- `npx tavily-mcp@0.2.9` without the required parameter - TAVILY_API_KEY) , I am presented with this page:
> 
[...]

PR #4611, opened 2025-10-13
enhance: remove auto delete and see logs of mcp failure on setup

ivyjeong13:
Addresses #4182

### DIFF
--- a/pkg/mcp/details.go
+++ b/pkg/mcp/details.go
@@ -2,6 +2,7 @@ package mcp
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 
@@ -24,6 +25,18 @@ func (sm *SessionManager) GetServerDetails(ctx context.Context, userID, mcpServe
 	return sm.backend.getServerDetails(ctx, serverConfig.Scope)
 }
 
+// shouldAttemptLogStreaming checks if the error is one that should still have logs
+// These are errors where the server might be running but not healthy
+func shouldAttemptLogStreaming(err error) bool {
+	for unwrappedErr := err; unwrappedErr != nil; unwrappedErr = errors.Unwrap(unwrappedErr) {
+		switch unwrappedErr {
+		case ErrHealthCheckFailed, ErrHealthCheckTimeout, ErrPodConfigurationFailed:
+			return true
+		}
+	}
+	return false
+}
+
 // StreamServerLogs will stream the logs of a specific MCP server based on its configuration, if the backend supports it.
 // If the server is remote, it will return an error as remote servers do not support this operation.
 // If the backend does not support the operation, it will return an [ErrNotSupportedByBackend] error.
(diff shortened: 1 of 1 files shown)


ui/user/src/lib/components/mcp/MyMcpServers.svelte
```diff
@@ -401,6 +427,19 @@
 		if (!selectedEntryOrServer) return;
 		if (!configureForm) return;
 
+		console.log(selectedEntryOrServer, configureForm, relaunching);
```
g-linville: debug log left behind?

ivyjeong13:
Wanted to test this again before merging, but looks like /logs is no longer working for me when trying to run with the Tavily example case

After getting the 503 from /launch, trying to hit /logs gives me:
500 `server running, but readiness check failed: server readiness check failed: MCP server is not healthy`

Before, /logs would be successful and I could get the logs info showing that I am missing an environment variable. Example of what was happening before: https://www.loom.com/share/23d35958db214d99a48b99349b42e5e4

Wondering if there was a change with Tavily or on our side and holding off merge to see if I can try with another MCP server!

ivyjeong13:
> Wanted to test this again before merging, but looks like /logs is no longer working for me when trying to run with the Tavily example case
> 
> After getting the 503 from /launch, trying to hit /logs gives me: 500 `server running, but readiness check failed: server readiness check failed: MCP server is not healthy`
> 
> Before, /logs would be successful and I could get the logs info showing that I am missing an environment variable. Example of what was happening before: https://www.loom.com/share/23d35958db214d99a48b99349b42e5e4
> 
> Wondering if there was a change with Tavily or on our side and holding off merge to see if I can try with another MCP server!

Was able to determine there was a backend change that made it so /logs will return a 500 if the deployment fails for any reason. Waiting on a backend PR to merge https://github.com/obot-platform/obot/pull/4530 to verify that /logs can be hit if due to invalid/not provided environment variable.

pkg/mcp/details.go
```diff
@@ -24,6 +25,20 @@ func (sm *SessionManager) GetServerDetails(ctx context.Context, userID, mcpServe
 	return sm.backend.getServerDetails(ctx, serverConfig.Scope)
 }
 
+// shouldAttemptLogStreaming checks if the error is one that should still have logs
+// These are errors where the server might be running but not healthy
+func shouldAttemptLogStreaming(err error) bool {
```
ivyjeong13, line 30: Dropped review(s) to check this backend portion! Is this alright? We're looking to allow users to see logs when configuring a mcp server fails but logs can be pulled out of it (Are any other cases to add here as well or lemme know if these are the wrong ones! 😂) Thank you!

Screenshot example:
Screenshot 2025-10-23 at 1 10 40 PM

g-linville: This seems right to me!

g-linville (approved):
Backend lgtm

Issue #4563, opened 2025-10-08
Multiline sensitive field does not grow with entered text and using down arrow results in bad user experience.


sangee2004:
**Describe the bug**
Multiline sensitive field does not grow with entered text and using up/down arrow results in bad user experience.

**To Reproduce**
1. Connect with `BigQuery Toolbox` mcp server
2. Try to copy credentials to `Google Application Credentials` field which is more than 3 lines

Notice that the field itself shows only 3 lines . Using down arrow does not result in more lines being shown.
If i try to delete few lines , i start seeing some red dotted lines.

https://www.loom.com/share/7d8d57c580414807b3aba4464db88495?sid=b88d003f-3af4-4a97-9541-f7fdd8a5fad3

**Expected behavior**
Multiline sensitive field should  grow with entered text up to certain number of lines. After this user should be able to scroll thru the contents of this field.

sangee2004:
@ryu-man When testing with latest build- [v0.0.0-dev+75b14879](https://github.com/obot-platform/obot/commit/75b14879) , Copying multi line texts into sensitive field results in  UI to be  completely broken with text overflowing across many fields.

The text i tried to copy over in this case:
```
test1
test2
test3
test4
test5
test6
test7
test8
test9
```


<img width="708" height="734" alt="Image" src="https://github.com/user-attachments/assets/a95f5319-0714-469f-bfdd-3adaf95f2b80" />

ryu-man:
@sangee2004 This is how it looks on my side. works properly! Which browser are you using? Can you make sure to clear the cache and try again?

<img width="881" height="891" alt="Image" src="https://github.com/user-attachments/assets/1c611eef-47c1-4896-8acf-4817b33c5f11" />

sangee2004:
@ryu-man I am testing in chrome browser in MAC laptop.
Chrome Version - `141.0.7390.66 (Official Build) (arm64)`

sangee2004:
@ryu-man When testing with latest builds,  copying multi line texts into sensitive field results in text growing as expected and no overflows.

But i still see some red dotted lines show up.

<img width="892" height="754" alt="Image" src="https://github.com/user-attachments/assets/54f1e002-78ef-47c1-8b38-6137a713678a" />

Sending it over to you to follow up on this.

PR #4614, opened 2025-10-13
fix: improve multilines ux for sensitive textarea component

ryu-man:
Addresses #4563 

- Preserves all whitespace characters so original text and masked text break at the same point.
- Sync scroll position


ivyjeong13:
Sorry accidentally pushed commit to this branch, reset and removed it!

ryu-man:
No worries.

PR #4652, opened 2025-10-14
feat: support catalog yaml field `prefix`

tybalex:
https://github.com/obot-platform/mcp-catalog/issues/260
Add optional prefix field `prefix` to MCPHeader/MCPEnv for auto-prepending values like `Bearer `. This applies to multiple MCP Servers.

### DIFF
--- a/pkg/mcp/types.go
+++ b/pkg/mcp/types.go
@@ -64,7 +64,7 @@ func expandEnvVars(text string, credEnv map[string]string, fileEnvVars map[strin
 // applyPrefix adds a prefix to a value if the value doesn't already start with it.
 // Returns the original value if prefix is empty or if value already starts with the prefix.
 func applyPrefix(value, prefix string) string {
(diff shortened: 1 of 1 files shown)


pkg/mcp/types.go
```diff
@@ -60,6 +61,15 @@ func expandEnvVars(text string, credEnv map[string]string, fileEnvVars map[strin
 	})
 }
 
+// applyPrefix adds a prefix to a value if the value doesn't already start with it.
+// Returns the original value if prefix is empty or if value already starts with the prefix.
+func applyPrefix(value, prefix string) string {
+	if prefix == "" || value == "" || strings.HasPrefix(value, prefix) {
```
thedadams: nit:
```suggestion
	if value == "" || strings.HasPrefix(value, prefix) {
```

thedadams (approved):

Issue #4566, opened 2025-10-08
`Runs` link for task does not fetch any runs.


sangee2004:
**Describe the bug**
`Runs` link for task does not fetch any runs.

**To Reproduce**
1. As any user with Auditor role , access `Admin Dashboard`
2. Go to `Tasks` and click on `Runs` link for any Task that has runs.

Notice that user is taken to `Task Runs` page with no task runs.

https://www.loom.com/share/cdce678371d546298cfb782d5f38f7df?sid=6718fd33-15ee-4745-a07a-5d85fb68dabe

**Expected behavior**
Should be able to show all the task runs related to this task.

sangee2004:
This issue is addressed when testing with latest build.

PR #4653, opened 2025-10-14
fix: use proper task property to filter table data

ryu-man:
Addresses #4566

### DIFF
--- a/ui/user/src/routes/admin/task-runs/+page.svelte
+++ b/ui/user/src/routes/admin/task-runs/+page.svelte
@@ -39,7 +39,6 @@
 			.filter((entry) => entry[0] !== 'query')
 			.reduce(
 				(acc, [key, value]) => {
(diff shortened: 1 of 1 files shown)


ui/user/src/routes/admin/task-runs/+page.svelte
```diff
@@ -33,7 +33,19 @@
...
 
 	let query = $state(page.url.searchParams.get('query') || '');
-	let urlFilters = $state<Record<string, (string | number)[]>>({});
+	let urlFilters = $derived.by<Record<string, (string | number)[]>>(() => {
+		return page.url.searchParams
+			.entries()
+			.filter((entry) => entry[0] !== 'query')
+			.reduce(
+				(acc, [key, value]) => {
+					console.log(key, value);
```
ivyjeong13: Can drop the debug log here!

Issue #4651, opened 2025-10-14
Okta - User name is not reported (only email is reported)


sangee2004:
**Describe the bug**
OKTA - User name is not reported  (only email is reported)

Pre Req - Have OKTA auth provider set up

**To Reproduce**
1. Log in using OKTA auth provider 

Notice that user name is not reported 

<img width="1110" height="431" alt="Image" src="https://github.com/user-attachments/assets/ef881d58-96ab-4fbd-90a7-a59dba0000a8" />

From okta:

<img width="1036" height="393" alt="Image" src="https://github.com/user-attachments/assets/7cecaa28-b5b1-45db-8ae9-e231973173e3" />

**Expected behavior**
User name should be reported correctly in User profile and user list view

g-linville:
Just merged the fix for this.

sangee2004:
This issue is addressed when testing with latest build

<img width="964" height="386" alt="Image" src="https://github.com/user-attachments/assets/c691521a-0649-40df-95ea-159975a19edb" />

PR #4655, opened 2025-10-14
fix: okta: return proper display names

g-linville:
for https://github.com/obot-platform/obot/issues/4651


g-linville:
made a PR on the wrong repo lol

Issue #4580, opened 2025-10-08
Restarting single user MCP server does not result in MCP server restarting in docker installs


sangee2004:
**Describe the bug**
Restarting single user MCP server does not result in MCP server restarting.

Environment: local docker install
Consistently Reproduces: Yes .
Additional environments attempted in: main ( But this may be relating to capacity issue). Will be testing local k8s deployment to see if it works there in a bit.

**To Reproduce**
1. As any user, connect to Single user MCP server.
2. As admin user . go to MCP Servers page and go to `Deployments and Connections tab`.
3. Select the MCP server instance that was created in step1.
4. Use `Restart` option to restart MCP server instance

Notice that POST call made to `api/mcp-servers/<msid>/restart` from UI succeeds (204 status). But the `Last Restart` or the `Recent Events` reported in Server details page does not indicate that a restart was done.

https://www.loom.com/share/a56ef42a80804e699d283fb854ce5fc9?sid=8738922a-8679-46bd-ad73-f9b8e525029e

**Expected behavior**
MCP server should get restarted successfully.

StrongMonkey:
@sangee2004 I am not able to reproduce this. When clicking on restart, the pod was terminated correctly and new events show up.

<img width="1522" height="996" alt="Image" src="https://github.com/user-attachments/assets/c4e0f4da-512a-4afe-b8f7-94bf5dc2317d" />

StrongMonkey:
One scenario if you are testing with main:

When we do restart, we are doing a gracefully shutdown of deployment which will allow new pod to be created first before terminating old one. If main is running out of capacity, the new one could never come alive and you won't see events and it won't be propably restart due to resource constraints.

sangee2004:
@StrongMonkey I am seeing this issue when testing with my local docker install set up as well as in `main`.

sangee2004:
This issue is addressed when testing with latest build

1 . On docker install , MCP server instance  restart results in restart of the  MCP server docker process as expected.

```
docker ps
CONTAINER ID   IMAGE                                        COMMAND                  CREATED              STATUS             PORTS                       NAMES
5f1c67cf769f   ghcr.io/obot-platform/mcp-images/phat:main   "nanobot run --liste…"   About a minute ago   Up 8 seconds       127.0.0.1:57361->8099/tcp   mcp2f5600909d92b08e2151df489715c36b392846f60779a653c0fb5bb9303b
```
[...]

PR #4660, opened 2025-10-15
Fix: don't check for server.command

StrongMonkey:
For restart deployment, we don't need to check for server.command anymore. This is because for built-in images from mcp-image the command will be built-in. Checking for empty command will fail to restart these deployment.

https://github.com/obot-platform/obot/issues/4580

### DIFF
--- a/pkg/mcp/loader.go
+++ b/pkg/mcp/loader.go
@@ -230,10 +230,9 @@ func (sm *SessionManager) closeClients(id string) {
 // RestartServerDeployment restarts the server in the currently used backend, if the backend supports it.
 // If the backend does not support restarts, then an [ErrNotSupportedByBackend] error is returned.
 func (sm *SessionManager) RestartServerDeployment(ctx context.Context, server ServerConfig) error {
-	if server.Command == "" {
-		return nil
+	if server.Runtime == otypes.RuntimeRemote {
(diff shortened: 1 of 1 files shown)


pkg/mcp/loader.go
```diff
@@ -230,10 +230,6 @@ func (sm *SessionManager) closeClients(id string) {
 // RestartServerDeployment restarts the server in the currently used backend, if the backend supports it.
 // If the backend does not support restarts, then an [ErrNotSupportedByBackend] error is returned.
 func (sm *SessionManager) RestartServerDeployment(ctx context.Context, server ServerConfig) error {
-	if server.Command == "" {
```
thedadams, line 233: I think the original intent of this is to not call the below function for remote MCP servers.

Can we add this back, but change the check to `server.Runtime == types.RuntimeRemote`.

StrongMonkey: We actually check here for remote server
https://github.com/obot-platform/obot/blob/f46f93ced5600e6c9ad599c24f9d2929029df848/pkg/api/handlers/mcp.go#L2050

thedadams: We also need to check here in case this function is used somewhere else.

StrongMonkey: OK, I can add it here.

StrongMonkey: This is added

PR #4662, opened 2025-10-15
chore: make server migrations easier

thedadams:
Since we are using a hash of the server config for container and deployment names, it is possible to orphan these objects as the server configuration evolves.

This change will use a stable name for these objects so that they are updated rather than orphaned.

### DIFF
--- a/pkg/mcp/docker.go
+++ b/pkg/mcp/docker.go
@@ -30,10 +30,14 @@ type dockerBackend struct {
 	baseImage    string
 }
 
-func newDockerBackend(baseImage string) (backend, error) {
+func newDockerBackend(ctx context.Context, baseImage string) (backend, error) {
 	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
 	if err != nil {
 		return nil, fmt.Errorf("failed to create Docker client: %w", err)
+	}
+
+	if err = cleanupContainersWithOldID(ctx, cli); err != nil {
+		return nil, fmt.Errorf("failed to cleanup containers with old ID: %w", err)
 	}
 
 	return &dockerBackend{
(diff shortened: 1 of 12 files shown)


pkg/mcp/docker.go
```diff
@@ -30,23 +30,52 @@ type dockerBackend struct {
...
+
 	return &dockerBackend{
 		client:       cli,
 		containerEnv: os.Getenv("OBOT_CONTAINER_ENV") == "true",
 		baseImage:    baseImage,
 	}, nil
 }
 
-func (d *dockerBackend) ensureServerDeployment(ctx context.Context, server ServerConfig, id, mcpServerDisplayName, _ string) (ServerConfig, error) {
+func cleanupContainersWithOldID(ctx context.Context, client *client.Client) error {
```
g-linville, line 53: If this is a migration, I think there should be a comment saying so.

g-linville (approved):

pkg/mcp/kubernetes.go
```diff
@@ -47,25 +47,25 @@ func newKubernetesBackend(clientset *kubernetes.Clientset, client kclient.WithWa
...
-	objs, err := k.k8sObjects(id, server, mcpServerDisplayName, mcpServerName)
+	objs, err := k.k8sObjects(server, userID, mcpServerDisplayName, mcpServerName)
 	if err != nil {
-		return ServerConfig{}, fmt.Errorf("failed to generate kubernetes objects for server %s: %w", id, err)
+		return ServerConfig{}, fmt.Errorf("failed to generate kubernetes objects for server %s: %w", server.Scope, err)
 	}
 
-	if err := apply.New(k.client).WithNamespace(k.mcpNamespace).WithOwnerSubContext(id).Apply(ctx, nil, objs...); err != nil {
-		return ServerConfig{}, fmt.Errorf("failed to create MCP deployment %s: %w", id, err)
+	if err := apply.New(k.client).WithNamespace(k.mcpNamespace).WithOwnerSubContext(server.Scope).Apply(ctx, nil, objs...); err != nil {
```
StrongMonkey, line 63: is server.scope going to unique per mcp server?

thedadams: It's unique per MCP server and user.

pkg/mcp/docker.go
thedadams: Comment added.

Issue #4654, opened 2025-10-14
Using `MCP Server` breadcrumb link after creating MCP Server results in user being brought to create MCP server page.


sangee2004:
**Describe the bug**
Using `MCP Server` breadcrumb link after creating MCP Server results in user being brought to create MCP server page.

**To Reproduce**
1. As admin/owner, go to `Admin Dashboard`
2. Create MCP server
3. Once MCP server is created , use  `MCP Server` breadcrumb link 

Notice that user is brought to `Create MCP server` page.
Using  `MCP Server` breadcrumb link  from the this page results in empty MCP Server list page

https://www.loom.com/share/644b14c807454635ab3d286f2e875455?sid=8d6f1419-37cd-4e7c-b260-301733c874d2

**Expected behavior**
Using `MCP Server` breadcrumb link after creating MCP Server should take the user back to  MCP Server list page with all MCP servers listed.

sangee2004:
This issue is addressed when testing with latest build.
Using MCP Server breadcrumb link after creating MCP Server results in user being brought to MCP server page with all entries presented to the user (with no filters applied) as expected.

PR #4664, opened 2025-10-15
fix: navigate  properly and preserve searchparams when navigate back

ryu-man:
Addresses #4654

- Drop history-based navigation; use internal `openLink` function.
- Preserve page search params when navigating back

### DIFF
--- a/ui/user/src/routes/admin/chat-threads/+page.svelte
+++ b/ui/user/src/routes/admin/chat-threads/+page.svelte
@@ -15,7 +15,7 @@
 	import { profile } from '$lib/stores';
 	import { twMerge } from 'tailwind-merge';
 	import { tooltip } from '$lib/actions/tooltip.svelte';
-	import { clearUrlParams, setUrlParams } from '$lib/url';
+	import { clearUrlParams, setSearchParamsToLocalStorage, setUrlParams } from '$lib/url';
 
 	let threads = $state<ProjectThread[]>([]);
 	let projects = $state<Project[]>([]);
@@ -159,6 +159,7 @@
 						onClearAllFilters={clearUrlParams}
 						onClickRow={isAuditor
 							? (d, isCtrlClick) => {
+									setSearchParamsToLocalStorage(page.url.pathname, page.url.search);
 									const url = `/admin/chat-threads/${d.id}`;
 									openUrl(url, isCtrlClick);
 								}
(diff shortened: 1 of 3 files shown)


ui/user/src/lib/components/BackLink.svelte
```diff
@@ -109,14 +110,12 @@
 				const isTouchDevice = 'ontouchstart' in window || navigator.maxTouchPoints > 0;
 				const isCtrlClick = isTouchDevice ? false : e.metaKey || e.ctrlKey;
 
-				const stepsBack = links.length - index;
-				const hasHistory = history.length > stepsBack;
+				const searchParams = getSearchParamsLocalStorage(link.href);
```
ivyjeong13: I think this change will breaking keeping filters for rest of the pages that arent coming from Task Runs

Would check for multi breadcrumb case and filter history being retained:
Ex: Go to Admins MCP Servers
In Registry Entries, filter the table by name of a MCP server or another filter that will bring back a result with a deployed single mcp server.
Click a server.
Go to Server Details & click a user server.
In the deployed server page, click the MCP Server breadcrumb. It should go back to MCP Servers with the filter you had before. 


Screenshot_20251015_143259_Chrome~2

ryu-man: This issue can be addressed by saving the URL search params using the `setSearchParamsLocalStorage` function. before navigation away, going to push a fix for that

ryu-man: I pushed a fix for the MCP server registry view & deployment view breadcrumb navigations. Search params are preserved now when navigating back to `/mcp-servers`

ivyjeong13: Also need to take into account:

Chat Threads
Tasks
Access Control
Filters

Also was thinking, instead of trying to save to localstorage, couldn't we just drop the ?new=true` from the url params before navigating to the new item?

ivyjeong13: Im also just making these assumptions without really being able to check (still ooo) so if someone else can verify and approve, feel free to resolve!

ryu-man: > Also was thinking, instead of trying to save to localstorage, couldn't we just drop the ?new=true` from the url params before navigating to the new item?

It may work, but I still think that using history-based navigation with breadcrumbs is not a good practice.

cjellick:
@ivyjeong13 give this a proper review and tryout when youre back tomorrow please. im holding off on releasing until a few bugs including this one are squared away.

ui/user/src/lib/components/BackLink.svelte
ivyjeong13: From what I gathered, we were treating these back links are user historical rather than hierarchical but I understand what you mean. 

Pulled down and checked all the routes that came to mind :+1: approving!

PR #4675, opened 2025-10-15
fix: improve masked value display in SensitiveInput component

ryu-man:
Addresses #4563

### DIFF
--- a/ui/user/src/lib/components/SensitiveInput.svelte
+++ b/ui/user/src/lib/components/SensitiveInput.svelte
@@ -55,7 +55,7 @@
 					bind:this={textareaElement}
 					data-1p-ignore
 					id={name}
-					contenteditable
+					contenteditable="plaintext-only"
 					class={twMerge(
 						'text-input-filled base min-h-full w-full flex-1 pr-10 font-mono',
 						error && 'border-red-500 bg-red-500/20 text-red-500 ring-red-500 focus:ring-1',


ui/user/src/lib/components/SensitiveInput.svelte
```diff
@@ -112,12 +112,11 @@
 					bind:this={maskedTextarea}
 					tabindex="-1"
 					class={twMerge(
-						'text-input-filled layer-1 pointer-events-none absolute inset-0 w-full bg-transparent pr-10 font-mono'
+						'text-input-filled layer-1 pointer-events-none absolute inset-0 w-full overflow-auto bg-transparent pr-10 font-mono break-words whitespace-pre-wrap'
```
ivyjeong13, line 115: https://www.loom.com/share/e07a9254c3594c3da47605077ab809c4 Not seeing the overflow of the masked value but I am seeing wonkiness when hitting Enter multiple times before typing letters.

ryu-man: Pushed a fix

ivyjeong13: Still seeing the issue:

https://www.loom.com/share/05a1a058b2db4b9ea30f9fb79dcce968

ryu-man: I had to disable rich text features. to prevent this issue from happening

ivyjeong13: :+1: pulled and tested, see it fixed for me! And I don't think we'd expect rich text support for these file type fields so that should be okay? Tagging @cjellick just in case.

Issue #4670, opened 2025-10-15
404 error page presented when access Audit logs for MCP server entries for Multi user and Remote MCP servers


sangee2004:
**Describe the bug**
404 error page presented when access Audit logs for MCP server entries for Multi user and Remote MCP servers

**To Reproduce**
1. As any user , connect to any Multi user/Remote MCP server and chat with it to make some tool calls.
2. As owner , access `Admin Dashboard`
3. Go to `Deployment and Connections` tab.
4. Select the connected MCP instances from step 1.
5. Hover over `View Audit Logs`
Notice that thu
<img width="1015" height="824" alt="Image" src="https://github.com/user-attachments/assets/d4969f1a-cd81-4fd0-9c2d-98a9755b50fe" />

6. Use `View Audit Logs` option` from `Connected Users` section This results in 404 Page Not Found errors

<img width="1258" height="815" alt="Image" src="https://github.com/user-attachments/assets/81738292-dc73-49bf-8ed9-737cd2198086" />

**Expected behavior**
Should be able to view Audit logs .

Same error is also seen in the following case

1. Go to `Registry Entries` tab.
2. Select multi user/remote  MCP server
3. Go to Server Details tab.
4. Use `View Audit Logs` option from Connected Users section

Note - This issue was fixed for single user MCP servers - https://github.com/obot-platform/obot/issues/4575 . But still exists for Multi user and Remote server.

cjellick:
@sangee2004 - fyi - while fixing this, we noticed a bunch of other problems with accessing audit logs under various conditions. Check out #4676 for more details.

sangee2004:
This issue is addressed when testing with latest builld.

PR #4676, opened 2025-10-16
fix audit logging

cjellick:
For https://github.com/obot-platform/obot/issues/4670:

Fixed 404 errors when viewing audit logs for individual users by using
the correct server ID in URLs:

- Multi-user (K8s) servers: Use mcpServerId prop directly instead of
  mcpInstanceId, which contained the full instance ID (e.g.,
  "msi1-2-ms16k8wm") rather than just the server ID (e.g., "ms16k8wm")

- Remote servers: Add catalog entry route fallback and use mcpServerId
  directly instead of attempting to extract ID with string manipulation

The mcpServerId prop already contains the correct server ID value for
constructing audit log URLs, making per-user instance ID extraction
unnecessary.

Also, I noticed user id wasnt resolving properly: Changed the audit log URL query parameter from 'userId' to 'user_id'
to match the backend API expectations.

### DIFF
--- a/ui/user/src/lib/components/admin/McpServerInstances.svelte
+++ b/ui/user/src/lib/components/admin/McpServerInstances.svelte
@@ -2,6 +2,7 @@
 	import {
 		AdminService,
 		ChatService,
+		Group,
 		type MCPCatalogEntry,
 		type MCPCatalogServer,
 		type MCPServerInstance,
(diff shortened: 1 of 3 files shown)


thedadams (approved):
Looks good after linting is fixed.

ivyjeong13:
Tested views with View Audit Logs links:
* In a MultiUser Entry’s Server Details tab (under Connected Users)
* In a Global Registry Single User Instance page (under Connected Users)
* In a Global Registry Multi User page (under Connected Users)
* In a Admin Workspace Single User Instance page (under Connected Users)
* In a Admin Workspace Multi User page (under Connected Users)
* In MCP Servers, Deployment & Connections, ‘…’ > ‘View Audit Logs’ for a row item
    * Checked with Single, Multi-User, & Remote
* In MCP Publisher Single User Instance page for a PUP (under Connected Users)
* In MCP Publisher Remote User Instance page for a PUP (under Connected Users)
* In MCP Publisher MultiUser Entry's Server Details tab (under Connected Users)

For cases above, also clicked View Audit Logs and checked I was getting results back just in case.

Added additional following fixes around View Audit Logs:
* Noticed as a Basic User w/ Auditor Role, I did not have the View Audit Logs option in the Connected Users table (I believe this was omitted initially with initial auditor role impl but should be okay now) Added it back in.
* Noticed as a PUP, going into MCP Publisher, wasn’t seeing the View Audit Logs link (I believe this was before audit logs was impl into MCP Publisher) Removed the check and allowed showing the link, directing to MCP Publisher Audit Logs instead
* For admin, was seeing “Server information cannot be displayed at this time” for a poweruser Workspace Remote Deployed Server under Deployment & Connections; fixed!
* Fixed clicking ‘View Audit Logs’ in Deployment & Connections for Multi-User servers (needed to use mcp_server_display_name)

PR #4678, opened 2025-10-16
chore: ui: remove cline references

g-linville:
Cline currently does not work with Obot because they do not support the latest version of the MCP protocol.


ui/user/src/lib/components/mcp/HowToConnect.svelte
```diff
@@ -112,7 +108,7 @@
 		style="scroll-behavior: smooth;"
 	>
 		{#each options as option (option.key)}
-			<li class="w-36 flex-shrink-0">
+			<li class="w-49 flex-shrink-0">
```
g-linville, line 111: I adjusted the width here so that the Cursor, Claude, and VSCode tabs would fill the whole width of the tab bar.

Issue #4375, opened 2025-09-23
PUP/PU users are not allowed to update MCP server instances . 403 errors encountered.


sangee2004:
**Describe the bug**
PUP users are not allowed to update MCP server instances .

Environment: local
Consistently Reproduces: Yes 
Additional environments attempted in: Not attempted in main yet

**To Reproduce**
1. As PUP user , create MCP server (single user Tavily Search in my case)
2. As another user , deploy this MCP server.
3. As PUP user (from step1) , update MCP server config and add a new required config parameter.
4. Go to `Server Details` page and see that there is server update available notification as expected.

Click on `Update Server`

Notice thar 403 errors are presented to the user.

<img width="1064" height="769" alt="Image" src="https://github.com/user-attachments/assets/e0cab6cf-0aee-4292-a155-80b397d1d5df" />

Full error message `403 /mcp-servers/ms1f6p2m/trigger-update: forbidden`

**Note** This issue is also seen for MCP servers created by PU users.

**Expected behavior**
PUP/PU user should be able to update the server .

g-linville:
Merged the fix

sangee2004:
@g-linville I am testing with latest build from `main` - `[v0.0.0-dev+dff3476a](https://github.com/obot-platform/obot/commit/dff3476a)` and use PUP user for this testing.

When following the same steps mentioned in the issue , I am seeing different error - `failed to fetch` being presented to the user when `Update Server` option is used.


<img width="1221" height="781" alt="Image" src="https://github.com/user-attachments/assets/9571211b-4d46-465b-986e-f00e36fcd43d" />

Following is the api call that was made by  UI:
<img width="1017" height="334" alt="Image" src="https://github.com/user-attachments/assets/9ade6062-791c-4916-9c8c-6bcccae2e433" />

sangee2004:
Reopening this issue for followup.

sangee2004:
404 errors are seen when testing server updates with PU users which is tracked in https://github.com/obot-platform/obot/issues/4526

g-linville:
Moving this back to Next Up due to shifting priorities

g-linville:
Merged the fix

sangee2004:
This issue is addressed when testing with latest build.

PR #4681, opened 2025-10-16
fix: eliminate potential nil deref in authz

g-linville:
related to https://github.com/obot-platform/obot/issues/4375 (but doesn't fully fix it)

Figured we should get this in before the release.

### DIFF
--- a/pkg/api/authz/mcpserver.go
+++ b/pkg/api/authz/mcpserver.go
@@ -44,7 +44,7 @@ func (a *Authorizer) checkMCPServer(req *http.Request, resources *Resources, u u
 
 		resources.Authorizated.MCPServer = &mcpServer
 		return true, nil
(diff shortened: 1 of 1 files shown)


pkg/api/authz/mcpserver.go
```diff
@@ -44,7 +44,7 @@ func (a *Authorizer) checkMCPServer(req *http.Request, resources *Resources, u u
 
 		resources.Authorizated.MCPServer = &mcpServer
 		return true, nil
-	} else if mcpServer.Spec.MCPServerCatalogEntryName != "" {
+	} else if mcpServer.Spec.MCPServerCatalogEntryName != "" && resources.Authorizated.PowerUserWorkspace != nil {
```
g-linville, line 47: Two lines down, we deref `resources.Authorizated.PowerUserWorkspace`, assuming it is non-nil.

PR #4689, opened 2025-10-16
The default mcp runtime backend is docker

akesiraju:
The documentation says the default mcp runtime is docker, but the actual value is kubernetes


g-linville:
While the default in the actual Obot code is `docker`, we want the chart's default to be `kubernetes`, because users will likely always want to use kubernetes as the backend for deploying MCP servers. Closing.

akesiraju:
Got it. I missed this env as I tried to deploy on k8s without charts.

Issue #4552, opened 2025-10-07
Restrict user from being able to configure more than one auth provider.


sangee2004:
**Describe the bug**
Restrict user from being able to configure more than one auth provider.

Currently I am allowed to configure more than one auth provider and log in with using both auth providers.

Since this is not a supported configuration anymore ,  user should not be allowed to configure auth providers when there is already 1 auth provider configured.

cjellick:
@g-linville make it so the API will recheck a second one being configured
@ivyjeong13 make it so the UI only lets you enable one at a time.

g-linville:
Starting on the backend portion of this. It is unrelated to the UI portion, so these can be worked on simultaneously.

g-linville:
The backend part of this is completed and merged.

sangee2004:
This issue is addressed when testing with latest build.

We are not allowed to configure auth providers when there is 1 auth provider that is already configured.

<img width="1021" height="508" alt="Image" src="https://github.com/user-attachments/assets/718f7639-f285-43a1-8103-e022506298c9" />

PR #4696, opened 2025-10-17
fix: only allow one auth provider to be configured at once

g-linville:
for https://github.com/obot-platform/obot/issues/4552

### DIFF
--- a/pkg/api/handlers/authprovider.go
+++ b/pkg/api/handlers/authprovider.go
@@ -196,18 +196,19 @@ func (ap *AuthProviderHandler) Configure(req api.Context) error {
 	// Deconfigure it if that is not the case, and return a 400.
 	ap.configureLock.Lock()
 	configuredProviders = ap.dispatcher.ListConfiguredAuthProviders(req.Namespace())
-	ap.configureLock.Unlock()
 	for _, configuredName := range configuredProviders {
 		if configuredName != ref.Name {
 			// Delete the credential we just configured
 			_ = req.GPTClient.DeleteCredential(req.Context(), string(ref.UID), ref.Name)
+			ap.dispatcher.UpdateConfiguredAuthProviders(req.Context())
+			ap.configureLock.Unlock()
 			return types.NewErrBadRequest(
 				"only one authentication provider can be configured at a time. Please deconfigure %q first",
 				configuredName,
 			)
 		}
 	}
-
+	ap.configureLock.Unlock()
 	if ref.Annotations[v1.AuthProviderSyncAnnotation] == "" {
 		if ref.Annotations == nil {
 			ref.Annotations = make(map[string]string, 1)


pkg/api/handlers/authprovider.go
```diff
@@ -136,6 +142,18 @@ func (ap *AuthProviderHandler) Configure(req api.Context) error {
 		return types.NewErrBadRequest("%q is not an auth provider", ref.Name)
 	}
 
+	// Check if another auth provider is already configured
+	configuredProviders := ap.dispatcher.ListConfiguredAuthProviders(req.Namespace())
```
thedadams, line 144: This makes it less likely to have two auth providers configured, but there is still a race condition because we call `ap.dispatcher.UpdateConfiguedAuthProviders` in a goroutine.

Secondly, holding the lock during all of these io calls isn't great. I think a decent fix would be to:
1. Lock
2. Check
3. Unlock
4. 400 if necessary
5. Configure
6. Update
7. Lock
8. Check again
9. Deconfigure, if necessary and update again
10. Unlock

g-linville: Fixed - let me know if this looks correct to you

pkg/api/handlers/authprovider.go
```diff
@@ -136,6 +139,20 @@ func (ap *AuthProviderHandler) Configure(req api.Context) error {
...
+	for _, configuredName := range configuredProviders {
+		// Allow reconfiguring the same provider
+		if configuredName != ref.Name {
+			return types.NewErrBadRequest(
+				"only one authentication provider can be configured at a time. Please deconfigure %q first",
+				configuredName,
+			)
+		}
+	}
+
```
thedadams:
```suggestion
	for _, configuredName := range configuredProviders {
		// Allow reconfiguring the same provider
		if configuredName != ref.Name {
			ap.configureLock.Unlock()
			return types.NewErrBadRequest(
				"only one authentication provider can be configured at a time. Please deconfigure %q first",
				configuredName,
			)
		}
	}
	ap.configureLock.Unlock()
```

pkg/api/handlers/authprovider.go
```diff
@@ -176,8 +189,25 @@ func (ap *AuthProviderHandler) Configure(req api.Context) error {
...
+		if configuredName != ref.Name {
+			// Delete the credential we just configured
+			_ = req.GPTClient.DeleteCredential(req.Context(), string(ref.UID), ref.Name)
+			return types.NewErrBadRequest(
+				"only one authentication provider can be configured at a time. Please deconfigure %q first",
+				configuredName,
+			)
+		}
+	}
+
```
thedadams: Similar here, but also need to call update again.
```suggestion
	ap.configureLock.Lock()
	configuredProviders = ap.dispatcher.ListConfiguredAuthProviders(req.Namespace())
	for _, configuredName := range configuredProviders {
		if configuredName != ref.Name {
			// Delete the credential we just configured
			_ = req.GPTClient.DeleteCredential(req.Context(), string(ref.UID), ref.Name)
			ap.dispatcher.UpdateConfiguredAuthProviders(req.Context())
			ap.configureLock.Unlock()
			return types.NewErrBadRequest(
				"only one authentication provider can be configured at a time. Please deconfigure %q first",
				configuredName,
			)
		}
	}
	ap.configureLock.Unlock()
```

g-linville: Thanks 👍

Issue #4677, opened 2025-10-16
Support Affinity, Tolerations, and resource limits/requests for MCP Server Scheduling


cjellick:
#### Overview  
Adds support for Kubernetes affinity and tolerations in MCP server workload scheduling. Admins can specify affinity, tolerations, and limits/requests either in the Helm chart or through the UI, with clear rules for where these settings are managed and how they are displayed.

#### Key Changes

- **Affinity and Tolerations Support:**  
  - Admins can provide full affinity and tolerations YAML/JSON blobs for MCP server pods.
  - Only affinity (including anti-affinity) and tolerations are supported—no node selectors, node names, or pod topology spread constraints at this time.

- **Configuration Locations:**  
  - Affinity and tolerations can be set in either the Helm chart or the UI.
  - If set in the Helm chart, these settings are read-only in the UI and cannot be edited there (but are displayed for visibility).
  - If not set in the Helm chart, admins can configure them directly in the UI.

- **Resource Requests and Limits:**  
  - CPU and memory requests and limits can also be set via the Helm chart or UI, following the same management rules.
  - For now, these are applied as a single configuration for all pods in the deployment.

#### Rationale  
Provides essential Kubernetes scheduling controls for MCP server workloads, allowing for better placement and resource management while keeping configuration simple and clear.

cloudnautique:
As a possible temporary solution to set a global blanket value, we could add default resource limit settings to the namespace we or the user creates for running the MCP servers. If we are already handling this as a setting on the deployment through code, then just bubbling up as a setting in the admin panel would make the most sense.

g-linville:
Additional requirements: when the global k8s configuration changes, Obot should show a notice next to deployed servers using the old configuration (similar to when we show the "upgrade available" button when something has changed in the upstream catalog entry for a deployed server). The button should probably say "apply current Kubernetes settings and redeploy" or something like that, with a confirmation modal in the UI.

g-linville:
Moved to Triage for frontend work.

sangee2004:
When testing with latest build, admin is provided with the ability to set Affinity, Tolerations and Resource Requests and Limits for mcp server deployments.

Issues found when testing this feature is tracked in separate issues

https://github.com/obot-platform/obot/issues/4885
https://github.com/obot-platform/obot/issues/4888
https://github.com/obot-platform/obot/issues/4893
https://github.com/obot-platform/obot/issues/4895
https://github.com/obot-platform/obot/issues/4896

PR #4709, opened 2025-10-17
feat: configure k8s affinity, tolerations, and resources

g-linville:
for https://github.com/obot-platform/obot/issues/4677

See the Loom that I just shared in Slack.


pkg/api/handlers/k8ssettings.go
```diff
@@ -0,0 +1,143 @@
...
+
+func (h *K8sSettingsHandler) Update(req api.Context) error {
+	var input types.K8sSettings
+	if err := req.Read(&input); err != nil {
+		return err
+	}
+
+	// Retry loop to handle conflicts
+	var settings v1.K8sSettings
+	for i := 0; i < 5; i++ {
```
g-linville: I'm not 100% sure why I was experiencing 409s here to begin with, but the loop fixed it. If there is a better fix or something obviously wrong that was causing the 409s, please let me know.

pkg/mcp/k8s_settings_getter.go
```diff
@@ -0,0 +1,62 @@
...
+	mu        sync.RWMutex
+	cached    *v1.K8sSettingsSpec
+	lastFetch time.Time
+	cacheTTL  time.Duration
+}
+
+func newK8sSettingsGetter(client storage.Client) K8sSettingsGetter {
+	return &k8sSettingsGetter{
+		client:   client,
+		cacheTTL: 30 * time.Second, // Cache for 30 seconds
```
g-linville: Not sure that a cache is really necessary, but Claude added one, so I figured we might as well keep it.

pkg/mcp/k8s_settings_getter.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/obot-platform/obot/pkg/storage"
+	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
+	"github.com/obot-platform/obot/pkg/system"
+	kclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// K8sSettingsGetter provides access to K8s settings with caching
+type K8sSettingsGetter interface {
+	GetK8sSettings(ctx context.Context) (*v1.K8sSettingsSpec, error)
+}
```
g-linville: I'm really not sure that this interface is necessary. It seems the main reason it is here is for locking and caching purposes. Pretty sure caching is not needed, but I don't think locking is needed either. In the place where this Getter is used, I think we can just use an obot storage client directly instead...let me know if you have any thoughts on this.

g-linville:
Last thing I need to add is validation of the provided Helm values.

g-linville:
This is updated with the validation logic and is now properly ready for review.

pkg/api/authz/authz.go
```diff
@@ -93,6 +93,8 @@ var (
 		"/api/file-scanner-config",
 		"/api/user-default-role-settings",
 		"/api/setup/",
+		"GET /api/k8s-settings",
+		"PUT /api/k8s-settings",
```
thedadams: nit:
```suggestion
		"/api/k8s-settings",
```

pkg/storage/apis/obot.obot.ai/v1/k8ssetting.go
```diff
@@ -0,0 +1,41 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type K8sSettings struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec K8sSettingsSpec `json:"spec,omitempty"`
```
thedadams: nit: add an empty status here, just in case we need it later.

pkg/services/config.go
```diff
@@ -222,6 +228,49 @@ func buildLocalK8sConfig() (*rest.Config, error) {
 	return clientcmd.BuildConfigFromFlags("", kubeconfig)
 }
 
+// unmarshalJSONStrict unmarshals JSON with strict validation that rejects unknown fields
+func unmarshalJSONStrict(data []byte, v interface{}) error {
```
thedadams: nit:
```suggestion
func unmarshalJSONStrict(data []byte, v any) error {
```

pkg/mcp/k8s_settings_getter.go
thedadams: I don't think this abstraction is necessary.

pkg/mcp/kubernetes.go
```diff
@@ -506,26 +543,120 @@ func (k *kubernetesBackend) restartServer(ctx context.Context, id string) error
...
-						"kubectl.kubernetes.io/restartedAt": time.Now().Format(time.RFC3339),
-					},
+					"annotations": podAnnotations,
 				},
+				"spec": map[string]any{},
 			},
 		},
 	}
 
+	templateSpec := patch["spec"].(map[string]any)["template"].(map[string]any)["spec"].(map[string]any)
```
thedadams: nit: if you set `templateSpec := make(map[string]any)` above the patch, then you can set `"spec": templateSpec` in the patch and not do all the casting.

pkg/api/handlers/k8ssettings.go
thedadams: I don't see a reason why you would be getting conflicts with these current changes. I would prefer not to put in a loop here unless we absolutely need it. Can we investigate this a little more. Happy to help, if needed.

g-linville: Okay I will do some more investigation :+1:

g-linville: I think this was actually caused by a bug in Insomnia. It seems to have been firing off two (or more) requests in quick succession. I don't get 409s when I just use curl. So this seems to be a non-issue. I'm just gonna remove the loop.

g-linville: Donnie and I also talked in Slack and we decided to restructure this function a bit to make 409s slightly less likely (less time spent between get and update calls).

pkg/mcp/k8s_settings_getter.go
g-linville: wound up removing it

pkg/mcp/k8s_settings_getter.go
g-linville: Fixed

pkg/api/authz/authz.go
g-linville: Fixed

pkg/storage/apis/obot.obot.ai/v1/k8ssetting.go
g-linville: Fixed

pkg/services/config.go
g-linville: Fixed

pkg/mcp/kubernetes.go
g-linville: Fixed

pkg/mcp/kubernetes.go
```diff
@@ -506,26 +544,129 @@ func (k *kubernetesBackend) restartServer(ctx context.Context, id string) error
...
+
+func (k *kubernetesBackend) getK8sSettings(ctx context.Context) (*v1.K8sSettingsSpec, error) {
+	var settings v1.K8sSettings
+	err := k.obotClient.Get(ctx, kclient.ObjectKey{
+		Namespace: system.DefaultNamespace,
+		Name:      system.K8sSettingsName,
+	}, &settings)
+
+	return &settings.Spec, err
+}
```
thedadams: nit: this always returns a non-nil object, so probably OK to simplify.
```suggestion
func (k *kubernetesBackend) getK8sSettings(ctx context.Context) (v1.K8sSettingsSpec, error) {
	var settings v1.K8sSettings
	return settings.Spec, k.obotClient.Get(ctx, kclient.ObjectKey{
		Namespace: system.DefaultNamespace,
		Name:      system.K8sSettingsName,
	}, &settings)
}
```

StrongMonkey (approved):

Issue #4695, opened 2025-10-17
Accommodate all fields and action menu  to be visible in `Registry Entries` and `Deployment & Connection` tab in full screen mode without having to scroll horizontally.


sangee2004:
**Describe the bug**
Accommodate all fields and action menu  to be visible in `Deployment & Connection` tab in full screen mode without having to scroll horizontally.

**To Reproduce**

**Scenario 1**

1. As admin, Go to `MCP Servers` page and go to `Deployment & Connection' page

Notice that even in full screen mode , user is not presented with `Created` field and action menu.
<img width="1393" height="897" alt="Image" src="https://github.com/user-attachments/assets/09faec7a-5c57-439d-9f10-88f4e58ca9c4" />

These fields are visible only when scrolling to the right.

<img width="1348" height="912" alt="Image" src="https://github.com/user-attachments/assets/88134199-e127-411b-90fa-e3571e2ccccd" />

**Expected behavior**
Accommodate all fields and action menu  to be visible in `Deployment & Connection` tab in full screen mode without having to scroll horizontally.

**Scenario 2**

1. As admin user, Go to `MCP Servers` page
2.  Notice that `Name` field takes up lot of horizontal space in `Registry Entries` tab resulting in needing to scroll to get to the action menu drop down.

<img width="1311" height="741" alt="Image" src="https://github.com/user-attachments/assets/6478f2aa-35bc-4b64-af97-a3cfa95fb00a" />

Even all other fields can be made to take lesser space.

**Expected behavior**
Accommodate all fields and the action menu drop down  to be visible in full screen mode.

sangee2004:
This issue is addressed when testing with latest build.

All fields and action menu  in `Registry Entries` and `Deployment & Connection` are now visible without having to scroll horizontally.

Issue #4693, opened 2025-10-16
Providing independent Horizontal scrolling for title and contents in `MCP Servers` page is confusing.


sangee2004:
**Describe the bug**
Providing independent Horizontal scrolling for title and contents in `MCP Servers` page is confusing.

**To Reproduce**
1. As admin , Go to 'MCP servers' page
Notice that we are allowed to independently scroll  titles and contents  resulting in mis allignment.

<img width="1106" height="625" alt="Image" src="https://github.com/user-attachments/assets/1cbf0892-c97a-4d3f-9857-868029da7a47" />

https://www.loom.com/share/66ec309aa60a42b0bb270f3879e778be?sid=1dcd1033-371d-43eb-a9ef-873d663ba155

**Note** 
This behavior is also seen in `Deployment and Connections` and `Registry Sources` tab.
**Expected behavior**
Provide 1 horizontal scroll that will operate on both the the title and the contents.

cjellick:
@ivyjeong13 I'd like to do something better here. I do not have a great solution but here's what I want:

I never want any of our tables to have horizontal scrolling with columns that you can only see by scrolling to the sideways.

I want all the data visible. 


Solutions that I can think of:
1. Go wider with the tables.
2. Drop columns
3. Reduce column widith and just clip or wrap as needed
4. Go the github issues route and instead of doing column based tables, let the row be a bit more free form.

I know part of the issues is that we try to balance for the edgecase of `thisisareallylongnamewithoutabreakwhatareyougoingtodoaboutit` hopefully we can either force break or clip in those scenarios.

cjellick:
cc @ryu-man - just so that your aware of the above and my "edict" that i do not want tables to scroll horizontally

cjellick:
Related: https://github.com/obot-platform/obot/issues/4695

You can move these two issues along together

ryu-man:
@cjellick Roger

sangee2004:
This issue is addressed when testing with latest build.

All fields and action menu  in `Registry Entries` and `Deployment & Connection` are now visible without having to scroll horizontally.

There is no need for horizontal scrolling any more.

PR #4714, opened 2025-10-20
fix: table columns scaling & scrolling horizontally w/ fixed table header

ivyjeong13:
Addresses #4695 
Addresses #4693

### DIFF
--- a/ui/user/src/lib/components/table/Table.svelte
+++ b/ui/user/src/lib/components/table/Table.svelte
@@ -105,7 +105,6 @@
 	let headerScrollRef: HTMLDivElement | null = $state(null);
 	let bodyScrollRef: HTMLDivElement | null = $state(null);
 	let columnWidths = $state<number[]>([]);
-	let parentContainerWidth = $state<number>(0);
 
 	let tableData = $derived.by(() => {
 		let updatedTableData = data;
@@ -219,10 +218,7 @@
 			return naturalWidths;
 		}
 
-		const minWidths = naturalWidths.map((width, index) => {
-			return Math.max(width * 0.3, 100);
-		});
-
+		const minWidths = naturalWidths.map((width) => Math.max(width * 0.3, 100));
 		const totalMinWidth = minWidths.reduce((sum, width) => sum + width, 0);
 
 		if (totalMinWidth > availableWidth) {


ryu-man (approved):
I see, this works, but I would like to propose another approach to create table-like layouts with more flexibility over controlling column width. This is possible by using CSS Grid and creating custom components to define headers, rows, and columns. By using CSS Grid, we can specify flexible values at the `th` component level, such as this `width="1fr"` to make the column fill the available space when possible, or `width="auto"` to make the column fit the content, which would result in better responsiveness.

We should also consider conditionally renders columns based on screen width and hiding some columns on small devices to avoid horizontal scrolling behavior.

I was working on a UI library for Svelte, and I implemented this approach in the `DataGrid` component module. Please feel free to check it out at [https://github.com/svelte-atoms/core/blob/main/src/lib/components/datagrid/datagrid.stories.svelte](https://github.com/svelte-atoms/core/blob/main/src/lib/components/datagrid/datagrid.stories.svelte)

ivyjeong13:
> I see, this works, but I would like to propose another approach to create table-like layouts with more flexibility over controlling column width. This is possible by using CSS Grid and creating custom components to define headers, rows, and columns. By using CSS Grid, we can specify flexible values at the `th` component level, such as this `width="1fr"` to make the column fill the available space when possible, or `width="auto"` to make the column fit the content, which would result in better responsiveness.
> 
> We should also consider conditionally renders columns based on screen width and hiding some columns on small devices to avoid horizontal scrolling behavior.
> 
> I was working on a UI library for Svelte, and I implemented this approach in the `DataGrid` component module. Please feel free to check it out at https://github.com/svelte-atoms/core/blob/main/src/lib/components/datagrid/datagrid.stories.svelte | https://statuesque-boba-0fb888.netlify.app/?path=/story/atoms-datagrid--data-grid

I'll take a look!

> We should also consider conditionally renders columns based on screen width and hiding some columns on small devices to avoid horizontal scrolling behavior.

This was something I was considering! But was holding off in case we didn't want to start hiding columns if the intention is all columns should be available regardless of screen size or we could collapse columns but have a way to show the user [x] columns are hidden and can expand them if desired.

Issue #4386, opened 2025-09-23
Connectors are shown in `Configuration required` state in project view even before server instances are updated.


sangee2004:
**Describe the bug**
Connectors are shown in `Configuration required` state in project view even before server instances are updated.

**To Reproduce**
1. As admin, Create a single user MCP server  (`sangeetest2` - Tavily Search in my case)
2. As another user `user1` , connect with this MCP server.
3. As admin, update  MCP server created in step1 and add a new required configuration parameter.
This results in `server needs update` notification for server instance created in step2.
<img width="830" height="360" alt="Image" src="https://github.com/user-attachments/assets/944dd82c-8dab-48e0-ac9c-8b7dc4d8fe8d" />

4. Do not update server.
6. As `user1` , see that in project view ( from step2)  , that the connector is shown as being in configuration required state.<img width="1417" height="754" alt="Image" src="https://github.com/user-attachments/assets/4e7e82d0-eeb8-430b-882e-f4014d7895f6" />

Notice that overview page does not give user the option to `Edit Configuration`

6. Go to `My Connectors` view.
Notice that there is no configuration required indicator for the connector.

<img width="983" height="472" alt="Image" src="https://github.com/user-attachments/assets/f4deebef-b8a8-4bd9-a3c6-d2549839bf45" />

When using `Edit Configuration` user is presented only with existing configuration parameters. The new parameter that was added in step 3 is not presented to the user.

**Expected behavior**
User should not be presented with the `Configuration Required`  indicator in project view since server instance was never updated by the admin.

cjellick:
fyi grant we are discussing this here: https://acorn-io.slack.com/archives/C03FU91U4SX/p1758729274811009

sangee2004:
This issue is also seen in the following case:

1. As admin , create single use MCP server - Tavily search in my case using NPX runtime
2. As any other user , connect with this mcp server.
3. As admin , update the entry from step1 to use containerized runtime.
4. Go to `Server Details` tab

Sever instance from step2 is reported as needing update which is as expected.
<img width="937" height="428" alt="Image" src="https://github.com/user-attachments/assets/1c0251f5-b8ff-436b-907e-2168c9716324" />

Following is the drift in configuration.
<img width="1216" height="749" alt="Image" src="https://github.com/user-attachments/assets/4d857fc1-388a-44db-a440-9933919d3397" />

5. As user step2 , notice that this mcp server is reported in `Configuration required` state in project view.
[...]

PR #4718, opened 2025-10-20
Fix: Don't show Configuration Needed in chat

cjellick:
We have been incorrectly showing "Configuration Needed" for MCP servers
that have needsUpdate set to true. Users cannot update their own
servers. Only admins can, so this just drops the needsUpdate check.
Also, it drops the route for updating a server from the Basic user role.
They are not able to call that (and logic in the handler itself prevents
it).


cjellick:
Addresses https://github.com/obot-platform/obot/issues/4386

Issue #4698, opened 2025-10-17
Frontend for "feat: support catalog yaml field prefix"


cjellick:
In this PR:
https://github.com/obot-platform/obot/pull/4652

to address this issue:
https://github.com/obot-platform/mcp-catalog/issues/260

we introduced a prefix field. You can see it in use here:
https://github.com/obot-platform/mcp-catalog/blob/main/tavily_search.yaml#L52

The idea is to make it so that the end user only has to, for example, put their API key for tavily as input rather than
`Bearer APIKEY`.

I'd like frontend for this. The trick will be explaining it in a way that makes sense to the user. Good luck!

sangee2004:
@ivyjeong13 I am able to connect to `Tavily Search` MCP server that uses Prefix and able to connect successfully by providing only api key as expected.

<img width="888" height="219" alt="Image" src="https://github.com/user-attachments/assets/538c0d5e-77f2-4bd7-ab80-d7b135a606d6" />

Wanted to follow up on some issues i am seeing:

1. Wanted to check if we want to have the option to enter Prefix even for single user MCP servers when entering `User Supplied Configuration`

<img width="1138" height="702" alt="Image" src="https://github.com/user-attachments/assets/be5ad322-20f0-4e7e-a3f9-3e9611172018" />


I thought we would need the Prefix to be presented only for Remote servers when entering `User-Specified` headers?

2. When I create a remote MCP server and provide Prefix , i am not not able to see `tools` being fetched for the server because of `401 Unauthorized: Invalid authorization format` error.

Following is the configuration i use to connect to Pinecone MCP server
<img width="1142" height="765" alt="Image" src="https://github.com/user-attachments/assets/ae3a144c-10e4-4f2f-b0d5-e95c270a007f" />

I see `Bearer` being prefilled when promoted to enter the keys as expected.

<img width="835" height="369" alt="Image" src="https://github.com/user-attachments/assets/dd6f10f2-d13e-4325-886d-3a4a6b2507cf" />

Once i enter valid token value , I get connected to MCP server. Access `tools` call fails with following error
```
failed to create MCP client: failed to send request: failed to initialize HTTP Streaming client: authentication required: failed to initialize HTTP Streaming client: 401 Unauthorized: Unauthorized: authentication required: failed to initialize HTTP Streaming client: 401 Unauthorized: Invalid authorization format, expected 'Bearer [token]'
Failed to handle message: no response

oauth callback server is not configured
```

[...]

PR #4741, opened 2025-10-23
fix: make sure prefix is supplied on mcp server configure handleSave

ivyjeong13:
Addresses follow-up on #4698

### DIFF
--- a/ui/user/src/lib/components/mcp/CatalogConfigureForm.svelte
+++ b/ui/user/src/lib/components/mcp/CatalogConfigureForm.svelte
@@ -113,6 +113,19 @@
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


thedadams:
The backend already has logic that should take care of this. Perhaps there is a bug there?

In other words, the backend has logic to apply the prefix if one is set.

ivyjeong13:
> The backend already has logic that should take care of this. Perhaps there is a bug there?
> 
> In other words, the backend has logic to apply the prefix if one is set.

Ahh, I'll hold off this PR then! Will see if I can get some logs to verify!

ryu-man (approved):

ivyjeong13:
Yup, backend is adding the prefix during server configuration so frontend work shouldn't be needed! Gonna close this out

Issue #4707, opened 2025-10-17
Group based role assignments


cjellick:
#### Overview  
Extends existing group-based access control by allowing administrators to assign roles (e.g., admin, power user, auditor) to groups, not just to individual users. This streamlines permission management and supports scalable, organization-wide role assignments.

#### Key Changes

- **Role Assignment to Groups:**  
  - Admins can assign roles such as admin, power user, and auditor to groups, granting all group members the associated permissions.
  - Integrates with existing group-based access control rules for MCP Server catalogs.

- **Owner Role Consideration:**  
  - The ability to assign the owner role to a group is under review, due to its high level of privilege and its unique ability to assign auditors.

#### Rationale  
Simplifies and scales role management by leveraging group membership, reducing manual user-by-user assignments and aligning with organizational structures.

sangee2004:
Tested with obot build from `main` - [v0.0.0-dev+4549f850](https://github.com/obot-platform/obot/commit/4549f850)

I am not able to see the group-based assignments work as expected.

Steps to reproduce the problem:

Have an MS Entra auth-enabled setup. 

As the owner user, create a Group Role Assignment for `sangeetestgroup1`.

<img width="1048" height="273" alt="Image" src="https://github.com/user-attachments/assets/a49186b3-f6de-45e4-89e6-30c02fb8ceb3" />


Log in as an Entra user who is part of this group.

Notice that the user is created as "Basic User"

From `/api/me`:

```
{
  "id": "3",
  "created": "2026-01-05T17:55:47.077482Z",
  "username": "AQCn4CiRQUNEPSsILDIrXY-oLFyjT-IVy4KWjT9BFh0",
  "role": 4,
  "effectiveRole": 4,
  "groups": [
    "basic",
    "authenticated"
  ],
  "email": "sangee20041@outlook.com",
  "timezone": "America/Los_Angeles",
  "currentAuthProvider": "entra-auth-provider",
  "lastActiveDay": "2026-01-05T00:00:00Z",
  "displayName": "sangee20041@outlook.com"
}
```

Expected Behavior:

The user should be created with the role assigned to the Group it belongs to, which is PUP + auditor in this case.

g-linville:
This is fixed

sangee2004:
Group-based role assignments work as expected when testing with the latest builds using the MS Entra provider.

Issues found when testing this feature are tracked in separate issues:

https://github.com/obot-platform/obot/issues/5412
[...]

PR #4778, opened 2025-10-28
feat: add group role assignments

g-linville:
for https://github.com/obot-platform/obot/issues/4707

This adds the ability to assign roles to groups from the auth provider.

### DIFF
--- a/pkg/gateway/client/grouproleassignment.go
+++ b/pkg/gateway/client/grouproleassignment.go
@@ -108,12 +108,11 @@ func (c *Client) GetGroupRoleAssignmentsForGroups(ctx context.Context, groupName
 // ResolveUserEffectiveRole computes the effective role for a user by combining:
 // 1. Individual role from users table
 // 2. Group-based roles from GroupRoleAssignments
-// Returns the merged role with the highest permissions.
+// Returns the highest base role plus Auditor (if present).
 func (c *Client) ResolveUserEffectiveRole(ctx context.Context, user *types.User, authGroupIDs []string) (types2.Role, error) {
 	// Start with user's individual role
 	effectiveRole := user.Role
 
-	// If no auth provider groups, return individual role
 	if len(authGroupIDs) == 0 {
 		return effectiveRole, nil
 	}
@@ -132,7 +131,36 @@ func (c *Client) ResolveUserEffectiveRole(ctx context.Context, user *types.User,
 		effectiveRole |= assignment.Role
 	}
 
-	return effectiveRole, nil
+	// Normalize to keep only the highest base role + Auditor
+	return normalizeToHighestRole(effectiveRole), nil
+}
+
+// normalizeToHighestRole takes a combined role bitmap and returns only the highest
+// base role (Owner > Admin > PowerUserPlus > PowerUser > Basic) plus the Auditor bit if present.
+func normalizeToHighestRole(combinedRole types2.Role) types2.Role {
+	// Check if Auditor bit is set
+	hasAuditor := combinedRole.HasAuditorRole()
+
+	// Find the highest base role in descending order of privilege
+	var highestRole types2.Role
+	if combinedRole.HasRole(types2.RoleOwner) {
+		highestRole = types2.RoleOwner
+	} else if combinedRole.HasRole(types2.RoleAdmin) {
+		highestRole = types2.RoleAdmin
+	} else if combinedRole.HasRole(types2.RolePowerUserPlus) {
+		highestRole = types2.RolePowerUserPlus
+	} else if combinedRole.HasRole(types2.RolePowerUser) {
+		highestRole = types2.RolePowerUser
+	} else {
+		highestRole = types2.RoleBasic
+	}
+
+	// Add Auditor bit back if it was present
+	if hasAuditor {
+		highestRole = highestRole | types2.RoleAuditor
+	}
+
+	return highestRole
 }
 
 // GetUsersInGroup returns all users who are members of the given group.
@@ -203,7 +231,8 @@ func (c *Client) ResolveUserEffectiveRolesBulk(ctx context.Context, users []type
 			}
 		}
 
-		effectiveRoles[user.ID] = effectiveRole
+		// Normalize to keep only the highest base role + Auditor
+		effectiveRoles[user.ID] = normalizeToHighestRole(effectiveRole)
 	}
 
 	return effectiveRoles, nil


pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
```diff
@@ -31,33 +32,30 @@ func (h *Handler) HandleRoleChange(req router.Request, _ router.Response) error
...
-			return err
-		}
-		if err := h.ensureWorkspaceForUser(req.Ctx, req.Client, req.Namespace, *user); err != nil {
-			return err
-		}
-	} else if oldPrivileged && !newPrivileged {
+	// Get the user
+	user, err := h.gatewayClient.UserByID(req.Ctx, userIDStr)
+	if err != nil {
+		// User might have been deleted - clean up any workspaces
```
njhale: nit: is it possible/helpful to check the error type/msg to determine if the user wasn't found vs. transient error and throw an error to requeue instead?

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
```diff
@@ -74,78 +72,143 @@ func (h *Handler) isPrivilegedRole(role types.Role) bool {
...
 		return err
 	}
 
-	for _, workspace := range workspaces.Items {
-		if err := client.Delete(ctx, &workspace); err != nil && !errors.IsNotFound(err) {
+	for _, acr := range acrs.Items {
+		if err := client.Delete(ctx, &acr); err != nil && !errors.IsNotFound(err) {
 			return err
 		}
 	}
```
njhale, line 189: I think you should be able to make use of `client.DeleteAllOf` here instead of issuing individual delete calls.

pkg/gateway/server/user.go
```diff
@@ -53,13 +61,39 @@ func (s *Server) getUsers(apiContext api.Context) error {
 		return fmt.Errorf("failed to get users: %v", err)
 	}
 
-	items := make([]types2.User, 0, len(users))
+	// Filter out bootstrap user and collect valid users with their IDs
+	var validUsers []types.User
```
njhale: nit: looks like this can be initialized with a capacity of `len(users)`, since we're appending alongside `userIDs`

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
g-linville: Yes! Good call

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
g-linville: Fixed

pkg/gateway/server/user.go
g-linville: Fixed

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
g-linville: Fixed

pkg/api/handlers/setup/confirm_owner.go
```diff
@@ -94,18 +92,17 @@ func (h *Handler) ConfirmOwner(req api.Context) error {
...
 	if err := req.Create(&v1.UserRoleChange{
 		ObjectMeta: metav1.ObjectMeta{
 			GenerateName: system.UserRoleChangePrefix,
 			Namespace:    system.DefaultNamespace,
 		},
 		Spec: v1.UserRoleChangeSpec{
-			UserID:  user.ID,
-			OldRole: originalUserRole,
-			NewRole: user.Role,
+			UserID: user.ID,
```
StrongMonkey: do we not need oldRole and newRole any more?

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
```diff
@@ -74,78 +80,119 @@ func (h *Handler) isPrivilegedRole(role types.Role) bool {
...
-		FieldSelector: fields.SelectorFromSet(map[string]string{
-			"spec.userID": userID,
-		}),
-	}); err != nil {
+// cleanupWorkspaceResources removes ACRs and MCPServers when demoting to PowerUser.
+func (h *Handler) cleanupWorkspaceResources(ctx context.Context, client kclient.Client, workspace *v1.PowerUserWorkspace) error {
+	namespace := workspace.Namespace
+
+	// Delete AccessControlRules in this workspace
+	if err := client.DeleteAllOf(ctx, &v1.AccessControlRule{}, kclient.MatchingLabels{"spec.powerUserWorkspaceID": workspace.Name}, kclient.InNamespace(namespace)); err != nil {
```
thedadams: I remember this not working in the past. Just calling it out in case it wasn't thoroughly tested.

g-linville: I did a quick test, and it worked, but I didn't do a thorough test. To be safe, I think I will undo this change and switch it back to the way it was before.

g-linville: Done

go.mod
```diff
@@ -213,7 +213,7 @@ require (
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/jackc/pgpassfile v1.0.0 // indirect
 	github.com/jackc/pgservicefile v0.0.0-20240606120523-5a60cdf6a761 // indirect
-	github.com/jackc/pgx/v5 v5.7.5 // indirect
+	github.com/jackc/pgx/v5 v5.7.5
```
thedadams: nit: move this to the direct dependency block.

pkg/storage/apis/obot.obot.ai/v1/userrolechange.go
```diff
@@ -17,8 +17,8 @@ type UserRoleChange struct {
 
 type UserRoleChangeSpec struct {
 	UserID  uint       `json:"userID,omitempty"`
-	OldRole types.Role `json:"oldRole,omitempty"`
-	NewRole types.Role `json:"newRole,omitempty"`
+	OldRole types.Role `json:"oldRole,omitempty"` // deprecated and unused
+	NewRole types.Role `json:"newRole,omitempty"` // deprecated and unused
```
thedadams: If we don't use them anymore, then deleting these fields is fine.

pkg/gateway/client/group.go
```diff
@@ -129,6 +132,38 @@ func (c *Client) ListGroupIDsForUser(ctx context.Context, userID uint) ([]string
 	return groupIDs, nil
 }
 
+// GetUserGroupMemberships fetches group memberships for multiple users in a single query.
+// Returns a map of userID to slice of groupIDs.
+func (c *Client) GetUserGroupMemberships(ctx context.Context, userIDs []uint) (map[uint][]string, error) {
+	if len(userIDs) == 0 {
+		return make(map[uint][]string), nil
```
thedadams: micro-nit:
```suggestion
		return nil, nil
```

pkg/gateway/client/group.go
```diff
@@ -129,6 +132,38 @@ func (c *Client) ListGroupIDsForUser(ctx context.Context, userID uint) ([]string
...
+		Select("user_id, group_id").
+		Where("user_id IN ?", userIDs).
+		Find(&results).Error
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch user group memberships: %w", err)
+	}
+
+	// Build map
+	memberships := make(map[uint][]string)
```
thedadams:
```suggestion
	memberships := make(map[uint][]string, len(userIDs))
```

pkg/gateway/types/users.go
```diff
@@ -36,12 +36,18 @@ type User struct {
...
 	if u == nil {
 		return nil
 	}
 
-	groups := u.Role.Groups()
+	// Determine which role to use for groups
+	role := u.Role
+	if len(effectiveRole) > 0 {
+		role = effectiveRole[0]
+	}
```
thedadams: I understand what is happening here, but this is the type of thing that gets lost or forgotten about over time.

My preference would be to have two functions: `ConvertUser` and `ConvertUserWithEffectiveRole` (this would have signature `ConvertUserWithEffectiveRole(u *User, roleFixed bool, authProviderName string, effectiveRole types2.Role) *types2.User`.

The `ConvertUser` would just call `ConverUserWithEffectiveRole` and pass `u.Role` for the `effectiveRole`.

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
```diff
@@ -2,6 +2,8 @@ package poweruserworkspace
 
 import (
 	"context"
+	goerrors "errors"
```
thedadams: nano-nit: leave this as `errors` and change the `k8s` one to `apierrors`.

pkg/gateway/client/grouproleassignment.go
```diff
@@ -0,0 +1,210 @@
...
+	// Auth group IDs look like: "github:org/team", "entra:group-uuid", etc.
+	// For GroupRoleAssignments, we'll match on the full group ID as the GroupName
+	assignments, err := c.GetGroupRoleAssignmentsForGroups(ctx, authGroupIDs)
+	if err != nil {
+		return effectiveRole, err
+	}
+
+	// Merge all group roles using bitwise OR
+	for _, assignment := range assignments {
+		effectiveRole |= assignment.Role
```
thedadams, line 131: This means it's possible for the `effectiveRole` to be something like `Admin | PowerUserPlus | Basic`.

This is probably fine for now, especially as it seems we aren't storing the effective role anywhere, but this is not the original intention of the design.

pkg/gateway/client/grouproleassignment.go
```diff
@@ -0,0 +1,210 @@
...
+	if err != nil {
+		// Don't fail - fall back to individual roles
+		for _, user := range users {
+			effectiveRoles[user.ID] = user.Role
+		}
+		return effectiveRoles, nil
+	}
+
+	// Build group -> role map for fast lookup
+	groupRoles := make(map[string]types2.Role)
```
thedadams:
```suggestion
	groupRoles := make(map[string]types2.Role, len(assignments))
```

pkg/gateway/server/grouproleassignment.go
```diff
@@ -0,0 +1,253 @@
...
+
+// extractBaseRole removes the Auditor flag from a combined role to get the base role
+func extractBaseRole(role types2.Role) types2.Role {
+	return role &^ types2.RoleAuditor
+}
+
+// hasAuditorRole checks if the Auditor flag is set in the role
+func hasAuditorRole(role types2.Role) bool {
+	return role&types2.RoleAuditor != 0
+}
```
thedadams: I think these two functions deserve to be with the other role functions.

pkg/gateway/server/grouproleassignment.go
```diff
@@ -0,0 +1,253 @@
...
+// to trigger workspace reconciliation based on their current effective role.
+func (s *Server) triggerReconciliationForGroup(ctx context.Context, apiContext api.Context, groupName string) error {
+	// Get all users in this group
+	users, err := apiContext.GatewayClient.GetUsersInGroup(ctx, groupName)
+	if err != nil {
+		return fmt.Errorf("failed to get users in group %s: %w", groupName, err)
+	}
+
+	// Create UserRoleChange event for each user to trigger reconciliation
+	for _, user := range users {
```
thedadams: nit: depending on the size of the system, this could put a strain on the database. Especially since these will soon be deleted, too.

Perhaps a `GroupChange` object, whose controller handler would list the users in the group and update the users directly?

pkg/gateway/server/grouproleassignment.go
```diff
@@ -0,0 +1,253 @@
...
+			types2.RolePowerUser,
+		}
+
+		isValid := false
+		for _, validRole := range validBaseRoles {
+			if baseRole == validRole {
+				isValid = true
+				break
+			}
+		}
```
thedadams, line 213: Similar comment here: I think this deserves to be its own function and put with the other role functions.

go.mod
g-linville: Huh, for some reason I thought that `go mod tidy` would do that automatically, but it didn't

pkg/gateway/types/users.go
g-linville: Good call

pkg/gateway/client/grouproleassignment.go
g-linville: Yeah, I tested it this way and it does work. But it did feel weird to me to do it this way, and if it seems weird to you, then I think I am going to change it so that it does not harm us in the future.

pkg/gateway/server/grouproleassignment.go
g-linville: Good idea, I will implement this

pkg/gateway/server/grouproleassignment.go
g-linville: I'm not really sure how this would be useful outside of this specific context. Are you looking for like a `func (u Role) BaseRoleIsOneOf(roles []Role) bool` function?

go.mod
g-linville: Fixed

pkg/storage/apis/obot.obot.ai/v1/userrolechange.go
g-linville: Fixed

pkg/gateway/client/group.go
g-linville: Fixed

pkg/gateway/client/group.go
g-linville: Fixed

pkg/gateway/types/users.go
g-linville: Fixed

pkg/controller/handlers/poweruserworkspace/poweruserworkspace.go
g-linville: Fixed

pkg/gateway/client/grouproleassignment.go
g-linville: Fixed

pkg/gateway/server/grouproleassignment.go
g-linville: Fixed

pkg/gateway/server/grouproleassignment.go
thedadams: I was imagining something like:
```go
func IsValidRole(role Role) bool {
    ...
}
```

g-linville: Oh, well "valid" in this case specifically means "valid for assigning to a group", which is different from "valid in any sense".

pkg/gateway/server/grouproleassignment.go
g-linville: Done (I called it GroupRoleChange)

pkg/gateway/client/grouproleassignment.go
g-linville: Done

PR #4780, opened 2025-10-28
fix: show catalog as syncing when refreshing

thedadams:


g-linville (approved):

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -885,7 +885,7 @@ func convertMCPCatalog(catalog v1.MCPCatalog) types.MCPCatalog {
 		},
 		LastSynced: *types.NewTime(catalog.Status.LastSyncTime.Time),
 		SyncErrors: catalog.Status.SyncErrors,
-		IsSyncing:  catalog.Status.IsSyncing,
+		IsSyncing:  catalog.Status.IsSyncing || catalog.Annotations[v1.MCPCatalogSyncAnnotation] == "true",
```
njhale, line 888: This is getting set somewhere I assume.

njhale: Ah, I see

njhale (approved):

PR #4782, opened 2025-10-28
feat: server scheduling admin page

ivyjeong13:
Addresses #4677

### DIFF
--- a/ui/user/src/routes/admin/server-scheduling/+page.svelte
+++ b/ui/user/src/routes/admin/server-scheduling/+page.svelte
@@ -112,7 +112,7 @@
 					<div class="notification-info p-3 text-sm font-light">
 						<div class="flex items-center gap-2">
 							<Info class="size-6" />
-							<h4 class="font-semibold">Configuration Notes</h4>
+							<p class="text-md font-semibold">Configuration Notes</p>
 						</div>
 						<ul class="list-disc px-8 py-1 text-sm">
 							<li>


ui/user/src/routes/admin/server-scheduling/+page.svelte
```diff
@@ -0,0 +1,284 @@
...
+						>
+					</div>
+				</div>
+				<div class="paper mt-1">
+					<div>
+						{@render headerContent('Resource Limits & Requests')}
+						<p class="text-sm">Set CPU memory requests and limits in the deployment.</p>
+					</div>
+
+					<h3 class="text-lg font-semibold">CPU Settings</h3>
```
ryu-man, line 177: - Users should not be able to type alphabetic values.
- Users should have the ability to select different units (MB, GB,...).
- Resource value needs to be collected from number inputs and/or sliders if values are predefined like 128 MB, 256 MB, 512 MB... Do we allow users to type custom values?

ivyjeong13: @cjellick Wanted to check how important is it to get this change in? From what I saw with Kubernetes settings, you can pass an integer only or with a unit (i.e. `1` and `1000m` are both valid; `134217728` and `256Mi` are valid -- 134217728 will default to bytes). Do we wanna lock them into selecting a unit? 

edit: Or if we wanna allow that freedom, a regex to validate they inputted a proper value into those inputs?

Issue #4763, opened 2025-10-25
Filters using webhook url - When one of configured filters report failure, other filters configured for the mcp servers are not invoked.


sangee2004:
**Describe the bug**
Filters using webhook url - When one of configured filters report failure, other filters configured for the mcp servers are not invoked.

**To Reproduce**
1. Create 1 filter for `/phrase-detector` and another filter for `/secrets-detector` endpoints for the same MCP server ( `@modelcontextprotocol/server-everything` in my case) using webhook application - https://github.com/thedadams/obot-webhook-example

 2. Chat with this MCP server and echo  a phrase (so that the `echo` tool gets called) that was set in `WEBHOOK_PHRASES`.

This results in filter reporting a failure as expected.

<img width="995" height="603" alt="Image" src="https://github.com/user-attachments/assets/cc65c1af-3b30-4873-9424-45cc3c0a57dc" />

3. Go to Audit logs, and open the message details for this tool call.

Notice that the  filter for `/secrets-detector` was not invoked for this tool call as reported by `Webhook Statuses` 

```						[
  {
    "url": "https://4ae2f3557144.ngrok-free.app/phrase-detector",
    "status": "400 Bad Request",
    "message": "webhook https://4ae2f3557144.ngrok-free.app/phrase-detector returned status code 400: found phrase \"print my keys\" in message parameters\n"
  }
]
```

Webhook application logs also shows only calls being received for `Webhook Phrase` and not for `Secret Detector` 

```
2025/10/24 16:30:29 Received webhook for Webhook Phrase: Method=tools/call, ID=1.761340372e+09
webhook Webhook Phrase failed: found phrase "print my keys" in message parameters
```

**Expected behavior**
All filters configured for the MCP server should be invoked (even when failures are encountered for one of them).

sangee2004:
This issue is addressed when testing with latest build from main.

Able to see `Secret Detector` webhook also get triggered after encountering a failure with `Webhook Phrase` filter as expected.

PR #4788, opened 2025-10-29
fix: fire all webhooks on failure

thedadams:
This change will fire all webhooks, regardless of the failures status to ensure that all webhooks are called.

Issue: https://github.com/obot-platform/obot/issues/4763

### DIFF
--- a/pkg/api/handlers/mcpgateway/handler.go
+++ b/pkg/api/handlers/mcpgateway/handler.go
@@ -443,12 +443,9 @@ func fireWebhooks(ctx context.Context, webhooks []mcp.Webhook, msg nmcp.Message,
 	}
 
 	auditLog.WebhookStatuses = make([]gatewaytypes.MCPWebhookStatus, 0, len(webhooks))
-	var (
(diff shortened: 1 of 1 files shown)


pkg/api/handlers/mcpgateway/handler.go
```diff
@@ -466,7 +463,23 @@ func fireWebhooks(ctx context.Context, webhooks []mcp.Webhook, msg nmcp.Message,
...
 
-	return nil
+	switch len(rpcErrors) {
+	case 0:
+		return nil
+	case 1:
+		return rpcErrors[0]
+	default:
+		var message strings.Builder
+		message.WriteString("failed to fire webhooks: ")
```
StrongMonkey, line 473: Nit: Would be better to return error that's associated with webhooks, something like 
`errorMsg from webhook foo`? This is so that user would know which webhook has problem.

thedadams: The individual error messages that are compiled below have that information on them.

PR #4794, opened 2025-10-29
enhance: add routes to list MCP servers that need k8s settings update

g-linville:


pkg/mcp/loader.go
```diff
@@ -249,37 +244,8 @@ func (sm *SessionManager) RestartServerDeployment(ctx context.Context, server Se
 	return sm.backend.restartServer(ctx, server.Scope)
```
g-linville, line 249: This function will no longer be needed. I am keeping track on the settings hash of the MCPServer's Status, so we don't need to go grab the k8s Deployment to check the annotation each time.

StrongMonkey (approved):

Issue #4790, opened 2025-10-29
Users count for MCP servers in `Registy Entries` does not get decremented after single server MP server instances are deleted.


sangee2004:
**Describe the bug**
Users count for MCP servers in `Registy Entries` does not get decremented after single server MP server instances are deleted.

Environment: local docker installs
Consistently Reproduces: Yes .
Additional environments attempted in: Not reproducible in main.

**To Reproduce**
1. As any user, Connect with any MCP Server (Duckduckgo in my case)
2. As admin , go to `Deployment and Connections` tab and delete the MCP server instance from step1.

Notice that it takes 30 secs for the docker process relating to the deleted the MCP server instance to get actually deleted.Is this expected?

3. Go to `Registry Entries` and search for MCP Server entry from step 1 (`Duckduckgo` entry in my case)

Notice that the `Users` count continues to show as 1 even when there are no server instances for this MCP server.

https://www.loom.com/share/830252572eb440359658f06d92771bf8

**Expected behavior**
`Users` count should be reported as 0 since there are no MCP server instances.

Also is it expected to  takes 30 secs for the docker process relating to the deleted the MCP server instance to get actually deleted.

sangee2004:
This issue is addressed when testing with latest build.

PR #4806, opened 2025-10-30
fix: stop including deleted servers in user count

thedadams:
This also "refreshes" the UI so the content is up to date after deleting a server.

Issue: https://github.com/obot-platform/obot/issues/4790


njhale (approved):

thedadams:
I figured out why the checks here are failing. Instead of opening another PR for such a small change, I will merge this one.

Issue #4800, opened 2025-10-29
Filters - Webhook status reports multiple failures when filters return error.


sangee2004:
**Describe the bug**
Filters - Webhook status reports multiple failures when filters return error.

**To Reproduce**
1. Create 1 filter for `/phrase-detector` and another filter for `/secrets-detector` endpoints for the same MCP server ( `@modelcontextprotocol/server-everything` in my case) using webhook application - https://github.com/thedadams/obot-webhook-example

 2. Chat with this MCP server and echo  a phrase (so that the `echo` tool gets called) that was set in `WEBHOOK_PHRASES`.

This results in `WEBHOOK_PHRASES` filter reporting a failure as expected.

3. Go to Audit logs, and open the message details for this tool call.

Notice that the  Webhook statuses field shows 2 entries  for `/phrase-detector` one which has the error message and one with empty message

<img width="549" height="478" alt="Image" src="https://github.com/user-attachments/assets/ed937dd3-1322-4428-aae8-83cc4702fc4a" />

**Expected behavior**
Webhook statuses field should not have duplicate error entries.

sangee2004:
This issue is addressed when testing with latest build.

PR #4807, opened 2025-10-30
fix: include one webhook status on failure

thedadams:
Issue: https://github.com/obot-platform/obot/issues/4800


njhale (approved):

thedadams:
I figured out why the checks here are failing. Instead of opening another PR for such a small change, I will merge this one.

PR #4812, opened 2025-10-30
docs: reorganize navigation

cloudnautique:


docs/docs/installation/enabling-authentication.md
```diff
@@ -0,0 +1,141 @@
+# Enabling Authentication
```
drpebcak, line 1: Might be worth mentioning in here that you can't just use obot normally with the bootstrap token - you have to configure an auth provider for everything to work right

docs/docs/installation/overview.md
```diff
@@ -0,0 +1,105 @@
...
+
+Deploy Obot on major cloud platforms with platform-specific guidance.
+
+- **Cloud-specific integrations**: Use managed databases, key management, and storage services
+
+For production-ready architectures on specific clouds:
+
+- [GCP GKE Reference Architecture](reference-architectures/gcp-gke)
+- [AWS EKS Reference Architecture](reference-architectures/aws-eks)
+- [Azure AKS Reference Architecture](reference-architectures/azure-aks)
```
drpebcak, line 56: This kind of makes it seem like a 'cloud platform deployments' is a different thing than a 'kubernetes deployment'

docs/docs/installation/overview.md
```diff
@@ -0,0 +1,105 @@
...
+
+## System Requirements
+
+### Minimum (Development/Testing)
+
+- **CPU**: 1 cores
+- **RAM**: 2 GB
+- **Storage**: 10 GB
+
+### Recommended (Production)
```
drpebcak, line 66: This feels like it's duplicated from the k8s documentation.

Issue #4817, opened 2025-10-31
Should not be allowed to update the role of the last `Owner` user.


sangee2004:
**Describe the bug**
Should not be allowed to update the role of the last `Owner` user.

Pre Req - Have an auth enabled setup with  1 owner user and atleast 1 admin user.

**To Reproduce**
1. As owner user , update the role of the current logged in user from `Owner` to any other role.

Update succeeds and now we are left with no `Owner` users in the setup.

The only way now to set up another user with `Owner` role , is to log in as bootstrap user that will have the ability to set owner roles to existing users.

**Expected behavior**
We should be able to restrict the  role update of the last `Owner` user like we do for the last `Admin` user.

sangee2004:
User is now present with following error message when updating the role of the last `owner`.

<img width="410" height="106" alt="Image" src="https://github.com/user-attachments/assets/89a1b91b-cd45-41e7-8705-a92f7b754ad1" />

PR #4819, opened 2025-10-31
fix: treat owners and admins separately when removing

thedadams:
We deleting or demoting a user that is an ownwer, we would allow it if there was another admin in the system. This is incorrect because it would be impossible for any users to be promoted to owner.

This change separates the owner check from the admin check to ensure there is always at least one owner.

Issue: https://github.com/obot-platform/obot/issues/4817

### DIFF
--- a/pkg/gateway/server/user.go
+++ b/pkg/gateway/server/user.go
@@ -136,6 +136,8 @@ func (s *Server) updateUser(apiContext api.Context) error {
 			status = http.StatusNotFound
 		} else if lae := (*client.LastAdminError)(nil); errors.As(err, &lae) {
 			status = http.StatusBadRequest
+		} else if loe := (*client.LastOwnerError)(nil); errors.As(err, &loe) {
+			status = http.StatusBadRequest
 		} else if ea := (*client.ExplicitRoleError)(nil); errors.As(err, &ea) {
 			status = http.StatusBadRequest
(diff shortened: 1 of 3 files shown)


pkg/gateway/server/user.go
```diff
@@ -136,6 +136,8 @@ func (s *Server) updateUser(apiContext api.Context) error {
 			status = http.StatusNotFound
 		} else if lae := (*client.LastAdminError)(nil); errors.As(err, &lae) {
 			status = http.StatusBadRequest
+		} else if lae := (*client.LastOwnerError)(nil); errors.As(err, &lae) {
```
g-linville: pico nit


```suggestion
		} else if loe := (*client.LastOwnerError)(nil); errors.As(err, &loe) {
```

pkg/gateway/server/user.go
```diff
@@ -219,6 +221,8 @@ func (s *Server) deleteUser(apiContext api.Context) (err error) {
...
```
g-linville: pico nit


```suggestion
		} else if loe := (*client.LastOwnerError)(nil); errors.As(err, &loe) {
```

g-linville (approved):

pkg/gateway/server/user.go
thedadams: Done.

pkg/gateway/server/user.go
thedadams: Done.

Issue #4635, opened 2025-10-08
Admin-Controlled Virtual MCP Servers for Fine-Grain Tool RBAC and Server Composition


cjellick:
#### Overview
This feature enables administrators to create **virtual MCP servers**—custom, composite servers that aggregate tools from multiple underlying MCP servers. These virtual servers provide admins with precise control over which tools and servers are exposed to different user groups.

#### Key Changes

- **Virtual MCP Server Creation:**
  - Admins can create virtual MCP servers that combine tools from multiple existing MCP servers.
  - Each virtual server can be shared via a single connection link, simplifying user access.

- **Fine-Grained Tool Access Control:**
  - Admins can curate which tools are available in each virtual MCP server.
  - Example: An admin can create an “Engineering GitHub MCP Server” with only the tools relevant to engineering, and a separate “Marketing GitHub MCP Server” for marketing-specific tools—even if both are based on the same underlying GitHub MCP server.

- **Server Composition:**
  - Virtual MCP servers can be composed of tools from multiple source servers, allowing for flexible, scenario-specific server configurations.

- **Customizable Tool Presentation:**
  - Admins can edit tool names and descriptions within a virtual MCP server to better fit team workflows or clarify tool purposes.

- **Access Control Integration:**
  - Standard access control rules apply to virtual MCP servers, ensuring that only authorized users can access the tools and servers intended for them.

#### Rationale
This feature gives organizations tighter control over tool exposure and server access, supporting both security and usability. It allows admins to tailor the toolset for each team or use case, and to simplify the user experience by exposing only a single, relevant MCP server per group.

#### Notes / Considerations
- This initial release is admin-only. A future update will allow end users to create their own virtual MCP servers for personal or project-specific needs.

ivyjeong13:
@sangee2004 FYI, noticed one bug after merging 🤦‍♀️ Created bug issue: https://github.com/obot-platform/obot/issues/4863 and working on it now!

sangee2004:
When testing with latest build , Composite MCP servers feature is available

Following workflows have been validated
1. As admin , Create  Composite MCP servers  with multiple MCP servers (single user, multi user and remote)
2. When Creating  Composite MCP servers , admin is able to choose specific tool sets that can be enabled in Composite MCP servers .
3. As regular user , able to connect and chat with the composite MCP servers . 
4. User is allowed to enable and disable individual MCP servers.
[...]

PR #4841, opened 2025-11-03
enhance: ui/ux updates for composite mcp servers

ivyjeong13:
Addresses #4635 UX/UI updates

### DIFF
--- a/ui/user/src/routes/admin/mcp-servers/DeploymentsView.svelte
+++ b/ui/user/src/routes/admin/mcp-servers/DeploymentsView.svelte
@@ -470,7 +470,7 @@
 					(s) => s.manifest.runtime !== 'remote' && s.configured
 				).length}
 				{@const upgradeableCount = Object.values(currentSelected).filter(
-					(s) => s.needsUpdate
+					(s) => s.needsUpdate && !s.compositeName
 				).length}
 				<div class="flex grow items-center justify-end gap-2 px-4 py-2">
 					<button


ui/user/src/lib/components/admin/McpServerCompositeInfo.svelte
```diff
@@ -0,0 +1,147 @@
...
+	let { name, connectedUsers, classes, entity, entityId, catalogEntry, mcpServerId }: Props =
+		$props();
+	let isAdminUrl = $derived(page.url.pathname.includes('/admin'));
+	let servers = $state<MCPCatalogServer[]>([]);
+	let serversMap = $derived(new Map(servers.map((s) => [s.catalogEntryID || s.id, s])));
+
+	onMount(async () => {
+		if (!mcpServerId || !catalogEntry?.id || !entityId) return;
+
+		const deployedCatalogEntryServers =
```
ivyjeong13, line 38: Wondering here if it'd be worth to have an endpoint that just returns the catalog entry servers/multi server tied to the composite? Right now just grabbing them all to find the right ones so I can link them (around line 81)

njhale: yeah, I think that makes sense and is a reasonable follow up for me to add here.

ui/user/src/lib/components/admin/McpServerCompositeInfo.svelte
```diff
@@ -0,0 +1,147 @@
...
+			...deployedWorkspaceCatalogEntryServers.filter((s) => s.compositeName === mcpServerId)
+		];
+	});
+
+	function getAuditLogUrl(d: OrgUser) {
+		if (!catalogEntry?.id) return null;
+		if (isAdminUrl) {
+			if (!profile.current?.hasAdminAccess?.()) return null;
+			return entity === 'workspace'
+				? `/admin/mcp-servers/w/${entityId}/c/${catalogEntry.id}?view=audit-logs&user_id=${d.id}`
```
njhale: so the entry/components should never be in a workspace at this point.

ivyjeong13: Got it, will axe that! Thank you! : )

ui/user/src/lib/components/admin/McpServerCompositeInfo.svelte
ivyjeong13: Awesome, thanks Nick! 🙏

Issue #4831, opened 2025-11-03
Deleting default MCP registry does not result in MCP entries from the registry being deleted.


sangee2004:
**Describe the bug**
Deleting default MCP registry does not result in MCP entries from the registry being deleted.

**To Reproduce**
1. As owner/admin , go to `MCP Servers` page and go to `Registry Sources`
2. Delete the default registry entry

Notice that after this , we continue to see all the MCP registries continue to be present in the `Registry Entries` tab.

<img width="1224" height="437" alt="Image" src="https://github.com/user-attachments/assets/c3d7a046-9a2b-47b0-9c60-58f3bdf68754" />

<img width="1235" height="717" alt="Image" src="https://github.com/user-attachments/assets/bbf2e820-df92-4e1a-adaa-694f4806be37" />

**Note** - Deleting a MCP registry that gets added as part of `Add MCP server` ->`Add servers from Git` option results in MCP server entries  getting deleted as expected.

**Expected behavior**
Deleting default MCP registry should  result in MCP entries from the default registry being deleted.

g-linville:
Merged the fix just now.

sangee2004:
This issue is addressed when testing with latest builds.

PR #4858, opened 2025-11-04
fix: catalog sync: prune when there are no sync errors

g-linville:
for https://github.com/obot-platform/obot/issues/4831


pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -126,7 +126,9 @@ func (h *Handler) Sync(req router.Request, resp router.Response) error {
 
 	// Don't run prune if there are sync errors
 	if len(mcpCatalog.Status.SyncErrors) > 0 {
-		app.WithNoPrune()
+		app = app.WithNoPrune()
+	} else {
```
thedadams, line 130: This is only a problem if `len(toAdd) == 0`. Probably worth always adding it, though.

g-linville: true

Issue #4869, opened 2025-11-04
Multi user MCP server is not presented in the composite MCP server list when connecting to it.


sangee2004:
**Describe the bug**
Multi user MCP server is not presented in the composite MCP server list when connecting to it.

**To Reproduce**
1. Create composite MCP server with multi user MCP server (`sanmulti` in my case)  and any other MCP server (remote mcp server - `Excel` in my case)
<img width="1228" height="309" alt="Image" src="https://github.com/user-attachments/assets/ea84dcf4-11c0-4cfa-b956-7de898849e42" /> 
 
2. Connect to this composite MCP server

Notice that multi user MCP server is not presented in the  composite MCP server list

<img width="1047" height="616" alt="Image" src="https://github.com/user-attachments/assets/b92a70a1-2eae-4054-822f-507ffae22acd" />

**Expected behavior**
Multi user MCP server should be  presented in the composite MCP server list when connecting to it.

sangee2004:
This issue is addressed when testing with latest build.

Enable/Disable option is now presented for multi user MCP server. But it is not working as expected -
https://github.com/obot-platform/obot/issues/4922

Issue #4866, opened 2025-11-04
Provide a text about "Fetching tools" when waiting for server deployment to be done when configuring tools as part of  composite MCP server creation workflow.


sangee2004:
**Describe the bug**
Provide a text about "Fetching tools" when waiting for server deployment to be done when configuring tools as part of  composite MCP server creation workflow.

**To Reproduce**
1. Create a composite MCP server with few MCP servers like `Exa Search` and `Duckduckgo Search` 
2. Configure tools when prompted for it

Notice that we are presented with "Fetching tools"  loading page for  `Duckduckgo Search` 

<img width="524" height="315" alt="Image" src="https://github.com/user-attachments/assets/bc080ec3-f615-4cd4-bc70-4ac9ecb4f952" />

But when  "Fetching tools"   for  `Exa Search`  which has required parameters to enter , notice that user is presented with loading icon and no extra helper text for a long time until server is deployed

watch the video from 10th second
https://www.loom.com/share/ddba081221ef491194dc5433831e975f

**Expected behavior**
Provide a text about "Fetching tools" when waiting for server deployment to be done when configuring tools as part of  composite MCP server creation workflow.

sangee2004:
This issue is addressed when testing with latest build

<img width="815" height="355" alt="Image" src="https://github.com/user-attachments/assets/0923eea5-7c0e-4233-93b6-d6e32f5ede39" />

Issue #4865, opened 2025-11-04
Typo in text presented when configuring tools for MCP server that requires oauth.


sangee2004:
**Describe the bug**
Typo in text presented when configuring tools for MCP server that requires oauth.

**To Reproduce**
1. Create a composite MCP server and add any MCP server that requires oauth like Gmail.
2. Choose to configure tools when promoted for it

Notice that following message is presented:

In order to request tools from the server , **some** oauth authentication is required first.

<img width="415" height="312" alt="Image" src="https://github.com/user-attachments/assets/422bca70-78c5-4050-a60b-17b0a47251bf" />

**Expected behavior**
There should be no need for "some" in this text.

sangee2004:
This issue is addressed when testing with latest build.

Issue #4863, opened 2025-11-04
Clicking Cancel for Populate Tools in Entry in Composte Does Nothing


ivyjeong13:
**Describe the bug**
Clicking Cancel in User Supplied Configuration of Populate Tools doesn't do anything.

**To Reproduce**
1. Create Composite MCP Server
2. Add in an entry that requires User Supplied Configuration (ex. Firecrawl)
3. Skip
4. Expand the entry from step 3 and click Populate Tools
5. Click Get Started
6. In the configuration screen, click Cancel. It does nothing. Clicking 'X', you see the Get Started dialog still behind it too. 

**Expected behavior**
Clicking cancel should close the dialog.

**Screenshots**
<img width="744" height="405" alt="Image" src="https://github.com/user-attachments/assets/67f8bae8-755e-46c3-bf6e-f8ea5ad5eb08" />

**Additional context**

sangee2004:
This issue is addressed when testing with latest build.

PR #4889, opened 2025-11-05
fix: composite mcp servers related bugfixes

ivyjeong13:
Addresses #4863 
Addresses #4862 
Addresses #4865 
Addresses #4869 
Addresses #4866


ivyjeong13:
> Some feedback relating to #4862:
> 
> I still see the option to view audit logs for component servers by following the link on the composite MCP server
> 
>  Screen.Recording.2025-11-05.at.5.47.05.PM.mov 
> but everything else looks good after playing with this.

Should be addressed now! Thanks Nick!

njhale (approved):

Issue #4860, opened 2025-11-04
Error message shown after successful 2nd level oauth for external clients of composite MCP servers


njhale:
**Describe the bug**
After completing the second level OAuth flow using an external client successfully, the redirect page shows an error message.

**To Reproduce**
1. Connect to a composite server w/ an OAuth enabled component MCP server using an external client (e.g. VSCode)
2. Complete the 2nd level OAuth flow by either authenticating or skipping all component servers that require OAuth

**Expected behavior**
A success message is shown on the page after all OAuth prompts have successfully completed.

**Screenshots**

<img width="600" height="275" alt="Image" src="https://github.com/user-attachments/assets/44a92708-3415-406d-b645-ff62e1e62265" />

**Additional context**
OAuth still completes successfully, and the external client can connect/use/list tools, but an error is shown anyway.

sangee2004:
This issue is addressed when testing with latest build.

PR #4890, opened 2025-11-05
fix: address CORS error when redirecting for composite MCP server

thedadams:
Issue: https://github.com/obot-platform/obot/issues/4860

### DIFF
--- a/pkg/api/handlers/mcpgateway/oauth/composite.go
+++ b/pkg/api/handlers/mcpgateway/oauth/composite.go
@@ -4,6 +4,7 @@ import (
 	"crypto/rand"
 	"crypto/sha256"
 	"fmt"
+	"net/url"
 	"strings"
 
 	"github.com/obot-platform/obot/apiclient/types"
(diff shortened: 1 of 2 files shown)


njhale (approved):

pkg/api/handlers/mcpgateway/oauth/composite.go
```diff
@@ -90,18 +91,30 @@ func (h *handler) checkCompositeAuth(req api.Context) error {
...
+			"state": {authRequest.Spec.State},
+		}
+		redirectURL := authRequest.Spec.RedirectURI + "?" + q.Encode()
+		return req.Write(map[string]string{
+			"redirect_uri": redirectURL,
+		})
 	}
 
-	return req.Write(pending)
+	return nil
```
njhale: OK, this line is causing the regression. The `nil` returned here causes the UI's `doGet` function to fail to parse the response body.

I suggest either returning `req.Write(pending)` here (which we know is empty at this point), or returning an empty `[]pendingComponentAuth{}` literal here.

njhale (changes requested):

pkg/api/handlers/mcpgateway/oauth/composite.go
thedadams: Thanks for testing it out. Fixed.

njhale (approved):

Issue #4882, opened 2025-11-05
User presented with empty update page when `Edit configuration` option is used after a server instance upgrade.


sangee2004:
**Describe the bug**
User presented with empty update page when `Edit configuration` option is used after a server instance upgrade.

**To Reproduce**
1. As admin, Create composite MCP server with few MCP servers.
2. As regular user, connect to this composite MCP server
3. Chat with this composite MCP server.
4. As admin , Update composite MCP server  and add another MCP server to this composite MCP server that has required configuration parameters.
5. Upgrade the composite MCP server instance created from step 2.
6. As regular user, refresh the chat view from step2 
Notice that user is presented with an empty page with `Save` option
<img width="1251" height="569" alt="Image" src="https://github.com/user-attachments/assets/988e0381-d844-42b8-bb1c-71e51a95484e" />

6. Close this page and navigate to `Overview` page for the composite mcp server entry and click on `Edit Configuration`

Notice that user is presented with empty update page

<img width="1187" height="640" alt="Image" src="https://github.com/user-attachments/assets/73f34f74-4cca-4ebb-9621-48c9eb6e2493" />

**Expected behavior**
User should be presented with configuration parameters that can be entered for configuring the new MCP server.

sangee2004:
This issue is addressed when testing with latest build.

Issue #4870, opened 2025-11-05
When required parameters are not provided when launching composite MCP server , no error message is presented to the user.


sangee2004:
**Describe the bug**
When required parameters are not provided when launching composite MCP server , no error message is presented to the user.

**To Reproduce**
1. Create composite MCP server with multiple MCP servers - `Exa Search` , `Duckduckgo` and `Tavily Search` MCP server
2. As regular user , connect to this composite MCP server
3. When prompted to enter credentials, click `Launch` without entering any credentials.
4. Notice that user is prompted to enter  `Exa Search` credentials.
5. Enter `Exa Search` credentials.
6. Click  `Launch`

Notice that nothing happens

https://www.loom.com/share/4fd82a04925b41f5ade6d3ff951fca16

**Expected behavior**
User should be presented to error indication to enter `Tavily Search`  credentials.

sangee2004:
This issue is addressed when testing with latest build.

PR #4894, opened 2025-11-05
fix: empty edit configuration in chat when requiring update

ivyjeong13:
Addresses #4882 
Addresses #4870


njhale (approved):
Played with this locally and LGTM.

I also noticed a pre-existing issue when upgrading composites where the edit form will always be built from the latest catalog entry instead of the existing server manifest, so you'll see the new servers and config fields from the latest entry even if you haven't upgraded the MCP server yet.

I [opened a PR against your branch](https://github.com/ivyjeong13/otto8/pull/1) with a fix that worked for me. If it looks good, maybe you could pull the changes into this PR before merging.

Issue #4191, opened 2025-09-09
Microsoft Entra - User prompted to re authenticate again with MCP servers (after about 1 hour) from VS code.


sangee2004:
**Describe the bug**
Microsoft Entra - User prompted to re authenticate again with MCP servers (after about 1 hour) from VS code.

**Pre Req** - Auth enabled obot setup with Microsoft Entra provider configured.

**To Reproduce**
1. As regular user connect to few MCP servers (Duckduck go search, Antv, gmail, Elastic Search in my case)
2. Get connection string for these MCP servers for Vs code and add them to VS code.
3. Start all the MCP servers and go thru oauth workflow when prompted for authentication.
4. Make sure tools calls for these MCP servers succeed from VS code
Wait for about 1 hour

Notice that VS code will prompt users to authenticate for some of the MCP servers . In my case I was prompted for authenticating with Duckduck go search and  Antv.

When I tried to char with gmail, Elastic Search mcp servers , I was prompted to get authenticated.
<img width="1248" height="780" alt="Image" src="https://github.com/user-attachments/assets/d18bedc7-0a65-4f7c-ab61-bb26daf382f6" />

Logs from `obot MCP Gateway` from VS code around the time I was prompted to authenticate with couple of existing MCP servers:

```
2025-09-08 17:17:53.925 [info] Token for session 1490392389 is about to expire, refreshing...
2025-09-08 17:17:53.934 [error] Failed to refresh token: Error: Invalid authorization token response: {"error":"invalid_request","error_description":"invalid session"}
2025-09-08 17:17:53.934 [info] Found 0 sessions for scopes: 
2025-09-08 17:18:37.867 [info] Creating session for scopes: 
2025-09-08 17:18:37.870 [info] Opening authorization URL for scopes: 
2025-09-08 17:18:39.165 [info] Authorization code received for scopes: 
2025-09-08 17:18:39.165 [info] Exchanging authorization code for token...
2025-09-08 17:18:39.175 [info] Successfully exchanged authorization code for token.
2025-09-08 17:18:39.175 [warning] Token scopes 'undefined' do not match requested scopes ''. Overwriting token with what was requested...
2025-09-08 17:18:39.176 [info] Created session for scopes: 
2025-09-08 17:18:40.638 [info] Getting sessions for scopes: 
2025-09-08 17:18:40.638 [info] Found 1 sessions for scopes: 
2025-09-08 17:18:40.638 [info] Token for session -291752459 is about to expire, refreshing...
[...]

PR #4905, opened 2025-11-06
chore: remove cached cookies for groups; delete GitHub groups

g-linville:
related to https://github.com/obot-platform/obot/issues/4191

We will no longer be fetching group information using users' session cookies, so this removes all the functionality around caching them. We are also dropping group support for GitHub for technical reasons, so this introduces a database migration to clear all of that out.

I also made a small UI change to add support for multiline configuration parameters in auth providers (needed for changes coming to Okta).

### DIFF
--- a/pkg/gateway/client/group.go
+++ b/pkg/gateway/client/group.go
@@ -44,7 +44,7 @@ func (c *Client) ListAuthGroups(ctx context.Context, authProviderURL, authProvid
 				u.RawQuery = q.Encode()
 			}
 
-			req, err := http.NewRequestWithContext(ctx, http.MethodPost, authProviderURL+"/obot-list-auth-groups", nil)
+			req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
 			if err == nil {
 				resp, err := http.DefaultClient.Do(req)
 				if err == nil {
--- a/pkg/gateway/db/remove_github_groups.go
+++ b/pkg/gateway/db/remove_github_groups.go
@@ -9,10 +9,7 @@ import (
 
 func removeGitHubGroups(tx *gorm.DB) error {
 	// Check if tables exist
-	if !tx.Migrator().HasTable(&types.GroupMemberships{}) {
-		return nil
-	}
-	if !tx.Migrator().HasTable(&types.Group{}) {
+	if !tx.Migrator().HasTable(&types.GroupMemberships{}) || !tx.Migrator().HasTable(&types.Group{}) {
 		return nil
 	}
 


pkg/gateway/client/group.go
```diff
@@ -45,12 +44,8 @@ func (c *Client) ListAuthGroups(ctx context.Context, authProviderURL, authProvid
 				u.RawQuery = q.Encode()
 			}
 
-			req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
+			req, err := http.NewRequestWithContext(ctx, http.MethodPost, authProviderURL+"/obot-list-auth-groups", nil)
```
thedadams: Is this a breaking change? Do we need to wait for the tools repo to be updated before merging this?

pkg/gateway/db/db.go
```diff
@@ -64,6 +64,14 @@ func (db *DB) AutoMigrate() (err error) {
 		return fmt.Errorf("failed to migrate mcp_audit_log client info: %w", err)
 	}
 
+	if err = migrateIfEntryNotFoundInMigrationsTable(tx, "drop_session_cookies", dropSessionCookiesTable); err != nil {
```
thedadams, line 67: This one probably doesn't need to use this migrations table thing, but fine if we do.

pkg/gateway/db/remove_github_groups.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+func removeGitHubGroups(tx *gorm.DB) error {
+	// Check if tables exist
+	if !tx.Migrator().HasTable(&types.GroupMemberships{}) {
+		return nil
+	}
+	if !tx.Migrator().HasTable(&types.Group{}) {
+		return nil
+	}
```
thedadams, line 14: nit:
```suggestion
	if !tx.Migrator().HasTable(&types.GroupMemberships{}) || !tx.Migrator().HasTable(&types.Group{}) {
		return nil
	}
```

pkg/gateway/client/group.go
g-linville: Thanks for catching this. That was an accidental change on this line

g-linville: Fixed

pkg/gateway/db/remove_github_groups.go
g-linville: Fixed

pkg/gateway/server/tokenreview.go
```diff
@@ -36,16 +29,13 @@ func (g *gatewayTokenReview) AuthenticateRequest(req *http.Request) (*authentica
...
 	if err != nil {
 		return nil, false, err
 	}
 
-	if hashedSessionID != "" {
-		err := HandleHashedSessionID(req, g.gatewayClient, g.dispatcher, hashedSessionID, namespace, name)
-		if err != nil {
-			return nil, false, err
-		}
+	if err := populateContext(req, g.dispatcher, namespace, name); err != nil {
```
njhale, line 37: nit: Doesn't look like using this function anywhere else, and it seems like it be small enough to inline the logic here.

StrongMonkey (approved):

PR #4910, opened 2025-11-06
enhance: add app version upgrade

ivyjeong13:
Screenshot 2025-11-06 at 11 30 54 AM

### DIFF
--- a/ui/user/src/lib/components/navbar/Profile.svelte
+++ b/ui/user/src/lib/components/navbar/Profile.svelte
@@ -32,6 +32,14 @@
 	let inAdminRoute = $state(false);
 	let loadingChat = $state(false);
 
+	let showUpgradeAvailable = $derived(
+		version.current.authEnabled
+			? profile.current.isAdmin?.()
+				? version.current.upgradeAvailable
+				: false
+			: version.current.upgradeAvailable
+	);
+
 	function getLink(key: string, value: string | boolean) {
 		if (typeof value !== 'string') return;
 
@@ -111,7 +119,7 @@
 	{#snippet icon()}
 		<div class="relative">
 			<ProfileIcon />
-			{#if version.current.upgradeAvailable}
+			{#if showUpgradeAvailable}
 				<CircleFadingArrowUp
 					class="absolute -right-0.5 -bottom-0.5 z-10 size-3 rounded-full bg-white text-blue-500 dark:bg-black"
 				/>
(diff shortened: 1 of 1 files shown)


ui/user/src/lib/components/navbar/Profile.svelte
```diff
@@ -223,6 +231,21 @@
...
+			{#if version.current.upgradeAvailable}
+				<div class="flex items-center gap-1 p-1 text-[11px] text-black dark:text-white">
+					<CircleFadingArrowUp class="size-4 flex-shrink-0 text-blue-500" />
+					<p>
+						Upgrade Available. <br /> Check out the
+						<a
+							rel="external"
+							target="_blank"
+							class="text-link"
+							href="https://github.com/obot-platform/obot/releases/tag/v0.12.0"
```
g-linville: Why is 0.12.0 hardcoded here?

ivyjeong13: :+1: Good catch! Should be /latest! Will update!

ryu-man (approved):

thedadams:
I would have expect this to integrate with these changes: https://github.com/obot-platform/obot/pull/4923

ivyjeong13:
> I would have expect this to integrate with these changes: #4923

Yup! Craig asked me to go ahead and make the UI changes ahead of backend change but I can also wait till the backend goes in.

cjellick:
@thedadams I don't understand your comment. This does integrate your changes, right? It looks for version.upgradeAvailable, which is what you're adding, correct? (well it's version.current.upgradeAvailable, but i assume thats just a svelte thing.

ui/user/src/lib/components/navbar/Profile.svelte
```diff
@@ -223,6 +231,20 @@
 		</div>
 
 		{#if version.current.obot}
+			{#if version.current.upgradeAvailable}
```
cjellick: I only want this to show for admins and owners because they are the only ones capable of upgrading. Is that do-able?

If auth is off (in that case im not sure what the user has), it should also show.

ivyjeong13: Yup definitely! Lemme quickly add that in!

ivyjeong13: updated!

PR #4915, opened 2025-11-06
Fix: Bump dependencies for cves

StrongMonkey:


go.mod
thedadams, line 1: It's better to determine where these indirect dependencies come from and bump the dependency that needs them. My IDE tells you why it is needed.

Image

So this `github.com/nwaples/rardecode/v2` dependency comes from gptscript. Running `go get github.com/gptscript-ai/gptscript@main` followed by `go mod tidy` bumps both of these (and others) to the correct versions.

StrongMonkey: Shoot, my bad. Updated.

thedadams (approved):

Issue #4896, opened 2025-11-05
Resource limits and request for cpu and memory get set "0" when changes are made to affinity/toleration fields.


sangee2004:
**Describe the bug**
Resource limits and request get set "0" when changes are made to affinity/toleration fields.

**To Reproduce**
1. As admin , Go to `Server Scheduling` page
2. Set some value for `Affinity Configuration` page and `Save`

Notice that Resource limits and request for cpu and memory get set "0"

<img width="1039" height="408" alt="Image" src="https://github.com/user-attachments/assets/1f9c5383-49a1-4c24-a123-60a303f1281f" />
3. Connect to any MCP Server

Notice that the deployment created for this MCP Server connection , has following `resources` set which is incorrect

```
    resources:
      limits:
        cpu: "0"
        memory: "0"
      requests:
        cpu: "0"
        memory: "0"
```

When no changes are made to Server scheduling , following are the resources with which mcp deployments are made.

```
      resources:
        requests:
          memory: 400Mi
```

**Expected behavior**
Resources for the mcp server pods should get deployed with the default `400Mi` for memory limit. There should be no other parameters set (they should not be set "0" like it is currently done)

g-linville:
I'm 99% sure this is a UI bug, so I'm gonna add that label.

sangee2004:
This issue is addressed when testing with latest build.

Issue #4873, opened 2025-11-05
User is presented with no feedback when enabling/disabling existing mcp servers and clicking on `Update` using `Edit Configuration` option


sangee2004:
**Describe the bug**
User is presented with no feedback when enabling/disabling existing mcp servers and clicking on `Update` using `Edit Configuration` option

**To Reproduce**
1. Create composite MCP server with multiple MCP servers.
2. As regular user , Connect to composite MCP server  and enable all services.
3. Go to `My Connectors` page
4. Select the composite MCP server  from `Enabled Connectors` list and use `Edit Configuration` to disable few of the MCP servers and click on `Update`

Notice that user is in this state after `Update` is clicked for about 1 minute after which this screen disappears.
<img width="938" height="612" alt="Image" src="https://github.com/user-attachments/assets/1b97c59f-c589-49a6-8e79-f9d19f9ede30" />

`/configure` api call made by UI takes a very long time to succeed in this case
<img width="1044" height="91" alt="Image" src="https://github.com/user-attachments/assets/605eaa9f-c671-477a-b5af-8c47d8ff52ba" />

The same behavior is also noted when enabling MCP servers.

**Expected behavior**
During the time update is happening user should be presented with some progress bar similar to what gets presented when user connected to composite MCP server for the first time.

sangee2004:
This issue is addressed when testing with latest build.

I see the same issue when using `Edit Configuration` option from with in the chat view for which I will log a different issue.

PR #4918, opened 2025-11-06
fix: show progress bar editing and auto supplying 0 in request/limits in server scheduling

ivyjeong13:
Addresses #4873 
Addresses #4896

### DIFF
--- a/ui/user/src/routes/admin/server-scheduling/+page.svelte
+++ b/ui/user/src/routes/admin/server-scheduling/+page.svelte
@@ -46,23 +46,75 @@
 				}
 			};
 
+		const result = {
+			requests: {
+				cpu: '',
+				memory: ''
+			},
+			limits: {
+				cpu: '',
+				memory: ''
+			}
+		};
+
 		const segments = resources.split('\n').map((segment) => segment.trim());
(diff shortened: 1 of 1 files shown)


ui/user/src/lib/components/mcp/CatalogConfigureForm.svelte
```diff
@@ -370,13 +370,15 @@
...
 														bind:value={comp.envs[i].value}
+														disabled={form.componentConfigs[compId].disabled}
 														textarea={env.file}
 														growable
 													/>
 												{:else if env.file}
 													<textarea
 														id={`${compId}-${env.key}`}
 														bind:value={comp.envs[i].value}
+														disabled={form.componentConfigs[compId].disabled}
```
ivyjeong13, line 381: Added these so that if the server is disabled, the inputs become disabled too to indicate they shouldn't fill this out unless it's enabled. 

Was confusing when I was filling out a required field on Edit Configuration, saved, then came back and saw my field wasn't saved. Realized it was cause my Tavily server was toggled off! 😅

njhale: yeah, just noticed that too recently. Thanks for fixing that

Issue #4913, opened 2025-11-06
Not able to update composite MCP servers after deleting one of the individual MCP servers.


sangee2004:
**Describe the bug**
Not able to update composite MCP servers after deleting one of the individual MCP servers.

**To Reproduce**
1. As admin, Create composite MCP servers with few MCP servers.
2. As regular user, connect to this MCP server
3. As admin, Delete of the individual MCP servers from step 1.
4. Select the composite MCP server from `MCP Servers` list
5. Notice that the deleted MCP server is still listed in the `Configuration` tab.
6. Use `Upgrade` option to upgrade

This results in 404 errors presented and to user and upgrade does bot succeed

<img width="968" height="451" alt="Image" src="https://github.com/user-attachments/assets/5c06301e-17ef-4774-92c1-e923bf3e05ff" />

https://www.loom.com/share/03baee9fd46a439ca37e860b0eeb3419

**Expected behavior**
Should be able to upgrade composite MCP server successfully.

sangee2004:
This issue is addressed when testing with latest build.

PR #4920, opened 2025-11-06
fix: handle removed component catalog entries on upgrade

njhale:
Addresses https://github.com/obot-platform/obot/issues/4913


ivyjeong13:
Just one small thing I noticed: If I am on the Configuration page and I click 'Upgrade', I don't see the deleted MCP server removed. If I switch to another tab and come back, it updates to remove the MCP server.

njhale:
> I don't see the deleted MCP server removed. If I switch to another tab and come back, it updates to remove the MCP server.

Thanks for catching that. I'll circle back to fix this as a quick follow up.

PR #4923, opened 2025-11-06
feat: add server upgrade check

thedadams:
This change integrates with our new upgrade server so we can indicate that a given installation is in need of an upgrade. We can also track the versions our users are on.

### DIFF
--- a/pkg/api/handlers/version.go
+++ b/pkg/api/handlers/version.go
@@ -28,7 +28,7 @@ const (
 	SessionStoreCookie SessionStore = "cookie"
 
 	installationIDPropertyKey   = "installation_id"
-	defaultUpgradeServerBaseURL = "https://upgrade.obot.ai"
+	defaultUpgradeServerBaseURL = "https://upgrade-server.obot.ai"
 )
 
 func sessionStoreFromPostgresDSN(postgresDSN string) SessionStore {


pkg/services/config.go
```diff
@@ -778,11 +809,22 @@ func New(ctx context.Context, config Config) (*Services, error) {
...
-		LocalK8sConfig:          localK8sConfig,
-		MCPServerNamespace:      config.MCPNamespace,
-		K8sSettingsFromHelm:     helmK8sSettings,
+		AccessControlRuleHelper:   acrHelper,
+		WebhookHelper:             mcp.NewWebhookHelper(mcpWebhookValidationInformer.GetIndexer()),
+		LocalK8sConfig:            localK8sConfig,
+		MCPServerNamespace:        config.MCPNamespace,
+		K8sSettingsFromHelm:       helmK8sSettings,
+		ServerUpdateCheckInterval: time.Duration(config.UpdateCheckIntervalMins) * time.Minute,
+		Distribution: func() string {
```
cjellick: I think you got confused on distribution here.

The engine is meant to be kubernetes vs docker (vs local) and elsewhere you properly pass MCPRuntimeBackend as the engine value.

Distribution is meant to be oss vs enterprise.

...sorry, probably a bit more work to figure that out.

cjellick (changes requested):
See above comment regarding distribution.

I reviewed everything else reasonably closely and looks good otherwise.

pkg/services/config.go
thedadams: No, that's easier. Thanks for correcting me.

thedadams: Fixed

PR #4952, opened 2025-11-07
fix: incorrect tooltip & filtering disabled composite descendant UI

ivyjeong13:
* filters out disabled composite descendant deployments
* fix on tooltip in wrong spot
* if tooltip doesn't have text, do not enable showing tooltip on hover

### DIFF
--- a/ui/user/src/routes/admin/mcp-servers/DeploymentsView.svelte
+++ b/ui/user/src/routes/admin/mcp-servers/DeploymentsView.svelte
@@ -458,13 +458,6 @@
 
 										toggle((restarting = false));
 									}}
-									use:tooltip={d.compositeName
-										? {
-												text: 'This is a disabled component of a composite server disabled and cannot be restarted.',
-												classes: ['w-md'],
-												disablePortal: true
-											}
-										: undefined}
 								>
 									{#if restarting}
 										<LoaderCircle class="size-4 animate-spin" /> Restarting...


ui/user/src/routes/admin/mcp-servers/DeploymentsView.svelte
```diff
@@ -439,7 +469,7 @@
 									}}
 									use:tooltip={d.compositeName
 										? {
-												text: 'Cannot directly update a descendant of a composite server; update the composite MCP server instead.',
+												text: 'This descendant of a composite server is currently disabled and cannot be restarted.',
```
njhale: nit:
```suggestion
												text: 'This is a disabled component of a composite server disabled and cannot be restarted.',
```

ui/user/src/routes/admin/mcp-servers/DeploymentsView.svelte
```diff
@@ -393,6 +416,13 @@
 												server: d
 											};
 										}}
+										use:tooltip={d.compositeName
+											? {
+													text: 'Cannot directly update a descendant of a composite server; update the composite MCP server instead.',
```
njhale: nit:

```suggestion
													text: 'This is a component of a composite server and cannot be updated independently; update the composite MCP server instead',
```

njhale (changes requested):
One more thing, a blank tool tip will show up for non-components

Image

ui/user/src/routes/admin/mcp-servers/DeploymentsView.svelte
ivyjeong13: Since we're filtering out disabled descendants, removing this tooltip instead!

ivyjeong13:
> One more thing, a blank tool tip will show up for non-components
> 
> Image

Should be addressed now! : )

cjellick (approved):

njhale (approved):

PR #4956, opened 2025-11-07
Chore: Add OBOT_SERVER_UPDATE_CHECK_INTERVAL_MINS to config docs

cjellick:


cjellick:
oh also, hide the right nav on this page so that the env var config table has more room to breath

cjellick:
cc @thedadams FYI

Issue #4834, opened 2025-11-03
Show the correct status of Export without having to do a manual refresh.


sangee2004:
**Describe the bug**
Show the correct status of Export without having to do a manual refresh.

Pre Req - Have a valid `Amazon S3`  Storage provider configured. 
**To Reproduce**
1. Go to Audit logs and use `Create Export` -> `Create One-time Export`  option to Export audit logs
2. Provide `Export Name` and valid `Bucket Name` and click on `Create Export`

Notice that user is brought to the `Audit Log Exports` page were status is shown as `running` state.

<img width="1097" height="553" alt="Image" src="https://github.com/user-attachments/assets/22294435-016a-4efa-8768-ba2a94069d40" />

User needs to manually refresh the page to see the status get to `completed` state.

<img width="1093" height="513" alt="Image" src="https://github.com/user-attachments/assets/f662d16a-5061-41db-8000-35ef3fceb624" />

**Expected Behavior:**
Present the most recent state of export without having to do a manual refresh.

sangee2004:
This issue is addressed when testing with latest build.

PR #4978, opened 2025-11-10
fix: update export entry with running state automatically

ryu-man:
Addresses #4834

### DIFF
--- a/ui/user/src/lib/components/admin/audit-log-exports/CreateAuditLogExportForm.svelte
+++ b/ui/user/src/lib/components/admin/audit-log-exports/CreateAuditLogExportForm.svelte
@@ -11,7 +11,7 @@
 
 	interface Props {
 		onCancel: () => void;
-		onSubmit: (result?: Record<string, never>) => void;
+		onSubmit: (result?: AuditLogExport) => void;
 		mode?: 'create' | 'view' | 'edit';
 		initialData?: AuditLogExport;
 	}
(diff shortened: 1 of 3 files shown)


ui/user/src/lib/components/admin/audit-log-exports/CreateAuditLogExportForm.svelte
```diff
@@ -11,7 +11,7 @@
 
 	interface Props {
 		onCancel: () => void;
-		onSubmit: () => void;
+		onSubmit: (result?: Record<string, never>) => void;
```
g-linville: Can you explain why the result is a `Record<string, never>`? I am a TypeScript noob and don't understand this.

ryu-man: Sure, the `result` represent the created export entry, so it could an key-value object that's why I am typing it as a Record, the key is of type `string` and the value here is of type `never`,  `unknown` also would work, but since I am not going to do type checking on values later I used `never` just to prevent TS complaints.

g-linville: The explanations I've seen of `Record<string, never>` makes me think that an empty dictionary `{}` is the only value that this type can have.

ryu-man: I see what you mean, will make it less confusing

g-linville (approved):

Issue #4842, opened 2025-11-03
Provide drop down menu for  all fields presented in `Advanced Options` when creating one time export or scheduled exports.


sangee2004:
**Describe the bug**
Provide drop down menu for  all fields presented in `Advanced Options` when creating one time export or scheduled exports.

Pre Req - Have a valid audit storage provider configured.

**To Reproduce**
1. Go to 'Audit Logs` and Create export using `Create One-time Export` or `Create Export Schedule` 
2. Access  `Advanced Options` .

Notice that all the fields presented here does not have any drop down option that makes is very difficult for admin to provide correct values

<img width="1237" height="705" alt="Image" src="https://github.com/user-attachments/assets/7ab0d17a-e766-4f1c-96ae-292ff51fb0ff" />

**Expected behavior**

Provide drop down menu for  all fields presented in `Advanced Options` when creating one time export or scheduled exports. This should be very similar to how values are presented in `Filter` page in `Audit Logs` page.

<img width="622" height="797" alt="Image" src="https://github.com/user-attachments/assets/02a2af6f-2a49-430d-8bac-5cbb17dece5e" />

Also make the field names to match with what is presented in `Filter` page .

cjellick:
dont put this in the v0.13 release

sangee2004:
This issue is addressed when testing with latest build.
Issues found when testing this fix are tracked in https://github.com/obot-platform/obot/issues/5073 and https://github.com/obot-platform/obot/issues/5074

PR #4990, opened 2025-11-11
fix: use multi values select component in create audit log export form

ryu-man:
Addresses #4842


ui/user/src/lib/components/Select.svelte
```diff
@@ -202,7 +204,9 @@
 			{/if}
 
 			{#if multiple}
-				{@render searchInput()}
+				{#if !readonly}
```
ryu-man, line 207: In `reasonly` mode we don't have to show the search input

Issue #5021, opened 2025-11-14
Allowed to add existing composite MCP servers when creating composite MCP servers


sangee2004:
**Describe the bug**
Allowed to add existing composite MCP servers when creating composite MCP servers

**To Reproduce**
1. Create a composite MCP server and add few MCP servers to it.
2. Create another composite MCP server
3. Attempt to add MCP servers to it
Notice that UI presents the composite MCP server added in step1 and we are allowed to add this composite server successfully.
4. Trying to configure tools for this MCP server fails.
5. As regular user, trying to connect to this composite  MCP server (created in step2 fails)

<img width="1320" height="508" alt="Image" src="https://github.com/user-attachments/assets/04ff1e5e-591f-4e45-bf00-367c60657cb2" />

**Expected behavior**
Should not be dllowed to add existing composite MCP servers when creating composite MCP servers

njhale:
I'm adding validation for this and and preventing the UI from showing composites in the "Add Server" dialog.

sangee2004:
This issue is addressed when testing with latest build.
User is not allowed not add other existing composite servers when adding MCP servers to a composite server.

PR #5025, opened 2025-11-14
fix: prevent users from creating "nested" composite MCP servers

njhale:
- Hide composite catalog entries from the configuration search dialog (this was a regression)
- Add API-level validation logic to prevent users from creating nested composite MCP servers
- Add controller handlers to cleanup nested composite servers that were created before API-level validation was added

Addresses https://github.com/obot-platform/obot/issues/5021

### DIFF
--- a/ui/user/src/lib/components/admin/SearchMcpServers.svelte
+++ b/ui/user/src/lib/components/admin/SearchMcpServers.svelte
@@ -73,13 +73,16 @@
 			},
 			...mcpServerAndEntries.entries
 				.filter((entry) => {
+					if (skipComposite && entry.manifest.runtime === 'composite') {
+						return false;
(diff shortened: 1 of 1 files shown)


ui/user/src/lib/components/admin/SearchMcpServers.svelte
```diff
@@ -73,13 +73,16 @@
 			},
 			...mcpServerAndEntries.entries
 				.filter((entry) => {
+					if (skipComposite && entry.manifest.runtime === 'composite') {
+						return false;
+					}
+
 					if (type === 'filter') {
```
njhale, line 80: This line caused the regression to search that allowed users to create "nested" composites from the UI.

Issue #4974, opened 2025-11-10
Confusing user experience when updating composite MCP server fails when deleting the only mcp server but the deleted mcp server is not shown in the mcp server list after this.


sangee2004:
**Describe the bug**
Confusing user experience when updating composite MCP server fails when deleting the only mcp server but the deleted mcp server is not shown in the mcp server list after this.

**To Reproduce**
1. Create composite MCP server with 1 mcp server.
2. Access this composite MCP server , go to the configuration tab , remove this mcp server and update

Notice that User will be presented an 400 error message which can be improved and tracked in https://github.com/obot-platform/obot/issues/4965
 3. Dismiss the error message

Notice that the deleted MCP server is not present in `Component Servers` list.

Even when user goes to a different tab and revisits `Configuration` tab , the deleted mcp sever (which failed to get deleted as expected) will not be show in the `Component Servers` list.

This mcp sever shows up when a manual refresh is done.

**Expected behavior**
Show the deleted mcp server in the `Component Servers` list as soon as the user dismisses the error message presented when server update fails.

njhale:
Seems like we'll want the UI to refetch the catalog entry in the event of a failure. I'll poke around; should be a simple fix either way.

njhale:
I opened a PR to disable the save/update button until you have at least one MCP server in the list. This should make things less confusing IMO.

sangee2004:
@njhale When testing with latest build , I think the user experience can still be improved even when `Update` button is being disabled when last mcp server in the composite server is deleted.

1. Create composite mcp server with 1 mcp server - `Antv Charts` in my case
2. From `MCP Servers` page access in entry and go to  `Configuration` tab 
3. Delete  `Antv Charts` entry . `Update` option is disabled as expected.
4. Go to any other tab (like `Server Details`/`Usage`) and then revisit Configuration` tab 

When revisiting the `Configuration` tab , i am brought back to the page with no mcp servers and `Update` button disabled which is not the correct representation of the composite server at this time since we still have 1 mcp server for this composite server. There is an `/entries` call being made by the UI which shows 1 mcp server (`Antv Charts` in my case). But UI does not show this MCP server.

<img width="1339" height="1002" alt="Image" src="https://github.com/user-attachments/assets/37b0096c-cd1f-4db1-ba5c-1b04ced6a6ef" />
<img width="1215" height="640" alt="Image" src="https://github.com/user-attachments/assets/786065d7-7c68-4bcd-8875-730136c3c177" />


[...]

Issue #4965, opened 2025-11-08
Provide better error error message when trying to update composite mcp server with no mcp server fails


sangee2004:
**Describe the bug**
Provide better error error message when trying to update composite mcp server with no mcpserver fails

**To Reproduce**
1. Create composite MCP server with 1 mcp server.
2. Access this composite MCP server , go to the `configuration` tab , remove the mcp server and `update`

User will be presented with the following error message:

<img width="1154" height="704" alt="Image" src="https://github.com/user-attachments/assets/a3ef36e0-2c76-4b6d-ae18-15fee7409bc1" />

Full error message:
```
400 /mcp-catalogs/default/entries/default-dddd-cf78efd2w4qpp: failed to validate entry manifest: runtime composite validation error for field compositeConfig.componentServers: at least one component server is required
```
**Expected behavior**
Provide better error error message when trying to update composite mcp server with no mcp server fails.

sangee2004:
This issue is addressed when testing with latest build.
User is not allowed to update composite MCP server with no MCP server entries. `Update` button is disabled in this case.

Issue #4964, opened 2025-11-08
Provide better error error message when trying to add composite mcp server with no mcpserver fails.


sangee2004:
**Describe the bug**
Provide a better error error message when trying to add composite mcp server with no mcpserver fails.

**To Reproduce**
1.  Create a composite mcp server without adding any mcpservers.

User is presented with the following error message
<img width="1163" height="782" alt="Image" src="https://github.com/user-attachments/assets/0580d2e9-a40b-4238-b7f1-91a8e7b6419b" />

This is the full error when user tries to copy the error message to clipboard.
```
400 /mcp-catalogs/default/entries: failed to validate entry manifest: runtime composite validation error for field compositeConfig.componentServers: at least one component server is required
```

**Expected behavior**
Provide a better user friendly error message to the user in this case.

sangee2004:
This issue is addressed when testing with latest build.
User is not allowed to create composite MCP server with no MCP server entries. `Save` button is disabled in this case.

PR #5026, opened 2025-11-14
fix: disable save button when composite has no mcp servers

njhale:
Disable the save/update button when configuring composite catalog entries until at least one MCP server is present in the form.

Addresses:
- https://github.com/obot-platform/obot/issues/4974
- https://github.com/obot-platform/obot/issues/4965
- https://github.com/obot-platform/obot/issues/4964

### DIFF
--- a/ui/user/src/lib/components/admin/CatalogServerForm.svelte
+++ b/ui/user/src/lib/components/admin/CatalogServerForm.svelte
@@ -946,7 +946,14 @@
 			<span class="text-sm font-medium text-red-500">Fill out all required fields</span>
 		{/if}
 		<button class="button flex items-center gap-1" onclick={() => onCancel?.()}> Cancel </button>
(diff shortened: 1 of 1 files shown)


ui/user/src/lib/components/admin/CatalogServerForm.svelte
```diff
@@ -946,7 +946,14 @@
 			<span class="text-sm font-medium text-red-500">Fill out all required fields</span>
 		{/if}
 		<button class="button flex items-center gap-1" onclick={() => onCancel?.()}> Cancel </button>
-		<button class="button-primary flex items-center gap-1" onclick={handleSubmit}>
+		<button
```
njhale, line 949: This was the simplest fix that made sense to me. Something more complex would be highlighting the "Add MCP Server" section of the `CompositeRuntimeForm` when this button is clicked and no MCP servers are present.

Issue #4917, opened 2025-11-06
Test Connection button not visible for Custom S3 Compatible provider


shwetapsdet:
**Describe the bug**
When selecting the Custom S3 Compatible storage provider, the Test Connection button is not displayed.

**To Reproduce**
1. Go to Manage Exports-> Configure Storage
2. Select Custom S3 Compatible from the provider list.
3. Observe that the Test Connection button is missing.

**Expected behavior**
The Test Connection button should be visible for the Custom S3 Compatible provider.

**Screenshots**

https://www.loom.com/i/3ff8d39216dd442d904b377f827ec9a4

ryu-man:
Currently testing connection for custom S3 storage providers is not yet supported in the backend

PR #5028, opened 2025-11-14
fix: show test connection button for custom S3 storage provider

ryu-man:
Addresses #4917


StrongMonkey:
@ryu-man We don't have test connection backend functionlities for custom s3, so we are intentionally hiding it.

ryu-man:
@StrongMonkey I see. I'm going to close this PR for now

Issue #4849, opened 2025-11-04
When audit logs are exported by users with no auditor role provide a message about exported logs not having information relating to request/response headers and body.


sangee2004:
**Describe the bug**
When audit logs are exported by users with no auditor role provide a message about exported logs not having information relating to request/response headers and body.

**To Reproduce**
1. As owner/admin user with no audit role , create a one-time or scheduled export.

Notice that there is no indication to the user that the exported audit logs will not have have  information relating to request/response headers and body.

**Expected behavior**
When audit logs are exported by users with no auditor role provide a message about exported logs not having information relating to request/response headers and body.

sangee2004:
When testing with latest build , following message is presented to user when creating a one-time export as expected.
<img width="1096" height="720" alt="Image" src="https://github.com/user-attachments/assets/7da32894-a22e-4de5-ba34-0ad2059e6407" />

But this notice is not presented when creating a scheduled export.
<img width="1089" height="729" alt="Image" src="https://github.com/user-attachments/assets/5d494427-4f2b-486e-be11-91b298c1fd01" />

@ryu-man Reopening this issue for following up on the scheduled export usecase.

sangee2004:
This issue is addressed when testing with latest build

<img width="1261" height="208" alt="Image" src="https://github.com/user-attachments/assets/88ed921a-da81-4c99-860c-6619dcdd0fde" />

PR #5052, opened 2025-11-17
fix: add warning message for users without auditor permissions in scheduled export form

ryu-man:
Addresses #4849


ui/user/src/lib/components/admin/audit-log-exports/CreateScheduleForm.svelte
```diff
@@ -52,6 +53,8 @@
 	let creating = $state(false);
 	let error = $state('');
 
+	const hasAuditorPermissions = $derived(profile.current.groups.includes('auditor'));
```
ivyjeong13: `.includes(Group.AUDITOR)`

Issue #4967, opened 2025-11-08
Deleting multi user MCP server entries that are part of composite MCP server , results in them being removed from connected composite MCP server instances.


sangee2004:
**Describe the bug**
Deleting multi user MCP server entries that are part of composite MCP server , results in them being removed from connected composite MCP server instances.

**To Reproduce**
1. Create a Composite MCP server with 1 multiuser mcp server say `mcp1` and another single/remote mcp server say `mcp2`.
2. As regular user , connect to this MCP server.
Notice that user gets access to tools from both mcp servers as expected.
3. As admin user, delete multi user MCP server - `mcp1`  (from `Registry Entries` or `Deployments and Connections` tab)
4. As regular user , go to the tools tab on Composite MCP server entry  in Chat view from step2

Notice that user is presented with only tools from `mcp2` 
 
**Expected behavior**
Connected composite MCP server instances should not loose access to multiuser mcp server instance in this case.

From @njhale I think the best way to handle it is to return a validation error if the multi-user server is used by a composite and prevent its deletion.

sangee2004:
This issue is addressed when testing with latest build.

User is now not allowed to delete Multi user MCP servers that are added as part of composite MCP servers.

User is presented with following error message:

<img width="1114" height="714" alt="Image" src="https://github.com/user-attachments/assets/447560b2-dcda-4b90-8963-c953cfc26858" />

When  Multi user MCP server  gets deleted from Composite MCP server and then there is an attempt made to delete the Multi user MCP server , user will still be prevented from deleting Multi user MCP server as expected.

When all the connected  Composite MCP server instances are upgraded , we are allowed to delete Multi user MCP server .

PR #5053, opened 2025-11-17
fix: block deletion of multi-user servers with dependent composites

njhale:
Return a `409 Conflict` error when a user tries to delete a multi-user server that is in use by any composite catalog entry or running server.
Include a list of reference to the dependents in the response to make it easier for the UI to construct actionable dialogs based on the error.

Addresses backend changes for: https://github.com/obot-platform/obot/issues/4967

### DIFF
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -375,11 +375,111 @@ func (m *MCPHandler) DeleteServer(req api.Context) error {
 		)
 	}
 
+	// Prevent deletion of multi-user servers that are referenced by running composite MCP servers or catalog entries.
+	dependencies, err := m.listCompositeDeletionDependencies(req, server)
+	if err != nil {
+		return fmt.Errorf("failed to list composite deletion dependencies: %w", err)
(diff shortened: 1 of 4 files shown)


pkg/api/handlers/mcp.go
```diff
@@ -375,13 +375,114 @@ func (m *MCPHandler) DeleteServer(req api.Context) error {
 		)
 	}
 
+	// Prevent deletion of multi-user servers that are referenced by running composite MCP servers or catalog entries.
+	dependencies, err := m.listCompositeDeletionDependencies(req, server)
```
njhale, line 379: There's the potential for this to race with composite catalog/server create and update. The other option I can think of is a soft delete, where we'd add a `deleteScheduled` flag to the multi-user server so that it doesn't show up for users when they're adding MCP servers, and then only really delete it once it's been removed from every dependent.

I went with this since it was simple, quick, and didn't add more cruft to the MCP server type and other logic. LMK if that seems reasonable.

thedadams (approved):

Issue #4796, opened 2025-10-29
Composite MCP servers - Not able to generate tool preview.


sangee2004:
**Describe the bug**
Composite MCP servers - Not able to generate tool preview.

**To Reproduce**
1. Create a composite MCP server with `Gmail` and `Tavily search` MCP server.
2. Go to `Tools` tab and click on `Populate Tool Preview`
3. When prompted to enter Tavily keys , enter a key and click on `Launch`

Notice that nothing happens
<img width="1057" height="804" alt="Image" src="https://github.com/user-attachments/assets/3c040fa2-b132-4512-b06e-6e0f779f25ef" />

Following are the api calls being made:

<img width="846" height="209" alt="Image" src="https://github.com/user-attachments/assets/cf727d5f-5e66-4160-8d24-d1e2c198e64c" />

Error message for  `oauth-url`  api call `MCPServer.obot.obot.ai "default-mytestcomposite1-2aed79a3kjcgt" not found` 
Error message for  `generate-tool-previews` api call `MCP server requires OAuth authentication` 

**Expected behavior**
Should be prompted to go over the oauth workflow for gmail after which we should be presented with tools from both Tavily and Gmail MCP server.

cjellick:
marking this as v0.13.0, but @njhale just lmk if you cant get it in

njhale:
> marking this as v0.13.0, but @njhale just lmk if you cant get it in

@cjellick I don't think this will make it in at this point.

sangee2004:
This issue is addressed when testing with latest build.

Able to generate tool preview for composite MCP servers.

Generated tool preview is also rendered as expected when trying to access the `Tools` tab after selecting the composite MCP servers from `Available Connectors` list.

PR #5062, opened 2025-11-18
enhance: enable tool previews for composites

njhale:
- Enable tool previews for composite servers
- Refactor composite MCPServer creation logic
- Simplify composite MCPServer update logic

Addresses: https://github.com/obot-platform/obot/issues/4796

### DIFF
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -22,6 +22,7 @@ import (
 	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
 	"github.com/obot-platform/obot/pkg/system"
 	"github.com/obot-platform/obot/pkg/validation"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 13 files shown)


pkg/api/handlers/mcp.go
```diff
@@ -3542,182 +3517,116 @@ func (m *MCPHandler) triggerCompositeUpdate(req api.Context, server v1.MCPServer
...
-			// Instance was removed, delete it
-			if err := req.Delete(existingInst); err != nil {
-				return fmt.Errorf("failed to delete removed component instance %s: %w", existingInst.Name, err)
-			}
-		}
-	}
+	// Updating a composite server has 3 steps:
+	// 1. Create new component servers and instances
+	// 2. Update existing component servers (no-op on existing instances, since there's nothing to change)
+	// 3. Delete removed component servers and instances
```
g-linville, line 3606: Very helpful comment!

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -1151,6 +1220,7 @@ func convertServerManifestToCatalogManifest(serverManifest types.MCPServerManife
 		}
 	case types.RuntimeComposite:
 		if serverManifest.CompositeConfig != nil {
+			// TODO(njhale): Figure out if this can be removed
```
g-linville, line 1347: Just calling this out in case you wanted to do something here prior to merging.

g-linville (approved):
Backend lgtm

pkg/api/handlers/mcpcatalogs.go
njhale: I don't think this branch is actually used anywhere, but going to keep it around for now just in case

Issue #5059, opened 2025-11-18
Not able to get access to MCP servers that have been granted access to groups that user belongs to after upgrading from 0.13.0.


sangee2004:
**Describe the bug**
Not able to get access to MCP servers that have been granted access to groups that user belongs to after upgrading from 0.13.0.

**Pre Req** - Have Entra set up with few users and groups set. Have app registration configured with api permission from obot docs relating to `v0.13.0`.
Have `User1` be part of `Group1`

**Steps to reproduce the problem:**

1. Fresh install of v0.13.0
2. Enable Microsoft Entra provider.
3. Log in as admin user , deleted the existing default ACR that provides access to all MCP servers to all users in global registry.
4. Create ACR1 , that will allow `Group1` to some MCP servers say MCP1 and MCP2
5. Create ACR2 , that will allow `User1` to MCP3
6. Log in as `User1`.
User is able to see and connect (and chat) with MCP1 and MCP2 (from `Group1`) and MCP3. He does not have access to any other MCP server as expected.

7. Upgrade Obot server to latest build from main
As `User1` , I was still connected to  MCP1, MCP2 and MCP3 and was able to chat with all the connected MCP servers as expected.

8. Make necessary changes to the existing Entra app to get it configured to the latest requirements  as described in obot docs relating to `Next` release. I deleted the existing permission that was not needed for the next release (GroupMember.Read.All delegated permission)  and then added  the 2 new application permissions that was needed.

<img width="863" height="311" alt="Image" src="https://github.com/user-attachments/assets/26626145-c5f5-47a3-bcb1-00af3981e386" />

9. Log in a `User1`

Notice that we dont have access to MCP1 and MCP2 servers. The connected MCP server instances get deleted and also these MCP servers are not available in the `Available Connectors` list.
User still has access to MCP3 that was from ACR that provided access to the user directly.

10. Add new ACR that grants access to any mcp server say MCP4 from `Group1`

Notice that `User1` does not get access to MCP4 as well

**Note**

`group_memberships` table does not have group information for `User1`

User-id 4 and 5  that have group info are users that I have not logged in after upgrade.
```
 select * from group_memberships;
 user_id |                  group_id                  |          created_at           
---------+--------------------------------------------+-------------------------------
       4 | entra/ae868713-8b9b-4fd5-9bf3-ef3a0134fe31 | 2025-11-17 21:10:27.876715+00
[...]

PR #5076, opened 2025-11-19
fix: return a 500 when failing to fetch a user's groups

g-linville:
Previously, if Obot failed to fetch a user's groups, it would fail silently and delete all of the user's existing `group_memberships`. This fixes that, so that we back out of the entire authentication chain and return a custom error type that gets translated into a 500.

related to https://github.com/obot-platform/obot/issues/5059

depends on https://github.com/obot-platform/tools/pull/765

### DIFF
--- a/pkg/gateway/types/grouproleassignment.go
+++ b/pkg/gateway/types/grouproleassignment.go
@@ -1,3 +1,4 @@
+// nolint:revive
 package types
 
 import (


pkg/services/config.go
```diff
@@ -654,19 +654,19 @@ func New(ctx context.Context, config Config) (*Services, error) {
 		proxyManager = proxy.NewProxyManager(ctx, providerDispatcher)
 
 		// Token Auth + OAuth auth
-		authenticators = union.New(authenticators, proxyManager)
+		authenticators = union.NewFailOnError(authenticators, proxyManager)
```
g-linville, line 657: The difference between `union.New` and `union.NewFailOnError` is as follows.

- `union.New` will progress through the authentication chain until one authenticator returns successfully. If no authenticators return successfully, then all errors will be aggregated and returned.
- `union.NewFailOnErr` will progress through the authentication chain until one authenticator returns successfully, or until one authenticator returns an error. If an error happens, it will immediately return that error.

We need the error to return immediately, otherwise we just progress down through the chain and eventually to the anonymous authenticator, which never errors, causing the custom error type to never be processed, and the user to be unable to log in but with no error message.

I checked on the later authenticators in the chain. `ephemeralTokenServer` never returns errors, so the behavior is unchanged there. `bootstrapper` can return errors, but only for system-level things that go wrong, so it's probably good that those would get bubbled up. `authn.Anonymous{}` also never returns errors. So basically, I don't think that this change can impact any existing behavior in the authentication chain.

Issue #4977, opened 2025-11-10
Overlap between Selection fields and Name(or URL fields) in `Registries Entries`/ "Deployments and Connections`/`Registry Sources` with Search string that results in no match.


sangee2004:
**Describe the bug**
Overlap between Selection fields and Name in `Registries Entries` and Deployments and Connections` with Search string that results in no match.

**To Reproduce**
1. Go to `MCP servers` page 
2. Enter a Search string that will not fetch any matching entries for `Registries Entries`/ "Deployments and Connections`/`Registry Sources` tab.
3. Access `Deployments & Connections` tab .

Notice overlap between Selection fields and Name
<img width="1163" height="322" alt="Image" src="https://github.com/user-attachments/assets/2bdb75da-f08a-4a2e-b682-e0e3a6500dd7" />

4. Access `Registry Sources` tab .

Notice overlap between Selection fields and Url

<img width="1194" height="333" alt="Image" src="https://github.com/user-attachments/assets/0156dd08-f2be-4672-b6fc-793b896831df" />

5. Access `Registry Entries` tab .
Notice overlap between Selection fields and Name

<img width="1183" height="290" alt="Image" src="https://github.com/user-attachments/assets/e1e7483d-7640-4c41-972d-e03e1cf9d7f5" />

**Expected behavior**
There should be no overlap between Selection fields and other fields.

sangee2004:
@ryu-man When testing with latest build , I see the search string getting cleared when switching between different tabs.

It is useful to have the search string not getting cleared when switching tabs between `Registry Entries` and `Deployment and Connections` 


Also the behavior seen withe overlap still exists when revisiting the tab with no entries

https://www.loom.com/share/e8bc6335e8ec4b22ac9083be5ba00703


Reopening for followup

ryu-man:
> It is useful to have the search string not getting cleared when switching tabs between Registry Entries and Deployment and Connections

@sangee2004  This behavior is confusing, because the user would expect to see the result of what is typed in the search input! unless if the search input is supposed to filter only `Registry Entries`, is that correct?

sangee2004:
@sangee2004 This behavior is confusing, because the user would expect to see the result of what is typed in the search input! unless if the search input is supposed to filter only Registry Entries, is that correct?

Yes. My expectation was the for the entered search string to be present (and applied ) when switching between tabs.

sangee2004:
This issue is addressed when testing with latest build.

PR #5077, opened 2025-11-19
fix: improve cross tab filtering in mcp servers page

ryu-man:
Addresses #4977

- Sync the search query with the URL.
- Save the search query for each view in the local storage.
- Load view search query on change.

### DIFF
--- a/ui/user/src/routes/admin/mcp-servers/+page.svelte
+++ b/ui/user/src/routes/admin/mcp-servers/+page.svelte
@@ -174,7 +174,7 @@
 	// Helper function to persist query to local storage
 	function persistQueryToLocalStorage(view: View, queryValue: string): void {
 		if (!localStorageViewQuery.current) {
-			// Do nothing if local value has not loaded yets
+			// Do nothing if local value has not loaded yet
 			return;
 		}
 
@@ -183,6 +183,7 @@
 
 	function clearQueryFromLocalStorage(view?: View): void {
 		if (!localStorageViewQuery.current) {
+			// Do nothing if local value has not loaded yet
 			return;
 		}
 


ui/user/src/routes/admin/mcp-servers/+page.svelte
```diff
@@ -34,13 +34,24 @@
...
 
 	let view = $state<View>((page.url.searchParams.get('view') as View) || 'registry');
 	const defaultCatalogId = DEFAULT_MCP_CATALOG_ID;
 
+	const query = $derived(page.url.searchParams.get('query') || '');
+
+	type LocalStorageViewQuery = Record<View, string>;
+	const localStorageViewQuery = localState<LocalStorageViewQuery>(
+		'@obot/admin/mcp-servers/search-query',
+		{} as LocalStorageViewQuery
```
ivyjeong13: Wondering if the default value ` { registry: '', deployments: '', urls: '' }` can be added here?

ui/user/src/routes/admin/mcp-servers/+page.svelte
```diff
@@ -146,11 +163,50 @@
...
+		if (value) {
+			url.searchParams.set(key, value);
+		} else {
+			url.searchParams.delete(key);
+		}
+	}
+
+	// Helper function to persist query to local storage
+	function persistQueryToLocalStorage(view: View, queryValue: string): void {
+		if (!localStorageViewQuery.current) {
```
ivyjeong13, line 176: Tagging to same thought of the default value, if default value is set, does this `if` check and setting of the default value need to happen here?

ui/user/src/routes/admin/mcp-servers/+page.svelte
```diff
@@ -92,7 +109,7 @@
 	let selectServerTypeDialog = $state<ReturnType<typeof SelectServerType>>();
 	let selectedServerType = $state<'single' | 'multi' | 'remote' | 'composite'>();
 	let selectedEntryServer = $state<MCPCatalogEntry | MCPCatalogServer>();
-	let query = $state('');
+	// let query = $state('');
```
ivyjeong13: Can remove this line!

ui/user/src/routes/admin/mcp-servers/+page.svelte
ryu-man: Indeed, it can be set initially

ui/user/src/routes/admin/mcp-servers/+page.svelte
ryu-man: By design Yes, initially `localStorageViewQuery.current` will be `undefined` until `localState` rune's effect runs, where it either returns the value from localStorage if defined, or otherwise it will return the default value supplied.

ryu-man: Going to change the code to do nothing instead of intilizing the value

Issue #5080, opened 2025-11-19
Obot should work in HA mode


StrongMonkey:

StrongMonkey:
@sangee2004 To test HA with obot, you can run postgres locally in cluster and having obot point to it.

To install postgres:
```
helm repo add bitnami https://charts.bitnami.com/bitnami
helm repo update
helm install pg bitnami/postgresql
```

To get the secret of db

```
kubectl get secret --namespace default pg-postgresql -o jsonpath="{.data.postgres-password}" | base64 -d
```

Run obot to point to the db

```
helm install obot chart/ --set image.repository=ghcr.io/obot-platform/obot-enterprise  --set image.tag=main --set config.OBOT_SERVER_ENABLE_AUTHENTICATION=true --set config.OBOT_BOOTSTRAP_TOKEN=bootstrap  --set config.OBOT_SERVER_DSN=postgresql://postgres:${password}@pg-postgresql.default.svc/postgres
```

shwetapsdet:
Replica Scaling:

- 1 → 3 Pods: 1 pod initially failed (Postgres recovery/TLS), stabilized in ~1 min; UI fully functional.
- 3 → 5 Pods: 3 new pods temporarily in CrashLoopBackOff; existing pods ran fine; UI remained functional.
- 5 → 2 Pods & 2 → 1 Pod: Scale down successful; no downtime; UI/API stable.

Failover / HA Validation:

- Deleted 1 pod while running 3 replicas; traffic served by remaining pods; no downtime.

Database & Service Stability:

- Postgres initially refused TLS and was in recovery mode; recovered automatically.
- UI remained up during all scaling tests.

Issues Observed at Higher Replica Counts:

- CrashLoopBackOff for new pods (5–10 replicas).
- 500 /projects, slow MCP connector loading, 409 Conflict on task edits, 401 Session Expired.
- Multi-window usage (3–4 browsers) caused UI lag.
- 15 pods: UI initially failed to load; chat lag; multiple pods in CrashLoopBackOff; cluster unstable.

- Re-tested all scenarios using Okta authentication provider.

Issue Tracking:

https://github.com/obot-platform/obot/issues/5176
https://github.com/obot-platform/obot/issues/5175
https://github.com/obot-platform/obot/issues/5177
https://github.com/obot-platform/obot/issues/5178

PR #5081, opened 2025-11-19
Fix: Use persistent token to generate for runs

StrongMonkey:
The ephemaral token is causing in HA setup where different replicas hold different signing secrets. Instead, we should switch all run token to be persistent token signed by a global secret stored in database. This PR changed the implementation of that.

https://github.com/obot-platform/obot/issues/5080

### DIFF
--- a/pkg/jwt/persistent/persistent.go
+++ b/pkg/jwt/persistent/persistent.go
@@ -47,7 +47,7 @@ const (
 	TokenTypeRun TokenType = "run"
 )
 
-// EnsureJWK ensures that the JWK is created and stored in the GPTScript client. It should only be called in a controller pre-start hook which only allows one to be run at a time.
+// EnsureJWK ensures that the JWK is created and stored in the GPTScript client. It should only be called in a controller post-start hook which only allows one to be run at a time.
 func (t *TokenService) EnsureJWK(ctx context.Context) error {
 	// Read the credential, if it exists, then use it.
 	cred, err := t.credOnlyGPTClient.RevealCredential(ctx, []string{system.JWKCredentialContext}, system.JWKCredentialContext)


pkg/controller/controller.go
```diff
@@ -64,6 +64,10 @@ func (c *Controller) PreStart(ctx context.Context) error {
 		return fmt.Errorf("failed to apply data: %w", err)
 	}
 
+	if err := c.services.PersistentTokenServer.EnsureJWK(ctx); err != nil {
```
thedadams: I think the intention of moving this here was to ensure that it is only run by the leader, but `PreStart` is run by every container.

pkg/jwt/persistent/persistent.go
```diff
@@ -30,28 +29,52 @@ type TokenService struct {
...
-		return nil, err
+type TokenType string
+
+const (
+	TokenTypeRun TokenType = "run"
+)
+
+// EnsureJWK ensures that the JWK is created and stored in the GPTScript client. It should only be called in a controller pre-start hook which only allows one to be run at a time.
+func (t *TokenService) EnsureJWK(ctx context.Context) error {
+	_, err := ensureJWK(ctx, t.credOnlyGPTClient)
```
thedadams: nit: I believe this is the only place `ensureJWK` is used and we don't need the first returned parameter. It can be removed.

pkg/controller/controller.go
StrongMonkey: yeah, didn't know preStart is called on every container. Moved to postStart, which should be called in leader election context.

thedadams (approved):

pkg/jwt/persistent/persistent.go
```diff
@@ -30,28 +29,83 @@ type TokenService struct {
...
 
-	if err = t.replaceKey(ctx, key); err != nil {
-		return nil, err
+type TokenType string
+
+const (
+	TokenTypeRun TokenType = "run"
+)
+
+// EnsureJWK ensures that the JWK is created and stored in the GPTScript client. It should only be called in a controller pre-start hook which only allows one to be run at a time.
```
g-linville:
```suggestion
// EnsureJWK ensures that the JWK is created and stored in the GPTScript client. It should only be called in a controller post-start hook which only allows one to be run at a time.
```

g-linville (approved):
Just one comment that needs to change.

Issue #5093, opened 2025-11-20
Not able to chat with `Failed with status 401: Invalid client credentials` message presented.


sangee2004:
**Describe the bug**

**To Reproduce**
1. Create composite Sever with Antv Chart  and Tavily Search
2. As regular user, Connect to this composite MCP server. Able to chat with both mcp servers.
3. As admin, add Exa Search
4. Update composite Sever instance that was created from step 2
5. As regular user , refresh chat view from step 2 and enter valid credentials when prompted for exa creds.
6. Even after this , I am not able to chat and see the following errors

<img width="443" height="506" alt="Image" src="https://github.com/user-attachments/assets/08545bcb-9a0f-4c15-916b-d35da3ede586" />

7. Access `Tools` tab for the composite server entry

Notice that there are no tools presented and `tools` call fails with following error
```
failed to list tools: failed to list MCP tools: error from server: JSON RPC internal error: error from server: JSON RPC internal error: error from server: JSON RPC internal error: failed to send request: failed to initialize client: failed to exchange token: token exchange failed with status 401: Invalid client credentials
```

<img width="1295" height="818" alt="Image" src="https://github.com/user-attachments/assets/790e4cfd-52b5-4c41-88f1-028aec0e48e2" />

**Note** - This issue is seen intermittently

When I tried to make a new connection to the same  Composite MCP server , i see the same `token exchange failed with status 401: Invalid client credentials` error happen.

I tried to make another new connection to the same  Composite MCP server and it succeeded this time. I was able to make tool calls to all 3 MCP servers.

**Expected behavior**
Should be able to chat successfully with all mcp servers .

I also saw the same error when I was trying to connect with a remote MCP server - Pinecone MCP server with the following configuration:

<img width="1253" height="750" alt="Image" src="https://github.com/user-attachments/assets/85464b49-c19e-488d-96e0-62d4f14a4893" />

When connecting to this MCP server , I was prompted to enter the credentials. After entering valid credentials , I see the following error when chatting 

<img width="988" height="712" alt="Image" src="https://github.com/user-attachments/assets/fd523bfa-f441-4b3e-aa6a-d589a2868478" />

`tools` call fails with following error:
```
failed to list tools: failed to list MCP tools: error from server: JSON RPC internal error: failed to send request: failed to initialize client: failed to exchange token: token exchange failed with status 401: Invalid client credentials
```

[...]

PR #5098, opened 2025-11-21
fix: trim prefix to parse client credentials

thedadams:
Trimming "Basic" instead of trimming the prefix would mean we would remove characters and decoding the base64 encoded string would fail.

Issue: https://github.com/obot-platform/obot/issues/5093


thedadams:
Closing. Wrong remote

Issue #5045, opened 2025-11-15
Implement Registry API with OAuth for Obot Integration


cjellick:
- **Registry API Endpoint**
  - When a user hits a predefined URL for an Obot installation, it should serve the registry API.
  - Only the read parts of the API are needed for now (e.g., list, get).
  - The endpoint should be protected by OAuth (see draft PR: [add registry authorization to specification](https://github.com/modelcontextprotocol/registry/pull/756/files)).
  - The authorization server will always be Obot itself (no external auth servers).
  - Should only need to support a single scope (e.g., registry:read).
  - No UI changes are needed for this pass; just present everything the user has access to at the endpoint so that clients can consume it.
  - The endpoint should ideally be at the root (e.g., `chat.obot.ai/`), but `chat.obot.ai/registry` is also acceptable.

- **Access Control**
  - For MVP: one global registry that aggregates all access control rules.
  - Post-MVP: support a registry URL for each access control rule (e.g., for power users to share specific registries).

- **Other Considerations**
  - Handling of local running servers (for developers) is out of scope for this release.
  - Handling of unconfigured "single user" servers needs thought through but I'm hoping we can get away with something somewhat hacky for them

### Deliverables

- [ ] Implement the registry API endpoint with OAuth protection.
- [ ] Support only read operations (list, get, whatever is in the spec).
- [ ] Use Obot as the OAuth authorization server.
- [ ] Expose everything the user has access to at a "global" endpoint at the root or `/registry` as appropriate.
- [ ] Document the endpoint and OAuth flow.

g-linville:
I forgot to move this to Testing before. @sangee2004 has already done a lot of validation on this.

sangee2004:
Validated basic workflow relating to MCP registry was tested with No-Auth Mode (Default) mode using `VS Client Insider` as MCP Registry client. Testing was done with both auth enabled and auth disabled obot setup using 0.14.0 rc builds.


Issues found when testing this feature is tracked in issues logged with label `MCP Registry` - https://github.com/obot-platform/obot/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22MCP%20Registry%22

Will continue to validate this feature in `main` branch from which the next release will be made.

PR #5100, opened 2025-11-21
feat: add MCP registry API support

g-linville:
for https://github.com/obot-platform/obot/issues/5045

### DIFF
--- a/pkg/api/handlers/mcpgateway/oauth/authorize.go
+++ b/pkg/api/handlers/mcpgateway/oauth/authorize.go
@@ -303,7 +303,7 @@ func (h *handler) callback(req api.Context) error {
 		return nil
 	}
 
-	if mcpID := req.PathValue("mcp_id"); mcpID != "" && mcpID != "registry" {
+	if mcpID := req.PathValue("mcp_id"); mcpID != "" {
 		// Check whether the MCP server needs authentication.
 		jwks, err := h.jwks(req.Context())
 		if err != nil {


chart/values.yaml
```diff
@@ -89,6 +89,8 @@ config:
   GITHUB_AUTH_TOKEN: ""
   # config.OBOT_SERVER_ENABLE_AUTHENTICATION -- Enables authentication for Obot
   OBOT_SERVER_ENABLE_AUTHENTICATION: false
+  # config.OBOT_SERVER_ENABLE_REGISTRY_AUTH -- Enables authentication for the MCP registry API. When false (default), registry is accessible without authentication and returns only default catalog items with wildcard access control rules.
+  OBOT_SERVER_ENABLE_REGISTRY_AUTH: false
```
drpebcak, line 93: im curious about this choice to turn the auth off by default, even if `OBOT_SERVER_ENABLE_AUTHENTICATION` is `true` - won't most people expect `OBOT_SERVER_ENABLE_REGISTRY_AUTH` to be `true` by default if they enabled auth?

drpebcak (approved):

chart/values.yaml
g-linville: This was by request from @ibuildthecloud

pkg/jwt/persistent/persistent.go
```diff
@@ -30,13 +30,15 @@ type TokenService struct {
 	gatewayClient     *client.Client
 	credOnlyGPTClient *gptscript.GPTScript
 	serverURL         string
+	authEnabled       bool
```
thedadams: This looks unused.

pkg/api/server/server.go
```diff
@@ -163,6 +165,11 @@ func (s *Server) Wrap(f api.HandlerFunc) http.HandlerFunc {
 				})
 			}
 
+			// Only set WWW-Authenticate if not in no-auth mode
+			if strings.HasPrefix(req.URL.Path, "/v0.1") && !s.registryNoAuth {
+				rw.Header().Set("WWW-Authenticate", fmt.Sprintf(`Bearer realm="MCP Registry", scope="mcp-registry:read", resource_metadata="%s/.well-known/oauth-protected-resource/registry"`, strings.TrimSuffix(s.baseURL, "/api")))
```
thedadams: From the perspective of the OAuth Protected Resource RFC, I believe this is fine.

However, the MCP spec is a little more prescriptive. For MCP servers, it states:
>Well-Known URI: Serve metadata at a well-known URI as specified in [RFC9728](https://datatracker.ietf.org/doc/html/rfc9728). This can be either:
At the path of the server’s MCP endpoint: https://example.com/public/mcp could host metadata at https://example.com/.well-known/oauth-protected-resource/public/mcp
At the root: https://example.com/.well-known/oauth-protected-resource

This is for registry, so a little different, but I don't know if we would run into compatibility issues with this. Perhaps appending `/v0/servers` instead of `/registry` would be the way to go.

pkg/api/handlers/wellknown/oauth.go
```diff
@@ -39,3 +41,20 @@ func (h *handler) oauthProtectedResource(req api.Context) error {
...
+		return &types.ErrHTTP{
+			Code:    http.StatusNotFound,
+			Message: "Registry OAuth is not available when registry authentication is disabled",
+		}
+	}
+
+	return req.Write(fmt.Sprintf(`{
+	"resource": "%s",
+	"authorization_servers": ["%[1]s"],
+	"scopes_supported": ["mcp-registry:read"],
```
thedadams: nit: I say we remove this until we actually implement scopes. It is optional in the spec.

apiclient/types/registry.go
```diff
@@ -0,0 +1,73 @@
...
+
+// RegistryServerResponse wraps a server with registry metadata
+type RegistryServerResponse struct {
+	Server ServerDetail `json:"server"`
+	Meta   RegistryMeta `json:"_meta,omitzero"`
+}
+
+// ServerDetail matches the Registry API ServerDetail schema
+// For Obot, configured servers always use Remotes (never Packages)
+type ServerDetail struct {
```
thedadams: nit: for consistency, I think the types related to servers in the registry should be prepended with `RegistryServer`. Having `ServerDetail` and `RegistryRemote` confused me.

docs/docs/concepts/mcp-gateway/registry-api.md
```diff
@@ -0,0 +1,323 @@
+# Registry API
+
+The Registry API provides a standardized, MCP-compliant interface for discovering and accessing MCP servers configured in your Obot instance. This API follows the [MCP Registry specification](https://github.com/modelcontextprotocol/registry/blob/main/docs/reference/api/generic-registry-api.md), enabling MCP clients to programmatically discover available servers based on user permissions.
+
+## Overview
+
+The Registry API exposes MCP servers through a unified `/v0.1/servers` endpoint. The servers returned depend on the authentication mode:
+
+**No-Auth Mode (Default)**: Returns servers from the default catalog with wildcard Access Control Rules, accessible without authentication. This is ideal for public Obot instances sharing a curated set of MCP servers.
```
thedadams: The phrase "wildcard Access Control Rules" seems like an implementation detail. Could we translate that into what it means for the admin/user for the docs?

docs/docs/concepts/mcp-gateway/registry-api.md
```diff
@@ -0,0 +1,323 @@
...
+GET /v0.1/servers
+```
+
+Returns a paginated list of MCP servers. In No-Auth Mode, returns servers from the default catalog with wildcard ACRs. In Auth Mode, returns all servers accessible to the authenticated user.
+
+#### Query Parameters
+
+| Parameter | Type | Default | Description |
+|-----------|------|---------|-------------|
+| `cursor` | string | - | Pagination cursor (server name from previous response) |
```
thedadams:
```suggestion
| `cursor` | string | - | Pagination cursor |
```

docs/docs/concepts/mcp-gateway/registry-api.md
```diff
@@ -0,0 +1,323 @@
...
+**Auth Mode**: Returns all MCP servers that an authenticated user has access to, including:
+- **Personal Servers**: Single-user servers deployed specifically for you
+- **Catalog Servers**: Multi-user servers shared across your organization
+- **Workspace Servers**: Servers available within your Power User workspace
+
+All servers are returned in a standardized format, regardless of their underlying runtime (npx, uvx, containerized, or remote).
+
+To enable Auth Mode, set the environment variable: `OBOT_SERVER_ENABLE_REGISTRY_AUTH=true`
+
+## API Endpoints
```
thedadams, line 20: Can we refer them to the MCP registry spec for this section instead of providing examples?

pkg/accesscontrolrule/helper.go
```diff
@@ -276,6 +276,76 @@ func (h *Helper) UserHasAccessToMCPServerCatalogEntry(user kuser.Info, entryName
 	return h.UserHasAccessToMCPServerCatalogEntryInCatalog(user, entryName, system.DefaultCatalog)
 }
 
+// HasWildcardAccessToMCPServerCatalogEntryInCatalog checks if there are ACRs with wildcard selector for an entry
+func (h *Helper) HasWildcardAccessToMCPServerCatalogEntryInCatalog(entryName, catalogID string) (bool, error) {
```
thedadams, line 280: Do we need the corresponding functions for workspaces?

pkg/api/handlers/registry/utils.go
```diff
@@ -0,0 +1,96 @@
...
+	// Reverse the parts
+	for i, j := 0, len(parts)-1; i < j; i, j = i+1, j-1 {
+		parts[i], parts[j] = parts[j], parts[i]
+	}
+
+	// Normalize each label to be DNS-compliant
+	normalizedParts := make([]string, len(parts))
+	for i, part := range parts {
+		normalizedParts[i] = normalizeDNSLabel(part)
+	}
```
thedadams: nit: slightly more efficient:
```suggestion
	// Normalize each label to be DNS-compliant
	normalizedParts := make([]string, 0, len(parts))
	for _, part := range slices.Backward(parts) {
		normalizedParts = append(normalizedParts, normalizeDNSLabel(part))
	}
```

pkg/api/handlers/registry/utils.go
```diff
@@ -0,0 +1,96 @@
...
+	// Convert to lowercase
+	label = strings.ToLower(label)
+
+	// Replace invalid characters with hyphens
+	label = dnsLabelRegex.ReplaceAllString(label, "-")
+
+	// Collapse multiple hyphens
+	for strings.Contains(label, "--") {
+		label = strings.ReplaceAll(label, "--", "-")
+	}
```
thedadams: nit: I believe this is correct (untested) and slightly more efficient.
```suggestion
	label = string(slices.CompactFunc([]byte(label), func(a, b byte) bool {
		return a == '-' && b == a
	}))
```

pkg/api/handlers/registry/utils.go
```diff
@@ -0,0 +1,96 @@
...
+// isIPAddress checks if a hostname is an IP address
+func isIPAddress(hostname string) bool {
+	// Simple check: contains only digits, dots, and colons (IPv4 or IPv6)
+	for _, char := range hostname {
+		if char != '.' && char != ':' && (char < '0' || char > '9') && (char < 'a' || char > 'f') && (char < 'A' || char > 'F') {
+			return false
+		}
+	}
+	return strings.Contains(hostname, ".") || strings.Contains(hostname, ":")
+}
```
thedadams, line 79: Seems like a good use for `net.ParseIP` instead of rolling our own.

pkg/api/handlers/registry/convert.go
```diff
@@ -0,0 +1,300 @@
...
+	credEnv map[string]string,
+	serverURL string,
+	slug string,
+	reverseDNS string,
+	userID string,
+) (obottypes.RegistryServerResponse, error) {
+	// Skip composite servers
+	if server.Spec.Manifest.Runtime == obottypes.RuntimeComposite {
+		return obottypes.RegistryServerResponse{}, fmt.Errorf("composite servers are not supported")
+	}
```
thedadams: Is there a reason we are skipping composite? Is there a meaningful difference?

pkg/api/handlers/registry/handler.go
```diff
@@ -0,0 +1,820 @@
...
+
+// ListServers handles GET /v0.1/servers
+func (h *Handler) ListServers(req api.Context) error {
+	// Parse query parameters
+	cursor := req.URL.Query().Get("cursor")
+	limit := parseLimit(req.URL.Query().Get("limit"))
+	search := req.URL.Query().Get("search")
+
+	// Get reverse DNS from base URL (strip /api suffix if present)
+	baseURL := strings.TrimSuffix(h.serverURL, "/api")
```
thedadams: `h.serverURL` won't have `/api` appended to it.

pkg/api/handlers/registry/handler.go
```diff
@@ -0,0 +1,820 @@
...
+	// Step 2: List catalog entries in default catalog with access
+	catalogEntries, err := h.listCatalogEntriesInCatalog(req, system.DefaultCatalog, addedCatalogEntries)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, entry := range catalogEntries {
+		converted, err := ConvertMCPServerCatalogEntryToRegistry(req.Context(), entry, h.serverURL, reverseDNS)
+		if err != nil {
+			continue
```
thedadams, line 112: Can we add comments to all places where we are continuing after error?

pkg/api/handlers/registry/handler.go
```diff
@@ -0,0 +1,820 @@
...
+	return result
+}
+
+func paginateServers(servers []types.RegistryServerResponse, cursor string, limit int) types.RegistryServerList {
+	// Simple cursor-based pagination using server name as cursor
+	startIdx := 0
+	if cursor != "" {
+		// Find the position after the cursor
+		for i, server := range servers {
+			if server.Server.Name == cursor {
```
thedadams, line 607: Using the server name as the cursor assumes that lists always return in the same order, which is not the case. We should be able to find something more stable.

pkg/api/handlers/registry/handler.go
```diff
@@ -0,0 +1,820 @@
...
+// ListServerVersions handles GET /v0.1/servers/{serverName}/versions
+func (h *Handler) ListServerVersions(req api.Context) error {
+	// Extract server name from path (should be in format: reverseDNS/serverName)
+	serverName := req.PathValue("serverName")
+	if serverName == "" {
+		return fmt.Errorf("serverName is required")
+	}
+
+	// Parse reverse DNS and actual server name
+	parts := strings.SplitN(serverName, "/", 2)
```
thedadams, line 651: Making callers url-encode the server names seems like a strange design. I know this is in the spec, but weird choice.

pkg/api/handlers/registry/handler.go
```diff
@@ -0,0 +1,820 @@
...
+		return h.notFoundError("Server not found")
+	}
+
+	return req.Write(server)
+}
+
+// findServerByName searches for a server by name and checks user access
+func (h *Handler) findServerByName(req api.Context, serverName string, reverseDNS string) (types.RegistryServerResponse, error) {
+	// Determine type based on prefix
+	if strings.HasPrefix(serverName, "ms1") {
```
thedadams: There are functions in the `system` package for this.

pkg/api/handlers/registry/handler.go
```diff
@@ -0,0 +1,820 @@
...
+	return h.findMCPServerCatalogEntry(req, serverName, reverseDNS)
+}
+
+// findMCPServer looks up an MCPServer and checks ownership or ACR permissions
+func (h *Handler) findMCPServer(req api.Context, serverName, reverseDNS string) (types.RegistryServerResponse, error) {
+	var server v1.MCPServer
+	err := req.Storage.Get(req.Context(), kclient.ObjectKey{
+		Namespace: system.DefaultNamespace,
+		Name:      serverName,
+	}, &server)
```
thedadams: nit:
```suggestion
	err := req.Get(&server, serverName)
```

pkg/api/handlers/registry/handler.go
```diff
@@ -0,0 +1,820 @@
...
+		credEnv map[string]string
+	)
+
+	if server.Spec.UserID == req.User.GetUID() && server.Spec.MCPCatalogID == "" && server.Spec.PowerUserWorkspaceID == "" {
+		// Personal server - user owns it
+		slug, err = handlers.SlugForMCPServer(req.Context(), req.Storage, server, req.User.GetUID(), "", "")
+		if err != nil {
+			return types.RegistryServerResponse{}, fmt.Errorf("failed to generate slug")
+		}
+		credEnv, _ = h.getCredentialsForServer(req, server, req.User.GetUID(), "", "")
```
thedadams, line 747: Can we add comments explaining why it's OK to ignore these errors?

apiclient/types/registry.go
g-linville: I will rename several of these types for consistency

pkg/accesscontrolrule/helper.go
g-linville: No, because this is only used for no-auth mode, and no-auth mode only returns things from the default catalog, and never things from workspaces.

pkg/api/handlers/registry/utils.go
g-linville: Nice, I didn't know about `slices.Backward`

pkg/api/handlers/registry/utils.go
g-linville: I tested it and this works

pkg/api/handlers/registry/convert.go
g-linville: There wasn't a good reason at least, lol. Fixed

pkg/api/handlers/registry/utils.go
g-linville: Fixed

pkg/api/handlers/registry/utils.go
g-linville: Fixed

pkg/api/handlers/registry/utils.go
g-linville: Fixed

pkg/accesscontrolrule/helper.go
g-linville: Fixed

docs/docs/concepts/mcp-gateway/registry-api.md
g-linville: Fixed

docs/docs/concepts/mcp-gateway/registry-api.md
g-linville: Fixed

docs/docs/concepts/mcp-gateway/registry-api.md
g-linville: Fixed

apiclient/types/registry.go
g-linville: Fixed

pkg/api/handlers/wellknown/oauth.go
g-linville: Fixed

pkg/api/server/server.go
g-linville: Fixed

pkg/jwt/persistent/persistent.go
g-linville: Fixed

pkg/api/handlers/registry/handler.go
g-linville: Is there something else you would prefer that we do?

pkg/api/handlers/registry/handler.go
g-linville: Done

pkg/api/handlers/registry/handler.go
g-linville: Fixed

pkg/api/handlers/registry/handler.go
g-linville: Well having cursors in general assumes that they always return in the same order. So far in my testing, they do always come back in the same order. Should I add an explicit sort based on server name in order to ensure this?

pkg/api/handlers/registry/handler.go
g-linville: Fixed

pkg/api/handlers/registry/handler.go
g-linville: Fixed

pkg/api/handlers/registry/handler.go
g-linville: Fixed

pkg/api/handlers/registry/handler.go
thedadams: It's in the spec that way. Nothing we can do. Just pointing out that it is strange.

pkg/api/handlers/registry/handler.go
thedadams: Cursors are typically much more advanced than this. For example, many will ensure that you only get the data that existed when the cursor was created. They won't always come back in the same order so we should do something more stable.

g-linville: Donnie and I talked about this in Slack, and we are going to keep cursors as they are, but always sort the results by creation time.

pkg/api/handlers/registry/handler.go
```diff
@@ -593,6 +594,11 @@ func filterServersBySearch(servers []types.RegistryServerResponse, search string
 }
 
 func paginateServers(servers []types.RegistryServerResponse, cursor string, limit int) types.RegistryServerList {
+	// Sort servers by creation timestamp
+	slices.SortStableFunc(servers, func(i, j types.RegistryServerResponse) int {
+		return int(i.CreatedAtUnix - j.CreatedAtUnix)
```
thedadams, line 599: Just in case.
```suggestion
		if i.CreatedAtUnix == j.CreatedAtUnix {
			if i.Name < j.Name {
				return -1
			}
			return 1
		}
		return int(i.CreatedAtUnix - j.CreatedAtUnix)
```

Issue #5088, opened 2025-11-20
Chatting with composite MCP server results in tool calls failing with tool not found errors and validation errors when `Configure Tools` option is used for MCP servers.


sangee2004:
**Describe the bug**
Chatting with composite MCP server results in tool calls failing with tool not found errors and validation errors when `Configure Tools` option is used for MCP servers.

**To Reproduce**
1. Create composite MCP server with `Antv Charts` , `gmail` and` Tavily_search`.
2.  When adding  Tavily_search` MCP server, use `Configure Tools` option and enable couple of tools (you can also enable all tools)

<img width="1038" height="812" alt="Image" src="https://github.com/user-attachments/assets/fff44f5a-7816-4bce-b560-9fed4254db2f" />

2. As regular user, Connect to this composite MCP server

3. Try to chat with composite MCP server

Outlook tool calls fail wiith tool not found error relating to `Antv Chart` MCP server
<img width="973" height="627" alt="Image" src="https://github.com/user-attachments/assets/cf4b7fcb-d42b-4265-bdd1-f947f43c92d7" />

Tavily search fails with validation errors:

<img width="999" height="645" alt="Image" src="https://github.com/user-attachments/assets/15ae34c4-0964-49db-b901-dc5b010c1354" />

 `Antv Chart` search tool calls fail with validation errors:

<img width="968" height="710" alt="Image" src="https://github.com/user-attachments/assets/6b6706f5-c731-47e3-a628-9f3176eb2dd7" />
<img width="1008" height="689" alt="Image" src="https://github.com/user-attachments/assets/8494f728-5856-48e5-8764-b8f329fe1c98" />

**Expected behavior**
Tool calls should succeed.

**Note**  - This behavior is seen even when `Configure Tool` option is used all the tools are enabled .

njhale:
@sangee2004 when you list the tools after connecting to the mcp server, what do you see?

sangee2004:
@njhale I see the right set of tools being returned

njhale:
But chat didn't ask you to authenticate? Or did you go through the auth flow before chatting?

sangee2004:
I was presented with normal auth flow when connecting to the composite MCP server as expected.

I see this behavior only when one of the MCP server has only some tools enabled (in my case for `Tavily Search` MCP server i had only couple of tools enabled.

sangee2004:
This issue is addressed when testing with latest build.

PR #5104, opened 2025-11-21
fix: ensure tool overrides are properly applied

thedadams:
This change also includes an image bump for the remote shim that makes tool overrides static. Meaning, only tools specified in the tool overrides are enabled for a sever. Empty overrides means all tools, including any new tools that are added by the server, are enabled.

Issue: https://github.com/obot-platform/obot/issues/5088

### DIFF
--- a/pkg/mcp/docker.go
+++ b/pkg/mcp/docker.go
@@ -208,7 +208,10 @@ func (d *dockerBackend) ensureDeployment(ctx context.Context, server ServerConfi
 	// Check if container already exists
 	existing, err := d.getContainer(ctx, server.MCPServerName)
 	if err == nil && existing != nil {
-		if existing.Labels["mcp.config.hash"] != configHash || existing.NetworkSettings == nil || existing.NetworkSettings.Networks[d.network] == nil {
+		if existing.Labels["mcp.config.hash"] != configHash ||
+			existing.NetworkSettings == nil ||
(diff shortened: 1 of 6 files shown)


pkg/mcp/docker.go
```diff
@@ -208,7 +209,7 @@ func (d *dockerBackend) ensureDeployment(ctx context.Context, server ServerConfi
 	// Check if container already exists
 	existing, err := d.getContainer(ctx, server.MCPServerName)
 	if err == nil && existing != nil {
-		if existing.Labels["mcp.config.hash"] != configHash || existing.NetworkSettings == nil || existing.NetworkSettings.Networks[d.network] == nil {
+		if existing.Labels["mcp.config.hash"] != configHash || existing.NetworkSettings == nil || existing.NetworkSettings.Networks[d.network] == nil || (server.Runtime == types.RuntimeRemote || server.Runtime == types.RuntimeComposite) && existing.Image != d.remoteShimBaseImage {
```
thedadams: Snuck this in here too: redeploy when the remote shim image is bumped.

njhale (approved):

pkg/mcp/docker.go
g-linville: Are the parentheses where you want them to be here? I can't tell.

thedadams: Yes they are. Thanks for double-checking.

PR #5113, opened 2025-11-21
Chore: Add API for accepting EULA

cjellick:
The installer/owner needs to accept Obot's EULA before they can use it.
On the frontend, we'll ensure that they can only proceed with
using/setting up Obot if the api returns true.


pkg/api/authz/authz.go
```diff
@@ -243,6 +243,8 @@ var (
 			"POST /api/logout-all",
 			"GET /api/version",
 			"GET /api/setup/oauth-complete",
+			"GET /api/eula",
+			"PUT /api/eula",
```
thedadams: Should this go in the admin section?

cjellick: yep!

cjellick: I knew that...claude didnt and it missed my review

thedadams (approved):

Issue #5125, opened 2025-11-22
`invalid session` presented when accessing admin user session after obot was restarted


sangee2004:
**Describe the bug**
`invalid session` presented when accessing admin user session after obot was restarted

```
obot: [v0.0.0-dev+90af18a1](https://github.com/obot-platform/obot/commit/90af18a1)
obot-platform/tools: 8a469b0
obot-platform/enterprise-tools: df53d5f
```

Pre Req - I had a obot setup ( Microsoft Entra auth enable) with owner and basic users setup.

**To Reproduce**

1. Tried to upgrade obot to latest version (There was no new version available ) by following these steps:

docker rm -f obot
docker run  -d --pull=always --name obot -p 8080:8080 -v /var/run/docker.sock:/var/run/docker.sock -v mydata:/data -e OPENAI_API_KEY=sk-proj-xxx -e OBOT_SERVER_ENABLE_AUTHENTICATION=true  ghcr.io/obot-platform/obot-enterprise:main

2. After obot started , i am not able to access obot as admin user. Accessing obot server results in `invalid session`  page presented to user

<img width="1313" height="635" alt="Image" src="https://github.com/user-attachments/assets/c124ae6f-7bee-4944-8460-43458c9063be" />

obot_access_token expires only few days later
<img width="1351" height="681" alt="Image" src="https://github.com/user-attachments/assets/2c7f2667-877c-4ddb-9e39-0a70c82bcbb5" />

Once i removed the obot_access_token , i am able to access obot server again

This issue was not seen with the basic user sessions which continued to work after upgrade.

**Note** This issue does not reproduce again. I tried to do the same steps of upgrading obot server (like in step 1) , admin access continues to be good after upgrade .

**Expected behavior**

User should not be presented with `invalid session`

g-linville:
Merged the fix

sangee2004:
This issue is addressed when testing with latest build.

PR #5132, opened 2025-11-24
fix: delete the cookie when the session is invalid

g-linville:
for https://github.com/obot-platform/obot/issues/5125

### DIFF
--- a/pkg/api/server/server.go
+++ b/pkg/api/server/server.go
@@ -92,17 +92,19 @@ func (s *Server) Wrap(f api.HandlerFunc) http.HandlerFunc {
 					Path:   "/",
 					MaxAge: -1,
 				})
+				// Refresh the page so that the cookie deletes.
+				http.Redirect(rw, req, req.URL.String(), http.StatusFound)
 				return
 			}
+
+			http.Error(rw, err.Error(), http.StatusUnauthorized)
 
 			// Check if this is a FetchUserGroupsError which indicates an auth provider configuration issue
 			var fetchGroupsErr *gclient.FetchUserGroupsError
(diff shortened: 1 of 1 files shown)


pkg/api/server/server.go
```diff
@@ -101,7 +92,19 @@ func (s *Server) Wrap(f api.HandlerFunc) http.HandlerFunc {
 					Path:   "/",
 					MaxAge: -1,
 				})
+				// Refresh the page so that the cookie deletes.
+				http.Redirect(rw, req, req.URL.String(), http.StatusFound)
```
g-linville, line 96: This feels like an abuse of HTTP but I'm not sure of any other way to delete the cookie and refresh, besides doing this.

pkg/api/server/server.go
```diff
@@ -101,7 +92,19 @@ func (s *Server) Wrap(f api.HandlerFunc) http.HandlerFunc {
...
+				return
 			}
+
+			http.Error(rw, err.Error(), http.StatusUnauthorized)
+
+			// Check if this is a FetchUserGroupsError which indicates an auth provider configuration issue
+			var fetchGroupsErr *gclient.FetchUserGroupsError
+			if errors.As(err, &fetchGroupsErr) {
+				http.Error(rw, fmt.Sprintf("Authentication provider configuration error: %s. Please contact an administrator to fix the auth provider configuration.", err.Error()), http.StatusInternalServerError)
```
njhale, line 105: Is there an issue with writing the error on line 100, and then writing it again here?

njhale: Oh, I guess this was already the case in the original code.

PR #5141, opened 2025-11-25
fix: cleanup nobody user when enabling auth

thedadams:
If auth was disabled and the nobody user deployed MCP servers, for example, we need to clean those up when auth is enabled. The nobody user was already being deleted, but their stuff was left behind.

### DIFF
--- a/pkg/gateway/client/identity.go
+++ b/pkg/gateway/client/identity.go
@@ -380,8 +380,9 @@ func (c *Client) RemoveIdentity(ctx context.Context, id *types.Identity) error {
 // RemoveIdentityAndUser deletes an identity and the associated user from the database.
 // The identity and user are deleted using UserID if set, otherwise ProviderUsername.
 // The method is idempotent and ignores not-found errors, returning only unexpected errors.
-func (c *Client) RemoveIdentityAndUser(ctx context.Context, id *types.Identity) error {
-	return c.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+func (c *Client) RemoveIdentityAndUser(ctx context.Context, id *types.Identity) (uint, error) {
+	var user types.User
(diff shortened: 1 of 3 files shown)


pkg/gateway/client/identity.go
```diff
@@ -400,13 +401,17 @@ func (c *Client) RemoveIdentityAndUser(ctx context.Context, id *types.Identity)
 			return err
 		}
 
-		// Attempt to delete the user
+		if err := userQuery.First(&user).Error; err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
+			return err
+		}
```
StrongMonkey, line 409: Nit: return earlier if err is ErrRecordNotFound?

thedadams: Done.

pkg/services/config.go
```diff
@@ -671,12 +671,25 @@ func New(ctx context.Context, config Config) (*Services, error) {
...
 		// is enabled.
-		if err := gatewayClient.RemoveIdentityAndUser(ctx, &types.Identity{
+		if id, err := gatewayClient.RemoveIdentityAndUser(ctx, &types.Identity{
 			ProviderUsername:     "nobody",
 			ProviderUserID:       "nobody",
 			HashedProviderUserID: hash.String("nobody"),
 		}); err != nil {
 			return nil, fmt.Errorf(`failed to remove "nobody" user and identity from database: %w`, err)
+		} else if id != 0 {
+			// Create this UserDelete object so that they're stuff gets deleted.
```
g-linville: mini nit


```suggestion
			// Create this UserDelete object so that their stuff gets deleted.
```

g-linville (approved):

pkg/services/config.go
thedadams: Done.

PR #5142, opened 2025-11-25
fix: prevent table columns overlap when data is empty

ryu-man:
Addresses #4977


ryu-man:
Re-apply search input mono-tab filtering in the MCP servers page

ivyjeong13 (approved):

PR #5156, opened 2025-11-26
fix: allow anyGroup to access /oauth/callback/{oauthrequestid}/{mcpid}

g-linville:


g-linville:
Nvm, this is the wrong approach.

Issue #5135, opened 2025-11-24
Running in Kubernetes mode with localhost:8080 doesn't work


thedadams:
This is because the MCP pods need to communicate with Obot, but they are given `localhost:8080`, which doesn't resolve to the Obot pod. The fix is to provide the service FQDN.

sangee2004:
Testing with latest version - `v0.0.0-dev+a44e2049`

I am setting  `--set config.OBOT_SERVER_HOSTNAME="http://localhost:8081"` when deploying obot using helm chart.

I have the following port forwarding done to get access to my local HA and am accessing obot as http://localhost:8081/.
 
` kubectl port-forward svc/obot-obot 8081:80`

I am seeing this error now for `/tools`  call made for gmail MCP server
```
failed to list tools: failed to list MCP tools: error from server: JSON RPC internal error: failed to send request: failed to initialize client: failed to exchange token: failed to call token exchange endpoint: Post "http://localhost:8081/oauth/token": dial tcp [::1]:8081: connect: connection refused
```
Following error when launching single user MCP servers.

<img width="1278" height="685" alt="Image" src="https://github.com/user-attachments/assets/3362f953-2371-4d73-baed-f072a88932bf" />

@thedadams Reopening this issue for followup

sangee2004:
The fixes involved changes to chart as well which was not in the latest version when I had validated the issue previously.

When latest chart fixes were used , I see this issue addressed.

Able to connect and chat with both remote and single user MCP servers.

PR #5157, opened 2025-11-27
fix: make Kubernetes backend work on localhost

thedadams:
The MCP containers now need to communicate with Obot for token exchange. This doesn't work when running on localhost. This change will pass the service FQDN as the token exchange endpoint.

Issue: https://github.com/obot-platform/obot/issues/5135

### DIFF
--- a/chart/templates/secret.yaml
+++ b/chart/templates/secret.yaml
@@ -18,6 +18,8 @@ data:
   {{- end }}
   {{- if eq .Values.config.OBOT_SERVER_MCPRUNTIME_BACKEND "kubernetes" }}
   OBOT_SERVER_MCPNAMESPACE: {{ include "obot.config.mcpNamespace" . | b64enc }}
+  OBOT_SERVER_SERVICE_NAME: {{ include "obot.fullname" . | b64enc }}
+  OBOT_SERVER_SERVICE_NAMESPACE: {{ .Release.Namespace | b64enc }}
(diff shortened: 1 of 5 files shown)


chart/values.yaml
```diff
@@ -142,6 +142,8 @@ config:
   OBOT_SERVER_DISALLOW_LOCALHOST_MCP: ""
   # config.OBOT_SERVER_MCPRUNTIME_BACKEND -- The runtime backend to use for MCP servers. Can be 'docker' or 'kubernetes'. Defaults to 'docker'. Setting this to 'kubernetes' will also create the necessary service account, role and rolebinding.
   OBOT_SERVER_MCPRUNTIME_BACKEND: "kubernetes"
+  # config.OBOT_SERVER_SERVICE_NAME -- The Kubernetes service name for the obot server. Automatically set when using kubernetes backend. Used to construct the internal service FQDN for token exchange endpoints.
+  # config.OBOT_SERVER_SERVICE_NAMESPACE -- The Kubernetes namespace where the obot server runs. Automatically set when using kubernetes backend. Used to construct the internal service FQDN for token exchange endpoints.
```
drpebcak: Not sure setting this config would do anything because you've set it to `.Release.Namespace` in the secret template. Do we want to support overriding that?

thedadams: I think it makes sense to remove this. I'll do that.

thedadams: Removed.

Issue #5096, opened 2025-11-21
Deleting MCP server entry that is part of composite MCP sever results in not being able to chat with composite MCP sever because of `tools` call failing.


sangee2004:
**Describe the bug**
Deleting MCP server entry that is part of composite MCP sever results in not being able to chat with composite MCP sever because of `tools` call failing.

**To Reproduce**
1. Create composite MCP server with few mcp servers say `mcp1` and `mcp2`
2. As regular user , connect to this composite MCP server
3. As admin , delete `mcp1` mcp server entry.
4.  As regular user , Chat in the project created in step 2.

Notice that user is presented with following error message:

<img width="1072" height="707" alt="Image" src="https://github.com/user-attachments/assets/d0261122-74de-4edc-8593-a6baee26571c" />

Accessing `tools` tab shows no tools with `tools` call failing with following errors
```
failed to list tools: failed to list MCP tools: error from server: JSON RPC internal error: error from server: JSON RPC internal error: failed to send request: failed to initialize client: failed to initialize HTTP Streaming client: 500 Internal Server Error: failed to ensure server is deployed: failed to get mcp server config: failed to get MCP server catalog entry: MCPServerCatalogEntry.obot.obot.ai "default-santestnew1-cde6739atkjwq" not found

failed to connect to SSE server url http://172.17.0.2:8080/mcp-connect/ms1djsrv: 500 Internal Server Error, failed to ensure server is deployed: failed to get mcp server config: failed to get MCP server catalog entry: MCPServerCatalogEntry.obot.obot.ai "default-santestnew1-cde6739atkjwq" not found

```
**Expected behavior**
Connected composite MCP servers should continue to get access to `mcp1` server until composite MCP server  is upgraded and then connected composite server instance is updated.

sangee2004:
This issue is addressed when testing with latest build.

When testing the scenario outlined in the issue, connected composite MCP servers continues to get access to `mcp1` server even after `mcp1` server entry is deleted as expected.

When composite MCP server is upgraded and then connected composite server instance is updated , User dose not get access to the deleted `mcp1` server  and is able to continue to use the other mpc servers in the composite server as expected.

PR #5159, opened 2025-11-27
fix: stop failing if component server catalog entry is deleted

thedadams:
For composite servers, a user is allowed to delete the catalog entry for a component server and have the composite server still function properly. If the catalog entry no longer exists, then it is not possible to retrieve the catalog ID from the entry. This change adds a field to the status to track the catalog ID of the component server, which was previously pulled from the catalog entry.

Issue: https://github.com/obot-platform/obot/issues/5096

### DIFF
--- a/pkg/controller/handlers/mcpserver/mcpserver.go
+++ b/pkg/controller/handlers/mcpserver/mcpserver.go
@@ -15,6 +15,7 @@ import (
 	"github.com/obot-platform/obot/pkg/system"
 	"github.com/obot-platform/obot/pkg/utils"
 	"golang.org/x/crypto/bcrypt"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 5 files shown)


pkg/controller/handlers/mcpserver/mcpserver.go
```diff
@@ -287,6 +292,22 @@ func (h *Handler) DeleteServersWithoutRuntime(req router.Request, _ router.Respo
 	return nil
 }
 
+func (h *Handler) EnsureMCPCatalogID(req router.Request, _ router.Response) error {
+	server := req.Object.(*v1.MCPServer)
+
+	if server.Status.MCPCatalogID == "" && server.Spec.MCPCatalogID == "" && server.Spec.MCPServerCatalogEntryName != "" {
+		var mcpCatalogEntry v1.MCPServerCatalogEntry
+		if err := req.Get(&mcpCatalogEntry, server.Namespace, server.Spec.MCPServerCatalogEntryName); err != nil {
```
g-linville, line 309: Do we want to ignore not found errors?

thedadams: In a typical situation, we wouldn't get not-found errors here because the MCP server would have just been created and we fetched the MCP catalog entry in the API.

g-linville: Don't controller handlers re-trigger every 10h or so though? And I thought we were specifically adding this since the catalog entry could get deleted, which I think would cause the handler to return an error the next time it gets triggered.

thedadams: The first part of the if-check handles that. If the server already has the status field set, then we don't check for the catalog entry.

g-linville: Oh, I missed that

PR #5165, opened 2025-12-01
fix: delete unauthorized servers for user when they lose group membership

g-linville:


pkg/controller/handlers/mcpcatalog/usergroupchange.go
```diff
@@ -0,0 +1,177 @@
...
+		return fmt.Errorf("failed to list MCP servers for user %s: %w", userID, err)
+	}
+
+	for _, server := range mcpServers.Items {
+		if !server.DeletionTimestamp.IsZero() {
+			continue
+		}
+
+		// Skip special server types that don't need access checks
+		if server.Spec.ThreadName != "" || server.Spec.UserID == "anonymous" || server.Spec.CompositeName != "" {
```
thedadams: There is a separate controller that deletes servers that belong to the anonymous user. No need to worry about them here.
```suggestion
		if server.Spec.ThreadName != "" || server.Spec.CompositeName != "" {
```

PR #5180, opened 2025-12-02
enhance: ui/ux overhaul "my connectors" "admin dashboard" & "mcp-publisher" into 1

ivyjeong13:
* overhaul and merging of "My Connectors" "MCP Publisher" with Admin "dashboard" experience


ivyjeong13:
Addressed #5278 in here as well since super quick fix!

cjellick (approved):
We have some small UX fixes to iterate on but this is can be merged and we can fix those separately

sangee2004:
Changes tracked in this PR is now available when testing with latest build from `main`
Issues found when testing these changes are logged as separate issues.

PR #5199, opened 2025-12-04
fix: auth providers: return correct error when sessions are missing from db

g-linville:
This is something I ran across when I was working on an auth provider issue. If the session was not found in the database (i.e. the user is using a cookie that doesn't map to an existing session, but is formatted correctly), they would receive a cryptic JSON parsing error attached to the 401. This fixes that by checking the body first before we attempt to decode it as JSON.

### DIFF
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -349,7 +349,7 @@ func (p *Proxy) authenticateRequest(req *http.Request) (*authenticator.Response,
 	}
 
 	var ss serializableState
-	if err = json.NewDecoder(stateResponse.Body).Decode(&ss); err != nil {
+	if err = json.Unmarshal(body, &ss); err != nil {
 		return nil, false, err
 	}
 


pkg/proxy/proxy.go
```diff
@@ -338,6 +339,14 @@ func (p *Proxy) authenticateRequest(req *http.Request) (*authenticator.Response,
...
+	if err != nil {
+		return nil, false, err
+	}
+
+	if stateResponse.StatusCode == http.StatusInternalServerError && strings.Contains(string(body), "record not found") {
+		return nil, false, ErrInvalidSession
+	}
 
 	var ss serializableState
 	if err = json.NewDecoder(stateResponse.Body).Decode(&ss); err != nil {
```
njhale: if we do `io.ReadAll(stateResponse.Body)` above, doesn't that consume the body reader before we can decode it here?

g-linville: Not sure, but I pushed a change to just use the existing `body` far instead so that we don't have to worry about it.

PR #5210, opened 2025-12-05
fix: don't store picture in token if it is a base64 encoded image

thedadams:


pkg/jwt/persistent/persistent.go
```diff
@@ -294,6 +294,11 @@ func (t *TokenService) DecodeToken(ctx context.Context, token string) (*TokenCon
 }
 
 func (t *TokenService) NewToken(ctx context.Context, context TokenContext) (string, error) {
+	var picture string
+	if !strings.HasPrefix(picture, "data:") {
+		// Don't store the picture in the token if it is a base64 encoded image
+		picture = context.Picture
+	}
```
g-linville, line 301: This doesn't seem right? We declare picture, then check it (when it will always be empty), and never use it.

g-linville (changes requested):

Issue #5212, opened 2025-12-05
Not able to connect and chat with  with single user MCP servers that have required configuration parameters.


sangee2004:
**Describe the bug**
Not able to connect and chat with  single user MCP servers that have required configuration parameters.

Pre req - obot needs to be installed using helm charts

**To Reproduce**
1. Connect to `Brave Search` mcp server and provide valid api key

Notice that user is presented with following error and mcp server fails to launch:

<img width="1335" height="724" alt="Image" src="https://github.com/user-attachments/assets/60b61e68-5b3b-4ced-8305-e67902cffd1f" />

2. Connect to `Exa Search` mcp server and provide valid api key
3. Once connected , chat with this  mcpserver

Notice that tool calls fail with `400 errors`

<img width="1039" height="605" alt="Image" src="https://github.com/user-attachments/assets/8af78602-a9fb-441d-b7ef-4d4e33f7795d" />

**Expected behavior**
Should be able to connect and chat with with single user MCP servers that have required configuration parameters.

**Note** I am able to connect and chat successfully with both `Brave Search` and  `Exa Search` mcp servers from local docker installs.

sangee2004:
`Firecrawl` MCP server also reports this problem when launching it

<img width="1416" height="704" alt="Image" src="https://github.com/user-attachments/assets/7b10072f-07fc-437d-800e-9c9d52e4aea2" />

shwetapsdet:
Getting a 503 error while connecting to the GitLab MCP server.
Log:
```
2025-12-05T22:27:53.621777372Z [94m->(server) │[0m {"jsonrpc":"2.0","id":1764973615,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{"roots":{"listChanged":false},"sampling":{},"elicitation":{}},"clientInfo":{"name":"nanobot","version":"v0.0.0-dev+9ed86f52"}}}
2025-12-05T22:27:53.981192431Z [35m<-(server:stderr)│[0m [31mERROR[39m [22:27:53.918] (68): [36mGITLAB_PERSONAL_ACCESS_TOKEN environment variable is not set[39m
2025-12-05T22:27:53.981232977Z [35m<-(server:stderr)│[0m [32mINFO[39m [22:27:53.918] (68): [36mEither set GITLAB_PERSONAL_ACCESS_TOKEN or enable OAuth with GITLAB_USE_OAUTH=true[39m
[...]

PR #5213, opened 2025-12-05
fix: dont strip nanobot meta env from real deployment

njhale:
This was breaking nanobot wrapper images that expect `NANOBOT_META_ENV` to carry required env vars for the MCP server they wrap.

See https://github.com/obot-platform/mcp-images/blob/628eab173e6932a5bfe94d9fd677dd20c9c7a6ea/scripts/nanobot.sh#L36

Addresses http://github.com/obot-platform/obot/issues/5212

### DIFF
--- a/pkg/mcp/kubernetes.go
+++ b/pkg/mcp/kubernetes.go
@@ -510,7 +510,7 @@ func (k *kubernetesBackend) k8sObjects(ctx context.Context, server ServerConfig,
 						if server.Runtime != types.RuntimeComposite {
 							delete(secretEnvStringData, k)
 						}
-					} else if strings.HasPrefix(k, "NANOBOT_") {
+					} else if strings.HasPrefix(k, "NANOBOT_RUN_") {
 						vars[k] = v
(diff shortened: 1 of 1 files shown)


thedadams:
It would help me to understand the issue if we could make the smallest change possible fixes it.

pkg/mcp/kubernetes.go
```diff
@@ -503,18 +503,26 @@ func (k *kubernetesBackend) k8sObjects(ctx context.Context, server ServerConfig,
...
+						continue
+					}
+
 					if k == "NANOBOT_DISABLE_HEALTH_CHECKER" {
+						// Force disable health checks for shim
 						vars[k] = "true"
-						if server.Runtime != types.RuntimeComposite {
-							delete(secretEnvStringData, k)
-						}
-					} else if strings.HasPrefix(k, "NANOBOT_") {
```
thedadams, line 513: If I do understand the problem correctly, I believe changing `NANOBOT_` here to `NANOBOT_RUN_` will fix the problem.

njhale: You're right! Change pushed.

PR #5221, opened 2025-12-08
enhance: add internal flag to detect local IP

thedadams:
This fixes running Obot outside of a container with the docker backend.


thedadams:
Merging with Nick's review and Abdel's testing.

PR #5225, opened 2025-12-08
feat: add system MCP server API

g-linville:
This is the first step in supporting hooks.

SystemMCPServers are a new type that is deployed automatically by the controller. The purpose of these servers is for internal use within Obot itself, programmatically. The first (and so far, only) use case for this is hooks, which is a feature in Nanobot that we have not yet integrated into Obot. This will allow for things like PII detection, prompt injection detection, etc. The goal will be for the Nanobot shim containers to be able to reach out to these system MCP servers and call particular tools with the MCP payloads for analysis and acceptance/rejection.

Rather than continue to overload our existing MCPServer type, I decided to define a completely new type for these. We convert them directly to ServerConfig for the backend to be able to deploy and manage. This avoids needing to filter these out of all the API handlers for MCPServers, and avoids adding yet another Spec field to check for constantly.

### DIFF
--- a/pkg/api/handlers/systemmcpserver.go
+++ b/pkg/api/handlers/systemmcpserver.go
@@ -297,7 +297,9 @@ func (h *SystemMCPServerHandler) Logs(req api.Context) error {
 
 	// Stream logs using the helper (handles SSE formatting, Docker header stripping, etc.)
 	return StreamLogs(req.Context(), req.ResponseWriter, logs, StreamLogsOptions{
-		SendEnded: true,
+		SendKeepAlive:  true,
+		SendDisconnect: true,
+		SendEnded:      true,
 	})
 }
 


pkg/mcp/loader.go
```diff
@@ -231,6 +231,12 @@ func (sm *SessionManager) ShutdownServer(ctx context.Context, serverName string)
 	return sm.backend.shutdownServer(ctx, serverName)
 }
 
+// EnsureServerDeployment ensures that a server is deployed with the given configuration and webhooks.
+// This is primarily used by controllers to deploy system-level servers.
+func (sm *SessionManager) EnsureServerDeployment(ctx context.Context, serverConfig ServerConfig, webhooks []Webhook) (ServerConfig, error) {
```
thedadams: Can we just use `LaunchServer` here? I think having fewer methods will avoid confusion as we build this out.

pkg/api/router/router.go
```diff
@@ -570,6 +571,20 @@ func Router(ctx context.Context, services *services.Services) (http.Handler, err
 	mux.HandleFunc("DELETE /api/mcp-webhook-validations/{mcp_webhook_validation_id}", mcpWebhookValidations.Delete)
 	mux.HandleFunc("DELETE /api/mcp-webhook-validations/{mcp_webhook_validation_id}/secret", mcpWebhookValidations.RemoveSecret)
 
+	// System MCP Servers (admin only)
+	mux.HandleFunc("GET /api/system-servers", systemMCPServers.List)
```
thedadams: micro-nit:
```suggestion
	mux.HandleFunc("GET /api/system-mcp-servers", systemMCPServers.List)
```

pkg/controller/handlers/systemmcpserver/systemmcpserver.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// CleanupDeployment handles cleanup when SystemMCPServer is deleted
+func (h *Handler) CleanupDeployment(req router.Request, _ router.Response) error {
+	systemServer := req.Object.(*v1.SystemMCPServer)
+
+	// Only cleanup if being deleted
+	if systemServer.DeletionTimestamp.IsZero() {
+		return nil
+	}
```
thedadams: This is handled by `nah`.
```suggestion
```

pkg/api/handlers/systemmcpserver.go
```diff
@@ -0,0 +1,660 @@
...
+	return req.Write(types.SystemMCPServerList{Items: servers})
+}
+
+// Get returns a specific system MCP server
+func (h *SystemMCPServerHandler) Get(req api.Context) error {
+	var systemServer v1.SystemMCPServer
+	if err := req.Get(&systemServer, req.PathValue("id")); err != nil {
+		if apierrors.IsNotFound(err) {
+			return types.NewErrNotFound("system MCP server not found")
+		}
```
thedadams: Returning the error if it is a "not found" error will do the right thing. We don't need to convert it here. Same for all instances below.

pkg/api/handlers/systemmcpserver.go
```diff
@@ -0,0 +1,660 @@
...
+
+	// Validate manifest
+	if err := validation.ValidateSystemMCPServerManifest(manifest); err != nil {
+		return types.NewErrBadRequest("validation failed: %v", err)
+	}
+
+	systemServer := v1.SystemMCPServer{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: system.SystemMCPServerPrefix,
+			Namespace:    req.Namespace(),
```
thedadams, line 79: Add the finalizer here, too.

pkg/api/handlers/systemmcpserver.go
```diff
@@ -0,0 +1,660 @@
...
+		Env:      envVars,
+	}); err != nil {
+		return fmt.Errorf("failed to create credential: %w", err)
+	}
+
+	// Update annotation to track configuration timestamp
+	if systemServer.Annotations == nil {
+		systemServer.Annotations = make(map[string]string)
+	}
+	systemServer.Annotations["obot.obot.ai/configured-at"] = metav1.Now().Format("2006-01-02T15:04:05Z")
```
thedadams: We should use one of the constant formats, if possible.

pkg/api/handlers/systemmcpserver.go
```diff
@@ -0,0 +1,660 @@
...
+		ToolName: systemServer.Name,
+		Type:     gptscript.CredentialTypeTool,
+		Env:      envVars,
+	}); err != nil {
+		return fmt.Errorf("failed to create credential: %w", err)
+	}
+
+	// Update annotation to track configuration timestamp
+	if systemServer.Annotations == nil {
+		systemServer.Annotations = make(map[string]string)
```
thedadams: nit:
```suggestion
		systemServer.Annotations = make(map[string]string, 1)
```

pkg/api/handlers/systemmcpserver.go
```diff
@@ -0,0 +1,660 @@
...
+			Reason:  cond.Reason,
+			Message: cond.Message,
+		})
+	}
+
+	// Find missing required env vars and headers
+	credCtx := server.Name
+	creds, _ := gptClient.ListCredentials(ctx, gptscript.ListCredentialsOptions{
+		CredentialContexts: []string{credCtx},
+	})
```
thedadams: I've received push-back in the past putting such things in these "convert" functions. The feedback I received is that these should just be converting some data into other data.

pkg/api/handlers/systemmcpserver.go
thedadams, line 1: There is a lot of overlap between this and the "regular" MCP server stuff. I would like to see us DRY this up, if we can.

pkg/api/handlers/systemmcpserver.go
g-linville: Good catch!

pkg/mcp/loader.go

pkg/api/router/router.go
g-linville: Fixed

pkg/controller/handlers/systemmcpserver/systemmcpserver.go
g-linville: Fixed

pkg/api/handlers/systemmcpserver.go
g-linville: Fixed

pkg/api/handlers/systemmcpserver.go
g-linville: Fixed

pkg/api/handlers/systemmcpserver.go
g-linville: Fixed

pkg/api/handlers/systemmcpserver.go
g-linville: Fixed

pkg/api/handlers/systemmcpserver.go
g-linville: Fixed. This dried it up a bit too.

pkg/api/handlers/systemmcpserver.go
g-linville: Just pushed a change to DRY it a bit. Mostly around the logs stuff. I don't think there's much else I could do reasonably to reduce code reuse. But I think the tradeoff is worth it to avoid continuing to overload the MCPServer type.

thedadams (approved):

Issue #5226, opened 2025-12-08
409 errors seen when launching composite MCP server.


sangee2004:
**Describe the bug**
409 errors seen when launching composite MCP server.

**To Reproduce**
1. Create a composite MCP server with Gmail MCP server.
2. Select few tools (all list tools) when configuring tools for this MCP server.
3. As regular user, connect to this MCP server

Notice that 409 errors are presented to the user.

<img width="1296" height="777" alt="Image" src="https://github.com/user-attachments/assets/b30593f1-10fa-45c4-a802-414e1cef2f1e" />

Full error:
```
409 /mcp-servers/ms122bz9/configure: Operation cannot be fulfilled on MCPServer.obot.obot.ai "ms122bz9": the object has been modified; please apply your changes to the latest version and try again
```

Retrying to connect with this MCP server results in the same error again.

Even though, these errors are presented to the user , I see the MCP server is launched and is available in `Enabled Connectors` . Able to connect and chat with it.
**Expected behavior**

**Note** I was able to connect to this composite MCP sever as another user. 

**Expected Behavior:**
409 errors should not be seen when launching composite MCP server.

sangee2004:
I have also  encountered  409 errors seen sometimes when updating composite MCP server instances which is tracked in  https://github.com/obot-platform/obot/issues/5127

sangee2004:
This issue is not seen anymore when testing with latest build from `main`

PR #5228, opened 2025-12-08
fix: retry composite server update after configuring components

njhale:
Retry composite server update if it fails with conflicts after configuring components.

This fixes a race condition between the configure API and the following controller that results in 409s when initially configuring composite servers:
https://github.com/obot-platform/obot/blob/e402df32ea62285251e0e2eb82fb9f084d2036dc/pkg/controller/handlers/mcpserver/mcpserver.go#L332

Addresses https://github.com/obot-platform/obot/issues/5226

### DIFF
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/gptscript-ai/go-gptscript"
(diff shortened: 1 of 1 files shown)


njhale:
~Took this back to draft. Need to go with the retry after all.~

Done

thedadams (approved):

PR #5229, opened 2025-12-08
chore: update docs dependencies

drpebcak:


thedadams (approved):
I'm astonished that those changes to the `package.json` caused 4,000+ lines to change in the lock file.

PR #5239, opened 2025-12-09
fix: improve popovers display

ryu-man:
The issue:
Currently popovers are displayed using absolute positioning only, which has limitations like conflicting with other higher-level elements having a higher z-index (like in the screenshot), as well as the current popover having a fixed position and not adjusting its position based on the available space. This makes popovers rendered near edges get clipped on smaller screens. 

image

The Calendar popover got covered by the sidebar.

image

The calendar got clipped by the edge in small screen

This PR solves the issue described above by rendering popovers in a top layer defined in the app tree to ensure they are always displayed above all other app elements. Use a modular approach to create customizable and easy-to-maintain popovers with accessibility in mind.

image

image

### DIFF
--- a/ui/user/src/lib/components/Calendar.svelte
+++ b/ui/user/src/lib/components/Calendar.svelte
@@ -178,7 +178,6 @@
 	function handleApply() {
 		onChange({ start, end });
 
-		// calendarPopover?.close();
 		open = false;
 	}
 
@@ -188,7 +187,6 @@
 		start = initialValue.start;
 		end = initialValue.end;
 
-		// calendarPopover?.close();
 		open = false;
 	}
 
@@ -229,9 +227,7 @@
 
 	// Sync open state with popover
 	$effect(() => {
-		if (calendarPopover.open !== open) {
-			calendarPopover.toggle(open);
-		}
+		calendarPopover.toggle(open);
 	});
 </script>
 


ui/user/src/lib/components/Calendar.svelte
```diff
@@ -47,11 +48,11 @@
 		end = $bindable(initialValue.end),
 		placeholder = 'Select date range',
 		format = 'MMM dd, yyyy',
-		compact
+		compact,
+		open = $bindable(false)
 	}: Props = $props();
 
 	let currentDate = $state(new Date());
-	let calendarPopover = $state<HTMLDialogElement>();
```
ivyjeong13, line 54: Curious if the popover action didn't work here for you? I believe that action portals the popover to the document level unless `disablePortal` is set.

```
import popover from '$lib/actions/popover.svelte';
let {
		ref,
		tooltip: popoverRef,
		toggle
	} = popover({
		placement: 'bottom-end',
		offset: -52
	});

// ...

     function handleApply() {
		onChange({ start, end });

		toggle(false); 
	}

	function handleCancel() {
		// Reset local value to initial value
		// localValue = { ...initialValue };
		start = initialValue.start;
		end = initialValue.end;

		toggle(false);
	}

// ...

<button
         // ...
		onmousedown={() => {
			if (disabled) return;
			toggle();
		}}
		use:ref
	>
		// ...
	</button>

    // NOTE: did have to change the <dialog> into a <div> here
	<div use:popoverRef class={twMerge('default-dialog min-w-[320px] p-4', classes?.calendar)}> 
        <!-- Calendar Header -->
		<div class={twMerge('mb-4 flex items-center justify-between', classes?.header)}>
// rest of dialog content
```

ivyjeong13: I do think it still has an issue with smart positioning since I had to add an offset but mostly hoping to avoid having add the popover component/files if the action can address the issues being faced!

ryu-man: I have not known such functionality existed for displaying popovers; I am going to try it

ryu-man: I see it work, so I have reverted the code to use the action instead; however, I am not convinced the action approach is the best way to do it. To me, it sounds more verbose. Just sharing my perspective, not criticizing the solution.

ivyjeong13: I'm totally okay with exploring an alternative route -- was more wondering if you experienced issues with it that led you to not using it! 

I think my concern with the prior solution is it seems like maybe too many components/modularity to address the root issue? Maybe I'm not following why all of them are necessary.

I'll go ahead and approve but happy to sync if you wanna discuss any of the above!

ui/user/src/lib/components/Calendar.svelte
```diff
@@ -177,7 +178,8 @@
 	function handleApply() {
 		onChange({ start, end });
 
-		calendarPopover?.close();
+		// calendarPopover?.close();
```
ivyjeong13: nit: super minor, can remove commented out

ui/user/src/lib/components/Calendar.svelte
```diff
@@ -186,7 +188,8 @@
 		start = initialValue.start;
 		end = initialValue.end;
 
-		calendarPopover?.close();
+		// calendarPopover?.close();
```
ivyjeong13: nit: same here, can remove commented out line

ivyjeong13 (approved):

ui/user/src/lib/components/Calendar.svelte
ryu-man: Sure, I would love to discuss this when you have a chance

PR #5242, opened 2025-12-09
chore: update package.json based on dependabot security alerts

ivyjeong13:
* upgrading packages to latest for related dependabot security issues

https://github.com/obot-platform/obot/security/dependabot

Notes:
* seems like the prefer-reactivity rule pops improperly for `Map` `Set` used in functions that don't need to be reactive -- had to add disable lines where seemed appropriate
* `unexpected without resolve()` lint for `href` and `goto` was most common lint error needed to be fixed with the package upgrades
* had to add override for `glob` and `cookie` -- their related packages are still on the alerted version

### DIFF
--- a/ui/user/package.json
+++ b/ui/user/package.json
@@ -68,7 +68,7 @@
 		"prettier-plugin-svelte": "^3.3.3",
 		"prettier-plugin-tailwindcss": "^0.6.11",
 		"svelte": "^5.45.6",
-		"svelte-check": "^4.2.2",
+		"svelte-check": "^4.3.4",
 		"tailwind-merge": "^2.6.0",
 		"tailwind-scrollbar": "^4.0.1",
 		"tailwindcss": "^4.0.14",
--- a/ui/user/pnpm-lock.yaml
+++ b/ui/user/pnpm-lock.yaml
@@ -29,7 +29,7 @@ importers:
         version: 1.39.7
       layerchart:
         specifier: 1.0.12
-        version: 1.0.12(svelte@5.35.5)(typescript@5.9.3)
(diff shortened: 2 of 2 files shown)


ui/user/src/lib/url.ts
```diff
@@ -80,10 +81,22 @@ export function setSearchParamsToLocalStorage(pathname: string, searchParams: st
...
+
 export function getSearchParamsFromLocalStorage(pathname: string): string | null {
 	if (browser) {
 		return localStorage.getItem(`page.searchParams.${pathname}`);
 	}
 
 	return null;
 }
+
+export function goto(url: string | URL, state?: object) {
```
ivyjeong13, line 98: to take care of the `unexpected goto without resolve()` lint errors that popped up everywhere

ui/user/src/lib/url.ts
```diff
@@ -80,10 +81,22 @@ export function setSearchParamsToLocalStorage(pathname: string, searchParams: st
 	}
 }
 
+export function replaceState(url: string | URL, state: object) {
```
ivyjeong13, line 84: same thing as the `goto`

ui/user/src/lib/components/admin/usage/UsageGraphs.svelte
```diff
@@ -341,6 +345,7 @@
 			},
 			formatTooltipText: (data) => `${data.errorCount} errors • ${data.serverDisplayName}`,
 			transform: (stats) => {
+				// eslint-disable-next-line svelte/prefer-svelte-reactivity
```
ivyjeong13, line 348: had to add disable-line for instances of `Map` or `Set` used that don't need to be reactive

ryu-man (approved):

Issue #4989, opened 2025-11-11
User is presented with empty  `Launch` page when connecting to remote MCP servers with static header.


sangee2004:
**Describe the bug**
User is presented with empty  `Launch` page when connecting to remote MCP servers with static headers.

**To Reproduce**
1. Create a remote MCP server entry with static header :

<img width="1138" height="606" alt="Image" src="https://github.com/user-attachments/assets/f1e8f91e-0b7c-47ad-a1ea-654fd5133fb0" />

2. Connect to this MCP server.

Notice that user is presented with empty `Launch` page
<img width="820" height="302" alt="Image" src="https://github.com/user-attachments/assets/8295f651-9430-4cab-b813-e04e09211884" />

**Expected behavior**
User should not be presented with this  empty `Launch` page

ryu-man:
<img width="942" height="218" alt="Image" src="https://github.com/user-attachments/assets/734fe281-1fd0-4bec-8eeb-212afd245076" />

@sangee2004 Should the input of a static header strictly accept one of the values defined in the configuration?

sangee2004:
@ryu-man Static headers should not be presented to the user when configuring MCP servers.

ryu-man:
@sangee2004 So the issue is only about improving how the UI is looking when there is no header to display, right?

sangee2004:
@ryu-man Yes

ivyjeong13:
Sorry just looking at this, and wondering for this case, wouldn't a better experience be this just goes straight to launching?

ryu-man:
@ivyjeong13 Thinking of this from a user perspective I would say the only issue with this experience is I would be confused why sometine I get a dialog and sometime it go staight to the launch! In my opinion, the best experience would be to add a checkbox on the launch dialog that allows the user to skip this step in the future and go straight to launch, which will help keep the user informed about the process. What do you think?

ivyjeong13:
@ryu-man We already do this for multi-user and single-user servers -- if there's nothing to configure, it'll just go ahead and launch it.

ivyjeong13:
<img width="1442" height="996" alt="Image" src="https://github.com/user-attachments/assets/ed85c223-b5c8-4ca6-beac-04b3a78225be" />

Also FYI, I think this issue might still be occurring (on main with latest change from an hour ago)

ryu-man:
> Also FYI, I think this issue might still be occurring (on main with latest change from an hour ago)

[...]

PR #5257, opened 2025-12-10
fix: enhance MCP connect dialog when there is no header to configure

ryu-man:
Addresses #4989

- show a message tell the user there no header to configure

### DIFF
--- a/ui/user/src/lib/components/mcp/CatalogConfigureForm.svelte
+++ b/ui/user/src/lib/components/mcp/CatalogConfigureForm.svelte
@@ -8,6 +8,7 @@
 	import SensitiveInput from '../SensitiveInput.svelte';
 	import { twMerge } from 'tailwind-merge';
 	import Confirm from '../Confirm.svelte';
+	import type { LaunchServerType } from '$lib/services';
 
 	export type LaunchFormData = {
 		envs?: MCPServerInfo['env'];
@@ -55,6 +56,7 @@
 		showAlias?: boolean;
 		disableOutsideClick?: boolean;
 		animate?: 'slide' | 'fade' | null;
+		type?: LaunchServerType;
 	}
 	let {
 		form = $bindable(),
@@ -71,7 +73,8 @@
 		isNew,
 		showAlias,
 		disableOutsideClick,
-		animate = 'slide'
+		animate = 'slide',
+		type
 	}: Props = $props();
 	let configDialog = $state<ReturnType<typeof ResponsiveDialog>>();
 	let highlightedFields = $state<Set<string>>(new Set());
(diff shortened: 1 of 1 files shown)


ui/user/src/lib/components/mcp/CatalogConfigureForm.svelte
```diff
@@ -83,6 +83,18 @@
 	let isOpen = $state(false);
 	let localError = $state<string | undefined>();
 
+	const headers = $derived.by(() => {
+		if (form && 'headers' in form) {
+			return (
+				form.headers
+					?.map((header, i) => ({ index: i, data: header }))
+					?.filter((item) => item.data.required) ?? []
```
g-linville: I don't think we want to filter out all non-required headers, right? Only the ones that already have a static value set on them.

ryu-man: Yes, you are right, and as far as I remember I could find another field to filter header based on whether it is static or not, but I need to have a second look

ryu-man: image


As I said above, currently a header item does not contain solid information on whether it is static or user-supplied except those created as static are marked as not required!

g-linville: @ryu-man When you first call the backend to get the information for an MCPServerCatalogEntry, which includes its manifest with the runtime information and headers, any that already have a `value` set are static.

ryu-man: Had to add 'isStatic' boolean attribute calculated based on your explanation, value was explicitly overwriten with an empty string

ui/user/src/lib/components/mcp/CatalogConfigureForm.svelte
```diff
@@ -537,33 +552,38 @@
...
+									bind:value={form!.headers![header.index].value}
+									class="text-input-filled"
+								/>
 							{/if}
-						{/each}
-					{/if}
+						</div>
+					{:else}
+						<div class="flex h-32 w-full items-center bg-surface1 rounded-md p-8 text-on-surface1">
+							<div>There are no headers to configure for this MCP server.</div>
```
njhale: As written, it looks like this will be displayed for all runtimes. I think we should only display this message for remote runtimes/components.

ryu-man: Yeah, you're right, nice catch

ryu-man: Pushed changes to show message only for remote server

Issue #4983, opened 2025-11-10
Not able to use `Configure Tools` option for for deleted MCP servers which are still part of composite MCP servers.


sangee2004:
**Describe the bug**
Not able to use `Configure Tools` option for for deleted MCP servers which are still part of composite MCP servers.

**To Reproduce**
1. Create composite MCP server with couple of MCP servers - `santest1` and `Asana` in my case.
2. Delete  `santest1` mcp server entry
3. Go to composite MCP server entry which will have a `upgrade` notice
4. Do not Upgrade the server
5. Try to use `Configure Tools` option for  `santest1`

Notice that user is presented with MCP server list.

<img width="523" height="528" alt="Image" src="https://github.com/user-attachments/assets/22bf02fd-0de7-44d8-a394-f3e2b1f3c11b" />

6. Go to `Asana` entry and click on `Configure Tools` option
7. When presented with `Configure Asana Tools` page , close it
8. Go to `santest1` entry and click on `Configure Tools` option

Notice that user is presented with  `Configure Asana Tools` workflow.

**Expected behavior**
`Configure Tools` option for  `santest1` should present user with workflow to configure `santest1`  tools

njhale:
This is part of a broader set of related issues I'm currently working on.

sangee2004:
@njhale I am still not able to configure tools for deleted MCP servers that are configured as part of composite server

I am testing with latest build - [v0.0.0-dev+dda4b7a7](https://github.com/obot-platform/obot/commit/dda4b7a7)

I see this error reported:
<img width="1040" height="389" alt="Image" src="https://github.com/user-attachments/assets/b88bedda-2d9d-4e01-a2f1-d74a5a495ce5" />

njhale:
OK, something must have broken with additional changes/rebase in that PR. Will fix.

sangee2004:
This issue is addressed when testing with latest build from `main`

PR #5271, opened 2025-12-11
Oauth enhancements

thedadams:


pkg/api/handlers/mcpcatalogs.go
```diff
@@ -1164,19 +1161,10 @@ func (h *MCPCatalogHandler) GenerateComponentToolPreviews(req api.Context) error
...
-			Namespace: composite.Namespace,
-		},
-		Spec: v1.MCPServerCatalogEntrySpec{
-			MCPCatalogName: composite.Spec.MCPCatalogName,
-			Manifest:       component.Manifest,
-		},
+	// Load the standalone catalog entry for shape / metadata and attach the preview.
+	var entry v1.MCPServerCatalogEntry
+	if err := req.Get(&entry, componentID); err != nil {
+		return fmt.Errorf("failed to get component catalog entry: %w", err)
```
njhale: This is undoing a change I recently made that fixes https://github.com/obot-platform/obot/issues/4983.

We want to be able to configure tool overrides for components whose original catalog entry no longer exists.

In general, we should be able to make changes to the composite entry and continue using it (creating servers with it, updating them, etc) even if the original component entries do not exist anymore; at least until the entry is "upgraded" and the component with missing entries are removed.

That being said, this feature may be more trouble than it's worth (especially since there are a few other bugs open that relate to missing parent entries for component servers).

njhale (approved):

Issue #5042, opened 2025-11-14
`Filter By Date` option is not usable when creating one-time export.


sangee2004:
**Describe the bug**
`Filter By Date` option is not usable when creating one-time export.

Pre Req - Configure storage provider for exporting audit logs
**To Reproduce**
1. Go to `Audit Logs` and create one-time export
2. Fill `Export Name` and `Bucket Name` 
3. Click on `Filter By Date` option 
Notice that calendar view appears to truncated with no left margin

<img width="1683" height="1068" alt="Image" src="https://github.com/user-attachments/assets/a34c2586-944c-434e-9b6e-3a1fa1803c9e" />

In smaller screen size , calendar is not fully presented and user needs to scroll down to get to see it fully. Left side of the calendar is truncated and first column and option to view pervious month is not presented.

<img width="1480" height="717" alt="Image" src="https://github.com/user-attachments/assets/812a4ae8-5100-409a-bca5-78706771affb" />

4. Click on any date

Notice that this results in export being created with the default 1 week time range

https://www.loom.com/share/5de0c23606554936be51eaf1b9bd8df1

**Expected behavior**
Should be able to set a time range using the calendar view that shows up when using `Filter By Date` option

ryu-man:
Already solved by #5239

shwetapsdet:
Verified it on Obot main it's working fine as expected

<img width="1590" height="787" alt="Image" src="https://github.com/user-attachments/assets/44247c8d-f71a-4253-bad5-18ad3b9b7886" />

PR #5300, opened 2025-12-12
fix: improve selecting time range in audit log export

ryu-man:
Addresses #5042

- on small devices both quick actions and the calendar are displayed as dialog rather than a popover


ui/user/src/lib/components/Calendar.svelte
```diff
@@ -226,6 +229,20 @@
 		}
 	});
 
+	const isSmallScreen = $derived(viewport.screen === 'sm' || viewport.screen === 'xs');
```
ivyjeong13: We have `responsive.isMobile` to handle this check?

ui/user/src/routes/+layout.svelte
```diff
@@ -15,11 +16,17 @@
...
 	let { children, data }: Props = $props();
 
+	new Viewport().share();
+
 	untrack(() => {
 		if (data.appPreferences) {
 			appPreferences.initialize(data.appPreferences);
 		}
 
+		if (data.appPreferences) {
```
ivyjeong13: Does this need to be called again?

ryu-man: No, I'm not sure how I ended up adding that!

ui/user/src/lib/components/Calendar.svelte
ryu-man: I am going to use that

ryu-man: Made the change; it does use `responsive.isMobile` now

g-linville (approved):

PR #5303, opened 2025-12-12
chore: allow users to list all PowerUserWorkspaces

g-linville:
requested by @cjellick


g-linville:
Turns out we don't need this.

PR #5313, opened 2025-12-12
chore: give basic and PU users access to only user IDs and display names

g-linville:
This is needed for some new work going into the frontend.

### DIFF
--- a/pkg/gateway/server/user.go
+++ b/pkg/gateway/server/user.go
@@ -73,7 +73,7 @@ func (s *Server) getUsers(apiContext api.Context) error {
 		}
 	}
 
-	// Basic and Power users are only allowed to access IDs and emails, so we have all the information needed for that.
+	// Basic and Power users are only allowed to access IDs and display names, so we have all the information needed for that.
 	if userIsBasicOrPower(apiContext.User) {
 		trimmedUsers := make([]types2.User, 0, len(validUsers))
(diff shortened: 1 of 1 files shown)


apiclient/types/types.go
```diff
@@ -9,7 +9,7 @@ type List[T any] struct {
 
 type Metadata struct {
 	ID       string            `json:"id,omitempty"`
-	Created  Time              `json:"created,omitempty"`
+	Created  Time              `json:"created,omitzero"`
```
g-linville, line 12: `omitempty` didn't do anything for zero-value Time object.

pkg/gateway/server/user.go
```diff
@@ -361,3 +386,13 @@ func (s *Server) restrictGroups(ctx context.Context, gptscriptClient *gptscript.
 
 	return groups, nil
 }
+
+func userIsBasicOrPower(u user.Info) bool {
+	for _, group := range u.GetGroups() {
+		switch group {
+		case types2.GroupPowerUserPlus, types2.GroupAuditor, types2.GroupAdmin, types2.GroupOwner:
```
g-linville, line 393: I'm considering Auditor as "higher" than PU (even though it is orthogonal) here, because it already has permissions to list all users.

cjellick:
looks good. please merge so ivy can move forward

Issue #5272, opened 2025-12-11
Not able to connect to composite MCP server when it has an MCP server that has been deleted from registry and composite MCP server has not been upgraded yet.


sangee2004:
**Describe the bug**
Not able to connect to the composite MCP server when it has an MCP server that has been deleted from the registry, and the composite MCP server has not been upgraded yet.

**To Reproduce**
1. Create composite MCP server with couple of MCP servers - `sancheckdelete` and `Asana` in my case.
2. Delete  `sancheckdelete` mcp server entry
3. Go to the composite MCP server entry, which will have an `upgrade` notice
4. Do not Upgrade the server
5. As a regular user, connect to this composite MCP server

User will be presented with the following error message:

<img width="1388" height="351" alt="Image" src="https://github.com/user-attachments/assets/d55ffb9f-fd77-42fa-939a-ddf8d3e5cf05" />

**Expected behavior**
User should be able to connect to this MCP server successfully.

shwetapsdet:
Verified it in obot main and it's working as expected

https://github.com/user-attachments/assets/260b858f-11f0-427a-a6e5-0821e3fb3068

Issue #5214, opened 2025-12-06
Not able to update composite MCP server instance after `Github` MCP server is added to it because of validating errors relating to hostname.


sangee2004:
**Describe the bug**
Not able to update composite MCP server instance after `Github` MCP server is added to it because of validating errors relating to hostname. 

**To Reproduce**
1. Create a composite MCP server with any MCP server.
2. Connect to this MCP server.
3. Update the composite MCP server and add  `Github` MCP server 
4. Try to  use `Update Server` option to update the composite MCP server instance (that was created from step 2)

Notice that user is presented with following error message:

<img width="1148" height="482" alt="Image" src="https://github.com/user-attachments/assets/3471266d-4d61-46ca-b410-1011049ffd11" />

`/trigger-update` api fails with `500 Internal Server Error` and following error message:
```
failed to convert component manifest: runtime remote validation error for field URL: user URL is required when catalog entry specifies hostname constraint
```
**Expected behavior**
Updating composite MCP server instance should succeed in this case.

njhale:
Looked into this briefly:
- We're validating hostname constraints on `trigger-update` for components
- I have an idea about how to avoid doing this without tackling https://github.com/obot-platform/obot/issues/5012 first (that issue also affects hostname constraints b/c the runtime configuration is also missing the original hostname constraint from the catalog entry it was created from)
- IIUC, we basically need to avoid validating for this use case and set the `Spec.NeedsURL` field of the newly added component server on update

shwetapsdet:
Verified it on Obot and it's working as expected

https://github.com/user-attachments/assets/f6accbf4-5221-46c3-a50c-118c3337e5b6

Issue #5127, opened 2025-11-22
409 errors seen sometimes when updating composite MCP server instances


sangee2004:
**Describe the bug**
409 errors seen sometimes when updating composite MCP server instances

**To Reproduce**
1. Create composite MCP sever with few MCP servers added to it
2. As regular user, connect to this composite MCP sever and chat with tools from all MCP servers.
3. As admin , update composite MCP sever and make some changes ( add a new MCP server or change tool selection for any existing MCP server)
4. Update the connected composite MCP server instance (created from step2)
Notice that trigger-update api call takes very long time ( ~ 30+ seconds) to succeed  which is tracked in https://github.com/obot-platform/obot/issues/5126
Sometimes, it reports the following error:
<img width="987" height="613" alt="Image" src="https://github.com/user-attachments/assets/8a324382-1bd7-42d8-bb15-ea75b04955e9" />

Full error message:

```
409 /mcp-servers/ms1fw2rb/trigger-update: Operation cannot be fulfilled on MCPServer.obot.obot.ai "ms1vqpdp": the object has been modified; please apply your changes to the latest version and try again
```

**Note** - Server update has actually succeeded in this case .

**Expected behavior**
409 errors should not be seen when updating composite MCP server instances

njhale:
Note: I'd expect to see this if the composite MCP server or any of its component servers/instances changed during the update. This can happen for any number of reasons, but I think the takeaway is that the update operation needs to be more robust and either happen in a controller or have some retry logic when updating any of the affected resources.

shwetapsdet:
Verified it on Obot main and it's working fine as expected

<img width="1412" height="890" alt="Image" src="https://github.com/user-attachments/assets/a13a09d7-6335-49e4-bb36-a228b69b528a" />

https://github.com/user-attachments/assets/28277ff6-d8a2-4a4b-b5bd-2a76446bcdfd

Issue #5084, opened 2025-11-20
No `Update Server` option presented for connected MCP server instance after upgrading composite MCP server for changes made to remote MCP server.


sangee2004:
**Describe the bug**
No `Update Server` option presented for connected MCP server instance after upgrading composite MCP server for changes made to remote MCP server.

**To Reproduce**
1. Create a remote MCP server  -`mcp1` with remote url say - https://gmail-mcp.obot.ai/mcp
2. Create composite MCP server with this remote MCP server - `mcp1`
3. Connect to this composite MCP server. User is prompted to go thru oauth workflow and is able to chat with gmail MCP server as expected.
4. As admin user, update remote MCP server - `mcp1`  created in step1 and update the remote url to say - https://outlook-mcp.obot.ai/mcp
5. Upgrade composite MCP server.

After this , i dont see `Update Server` option presented for the connected composite MCP server instance

**Expected behavior**
 `Update Server` option  option should be presented for the connected composite MCP server instance.

cjellick:
Thoughts here @njhale ?

njhale:
@cjellick server drift detection needs to be updated to handle this case. One of the things in the change I'm working on, although it depends on some larger changes to make it happen

shwetapsdet:
Verified it on Obot main and it's working as expected

https://www.loom.com/share/3f5f46121a084fe0a53e17cb146581a5

https://www.loom.com/share/8e59f1603e09486fade63d3c32cc5d23

PR #5317, opened 2025-12-13
fix: composite mcp servers

njhale:
Addresses:
- https://github.com/obot-platform/obot/issues/4975
- https://github.com/obot-platform/obot/issues/5272
- https://github.com/obot-platform/obot/issues/4929
- https://github.com/obot-platform/obot/issues/5084
- https://github.com/obot-platform/obot/issues/5127
- https://github.com/obot-platform/obot/issues/5214

### DIFF
--- a/apiclient/types/mcpserver.go
+++ b/apiclient/types/mcpserver.go
@@ -47,7 +47,6 @@ type RemoteRuntimeConfig struct {
 	URLTemplate string      `json:"urlTemplate,omitempty"` // URL template for user URLs
 	Hostname    string      `json:"hostname,omitempty"`    // Optional: Hostname constraint the URL conforms to
 	Headers     []MCPHeader `json:"headers,omitempty"`     // Optional
-
 }
 
 // RemoteCatalogConfig represents template configuration for remote servers in catalog entries
--- a/go.mod
+++ b/go.mod
@@ -62,6 +62,8 @@ require (
 	golang.org/x/crypto v0.46.0
 	golang.org/x/exp v0.0.0-20250813145105-42675adae3e6
 	golang.org/x/mod v0.30.0
+	golang.org/x/oauth2 v0.30.0
+	golang.org/x/sync v0.19.0
 	golang.org/x/term v0.38.0
 	golang.org/x/text v0.32.0
 	google.golang.org/api v0.197.0
@@ -311,8 +313,6 @@ require (
 	go.yaml.in/yaml/v2 v2.4.2 // indirect
 	go4.org v0.0.0-20230225012048-214862532bf5 // indirect
 	golang.org/x/net v0.47.0 // indirect
-	golang.org/x/oauth2 v0.30.0
-	golang.org/x/sync v0.19.0
 	golang.org/x/sys v0.39.0 // indirect
 	golang.org/x/time v0.11.0 // indirect
 	golang.org/x/tools v0.39.0 // indirect
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -385,7 +385,7 @@ func (m *MCPHandler) DeleteServer(req api.Context) error {
 	}
 
 	// Prevent deletion of multi-user servers that are referenced by running composite MCP servers or catalog entries.
-	dependencies, err := m.listCompositeDeletionDependencies(req, server)
+	dependencies, err := listCompositeDeletionDependencies(req, server)
 	if err != nil {
 		return fmt.Errorf("failed to list composite deletion dependencies: %w", err)
 	}
@@ -417,7 +417,7 @@ type compositeDeletionDependency struct {
 }
 
 // listCompositeDeletionDependencies lists the composite MCP servers and catalog entries that depend on the given multi-user server.
-func (*MCPHandler) listCompositeDeletionDependencies(req api.Context, server v1.MCPServer) ([]compositeDeletionDependency, error) {
+func listCompositeDeletionDependencies(req api.Context, server v1.MCPServer) ([]compositeDeletionDependency, error) {
 	if server.Spec.MCPServerCatalogEntryName != "" {
 		// Not a multi-user server, skip dependency check
 		return nil, nil
@@ -1932,9 +1932,9 @@ func (m *MCPHandler) configureCompositeServer(req api.Context, compositeServer v
 	}
 
 	var (
-		manifestChanged    bool
-		oldCompositeServer = compositeServer.DeepCopy()
-		componentCreds     = make(map[string]gptscript.Credential, len(existingServers))
+		manifestChanged bool
+		oldManifestHash = hash.Digest(compositeServer.Spec.Manifest)
+		componentCreds  = make(map[string]gptscript.Credential, len(existingServers))
 	)
 	for i, component := range compositeServer.Spec.Manifest.CompositeConfig.ComponentServers {
 		componentID := component.ComponentID()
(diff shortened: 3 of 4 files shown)


apiclient/types/mcpserver.go
```diff
@@ -42,9 +42,12 @@ type ContainerizedRuntimeConfig struct {
...
 type RemoteRuntimeConfig struct {
-	URL        string      `json:"url"`               // Required: Full URL to remote MCP server
-	Headers    []MCPHeader `json:"headers,omitempty"` // Optional
-	IsTemplate bool        `json:"isTemplate"`        // Optional: Whether the URL is a template
+	URL         string      `json:"url"`                   // Required: Full URL to remote MCP server
+	IsTemplate  bool        `json:"isTemplate"`            // Optional: Whether the URL is a template
+	URLTemplate string      `json:"urlTemplate,omitempty"` // URL template for user URLs
+	Hostname    string      `json:"hostname,omitempty"`    // Optional: Hostname constraint the URL conforms to
+	Headers     []MCPHeader `json:"headers,omitempty"`     // Optional
+
```
thedadams:
```suggestion
```

go.mod
```diff
@@ -312,7 +312,7 @@ require (
 	go4.org v0.0.0-20230225012048-214862532bf5 // indirect
 	golang.org/x/net v0.47.0 // indirect
 	golang.org/x/oauth2 v0.30.0
-	golang.org/x/sync v0.19.0 // indirect
+	golang.org/x/sync v0.19.0
```
thedadams: I'm surprised that `go mod tidy` doesn't do this, but if these are direct dependencies, then they should be in the first `require` block.

pkg/api/handlers/mcphelpers.go
```diff
@@ -152,3 +152,46 @@ func DeleteCredentialIfExists(ctx context.Context, gptClient *gptscript.GPTScrip
...
+	}
+	if !modified && len(cred.Env) == len(existing.Env) {
+		// No update needed
+		return false, nil
+	}
+
+	// Remove and recreate the credential to update it
+	if err := gptClient.DeleteCredential(ctx, existing.Context, cred.ToolName); err != nil && !errors.As(err, &gptscript.ErrNotFound{}) {
+		return false, fmt.Errorf("failed to remove existing credential: %w", err)
+	}
```
thedadams: nit:
```suggestion
```

pkg/controller/handlers/mcpserver/mcpserver.go
```diff
@@ -439,3 +404,168 @@ func (h *Handler) CleanupNestedCompositeServers(req router.Request, _ router.Res
...
+				if err := req.Get(&latestServer, compositeServer.Namespace, existingServer.Name); err != nil {
+					return err
+				}
+
+				latestServer.Spec.Manifest = component.Manifest
+				latestServer = withNeedsURL(latestServer)
+				return req.Client.Update(req.Ctx, &latestServer)
+			}); err != nil {
+				return fmt.Errorf("failed to update existing component server: %w", err)
+			}
```
thedadams, line 540: If the updating of component servers was done in a different handler (that operated on the component itself), then we wouldn't need to retry here. We could just return the error and have the controller framework handle the retrying.

pkg/api/handlers/mcp.go
```diff
@@ -414,7 +417,7 @@ type compositeDeletionDependency struct {
 }
 
 // listCompositeDeletionDependencies lists the composite MCP servers and catalog entries that depend on the given multi-user server.
-func (m *MCPHandler) listCompositeDeletionDependencies(req api.Context, server v1.MCPServer) ([]compositeDeletionDependency, error) {
+func (*MCPHandler) listCompositeDeletionDependencies(req api.Context, server v1.MCPServer) ([]compositeDeletionDependency, error) {
```
```suggestion
func listCompositeDeletionDependencies(req api.Context, server v1.MCPServer) ([]compositeDeletionDependency, error) {
```

pkg/api/handlers/mcp.go
```diff
@@ -1976,140 +1922,134 @@ func (m *MCPHandler) configureCompositeServer(req api.Context, compositeServer v
...
 
-	// Configure each component with its env vars and persist enabled flags
-	// Build quick map for parent composite components by catalog entry ID
-	parentComps := map[string]int{}
-	if compositeServer.Spec.Manifest.CompositeConfig != nil {
-		for i, comp := range compositeServer.Spec.Manifest.CompositeConfig.ComponentServers {
-			parentComps[comp.ComponentID()] = i
+	var (
+		manifestChanged    bool
+		oldCompositeServer = compositeServer.DeepCopy()
```
thedadams: If there is a reason why we are deep copying here, then can we add a comment about that? If there isn't, then can we remove it?

pkg/api/handlers/mcp.go
```diff
@@ -3599,146 +3515,84 @@ func (m *MCPHandler) TriggerUpdate(req api.Context) error {
 }
 
 // triggerCompositeUpdate upgrades a composite server and all its component servers from the latest catalog entry
-func (m *MCPHandler) triggerCompositeUpdate(req api.Context, server v1.MCPServer, entry v1.MCPServerCatalogEntry) error {
+func (m *MCPHandler) triggerCompositeUpdate(req api.Context, compositeServer v1.MCPServer, entry v1.MCPServerCatalogEntry) error {
+	oldServer := compositeServer.DeepCopy()
```
thedadams: Same comment here about deep copying.

go.mod
njhale: yup, it does not. could be an issue caused by a switch from existing indirect to direct.

pkg/controller/handlers/mcpserver/mcpserver.go
njhale: I think a single handler that manages all components for a composite server is more manageable. It's easier to tell when all of the components are up to date.

> We could just return the error and have the controller framework handle the retrying.

We can still do this here with the current setup and it should retry just fine. I just figured this would speed things up a bit.

pkg/api/handlers/mcp.go
njhale: `serverManifestFromCatalogEntryManifest` can have side-effects on some fields in the manifest. This prevents that from being an issue. I'll add a comment.

pkg/api/handlers/mcphelpers.go
njhale: don't we need to remove the credential before recreating it?

go.mod
njhale: moved.

njhale:
@thedadams feedback should be addressed. LMK if you'd like me to change anything further.

pkg/api/handlers/mcphelpers.go
thedadams: No, we don't. You can keep this in if you want, but it's not necessary.

pkg/controller/handlers/mcpserver/mcpserver.go
thedadams: If you do return the conflict error, then you need to use an uncached list for the component servers.

pkg/api/handlers/mcp.go
thedadams: Why would some of the fields in the manifest cause issues?

njhale: In `updateCompositeManifest` we're generating a hash of the old server's manifest to compare it to when determining if we should fail on conflict or retry. Changing those fields impacts the hash and throws off the comparison.

It seems like you're against a deepcopy here, so I can refactor this so that we take the hash up-front and pass it into `updateCompositeManifest` instead if that's sounds better to you.

pkg/api/handlers/mcphelpers.go
njhale: ah, ok. I just saw the delete happening in most places and a matching comment stating that we need to remove and recreate to update. Maybe I misinterpreted. I'll remove it.

pkg/api/handlers/mcp.go
thedadams: Yeah, let's try that.

njhale:
@thedadams addressed your second round of feedback

PR #5318, opened 2025-12-13
Fix: Drop tools/list from audit log default

cjellick:
Our default API call for audit logs is too big. A big contributor to
this is the tools/list call which gets called a lot and is typically
pertty big. This just drops that callType from the default set of
callTypes we show on the audit log page.


njhale:
May need to fix the user lint before merging

Issue #5314, opened 2025-12-12
MCP servers created by a PUP user get reported as being in the global registry in the `MCP servers` view.


sangee2004:
**Describe the bug**
MCP servers created by a PUP user get reported as being in the global registry in the `MCP servers` view.

**To Reproduce**
1. As a PUP user, create an MCP server.
Notice that these MCP servers get reported as being in global registry in `MCP servers` view for this PUP user.

<img width="964" height="456" alt="Image" src="https://github.com/user-attachments/assets/59925d2d-b0df-4fe2-a2ec-3e70bd0464bd" />

These MCP servers are correctly reported as being in PUP's workspace for owner users.

<img width="896" height="556" alt="Image" src="https://github.com/user-attachments/assets/327a4912-3af5-466e-8a63-d4e9a3a36722" />

**Expected behavior**
MCP servers created by PUP user should get reported as being in  PUP's workspace for PUP users.

cjellick:
@njhale can you validate this one?

njhale:
So it looks like the `Registry` is now reported correctly for PuP users (and Admins, that still works correctly), but the registry pill for PuP users still shows `Global Registry` for single and remote MCP servers (but is correct for multi-user servers created by the PuP user interestingly enough).

Here's the current PuP user view for a remote created and assigned to their default "registry" as of `9aadb90f`:
 
<img width="540" height="157" alt="Image" src="https://github.com/user-attachments/assets/a3f47b8c-ea66-4ff9-8b53-1503ad83c2b8" />

<img width="1166" height="62" alt="Image" src="https://github.com/user-attachments/assets/a05185ed-089a-4efc-9063-a235066d7c3f" />

Going to kick this back to triage.

ivyjeong13:
> So it looks like the `Registry` is now reported correctly for PuP users (and Admins, that still works correctly), but the registry pill for PuP users still shows `Global Registry` for single and remote MCP servers (but is correct for multi-user servers created by the PuP user interestingly enough).
> 
> Here's the current PuP user view for a remote created and assigned to their default "registry" as of `9aadb90f`:
> 
[...]

PR #5327, opened 2025-12-15
fix: user registry pill for basic/pup viewing a registry entry

ivyjeong13:
Addresses follow up on #5314 
* use common util fn `getUserRegistry`
* removes restriction on getting back powerUserWorkspaceID in all-mcps-/entries/{entry_id}

### DIFF
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -88,13 +88,7 @@ func (m *MCPHandler) GetEntryFromAllSources(req api.Context) error {
 		return types.NewErrForbidden("user is not authorized to access this catalog entry")
 	}
 
-	// Only include powerUserWorkspaceID if it matches the user's workspace
-	var powerUserWorkspaceID string
-	if entry.Spec.PowerUserWorkspaceID == system.GetPowerUserWorkspaceID(req.User.GetUID()) {
-		powerUserWorkspaceID = entry.Spec.PowerUserWorkspaceID
-	}
-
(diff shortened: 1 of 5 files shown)


pkg/api/handlers/mcp.go
```diff
@@ -88,13 +88,7 @@ func (m *MCPHandler) GetEntryFromAllSources(req api.Context) error {
...
 	}
 
-	// Only include powerUserWorkspaceID if it matches the user's workspace
-	var powerUserWorkspaceID string
-	if entry.Spec.PowerUserWorkspaceID == system.GetPowerUserWorkspaceID(req.User.GetUID()) {
-		powerUserWorkspaceID = entry.Spec.PowerUserWorkspaceID
-	}
-
-	return req.Write(ConvertMCPServerCatalogEntryWithWorkspace(entry, powerUserWorkspaceID, ""))
+	return req.Write(ConvertMCPServerCatalogEntryWithWorkspace(entry, entry.Spec.PowerUserWorkspaceID, ""))
```
njhale, line 91: this is the only change I'm wondering may break other things, but other than that LGTM

ivyjeong13: :+1: Thanks Nick! Tagging on @g-linville just to make sure!

g-linville: Yep this looks fine to me

ivyjeong13: Got okay from Grant 🎉

PR #5335, opened 2025-12-16
fix: directly launch remote server when there is no header to configure

ryu-man:
Addresses #4989

- apply a unified experience to show configuration only when there is somthing to configure


ivyjeong13:
Remote server going straight to launch when just having static header(s) :+1: 

But noticed when I add it to a composite server and connect to it, I see it show up as having something optional to configure when it shouldn't. 

Screenshot 2025-12-16 at 11 53 35 AM

ryu-man:
This should be a small change I will address it in this PR. Thanks for the mention

PR #5346, opened 2025-12-17
fix: cancel correctly redirects for pup mcp-registries

ivyjeong13:
* correct goto route for /mcp-registries depending if user has admin access
* for both canceling on creating new registry or for create new registry during catalog entry creation

### DIFF
--- a/ui/user/src/lib/components/admin/AccessControlRuleForm.svelte
+++ b/ui/user/src/lib/components/admin/AccessControlRuleForm.svelte
@@ -437,8 +437,10 @@
 						onclick={() => {
 							if (redirect) {
 								goto(redirect);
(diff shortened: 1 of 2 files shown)


cjellick:
No issue for this one as i was lazy and just reported in slack - https://acorn-io.slack.com/archives/C07P0EZR917/p1765936415550669

WIll just use the pr for qa

shwetapsdet:
Tested on both main and local Docker environments.
Verified correct routing for /mcp-registries based on admin access during:

- Creating a new registry from catalog entry
- Canceling registry creation

All flows are working as expected.

https://www.loom.com/share/20f986839d6748528413a3fb5008647a

Issue #5347, opened 2025-12-17
Issue: Multi-Cookie Session Authentication Failure


gazzadownunder:
## Summary

OAuth authentication fails with an infinite login loop when sessions exceed 4KB, causing users to be repeatedly redirected to the login page despite successful authentication.

## Environment

- **Obot Version:** Current main branch (as of 2025-12-17)
- **OAuth Provider:** oauth2-proxy (all versions)
- **Affected Auth Providers:** Keycloak, Okta, Entra ID, or any provider with large ID tokens
- **Trigger Condition:** Session size > 4KB 

## Reproduction Steps

1. Configure an OAuth provider  with a user who has many roles/groups
2. Start Obot with the configured auth provider
3. Navigate to `http://localhost:8080` and initiate login
4. Successfully authenticate with the OAuth provider
5. Observe: Browser redirects back to login page instead of /admin
6. Check browser cookies: Multiple cookies present (`obot_access_token_0`, `obot_access_token_1`, etc.)
7. Check browser network trace: Cookies are sent with requests
8. Result: Infinite redirect loop - user cannot log in

## Expected Behaviour

After successful OAuth authentication:
1. Session cookie(s) stored in browser
2. User authenticated on subsequent requests
3. User redirected to /admin or requested page
4. No further login prompts

## Actual Behaviour

After successful OAuth authentication:
1. Multiple session cookies stored (`obot_access_token_0`, `obot_access_token_1`, ...)
2. Browser sends cookies with subsequent requests
3. Obot `AuthenticateRequest()` returns "not authenticated"
4. User redirected back to login page
5. Infinite loop - login never succeeds

## Root Cause

**File:** [pkg/proxy/proxy.go:54](pkg/proxy/proxy.go#L54)

When OAuth sessions exceed 4KB, oauth2-proxy automatically splits the session across multiple cookies:
- Cookie names: `obot_access_token_0`, `obot_access_token_1`, `obot_access_token_2`, etc.
- The base cookie `obot_access_token` **does NOT exist** in multi-cookie mode

The `AuthenticateRequest()` function only checks for the base cookie name:

```go
func (pm *Manager) AuthenticateRequest(req *http.Request) (*authenticator.Response, bool, error) {
    // Check for the access token cookie.
    if _, err := req.Cookie(ObotAccessTokenCookie); errors.Is(err, http.ErrNoCookie) {
        return nil, false, nil  // ← Returns "not authenticated"
    }
    // ... rest of authentication logic
}
```

[...]

PR #5348, opened 2025-12-17
Fix: Multi-Cookie Session Authentication Failure

gazzadownunder:
Issue: Multi-Cookie Session Authentication Failure #5347

## Root Cause
[pkg/proxy/proxy.go:54](https://github.com/obot-platform/obot/issues/pkg/proxy/proxy.go#L54)

When OAuth sessions exceed 4KB, oauth2-proxy automatically splits the session across multiple cookies:

Cookie names: obot_access_token_0, obot_access_token_1, obot_access_token_2, etc.
The base cookie obot_access_token does NOT exist in multi-cookie mode
The AuthenticateRequest() function only checks for the base cookie name:

```
func (pm *Manager) AuthenticateRequest(req *http.Request) (*authenticator.Response, bool, error) {
    // Check for the access token cookie.
    if _, err := req.Cookie(ObotAccessTokenCookie); errors.Is(err, http.ErrNoCookie) {
        return nil, false, nil  // ← Returns "not authenticated"
    }
    // ... rest of authentication logic
}
```
When the base cookie doesn't exist:

1. req.Cookie() returns http.ErrNoCookie
2. Function returns nil, false, nil (not authenticated)
3. Middleware redirects to login page
4. User already has valid session but cannot be authenticated

## Solution
Check for both the base cookie and the first multi-cookie:

```
func (pm *Manager) AuthenticateRequest(req *http.Request) (*authenticator.Response, bool, error) {
    // Check for the access token cookie.
    // When sessions exceed 4KB, oauth2-proxy splits them into multiple cookies (_0, _1, _2, etc.)
    // so we check for either the base cookie or the _0 cookie.
    _, err := req.Cookie(ObotAccessTokenCookie)
    if errors.Is(err, http.ErrNoCookie) {
        // Try the _0 cookie for multi-cookie sessions
        if _, err := req.Cookie(ObotAccessTokenCookieZero); errors.Is(err, http.ErrNoCookie) {
            return nil, false, nil
        }
    }
    // ... rest of authentication logic
}


cjellick:
@njhale @g-linville review this when you have a chance. I know it's older, but it might not be a bad thing to get in.

ivyjeong13:
Stalled out -- closing for now. Bug was noted as irreproducible but please feel free to reopen if this issue is still encountered. And if could provide more details such as number of role/groups, what auth provider used, etc. it would be greatly appreciated.

Issue #5224, opened 2025-12-08
Make Nanobot UI OAuth work with Obot


thedadams:

thedadams:
This is no longer relevant based on the way we have incorporated Nanobot chat.

PR #5357, opened 2025-12-19
chore: adjust nanobot configuration for new auth settings

thedadams:
This change removes the passing of the JWKS blob in favor of passing the JWKS URL.

This change adds a userinfo endpoint for retrieving user information from JWTs.

This change also updates the nanobot configuration after the OAuth integration there was updated.

Issue: https://github.com/obot-platform/obot/issues/5224


apiclient/types/mcpwebhookvalidation.go
```diff
@@ -92,7 +92,7 @@ func (f *MCPSelector) Matches(method, identifier string) bool {
 }
 
 func (f MCPSelector) Strings() []string {
-	s := "message:"
+	var s string
 	if f.Method != "" && f.Method != "*" {
 		s += f.Method
```
g-linville: mini nit


```suggestion
		s = f.Method
```

pkg/api/handlers/mcpgateway/oauth/userinfo.go
```diff
@@ -0,0 +1,84 @@
...
+	UpdatedAt         int64  `json:"updated_at,omitempty"`
+
+	// Email scope claims
+	Email         string `json:"email,omitempty"`
+	EmailVerified *bool  `json:"email_verified,omitempty"`
+}
+
+func (h *handler) userInfo(req api.Context) error {
+	scope := gtypes.FirstSet(req.User.GetExtra()["oauthScope"]...)
+	if !strings.Contains(scope, " profile ") {
```
g-linville: The `" profile "` with the spaces seems off to me. Is it guaranteed to be somewhere in the middle of the space-separated list? Would it be better to do a `strings.Split` and then a `slices.Contains`?

pkg/api/handlers/mcpgateway/oauth/userinfo.go
```diff
@@ -0,0 +1,84 @@
...
+			"invalid_scope", "Insufficient scope")
+	}
+
+	userID := req.User.GetUID()
+	user, err := req.GatewayClient.UserByID(req.Context(), userID)
+	if err != nil {
+		log.Debugf("User lookup failed for ID %s: %v", userID, err)
+		// Don't reveal whether user exists - return generic error
+		return h.writeUserInfoError(req, http.StatusUnauthorized,
+			"invalid_token", "Invalid token")
```
g-linville: By the time we reach this error, wouldn't we have already authenticated the user? And since we are just looking up the logged in user, I think it would be okay to return a 500 here.

apiclient/types/mcpwebhookvalidation.go
thedadams: Fixed.

pkg/api/handlers/mcpgateway/oauth/userinfo.go
thedadams: Fixed.

pkg/api/handlers/mcpgateway/oauth/userinfo.go
thedadams: This endpoint is intended to be used with the JWT tokens issued for MCP servers. It is possible to get to this point and the user not exist in the database. I think this should be a 403, so I changed it to that.

PR #5365, opened 2025-12-22
fix: ui: providers: support sensitive optional params

g-linville:


ui/user/src/lib/components/admin/ProviderConfigure.svelte
```diff
@@ -89,7 +89,8 @@
 		'OBOT_GITHUB_AUTH_PROVIDER_ALLOW_USERS',
 		'OBOT_GITHUB_AUTH_PROVIDER_TEAMS',
 		'OBOT_GITHUB_AUTH_PROVIDER_REPO',
-		'OBOT_AUTH_PROVIDER_EMAIL_DOMAINS'
+		'OBOT_AUTH_PROVIDER_EMAIL_DOMAINS',
+		'OBOT_AZURE_OPENAI_MODEL_PROVIDER_DEPLOYMENTS'
```
g-linville, line 93: This is a new field that I am soon adding to Azure OpenAI that is multi-value.

ui/user/src/lib/components/admin/ProviderConfigure.svelte
```diff
@@ -213,12 +214,27 @@
...
+											bind:value={form[parameter.name]}
+											disabled={readonly}
+											textarea={parameter.multiline}
+											growable={parameter.multiline}
+										/>
+									{:else if multipValuesInputs.has(parameter.name)}
 										<MultiValueInput
 											bind:value={form[parameter.name]}
 											id={parameter.name}
+											labels={parameter.name === 'OBOT_AUTH_PROVIDER_EMAIL_DOMAINS'
```
njhale: Is this parameter typically in required or optional? Is there a reason this condition should be in both sections here?

g-linville: whoops, this was a copy-paste error

g-linville: fixed

PR #5370, opened 2025-12-23
enhance: add pass-through LLM proxy for responses and anthropic

thedadams:


thedadams:
Darren wanted different functionality here, so moving to draft while I work it out.

pkg/gateway/server/dispatcher/dispatcher.go
```diff
@@ -224,17 +224,17 @@ func (d *Dispatcher) GetConfiguredAuthProvider(ctx context.Context) (string, err
...
 // Errors are ignored and reported as the auth provider is not configured.
 // Returns: isConfigured (bool)
-func (d *Dispatcher) isAuthProviderConfigured(ctx context.Context, gptscriptClient *gptscript.GPTScript, credCtx []string, toolRef v1.ToolReference) bool {
+func (d *Dispatcher) isAuthProviderConfigured(ctx context.Context, gptscriptClient *gptscript.GPTScript, toolRef v1.ToolReference) bool {
 	if toolRef.Status.Tool == nil {
 		return false
 	}
 
-	cred, err := gptscriptClient.RevealCredential(ctx, credCtx, toolRef.Name)
+	credEnv, err := CredentialEnvForModelProvider(ctx, gptscriptClient, toolRef)
```
g-linville: Why are we calling `CredentialEnvForModelProvider` when we are checking an auth provider? I think we end up using the wrong credential context if we do this.

thedadams:
Ready for review again.

pkg/gateway/server/llmproxy.go
```diff
@@ -241,3 +310,76 @@ func (r *responseModifier) Close() error {
...
+		l.lock.Lock()
+		l.modelProvider = modelProvider
+		l.lock.Unlock()
+	}
+
+	remainingUsage, err := req.GatewayClient.RemainingTokenUsageForUser(req.Context(), req.User.GetUID(), tokenUsageTimePeriod, l.dailyUserTokenPromptTokenLimit, l.dailyUserTokenCompletionTokenLimit)
+	if err != nil {
+		return err
+	} else if !remainingUsage.UnlimitedPromptTokens && remainingUsage.PromptTokens <= 0 || !remainingUsage.UnlimitedCompletionTokens && remainingUsage.CompletionTokens <= 0 {
+		return types2.NewErrHTTP(http.StatusTooManyRequests, fmt.Sprintf("no tokens remaining (prompt tokens remaining: %d, completion tokens remaining: %d)", remainingUsage.PromptTokens, remainingUsage.CompletionTokens))
```
g-linville, line 360: Why do we only error when they are out of both prompt and completion tokens? Shouldn't we error if outOfPrompt || outOfCompletion?

pkg/gateway/server/dispatcher/dispatcher.go
thedadams: Fixed.

pkg/gateway/server/llmproxy.go
thedadams: Based on the logic here, I believe that is what we are doing: (if the user doesn't have unlimited prompt tokens AND they don't have any prompt tokens left) OR (the user doesn't have unlimited completion tokens AND they don't have any completion tokens left)

g-linville: Ah sorry, I misread the code

PR #5380, opened 2025-12-24
Chore: Allow UI to hit remote API for dev

cjellick:
In local development it is useful to be able to hit a remote instance of
the Obot API, like hitting our "main" environment so that you can
development the frontend against a backend with a lot more data.

This change makes that possible. You need to set two environment
variables when running npm run dev:
- `VITE_API_TARGET=https://<hostname>`
- `VITE_API_TOKEN=<token>`

A few notes on those:
- Note that the API target does NOT include /api at the end, just the
  hostname
- The token will be set as the Authorization Bearer token and is used to
  bypass OAuth. You can obtain a token through the Obot CLI with:
`OBOT_BASE_URL=https://<hostname>/api obot token`
Note that you do include the /api when getting a token this way.

### DIFF
--- a/ui/user/src/lib/services/http.ts
+++ b/ui/user/src/lib/services/http.ts
@@ -2,10 +2,27 @@ import { UNAUTHORIZED_PATHS } from '$lib/constants';
 import { profile } from '$lib/stores';
 import errors from '$lib/stores/errors.svelte';
 
+// For SSR, use VITE_API_TARGET if set (for remote API development)
+// For browser, use window.location.origin (requests go through Vite proxy)
+const apiTarget = import.meta.env.VITE_API_TARGET as string | undefined;
 export let baseURL = 'http://localhost:8080/api';
 
 if (typeof window !== 'undefined') {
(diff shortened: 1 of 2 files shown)


ryu-man:
the app is running after obtaining the token but I don't see the data; I'm not able to access some routes 

image

image

and I am logged in as "anonymous"

image

ryu-man:
Closed the console and run dev server again and I do get 'unauthorized' blanc page, `OBOT_BASE_URL=https://<hostname>/api obot token` returns the same token

cjellick:
You are putting the "main" url in when you generate the token?

cjellick:
There's a flag to force the cli to regenerate your token. Assuming your cli is built from the latest commit, try that

cjellick:
Also I'm curious what command you use run the ui

ryu-man:
> Also I'm curious what command you use run the ui

I used `make dev`

ryu-man:
> You are putting the "main" url in when you generate the token?

Yeah, I used the main acornlabs url

Issue #3505, opened 2025-07-23
Shutdown idle MCP Servers


thedadams:
It should be possible to shutdown single-user, multi-user, and agent MCP servers when they are idle. Different configuration values will be available for these three types. Setting them to -1 will disable idle server shutdown for that type.

thedadams:
The work for this has been merged. There are three environment variables that control when single-user, multi-user and agent MCP servers are shutdown: `OBOT_SERVER_SINGLE_USER_IDLE_SERVER_SHUTDOWN_HOURS`, `OBOT_SERVER_MULTI_USER_IDLE_SERVER_SHUTDOWN_HOURS`, `OBOT_SERVER_IDLE_AGENT_SHUTDOWN_HOURS`, respectively. Setting any of those to -1 will disable this feature for that type.

sangee2004:
In my local helm install , I had the following env variables ```--set config.OBOT_SERVER_SINGLE_USER_IDLE_SERVER_SHUTDOWN_HOURS=1  --set config.OBOT_SERVER_MULTI_USER_IDLE_SERVER_SHUTDOWN_HOURS=1 --set config.OBOT_SERVER_IDLE_AGENT_SHUTDOWN_HOURS=1```


Had connected to gmail and antv chart MCP servers and did not use these MCP servers.

After 1 hour of inactivity, I see the MCP server deployments getting terminated as expected
```
%kubectl get pods -n obot-mcp                     
NAME                                   READY   STATUS        RESTARTS   AGE
ms189x5p-f74ff4cc6-qqwvn               2/2     Terminating   0          68m
ms1jvppk-577fcdc86-hsngg               2/2     Running       0          50m
ms1kf6n7-86958f6484-v7bsj              2/2     Terminating   0          68m
ms1nba1s26wl-58b95f59c8-cwf5n          1/1     Running       0          57m
ms1smtwc-7c6ffc7b7f-p7dpp              1/1     Running       0          48m
ms1v8g9q-74477cc8bf-9t5pc              0/1     Terminating   0          67m
sms1obot-mcp-server-6d585f8875-tncc2   2/2     Running       0          21s
sms1obot-mcp-server-7bd9cfbfbd-6nvxt   2/2     Terminating   0          70m
```

After this, I still saw the `Health`  and `Update Status`  continued to be shown as `Available` and `Up to date` in the `Deployments and Connections` tab.

When I tried to click this entry to open the detail page, I think it might have triggered a redeployment.

[...]

PR #5381, opened 2025-12-24
enhance: add idle MCP server shutdown

thedadams:
After this change, idle MCP servers will be automatically shutdown on the configured interval. The activity of each server will be tracked at the granularity of 1/4 of the configured interval or every 15 minutes, whichever is smaller. That is, if the shutdown interval is 30 minutes, the activity will be tracked every 7.5 minutes.

Issue: https://github.com/obot-platform/obot/issues/3505


cjellick:
Does this shutdown shims to, say, remote mcps too or just actual running servers?


Does it shutdown only STDIO servers or would it shutdown a "multi-user" server as well?

I don't think it is, but is this configurable on a per MCP server basis? If I were an ops team running a true multi-user server, I may not want it shutting down like this.

thedadams:
It's clear that I haven't thought through this completely, so I will gather requirements later and implement it at a later time.

Issue #5360, opened 2025-12-19
API Key Authentication for MCP Servers


cjellick:
## Overview
Add first-class API keys that customers can generate in the OBOT UI and use to call MCP servers through OBOT. API keys must be permissioned to specific MCP servers, validated via an OBOT authentication webhook (called by the nanobot shim), and supported as an input to token exchange

## Problem Statement
Currently, there's no way for users to authenticate external applications/scripts to access MCP servers without going through the standard OAuth flow. Users need a simple token-based authentication method for programmatic access.

### 1. API Key Generation (UI)
- Users can generate API keys through the UI
- Users can specify which MCP servers each API key has access to
- Users can view, revoke, and manage their API keys
- API keys should have appropriate metadata (name, creation date, last used, etc.)
- Admins should be able to manage and revoke API keys across all users.

### 2. Authentication Webhook
- Implement an authentication webhook endpoint in Obot
- MCP servers (via nanobot shim) will call this webhook with the API key token
- Webhook validates the token and returns authentication information about the user
- Similar to Kubernetes authentication webhook pattern
- Response should include user identity information
- **Unknown from Craig** - does the authenticating webhook handle authorization? I get that it does authn (return this user info for that API key) but does handle authz (indicating whether the api key/user is allowed to hit the server)

### 3. Token Exchange Modification
**Current state:** Token exchange uses JWT tokens

**Required changes:**
- Support API keys as the token type for exchange
- When an API key is presented for token exchange:
  - Use the client ID and secret to identify which MCP server is requesting the exchange
  - Verify that the API key has access to that specific MCP server ***Again, I'm not 100% if this is where the authz happens or not***
  - Only complete the token exchange if authorization is valid
- Maintain backward compatibility with existing JWT-based token exchange

### Critical Dependencies
All three components (API key generation, authentication webhook, and token exchange) must be implemented together for this feature to work. They are interdependent and cannot function independently.

### Component Interaction Flow
1. External client presents API key to MCP server
2. Nanobot shim intercepts request and calls Obot authentication webhook with API key
3. Obot validates API key and returns user authentication info
4. Nanobot shim r performs token exchange using the API key
  - ***Clarification from Craig*** - I believe its the shim and not the server that does the exchange. Please confirm.
6. Obot validates the API key has access to the requesting MCP server (identified by client ID/secret)
7. Token exchange completes and MCP server can process the request

### Security Considerations
- API keys should be treated as sensitive credentials
[...]

PR #5389, opened 2025-12-29
feat: add API keys

g-linville:
This feature allows users to generate API keys that they can use to connect to MCP servers by passing a header `Authorization: Bearer <API key>`.

Each key has the format `ok1-<user ID>-<key ID>-<32 byte secret>`.

The keys are stored as bcrypt hashes in the database.

An expiration date can be set, but it is optional.

Admins can see and delete keys for other users.

Audit logs track API keys that were used for particular MCP calls.

The only API besides mcp-connect routes that API keys can hit is `/api/me`. Everything else is forbidden.

for: https://github.com/obot-platform/obot/issues/5360

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -135,7 +135,7 @@ config:
   # config.OBOT_SERVER_MCPBASE_IMAGE -- Deploy MCP servers in the cluster using this base image. OBOT_SERVER_MCPNAMESPACE is automatically added to the secret if config.OBOT_SERVER_MCPBASE_IMAGE is set.
   OBOT_SERVER_MCPBASE_IMAGE: "ghcr.io/obot-platform/mcp-images/phat:main"
   # config.OBOT_SERVER_MCPREMOTE_SHIM_BASE_IMAGE -- Deploy MCP remote shim servers in the cluster using this base image.
-  OBOT_SERVER_MCPREMOTE_SHIM_BASE_IMAGE: "ghcr.io/nanobot-ai/nanobot:v0.0.47"
+  OBOT_SERVER_MCPREMOTE_SHIM_BASE_IMAGE: "ghcr.io/nanobot-ai/nanobot:v0.0.49"
   # config.OBOT_SERVER_MCPHTTPWEBHOOK_BASE_IMAGE -- Deploy MCP HTTP webhook servers in the cluster using this base image.
   OBOT_SERVER_MCPHTTPWEBHOOK_BASE_IMAGE: "ghcr.io/obot-platform/mcp-images/http-webhook-mcp-converter:main"
   # config.OBOT_SERVER_MCPCLUSTER_DOMAIN -- The cluster domain to use for MCP services. Defaults to cluster.local. Only matters if the above image is set.
--- a/docs/docs/configuration/server-configuration.md
+++ b/docs/docs/configuration/server-configuration.md
@@ -34,7 +34,7 @@ The Obot server is configured via environment variables. The following configura
 | `OBOT_SERVER_AUDIT_LOGS_COMPRESS_FILE` | Controls whether or not to compress audit log files | `true` |
 | `OBOT_SERVER_AUDIT_LOGS_USE_PATH_STYLE` | Whether to use path style for S3 | - |
 | `OBOT_SERVER_MCPBASE_IMAGE` | Deploy MCP servers in the kubernetes cluster or using docker with this base image. | `ghcr.io/obot-platform/mcp-images/phat:main` |
-| `OBOT_SERVER_MCPREMOTE_SHIM_BASE_IMAGE` | Deploy MCP remote shim servers in the cluster using this base image. | `ghcr.io/nanobot-ai/nanobot:v0.0.47` |
+| `OBOT_SERVER_MCPREMOTE_SHIM_BASE_IMAGE` | Deploy MCP remote shim servers in the cluster using this base image. | `ghcr.io/nanobot-ai/nanobot:v0.0.49` |
 | `OBOT_SERVER_MCPHTTPWEBHOOK_BASE_IMAGE` | Deploy MCP HTTP webhook servers in the cluster using this base image. | `ghcr.io/obot-platform/mcp-images/http-webhook-mcp-converter:main` |
 | `OBOT_SERVER_MCPRUNTIME_BACKEND` | The runtime backend to use for running MCP servers: docker, kubernetes, or local. | `kubernetes` in the helm chart, `docker` otherwise |
 | `OBOT_SERVER_MCPCLUSTER_DOMAIN` | The cluster domain to use for MCP services. Only matters if `OBOT_SERVER_MCPBASE_IMAGE` is set. | `cluster.local` |
--- a/docs/versioned_docs/version-v0.15.0/configuration/server-configuration.md
+++ b/docs/versioned_docs/version-v0.15.0/configuration/server-configuration.md
@@ -34,7 +34,7 @@ The Obot server is configured via environment variables. The following configura
 | `OBOT_SERVER_AUDIT_LOGS_COMPRESS_FILE` | Controls whether or not to compress audit log files | `true` |
 | `OBOT_SERVER_AUDIT_LOGS_USE_PATH_STYLE` | Whether to use path style for S3 | - |
 | `OBOT_SERVER_MCPBASE_IMAGE` | Deploy MCP servers in the kubernetes cluster or using docker with this base image. | `ghcr.io/obot-platform/mcp-images/phat:main` |
-| `OBOT_SERVER_MCPREMOTE_SHIM_BASE_IMAGE` | Deploy MCP remote shim servers in the cluster using this base image. | `ghcr.io/nanobot-ai/nanobot:v0.0.47` |
+| `OBOT_SERVER_MCPREMOTE_SHIM_BASE_IMAGE` | Deploy MCP remote shim servers in the cluster using this base image. | `ghcr.io/nanobot-ai/nanobot:v0.0.49` |
 | `OBOT_SERVER_MCPHTTPWEBHOOK_BASE_IMAGE` | Deploy MCP HTTP webhook servers in the cluster using this base image. | `ghcr.io/obot-platform/mcp-images/http-webhook-mcp-converter:main` |
 | `OBOT_SERVER_MCPRUNTIME_BACKEND` | The runtime backend to use for running MCP servers: docker, kubernetes, or local. | `kubernetes` in the helm chart, `docker` otherwise |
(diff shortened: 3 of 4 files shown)


pkg/mcp/docker.go
```diff
@@ -701,6 +699,8 @@ func (d *dockerBackend) createAndStartContainer(ctx context.Context, server Serv
 				"NANOBOT_RUN_OAUTH_SCOPES=profile",
 				"NANOBOT_RUN_FORCE_FETCH_TOOL_LIST=true",
 				"NANOBOT_DISABLE_HEALTH_CHECKER=true",
+				"NANOBOT_RUN_API_KEY_AUTH_URL=" + localhostURLRegexp.ReplaceAllString(server.Issuer+"/api/api-keys/auth", d.hostBaseURLWithPort),
+				"MCP_SERVER_ID=" + strings.TrimSuffix(server.MCPServerName, "-shim"),
```
g-linville: This variable is for use by the mcp-oauth-proxy library that Nanobot uses. Let me know if I should add a prefix to the front of the variable name.

thedadams: I commented on the other PR, but I think it makes sense for this to be nanobot configuration that is passed down to `mcp-oauth-proxy`. That would mean that it would have a `NANOBOT_RUN_` prefix like the others.

apiclient/types/mcpauditlog.go
```diff
@@ -10,6 +10,7 @@ type MCPAuditLog struct {
 	CreatedAt                 Time            `json:"createdAt"`
 	UserID                    string          `json:"userID"`
 	MCPID                     string          `json:"mcpID"`
+	APIKey                    string          `json:"apiKey"`
```
thedadams, line 13: Are we storing the API key used for these requests in the audit logs?

pkg/api/authz/authz.go
```diff
@@ -267,6 +276,19 @@ var (
...
+			"POST /api/api-keys",
+			"GET /api/api-keys",
+			"GET /api/api-keys/{id}",
+			"DELETE /api/api-keys/{id}",
+		},
+
+		// API key users have restricted access - they can only access MCP-connect routes and /api/me
+		// They get access to anyGroup routes automatically (health checks, OAuth flows, etc.)
+		types.GroupAPIKey: {
+			"GET /api/me",
```
thedadams, line 290: Why is this permission required?

pkg/api/handlers/mcpgateway/oauth/token.go
```diff
@@ -31,6 +32,7 @@ const (
 	tokenExpiration         = 10 * time.Minute
 	tokenTypeJWT            = "urn:ietf:params:oauth:token-type:jwt"
 	tokenTypeAccessToken    = "urn:ietf:params:oauth:token-type:access_token"
+	tokenTypeAPIKey         = "urn:obot:token-type:api-key"
```
thedadams, line 35: Should this be `rn:ietf:params:oauth:token-type:api-key`?

pkg/api/handlers/mcpgateway/oauth/token.go
```diff
@@ -438,25 +481,72 @@ func (h *handler) doTokenExchange(req api.Context, oauthClient v1.OAuthClient, r
 					}
 				}
 
-				tokenCtx.MCPID = componentMCPID
-				tokenCtx.Audience = fmt.Sprintf("%s/mcp-connect/%s", h.baseURL, audienceID)
+				if subjectTokenType == tokenTypeAPIKey {
+					// For API keys, we need to create a JWT for the component server
+					// since component servers may not support API key authentication yet
```
thedadams: We control the component servers, so they could support API key authentication.

I think it makes sense for an API key that is used for a composite server to also work for the components.

pkg/gateway/client/apikey.go
```diff
@@ -0,0 +1,183 @@
...
+
+		// Check expiration
+		if apiKey.ExpiresAt != nil && apiKey.ExpiresAt.Before(time.Now()) {
+			return fmt.Errorf("API key has expired")
+		}
+
+		// Update last used timestamp
+		now := time.Now()
+		apiKey.LastUsedAt = &now
+		return tx.Model(&apiKey).Update("last_used_at", now).Error
```
thedadams: Does it make sense to update this at a courser granularity?

apiclient/types/mcpauditlog.go
g-linville: We're storing a portion of it. Maybe I should rename this to `APIKeyPrefix` or something. Just the `ok1-<user ID>-<key ID>` portion.

pkg/api/authz/authz.go
g-linville: It's not required, but I figured it would give us an easy way to test whether an API key is working. Just `curl <host>/api/me -H "Authorization: Bearer <key>"`

pkg/api/handlers/mcpgateway/oauth/token.go
g-linville: AI told me to do a custom type here. I'll do a little bit more research into this and figure out what would be best.

pkg/api/handlers/mcpgateway/oauth/token.go
g-linville: Okay, I can do that

pkg/gateway/client/apikey.go
g-linville: Not sure, I will ask Craig to see what he wants

g-linville: we're gonna do it to the nearest minute

pkg/api/handlers/mcpgateway/oauth/token.go
g-linville: So `urn:ietf:params:oauth:token-type:api-key` is not an officially defined type, as far as I can tell. Google returns no results for it, and it is not listed in [the RFC](https://datatracker.ietf.org/doc/html/rfc8693#section-3). Let me know if you think we should use it anyway.

pkg/mcp/docker.go
g-linville: fixed

pkg/api/handlers/mcpgateway/oauth/token.go
g-linville: fixed

pkg/gateway/client/apikey.go
g-linville: fixed

pkg/gateway/client/apikey.go
```diff
@@ -0,0 +1,192 @@
...
+
+// DeleteAPIKeyByID removes an API key by ID without user filtering (for admin use).
+func (c *Client) DeleteAPIKeyByID(ctx context.Context, keyID uint) error {
+	result := c.db.WithContext(ctx).Where("id = ?", keyID).Delete(&types.APIKey{})
+	if result.Error != nil {
+		return fmt.Errorf("failed to delete API key: %w", result.Error)
+	}
+	if result.RowsAffected == 0 {
+		return gorm.ErrRecordNotFound
+	}
```
thedadams: nit: generally, making delete idempotent is preferable.

pkg/gateway/server/apikey.go
```diff
@@ -0,0 +1,356 @@
...
+			return types2.NewErrBadRequest("MCP server %q not found", serverID)
+		}
+
+		// Check if user has access to this server
+		hasAccess, err := s.userHasAccessToMCPServer(apiContext, &server)
+		if err != nil {
+			return types2.NewErrHTTP(http.StatusInternalServerError, fmt.Sprintf("failed to check access to MCP server: %v", err))
+		}
+		if !hasAccess {
+			return types2.NewErrBadRequest("MCP server %q not found", serverID)
```
thedadams: Do we want to check all of the `MCPServerIDs` before returning?

pkg/gateway/server/apikey.go
```diff
@@ -0,0 +1,356 @@
...
+	if err != nil {
+		return types2.NewErrHTTP(http.StatusInternalServerError, fmt.Sprintf("failed to create API key: %v", err))
+	}
+
+	return apiContext.WriteCreated(response)
+}
+
+// userHasAccessToMCPServer checks if the current user has access to the given MCPServer.
+func (s *Server) userHasAccessToMCPServer(apiContext api.Context, server *v1.MCPServer) (bool, error) {
+	userID := strconv.FormatUint(uint64(apiContext.UserID()), 10)
```
thedadams:
```suggestion
	userID := apiContext.User.GetUID()
```

pkg/gateway/server/apikey.go
```diff
@@ -0,0 +1,356 @@
...
+	authHeader := apiContext.Request.Header.Get("Authorization")
+	if authHeader == "" {
+		return apiContext.Write(apiKeyAuthResponse{
+			Allowed: false,
+			Reason:  "missing Authorization header",
+		})
+	}
+
+	bearer := strings.TrimPrefix(authHeader, "Bearer ")
+	if bearer == authHeader || !strings.HasPrefix(bearer, "ok1-") {
```
thedadams: nit: I feel this is slightly more understandable. Ignore if you disagree.
```suggestion
	bearer, ok := strings.CutPrefix(authHeader, "Bearer ")
	if !ok || !strings.HasPrefix(bearer, "ok1-") {
```

pkg/gateway/server/apikey.go
```diff
@@ -0,0 +1,356 @@
...
+		})
+	}
+
+	// Check scope restrictions - an MCP server must be specified
+	if !system.IsMCPServerID(req.MCPID) {
+		return apiContext.Write(apiKeyAuthResponse{
+			Allowed: false,
+			Reason:  "bad request: no MCP server was specified",
+		})
+	}
```
thedadams: nit: I think we can safely remove this check. Other checks will catch it and we have been talking about doing "deployment per catalog entry" which will break this.

thedadams (approved):

pkg/gateway/client/apikey.go
g-linville: fixed

pkg/gateway/server/apikey.go
g-linville: fixed

pkg/gateway/server/apikey.go
g-linville: fixed

pkg/gateway/server/apikey.go
g-linville: fixed

pkg/gateway/server/apikey.go
g-linville: fixed

thedadams (approved):

PR #5377, opened 2025-12-23
feat: introduce Model Permission Rules

cjellick:
Backend:
  - Add modelpermissionrule.Helper for checking user access to models
  - Enforce permissions in LLM proxy - return 403 if user lacks access
  - Filter allowedModels in assistants API based on user permissions
  - Return empty model in default model endpoints if user lacks permission

Frontend:
  - Add fallback model logic when default model returns empty
  - Show first allowed model instead of "No Default Model"
  - Display model selector when user has allowedModels even without default
  - Use fallback model when creating threads if default is unavailable

Allows users without default model access to chat using other permitted models.

Signed-off-by: Craig Jellick <craig@obot.ai>

Issue #5283, opened 2025-12-12
Allow assignment of available models to users/groups


cloudnautique:
In an organization you want the ability to control cost when using models. Some users might benefit from using the most advanced and expensive models, while others only need fast text generation or access to a preconfigured agent+model combination.

njhale:
hold off on testing this. have one more change to put in.

right now admins/owners have implicit access to all models. this needs to change to match the "MCP Registries" behavior, where an admin/user needs to be granted access with a ModelAccessPolicy to use the model.

sangee2004:
Tested with the latest build from main

Admin/Owners can create `Model Access Policies` now that will allow users/groups to get access to models that can be used when chatting.
All models the user has access to will be available as a drop-down option in the chatwell, so the user can choose from them.

`Allowed Models` that we were able to set previously at the `Chat Configuration` level are not available anymore.

Issues found when testing this feature are logged as separate issues:

https://github.com/obot-platform/obot/issues/5436
https://github.com/obot-platform/obot/issues/5435
https://github.com/obot-platform/obot/issues/5434
https://github.com/obot-platform/obot/issues/5433
https://github.com/obot-platform/obot/issues/5447

PR #5393, opened 2025-12-30
feat: introduce model access policies

njhale:
Backend:
  - Add modelaccesspolicy.Helper for checking user access to models
  - Enforce permissions in LLM proxy - return 403 if user lacks access
  - Generate allowedModels in assistants API based on user permissions
  - Filter list models API with MAPs
  - Migrate chat permissions and default model to Model Access
    Policies
  - Return empty model in default model endpoints if user lacks permission
  - Generate default model policy on first startup pointing to default model aliases (they will always track the effective target model)

Frontend:
  - Add fallback model logic when default model returns empty
  - Show first allowed model instead of "No Default Model"
  - Display model selector when user has allowedModels even without default
  - Use fallback model when creating threads if default is unavailable

Addresses https://github.com/obot-platform/obot/issues/5283

Supersedes https://github.com/obot-platform/obot/pull/5377

### DIFF
--- a/pkg/api/handlers/assistants.go
+++ b/pkg/api/handlers/assistants.go
@@ -16,7 +16,6 @@ import (
 	"github.com/obot-platform/obot/pkg/gateway/server/dispatcher"
 	"github.com/obot-platform/obot/pkg/invoke"
 	"github.com/obot-platform/obot/pkg/mcp"
-	"github.com/obot-platform/obot/pkg/modelaccesspolicy"
 	"github.com/obot-platform/obot/pkg/projects"
 	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
 	"github.com/obot-platform/obot/pkg/system"
@@ -33,17 +32,15 @@ type AssistantHandler struct {
 	events            *events.Emitter
 	dispatcher        *dispatcher.Dispatcher
 	cachedClient      kclient.WithWatch
-	mapHelper         *modelaccesspolicy.Helper
 }
 
-func NewAssistantHandler(dispatcher *dispatcher.Dispatcher, mcpSessionManager *mcp.SessionManager, invoker *invoke.Invoker, events *events.Emitter, cachedClient kclient.WithWatch, mapHelper *modelaccesspolicy.Helper) *AssistantHandler {
+func NewAssistantHandler(dispatcher *dispatcher.Dispatcher, mcpSessionManager *mcp.SessionManager, invoker *invoke.Invoker, events *events.Emitter, cachedClient kclient.WithWatch) *AssistantHandler {
 	return &AssistantHandler{
 		invoker:           invoker,
 		mcpSessionManager: mcpSessionManager,
 		events:            events,
 		dispatcher:        dispatcher,
 		cachedClient:      cachedClient,
-		mapHelper:         mapHelper,
 	}
 }
 
--- a/pkg/controller/handlers/modelaccesspolicy/migrate.go
+++ b/pkg/controller/handlers/modelaccesspolicy/migrate.go
@@ -42,7 +42,7 @@ func MigrateAgentAllowedModels(req router.Request, _ router.Response) error {
 		policyModels = convertToModelResources(agent.Spec.Manifest.AllowedModels)
 	)
 
-	// Make sure there are actually policys to copy before attempting to avoid producing an invalid MAP.
+	// Make sure there are actually policies to copy before attempting to avoid producing an invalid MAP.
 	if len(policyModels) > 0 {
 		if err := req.Get(&policy, agent.Namespace, policyName); err == nil {
 			// Rule already exists, determine if we should add new models.
--- a/pkg/api/router/router.go
+++ b/pkg/api/router/router.go
@@ -34,7 +34,7 @@ func Router(ctx context.Context, services *services.Services) (http.Handler, err
 	oauthChecker := oauth.NewMCPOAuthHandlerFactory(services.ServerURL, services.MCPLoader, services.StorageClient, services.GPTClient, services.GatewayClient, services.MCPOAuthTokenStorage)
 
 	agents := handlers.NewAgentHandler(services.ProviderDispatcher, services.MCPLoader, services.Invoker, services.ServerURL, services.InternalServerURL)
-	assistants := handlers.NewAssistantHandler(services.ProviderDispatcher, services.MCPLoader, services.Invoker, services.Events, services.Router.Backend(), services.ModelAccessPolicyHelper)
+	assistants := handlers.NewAssistantHandler(services.ProviderDispatcher, services.MCPLoader, services.Invoker, services.Events, services.Router.Backend())
 	tools := handlers.NewToolHandler(services.Invoker)
 	tasks := handlers.NewTaskHandler(services.Invoker, services.Events)
 	invoker := handlers.NewInvokeHandler(services.Invoker, services.MCPLoader)
--- a/ui/user/src/lib/components/admin/ModelAccessPolicyForm.svelte
+++ b/ui/user/src/lib/components/admin/ModelAccessPolicyForm.svelte
@@ -8,6 +8,8 @@
 		type AccessControlRuleSubject,
 		type OrgUser,
 		type OrgGroup,
+		ModelUsage,
+		ModelUsageLabels,
 		ModelAlias,
 		ModelAliasLabels,
 		type DefaultModelAlias
@@ -107,6 +109,7 @@
 				id: resource.id,
 				aliasName: aliasName,
 				aliasLabel: ModelAliasLabels[aliasName as ModelAlias] || aliasName,
+				usage: model?.usage,
 				effectiveModelName: model?.displayName || model?.targetModel || 'Not configured',
 				isConfigured: !!model
 			};
@@ -133,7 +136,8 @@
 				provider: effectiveModel?.modelProviderName || '-',
 				effectiveModel: alias.effectiveModelName,
 				isAlias: true,
-				isConfigured: alias.isConfigured
+				isConfigured: alias.isConfigured,
+				usage: alias.usage
 			};
 		});
 
@@ -142,6 +146,7 @@
 			aliasName: undefined,
 			name: model.name,
 			provider: model.provider,
+			usage: model.usage,
 			effectiveModel: null,
 			isAlias: false,
 			isConfigured: true
@@ -251,6 +256,7 @@
 			return {
 				id: model.id,
 				name: m?.displayName || m?.name || model.id,
+				usage: m?.usage,
 				provider: m?.modelProviderName || '-'
 			};
 		});
@@ -396,15 +402,27 @@
 					noDataMessage="No models added."
 				>
 					{#snippet onRenderColumn(field, d)}
-						{#if field === 'name' && d.isAlias}
-							<div class="flex items-center gap-2">
+						{#if field === 'name'}
+							{#if d.isAlias}
 								<div class="flex flex-col">
-									<span class="font-medium">{d.aliasName}</span>
-									<span class="text-on-surface1 text-xs" class:text-yellow-500={!d.isConfigured}>
-										{d.effectiveModel}
-									</span>
+									<div class="flex items-center gap-2">
+										<span class="font-medium">{d.aliasName}</span>
+										<span class="text-on-surface1 text-xs" class:text-yellow-500={!d.isConfigured}>
+											{d.effectiveModel}
+										</span>
+									</div>
+									<span class="text-on-surface1 text-xs">{d.name}</span>
 								</div>
-							</div>
+							{:else}
+								<div class="flex flex-col">
+									<span class="font-medium">{d.name}</span>
+									{#if d.usage}
+										<span class="text-on-surface1 text-xs">
+											{ModelUsageLabels[d.usage as ModelUsage] || d.usage}
+										</span>
+									{/if}
+								</div>
+							{/if}
 						{:else}
 							{d[field as keyof typeof d]}
 						{/if}
--- a/ui/user/src/lib/components/admin/SearchModels.svelte
+++ b/ui/user/src/lib/components/admin/SearchModels.svelte
@@ -206,12 +206,16 @@
 								onclick={() => toggleSelection(alias.id)}
 							>
 								<div class="flex flex-col text-left">
-									<span>{alias.aliasName}</span>
-									<span class="text-on-surface1 text-xs">
-										<span class={alias.isConfigured ? '' : 'text-yellow-500'}>
+									<div class="flex items-center gap-2">
+										<span>{alias.aliasName}</span>
+										<span
+											class="text-on-surface1 text-xs"
+											class:text-yellow-500={!alias.isConfigured}
+										>
 											{alias.effectiveModelName}
 										</span>
-									</span>
+									</div>
+									<span class="text-on-surface1 text-xs">{alias.label}</span>
 								</div>
 								{#if selectedSet.has(alias.id)}
 									<Check class="text-primary size-4" />


pkg/api/handlers/assistants.go
```diff
@@ -133,16 +152,73 @@ func (a *AssistantHandler) List(req api.Context) error {
...
+	// If no rules apply to the user, they have no permissions - return empty list
+	if len(userAllowedModels) == 0 {
+		return []string{}
+	}
+
+	// Create a set for O(1) lookup
+	allowedSet := make(map[string]struct{}, len(userAllowedModels))
+	for _, modelID := range userAllowedModels {
+		allowedSet[modelID] = struct{}{}
+	}
```
g-linville: pico nit: renaming `allowedSet` to `userAllowedSet` might be nice for consistency

pkg/gateway/server/llmproxy.go
```diff
@@ -97,10 +101,46 @@ func (s *Server) dispatchLLMProxy(req api.Context) error {
...
+		}
+		modelID = m.Name
+
 		credEnv = cred.Env
 		personalToken = true
 	}
 
+	// Check if the user has permission to use this model
+	if s.mprHelper != nil && modelID != "" && token.UserID != "" {
+		userID, err := strconv.ParseUint(token.UserID, 10, 64)
```
g-linville, line 112: I feel like we should return this error? Is there a reason we're not?

pkg/modelpermissionrule/helper.go
```diff
@@ -0,0 +1,294 @@
...
+	// Check rules with wildcard subject selector (*) that include the model
+	wildcardSubjectRules, err := h.GetModelPermissionRulesWithWildcardSubject(system.DefaultNamespace)
+	if err != nil {
+		return false, err
+	}
+
+	for _, rule := range wildcardSubjectRules {
+		// Check if any subject is a wildcard selector
+		for _, subject := range rule.Spec.Manifest.Subjects {
+			if subject.Type == types.SubjectTypeSelector && subject.ID == "*" {
```
g-linville: Isn't this check redundant?

pkg/modelpermissionrule/helper.go
```diff
@@ -0,0 +1,294 @@
...
+	// Check rules that the user is directly included in
+	userRules, err := h.GetModelPermissionRulesForUser(system.DefaultNamespace, userID)
+	if err != nil {
+		return false, err
+	}
+
+	for _, rule := range userRules {
+		// Check if user is a subject in this rule
+		for _, subject := range rule.Spec.Manifest.Subjects {
+			if subject.Type == types.SubjectTypeUser && subject.ID == userID {
```
g-linville: And here too

pkg/modelpermissionrule/helper.go
```diff
@@ -0,0 +1,294 @@
...
+	// Check rules based on group membership
+	for groupID := range groups {
+		groupRules, err := h.getModelPermissionRulesForGroup(system.DefaultNamespace, groupID)
+		if err != nil {
+			return false, err
+		}
+
+		for _, rule := range groupRules {
+			for _, subject := range rule.Spec.Manifest.Subjects {
+				if subject.Type == types.SubjectTypeGroup && subject.ID == groupID {
```
g-linville: And here

pkg/modelpermissionrule/helper.go
```diff
@@ -0,0 +1,294 @@
...
+
+	// Check rules with wildcard subject selector (*)
+	wildcardSubjectRules, err := h.GetModelPermissionRulesWithWildcardSubject(system.DefaultNamespace)
+	if err != nil {
+		return nil, false, err
+	}
+
+	for _, rule := range wildcardSubjectRules {
+		for _, subject := range rule.Spec.Manifest.Subjects {
+			if subject.Type == types.SubjectTypeSelector && subject.ID == "*" {
```
g-linville: And here

pkg/modelpermissionrule/helper.go
```diff
@@ -0,0 +1,294 @@
...
+
+	// Check rules that the user is directly included in
+	userRules, err := h.GetModelPermissionRulesForUser(system.DefaultNamespace, userID)
+	if err != nil {
+		return nil, false, err
+	}
+
+	for _, rule := range userRules {
+		for _, subject := range rule.Spec.Manifest.Subjects {
+			if subject.Type == types.SubjectTypeUser && subject.ID == userID {
```
g-linville: And here

pkg/modelpermissionrule/helper.go
```diff
@@ -0,0 +1,294 @@
...
+	// Check rules based on group membership
+	for groupID := range groups {
+		groupRules, err := h.getModelPermissionRulesForGroup(system.DefaultNamespace, groupID)
+		if err != nil {
+			return nil, false, err
+		}
+
+		for _, rule := range groupRules {
+			for _, subject := range rule.Spec.Manifest.Subjects {
+				if subject.Type == types.SubjectTypeGroup && subject.ID == groupID {
```
g-linville: And here

pkg/api/handlers/assistants.go
```diff
@@ -133,16 +143,57 @@ func (a *AssistantHandler) List(req api.Context) error {
...
+		return nil, err
+	}
+
+	allowedModels = make([]string, 0, len(models.Items))
+	for _, model := range models.Items {
+		allowedModels = append(allowedModels, model.Name)
+	}
+
+	return allowedModels, nil
+}
```
njhale: This helper feels a bit clunky to me. Any thoughts? Maybe list in `GetAllowedModelsForRequest`?

pkg/gateway/server/llmproxy.go
```diff
@@ -186,7 +224,7 @@ func getModelProviderForModel(ctx context.Context, client kclient.Client, namesp
 		return respModel, nil
 	}
 
-	return nil, fmt.Errorf("model %q not found", model)
+	return nil, fmt.Errorf("model %q not found", modelReference)
 }
```
njhale, line 229: The original name of this function looks like it was probably a typo. Renamed and commented for clarity.

pkg/gateway/server/llmproxy.go
njhale: done

njhale:
I discussed a few additional changes with @cjellick that need to go in before this can merge:
1. Drop allowedModels from the assistant API and frontend entirely (either UI or backend will filter thread model selector to `usage = 11m`).
2. Add a removable `ModelAccessPolicy` that selects `Everyone` and tracks the current default models set in `DefaultModelAliases`
3. Lock thread chat box when no models are available

pkg/storage/apis/obot.obot.ai/v1/modelpermissionrule.go
```diff
@@ -0,0 +1,56 @@
...
+	_ fields.Fields = (*ModelAccessPolicy)(nil)
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type ModelAccessPolicy struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec ModelAccessPolicySpec `json:"spec,omitempty"`
```
thedadams: nit: other objects have an empty status instead of none.

pkg/storage/apis/obot.obot.ai/v1/modelpermissionrule.go
```diff
@@ -0,0 +1,56 @@
...
+	return slices.Contains(in.FieldNames(), field)
+}
+
+func (in *ModelAccessPolicy) Get(_ string) (value string) {
+	return ""
+}
+
+func (in *ModelAccessPolicy) FieldNames() []string {
+	return []string{}
+}
```
thedadams:
```suggestion
```

apiclient/types/modelaccesspolicy.go
```diff
@@ -0,0 +1,82 @@
...
+
+	models := make(map[ModelResource]struct{}, len(m.Models))
+	for _, model := range m.Models {
+		if err := model.Validate(); err != nil {
+			return fmt.Errorf("invalid model: %w", err)
+		}
+
+		if model.ID == "*" && len(m.Models) > 1 {
+			return fmt.Errorf("wildcard model (*) must be the only model")
+		}
```
thedadams: This can be in the `Validate` method.

pkg/api/handlers/assistants.go
```diff
@@ -133,16 +142,58 @@ func (a *AssistantHandler) List(req api.Context) error {
...
+	// Fetch the available chat models
+	var modelList v1.ModelList
+	if err := req.List(&modelList, &kclient.ListOptions{
+		FieldSelector: fields.SelectorFromSet(map[string]string{
+			"spec.manifest.usage": string(types.ModelUsageLLM),
+		}),
+		Namespace: req.Namespace(),
+	}); err != nil {
+		return assistant, err
+	}
```
thedadams: Can we only do this if `allowAll == true`?

pkg/api/handlers/threads.go
```diff
@@ -330,6 +333,9 @@ func (a *ThreadHandler) GetDefaultModelForThread(req api.Context) error {
 		model = alias.Spec.Manifest.Model
 	}
 
+	// Track the model ID before resolving to check permissions
+	modelID := model
+
 	if strings.HasPrefix(model, system.ModelPrefix) {
```
thedadams: Since you changed this in another place, can you change it here, too?
```suggestion
	if system.IsModelID(model) {
```

pkg/modelaccesspolicy/helper.go
```diff
@@ -0,0 +1,194 @@
...
+
+// userIsAdminOrOwner checks if the user is an admin or owner.
+func userIsAdminOrOwner(user kuser.Info) bool {
+	for _, group := range user.GetGroups() {
+		switch group {
+		case types.GroupAdmin, types.GroupOwner:
+			return true
+		}
+	}
+	return false
```
thedadams: This is enough.
```suggestion
	return slices.Contains(user.GetGroups(), types.GroupAdmin)
```

pkg/gateway/server/llmproxy.go
```diff
@@ -353,6 +412,53 @@ func (l *llmProviderProxy) proxy(req api.Context) error {
 		l.lock.Unlock()
 	}
 
+	// Reject chat completion and responses API requests that target models the user doesn't have access to
+	if path := req.URL.Path; req.Method == http.MethodPost &&
```
thedadams: Instead of checking the path, I think it makes sense to just look for a `model` in the request body. If one exists, then check it. If not, then just forward the request because it will fail if a model is expected.

My reasoning here is that this is supposed to work regardless of the model provider and the paths could be different. For example, for Anthropic, the path would end with `/messages`.

pkg/gateway/server/llmproxy.go
```diff
@@ -353,6 +412,53 @@ func (l *llmProviderProxy) proxy(req api.Context) error {
...
+		}
+
+		// Check if the user has been granted access to the model
+		hasAccess, err := l.mapHelper.UserHasAccessToModel(req.User, modelID)
+		if err != nil {
+			return fmt.Errorf("failed to check user access to model %q: %w", targetModel, err)
+		}
+		if !hasAccess {
+			return types2.NewErrForbidden("user does not have permission to use model %q", targetModel)
+		}
```
thedadams: Does it make sense to check all models returned instead of the first one and continue of any of them are allowed?

pkg/api/handlers/assistants.go
njhale: I'm working on a change that will use the user permission filtered `GET /api/models` endpoint to drive the UX, so this code (and the `AllowedModels` field on `types.Assistant`) is going away shortly anyway.

pkg/gateway/server/llmproxy.go
njhale: I had the same thought. Will do.

pkg/gateway/server/llmproxy.go
njhale: Although I don't think there's currently a case where we'll have more than one model per provider with the same targetName, your suggestion will make this more robust.

I'll do that.

pkg/modelaccesspolicy/helper.go
njhale: 🤦 thanks

thedadams:
I tested out the UX and things look good to me there, too!

pkg/gateway/server/llmproxy.go
njhale: done

pkg/gateway/server/llmproxy.go
njhale: done

pkg/modelaccesspolicy/helper.go
njhale: fixed

pkg/api/handlers/threads.go
njhale: done

pkg/api/handlers/assistants.go
njhale: this change is in

apiclient/types/modelaccesspolicy.go
njhale: the Validate method for `ModelResource`?

pkg/storage/apis/obot.obot.ai/v1/modelpermissionrule.go
njhale: Done

pkg/storage/apis/obot.obot.ai/v1/modelpermissionrule.go
njhale: Done

pkg/gateway/server/llmproxy.go
thedadams: It was not a typo.

pkg/api/handlers/assistants.go
```diff
@@ -32,15 +33,17 @@ type AssistantHandler struct {
...
 
-func NewAssistantHandler(dispatcher *dispatcher.Dispatcher, mcpSessionManager *mcp.SessionManager, invoker *invoke.Invoker, events *events.Emitter, cachedClient kclient.WithWatch) *AssistantHandler {
+func NewAssistantHandler(dispatcher *dispatcher.Dispatcher, mcpSessionManager *mcp.SessionManager, invoker *invoke.Invoker, events *events.Emitter, cachedClient kclient.WithWatch, mapHelper *modelaccesspolicy.Helper) *AssistantHandler {
 	return &AssistantHandler{
 		invoker:           invoker,
 		mcpSessionManager: mcpSessionManager,
 		events:            events,
 		dispatcher:        dispatcher,
 		cachedClient:      cachedClient,
+		mapHelper:         mapHelper,
```
thedadams: Looks like this is unused now.

apiclient/types/modelaccesspolicy.go
```diff
@@ -0,0 +1,98 @@
...
+
+	models := make(map[ModelResource]struct{}, len(m.Models))
+	for _, model := range m.Models {
+		if err := model.Validate(); err != nil {
+			return fmt.Errorf("invalid model: %w", err)
+		}
+
+		if model.IsWildcard() && len(m.Models) > 1 {
+			return fmt.Errorf("wildcard model (*) must be the only model")
+		}
```
thedadams, line 54: Why isn't this in the `Validate` method?

pkg/controller/handlers/modelaccesspolicy/migrate.go
```diff
@@ -0,0 +1,149 @@
...
+		policySubjects = []types.Subject{
+			{
+				Type: types.SubjectTypeSelector,
+				ID:   "*",
+			},
+		}
+		policyModels = convertToModelResources(agent.Spec.Manifest.AllowedModels)
+	)
+
+	// Make sure there are actually policys to copy before attempting to avoid producing an invalid MAP.
```
thedadams: nit:
```suggestion
	// Make sure there are actually policies to copy before attempting to avoid producing an invalid MAP.
```

pkg/api/handlers/assistants.go
njhale: good catch. fixed

pkg/controller/handlers/modelaccesspolicy/migrate.go
njhale: Done

thedadams (approved):

Issue #4997, opened 2025-11-12
Provide upgrade notification for existing MCP server instances when there are changes made to `Server Scheduling` settings.


sangee2004:
**Describe the bug**
Provide upgrade notification for existing MCP server instances when there are changes made to `Server Scheduling` settings.

**To Reproduce**
1. Connect to few single user MCP servers
2. Go to `Server Scheduling` section and change some settings ( like CPU Settings)
3. Access `Deployments and Connections` tab
Notice that there is no upgrade notification indication for existing connected MCP server instances.
4. Click on any of these   connected MCP server instances to get to the detail page
Notice that  this results in server being actually restarted.

**Expected behavior**
Provide upgrade notification for existing MCP server instances when there are changes made to `Server Scheduling` settings.
Opening detail page foe existing MCP server instances should not result in server getting restarted automatically.

cjellick:
@ryu-man as an experiment i had copilot take a crack at this one in this PR: https://github.com/obot-platform/obot/pull/5231

I never had time to test that PR so im just going to assign this to you. Figured id let you know about that PR in case you wanted to borrow code from it.

ryu-man:
thanks @cjellick I will checkout it out

sangee2004:
This issue is addressed when testing with the latest build.

After the admin changes some server settings, the status of existing MCP server instances shows as `Needs Scheduling Update` on the `Deployment and Connections` tab of the `MCP Servers` page. The user is also presented with the `Update Scheduling config` option to update the MCP server's scheduling config.

<img width="1155" height="269" alt="Image" src="https://github.com/user-attachments/assets/e766160f-4066-4a10-8ae5-3bf681d61d4f" />

Clicking any of the connected MCP server instances to open the detail page does not trigger server restarts as expected.


When there are server updates needed for an MCP server instance, then the status is shown as `Needs Config Update` and the user is provided with the `Update Server` option to update the server.

<img width="1172" height="422" alt="Image" src="https://github.com/user-attachments/assets/f055fb7c-4ad9-40ba-8344-ffcc9623eee5" />

When there are server updates and scheduling config updates needed for an MCP server instance, then the status is shown as `Needs Scheduling and Config Update` and the user is provided with the `Update Server` and `Update Scheduling config` options to update the server.

<img width="1175" height="531" alt="Image" src="https://github.com/user-attachments/assets/861223e6-cd01-49b7-b500-a3178bf33a85" />

Issues found when testing this feature will be tracked in separate issues

[...]

PR #5396, opened 2026-01-01
feat: add k8s settings update notification in deployment status

ryu-man:
Addresses #4997

- implement server-side logic
- show a yellow update icon to indicate update in k8s settings
- fix issue where server is re-deployed when navigating to `/admin/mcp-servers/c/[x]/instance/[y]/details route`

Screenshot 2025-12-31 181940


ryu-man:
@ivyjeong13 I am not sure if it is better to add an "Update k8s setting" item to the dots menu; however, I lean toward that. What do you think?

pkg/api/handlers/mcp.go
```diff
@@ -3046,6 +3048,14 @@ func (m *MCPHandler) RedeployWithK8sSettings(req api.Context) error {
 		return fmt.Errorf("failed to redeploy server: %w", err)
 	}
 
+	// Clear the NeedsK8sUpdate flag since user explicitly redeployed
+	if server.Status.NeedsK8sUpdate {
+		server.Status.NeedsK8sUpdate = false
+		if err := req.Storage.Status().Update(req.Context(), &server); err != nil {
+			return fmt.Errorf("failed to update server status: %w", err)
+		}
```
thedadams: Typically, we wouldn't want to do this in the API. Instead, let the controller update the field.

I think it's fine if you want to ensure the value is `true` when returning from the API, but better to let the controller update the database.

pkg/controller/handlers/k8ssettings/k8ssettings.go
thedadams, line 1: This file should be deleted.

The way our controller framework works, the two controllers you wrote (one for servers and one for catalog entries) will do the correct thing when the k8s settings change.

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
```diff
@@ -135,6 +136,76 @@ func DetectCompositeDrift(req router.Request, _ router.Response) error {
 	return nil
 }
 
+// DetectK8sSettingsDrift detects when servers created from this catalog entry need redeployment with new K8s settings
+func DetectK8sSettingsDrift(req router.Request, _ router.Response) error {
```
thedadams: This should be deleted.

The way our controller framework works, the controller you added for MCP servers will do the right thing.

pkg/mcp/details.go
```diff
@@ -5,10 +5,19 @@ import (
...
 	"github.com/obot-platform/obot/apiclient/types"
 )
 
+// isServerNotRunningError checks if the error indicates the server is not running
+func isServerNotRunningError(err error) bool {
+	if err == nil {
+		return false
+	}
+	return strings.Contains(err.Error(), "is not running")
+}
```
thedadams: It's better to create a type for this and use `errors.As` or `errors.Is` then match on the string.

pkg/storage/apis/obot.obot.ai/v1/mcpservercatalogentry.go
```diff
@@ -79,6 +79,8 @@ type MCPServerCatalogEntryStatus struct {
 	ManifestHash string `json:"manifestHash,omitempty"`
 	// NeedsUpdate indicates whether this composite catalog entry's component snapshots have drifted from their sources.
 	NeedsUpdate bool `json:"needsUpdate,omitempty"`
+	// NeedsK8sUpdate indicates whether servers created from this catalog entry need redeployment with new K8s settings
+	NeedsK8sUpdate bool `json:"needsK8sUpdate,omitempty"`
```
thedadams: Based on what I saw in the implementation, this can be removed. It seems like it was used to track when all servers deployed from this entry should be upgraded, but that is already handled by the MCP server controllers.

apiclient/types/mcpserver.go
```diff
@@ -122,6 +122,7 @@ type MCPServerCatalogEntry struct {
 	PowerUserWorkspaceID      string                        `json:"powerUserWorkspaceID,omitempty"`
 	PowerUserID               string                        `json:"powerUserID,omitempty"`
 	NeedsUpdate               bool                          `json:"needsUpdate,omitempty"`
+	NeedsK8sUpdate            bool                          `json:"needsK8sUpdate,omitempty"`
```
thedadams: I believe this can also be removed.

ryu-man: If I delete this I do get the following error in `mcp.go` line `159`
```
unknown field NeedsK8sUpdate in struct literal of type "github.com/obot-platform/obot/apiclient/types".MCPServerCatalogEntry
```

pkg/mcp/details.go
ryu-man: Please let me know if the change I made is what you mean

cjellick:
can I get a quick loom of this working end to end, including updating the k8s settings for the server and seeing it successfully redeploy

apiclient/types/mcpserver.go
thedadams: We don't need to track this for catalog entries. There are no deployments for catalog entries, so this field and all of its references should be removed.

ryu-man:
Yep, working on that

ryu-man:
@cjellick https://www.loom.com/share/5001ecf38d654f878498d488a0b01f69

cjellick:
@ryu-man - looking good. a couple things:
1. in the Status dropdown filter, can we add entries for the two different types of updates available? id like to be able to filter to servers that JUST need a k8s update or JUST need a catalog update
2. Somewhat related, I want to make sure we can bulk update by checking more that one row and clicking a button. 

Then end goal here is if i have a ton of seves and change the default k8s resources, i can easily update them all.

ryu-man:
@cjellick Working on that

ryu-man:
> @ryu-man - looking good. a couple things:
> 
> 1. in the Status dropdown filter, can we add entries for the two different types of updates available? id like to be able to filter to servers that JUST need a k8s update or JUST need a catalog update
> 2. Somewhat related, I want to make sure we can bulk update by checking more that one row and clicking a button.
> 
> Then end goal here is if i have a ton of seves and change the default k8s resources, i can easily update them all.


@cjellick loom for request changes.

https://www.loom.com/share/c5d286dd82fc4f8693fd2530fa45e941

ryu-man:
@cjellick Currently, if a user selects only "Kubernetes Upgrade Available," those servers with both server and Kubernetes will not be included, because currently the table filters allow matches only against a single value. if this PR can wait a little bit longer, I can work on a solution; otherwise, this can tracked in a separate PR if necessary

cjellick:
@ryu-man this pr can wait for you to find a solution.

Regarding your loom, here's some feedback:
- Instead of saying "Kubernetes Upgrade Available", let's say "Scheduling Update Available"
- Instead of saying "Upgrade Available", let's New Server Config Available"
And then make the action buttons match that, so like "Update Scheduling Configuration" and "Update Server"

Let me see a loom for that bc I'm not 100% sure itll look good.

Also, since we now have explicit text for the status, i think we can just drop the upgrade icons...they were there so that that column could do "double duty" of health status and upgrade status. so now I think we need two separate columns:
- Health
- Update Status

ryu-man:
@cjellick loom for the above requested changes + new status filteration approach

https://www.loom.com/share/884386f2ec324728a249792de4b1ed12

cjellick:
For the status field, I want to tweak the wording again. Let’s go with
    - Needs Scheduling Update
    - Needs Config Update
    - Needs Scheduling and Config Update

If the user hovers on a cell where an update is needed, can we explain more via a hover text:
- The default server scheduling rules have changed and can be applied to this server.
- The configuration for this server’s registry entry has changed and can be applied to this server.
- If both apply, just say both of the above with a new line in between

Regarding the text for updating scheduling being too long, you can shorten configuration to config

Filtering looks good.

Let me see the hover text in a short loom because Im note sure how well it will work out.

cjellick:
UX LGTM

cjellick (approved):
UX LGTM. I do want Ivy to review the code.

ui/user/src/app.css
```diff
@@ -45,6 +45,18 @@
 	--color-red-900: #7f1d1d;
 	--color-red-950: #450a0a;
 
+	--color-orange-50: oklch(0.98 0.016 73.684);
```
ivyjeong13: nit: looks like only orange-500 is used here, do we need the other orange values?

ivyjeong13 (approved):
Just one nit :+1:

pkg/controller/handlers/k8ssettings/k8ssettings.go
thedadams: This file should be deleted.

pkg/controller/routes.go
```diff
@@ -316,6 +318,9 @@ func (c *Controller) setupRoutes() {
 	// ScheduledAuditLogExport
 	root.Type(&v1.ScheduledAuditLogExport{}).HandlerFunc(scheduledAuditLogExportHandler.ScheduleExports)
 
+	// K8sSettings
+	root.Type(&v1.K8sSettings{}).HandlerFunc(k8ssettings.UpdateAllServerK8sSettingsDrift)
+
```
thedadams: This should be removed.

ryu-man:
@thedadams After I applied these changes, the 'k8s' update flag stopped being updated!!

thedadams:
The way our controller framework is designed, you don't need to update all the MCP servers when the K8S settings are updated. The handler function that operates on the MCP server object is all you need.

The logic in the two handler functions are different, so there may be something that needs to be copied.

Regardless, the `UpdateAllServerK8sSettingsDrift` function doesn't need to exist.

ryu-man:
@thedadams I tried to achieve that but can't get to work as expected; I would appreciate your help here

pkg/controller/handlers/k8ssettings/k8ssettings.go
```diff
@@ -0,0 +1,51 @@
...
+
+		// Check if the server needs update
+		needsK8sUpdate := server.Status.K8sSettingsHash != currentHash
+
+		if server.Status.NeedsK8sUpdate != needsK8sUpdate {
+			server.Status.NeedsK8sUpdate = needsK8sUpdate
+			if err := req.Client.Status().Update(req.Ctx, server); err != nil {
+				return err
+			}
+		}
```
thedadams, line 47: This is the difference in logic between the two controllers. This updates the flag if it is different than expected. The other handler only sets it to true.

I've pushed a change that removes this file and updates the other controller handler with this logic.

thedadams (approved):
This works as I understand and expect. Please test to make sure the changes I made work as you expect before merging.

ryu-man:
@thedadams I tested this with the latest changes, and it does not work as expected. When redeploying k8s settings it does not update the `needK8sUpdate` flag  in the MCP server

thedadams:
This is what I am seeing: https://www.loom.com/share/2cf95f8cbb37423f9877f4ebf4df0633

That is working as I expect. Please let me know if that is not what you expect.

ryu-man:
That's interesting. It does work on your side! checking what's happening

ryu-man:
I had to clear all the data and start fresh; apparently I had cache issues. This works as expected. Going to merge

Issue #5390, opened 2025-12-29
Improve search experience in the connectors dialog


ryu-man:
**Describe the bug**
In the connectors dialog, the user has to scroll to the top of the list to be able to use the search functionality.

**To Reproduce**
1. Got to chat page
2. Open connection dialog by clicking on the plus button on the left sidebar
3. Scroll down a little bit
4. Try narrow down result by typing in the search input

Notice that you have to scroll up to the top so you can use the search input

**Expected behavior**
The search input should be always displayed to the user during scroll

**Screenshots**

<img width="2304" height="949" alt="Image" src="https://github.com/user-attachments/assets/2b868986-3211-4856-85e4-1fb90a0ee411" />

ivyjeong13:
closing -- legacy chat related issue

PR #5400, opened 2026-01-01
fix: make search input sticky in connectors dialog (#5390)

srivtx:
## Summary
Fixes the search input scrolling out of view when scrolling through the connectors dialog.

## Problem
When opening the connectors dialog in the chat view, the search input at the top of the dialog would scroll away with the content, making it difficult for users to filter connectors after scrolling through the list.

## Solution
Added `sticky top-0 z-10 bg-background` CSS classes to the search header container in the `mainContent` snippet of `McpServerSetup.svelte`. This keeps the search input pinned to the top of the dialog while maintaining visual consistency with the background.

## Changes
- [McpServerSetup.svelte](ui/user/src/lib/components/chat/McpServerSetup.svelte): Modified the search header div to be sticky positioned

## Testing
- Verified the search input stays visible at the top while scrolling through the connectors list
- Confirmed the background color properly covers content scrolling behind it

### DIFF
--- a/ui/user/src/lib/components/chat/McpServerSetup.svelte
+++ b/ui/user/src/lib/components/chat/McpServerSetup.svelte
@@ -191,7 +191,7 @@
 {/snippet}
 
 {#snippet mainContent()}
-	<div class="w-full px-4 py-2">
+	<div class="bg-background sticky top-0 z-10 w-full px-4 py-2">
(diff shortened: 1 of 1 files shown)


ivyjeong13:
Marking as stale and closing for now.

Issue #5388, opened 2025-12-29
Make it possible to access MCP server catalog details from deployments tab


ryu-man:
**Describe the bug**
Same as the title.

**To Reproduce**
1. Go to MCP Server then Deployments & Connections

Notice if you need to access the MCP server catalog configuration, you have to: 
1. Switch to Server Entries tab
2. Search for the server
3. Open it

**Expected behavior**
The user should be able to directly access to the catalog configuration from `Deployments & Connections` tab

shwetapsdet:
Verified it on obot main and it's working as expected

https://github.com/user-attachments/assets/75fe47df-9a92-4b09-b788-676c58842b81

PR #5401, opened 2026-01-02
feat: add View Catalog Entry option in deployments tab (#5388)

srivtx:
## Summary
Adds a "View Catalog Entry" option in the Deployments & Connections tab dropdown menu, allowing users to directly access the MCP server catalog configuration.

## Problem
When users view a deployed server in the Deployments & Connections tab, they had to navigate to the Server Entries tab, search for the server, and open it to access the catalog configuration. This was a cumbersome workflow.

## Solution
Added a "View Catalog Entry" button in the dropdown menu for servers that were created from catalog entries. This provides direct navigation to the catalog entry configuration page.

## Changes
- [DeploymentsView.svelte](ui/user/src/lib/components/mcp/DeploymentsView.svelte): 
  - Added `FileBox` icon import
  - Added "View Catalog Entry" menu option that appears when a server has a `catalogEntryID`
  - Supports both regular and power user workspace URLs

## Testing
- Verified the option only appears for servers with a `catalogEntryID`
- Confirmed navigation works correctly for both admin and workspace paths
- Ctrl/Cmd+click opens in new tab

## Notes
This option will only be visible for servers that were created from catalog entries. Servers created directly (without a catalog entry) will not show this option since there's no catalog entry to link to.

### DIFF
--- a/ui/user/src/lib/components/mcp/DeploymentsView.svelte
+++ b/ui/user/src/lib/components/mcp/DeploymentsView.svelte
@@ -33,6 +33,7 @@
 		CircleAlert,
 		CircleFadingArrowUp,
 		Ellipsis,
+		FileBox,
 		GitCompare,
 		LoaderCircle,
 		MessageCircle,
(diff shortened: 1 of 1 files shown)


cjellick:
we are going to address this in a different pr from @ryu-man. Thanks for the PR.

Issue #4691, opened 2025-10-16
Upgrading helm setup with valid openai api key , does not result in model provider being reconfigured.


sangee2004:
**Describe the bug**
Upgrading helm setup with valid openai api key , does not result in auth provider being reconfigured.

**To Reproduce**
1. Install obot by setting an invalid OPENAI_API_KEY in helm command - 
```
helm install obot obot/obot --set config.OPENAI_API_KEY="sk-abc" --set config.OBOT_SERVER_ENABLE_AUTHENTICATION=true --set config.OBOT_BOOTSTRAP_TOKEN=bootstrap --set image.tag=main --set image.pullPolicy=Always --set dev.useEmbeddedDb=true
```
2. Set up auth provider. 
3. Log in as any user and chat

Following error is seen which is expected:

```
failed to stream: failed calling model for completion: request failed on non-retriable status-code 404: error, status code: 404 (404 Not Found), message: Model.obot.obot.ai "m1-openai-model-provider-5c51a66b0aff2f609fae358c5c9e6f71-54248" not found
```
4. upgrade the set up and set the correct value using following command -  `helm upgrade obot chart/ --reuse-values --set config.OPENAI_API_KEY="sk-proj-<valid-key>"`

After upgrading , i dont the see the new api key show up in the Model Provider configuration page. 
After a minute or so, i see the new api key value (that was set in step4) get reflected in the UI.

But even after this , i am not able to chat and get the same error 

```
failed to stream: failed calling model for completion: request failed on non-retriable status-code 404: error, status code: 404 (404 Not Found), message: Model.obot.obot.ai "m1-openai-model-provider-5c51a66b0aff2f609fae358c5c9e6f71-54248" not found
```

**Expected behavior**
Auth provider should get reconfigured it new OPENAI_API_KEY value that is set in the `helm upgrade` command.

g-linville:
I believe this was the deliberate choice for how the OPENAI_API_KEY env var works in Obot. It configures it the first time, and then it is ignored. So the question is whether we want to change that.

sangee2004:
This issue is addressed when testing with the latest build.

Upgrading the helm setup with a valid openai api key reconfigures the model provider as expected.

PR #5406, opened 2026-01-02
fix: stop running model provider when updating the credential

thedadams:
This will ensure that the models are populated with the correct credential on startup.

Issue: https://github.com/obot-platform/obot/issues/4691

### DIFF
--- a/pkg/controller/handlers/toolreference/toolreference.go
+++ b/pkg/controller/handlers/toolreference/toolreference.go
@@ -296,18 +296,19 @@ func (h *Handler) ensureModelProviderCredAndDefaults(ctx context.Context, c clie
 		}
 	} else if cred.Env[credentialEnvVarName] != apiKey {
 		// If the credential exists, but has a different value, then update it.
-		// The only way to update it is to delete the existing credential and recreate it.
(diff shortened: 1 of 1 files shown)


pkg/controller/handlers/toolreference/toolreference.go
```diff
@@ -296,18 +296,19 @@ func (h *Handler) ensureModelProviderCredAndDefaults(ctx context.Context, c clie
 		}
 	} else if cred.Env[credentialEnvVarName] != apiKey {
 		// If the credential exists, but has a different value, then update it.
-		// The only way to update it is to delete the existing credential and recreate it.
-		if err = h.gptClient.DeleteCredential(ctx, string(modelProviderRef.UID), modelProviderName); err != nil {
-			return fmt.Errorf("failed to delete credential: %w", err)
-		}
```
g-linville, line 302: I think we still need to delete it before we can create a new one?

thedadams: It's not necessary to delete before creating.

g-linville: oh, TIL

Issue #5438, opened 2026-01-07
Remove Horizontal Table Scroll


ivyjeong13:
When a table columns get large enough, there is horizontal scrolling.
Remove the horizontal scrolling.

* When it shrinks, hide the last column that isn't the actions column
* If the user wants to see that column on a smaller screen, provide a way to allow users to toggle what columns to hide/show

shwetapsdet:
On small screens, the filter column is not visible by default.
Apart from this, all other functionality works as expected.

https://www.loom.com/share/f201e40466cf4faab257712e541ed8ed

njhale:
> On small screens, the filter column is not visible by default

@shwetapsdet seems like will be covered in a followup: https://github.com/obot-platform/obot/issues/5438#issuecomment-3761549505

ivyjeong13:
> > On small screens, the filter column is not visible by default
> 
> [@shwetapsdet](https://github.com/shwetapsdet) seems like will be covered in a followup: [#5438 (comment)](https://github.com/obot-platform/obot/issues/5438#issuecomment-3761549505)

@njhale  If it's easier, I can close this out and move the follow-up into a bug ticket! edit: Oh actually have the fix, will be linking to this PR!

njhale:
I think we can reuse this @ivyjeong13 thanks for the quick fix

shwetapsdet:
I've completed testing and identified following issue

https://github.com/obot-platform/obot/issues/5581

PR #5443, opened 2026-01-07
fix: tables no horizontal scroll & filter display columns option

ivyjeong13:
Addresses #5438 

* remove overflow-x-hidden from Table
* start collapsing columns from the right if the parent container is smaller than the table container
* adds in option to filter what columns to display when it reaches that level
* also tiny fix to fix seeing the sidebar collapsing initially when loading on mobile size view

### DIFF
--- a/ui/user/src/lib/components/Layout.svelte
+++ b/ui/user/src/lib/components/Layout.svelte
@@ -374,7 +374,7 @@
 	<div class="relative flex w-full grow">
 		{#if layout.sidebarOpen && !hideSidebar}
 			<div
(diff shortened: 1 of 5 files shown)


ivyjeong13:
> Here is what I have noticed:
> 
> * in the small screens, there are some columns being cut by the overflow instead of being hidden
> 
> image
> It would be better to preselect which columns to show in small screens, like only the 'name' & 'type' columns
> 
> * On tablets and larger screens, deselecting columns from filters will not persist after reloading
> * Deselecting all columns from the filter will result in reselecting all the columns again, which is unexpected from the user's perspective! I think making some columns unselectable, like the "name" column, would have better UX
> * I think It would be better to use toggle pills for showing/hiding columns rather than closable chips, so user user can have better sight on which column to bring back without having to go through reset (in this case the search input is not needed)

Will be doing/looking through this as a follow-on!

PR #5446, opened 2026-01-07
docs: add model access policy docs

njhale:


docs/sidebars.ts
```diff
@@ -26,6 +26,7 @@ const sidebars = {
         "functionality/filters",
         "functionality/server-scheduling",
         "functionality/chat-management",
+        "functionality/model-access-policies",
         "functionality/user-management",
         "functionality/branding",
```
njhale, line 31: it feels like all 4 of these should go into the `Configuration and Operations` category.

g-linville (approved):

Issue #5361, opened 2025-12-19
Static OAuth Configuration for Remote MCP Servers


cjellick:
### Summary

Add first-class support for remote MCP servers that require static OAuth configuration (client ID and secret), fully replacing the existing OAuthApps mechanism. Catalog entries and UI-defined remote servers must be able to declare this requirement and block usage until an admin supplies the required OAuth configuration.

### Problem

There are two related user-facing problems:

1. Some OAuth-based MCP servers, such as the GitHub remote MCP, do not support dynamic client registration. Obot must be configured with a static client ID and secret generated by the MCP server's authorization server (ie https://github.com/login/oauth).
2. The current OAuthApps implementation is globally scoped, loosely tied to MCP servers, confusing to reason about, and insufficient for productized catalog.

This issue focuses on the remote-server side of the solution. Containerized MCP server OAuth is handled in a separate, related issue.

### Goals

* Allow remote MCP servers to explicitly declare that they require static OAuth configuration.
* Require admins to supply OAuth configuration before the server can be used.
* Support both catalog-defined remote servers and remote servers defined directly in the UI.
* Fully replace and deprecate the existing OAuthApps implementation.

### User Stories

* As an admin, I can easily discover what MCP servers shipped as part of Obot's catalog are remotes that require additional OAuth configuraiton before they can be used.
* As an admin when I enable a remote MCP server that requires OAuth, I am prompted to configure the client ID and secret before users can access it.
* As a basic and PU user, I cannot connect to a remote MCP server that requires OAuth until it has been configured by an admin. If I have access to it via Access Control Rules, but it is not configured, I should not see it at all.

#### Catalog-defined remote servers

* A catalog entry must be able to indicate that it requires static OAuth configuration.
* Required fields include at least client ID and client secret.
* The authorization server URL may be defined as part of the catalog entry. If so, it is treated as a default that an admin can override.
* Until OAuth configuration is supplied, the server is not usable and should be marked "Requires configuration" in the Status column on the MCP Servers page for Admins.
* For non-admins, it should not appear at all.
  * **TODO** This is a potentially contentious UX point. Maybe it should show up for PUP users as well so that they can configure it in their workspace if they want.

#### UI-defined remote servers

* When defining a remote MCP server via the UI, if static OAuth configuration is required, that configuration must be supplied before the server can be saved or enabled.
* The server cannot be turned on in a partially configured state.

### OAuth Flow (Remote)

[...]

PR #5450, opened 2026-01-08
feat: add static OAuth for remote catalog entries

g-linville:
for https://github.com/obot-platform/obot/issues/5361

I will add docs in a follow-up PR.

UX overview (video): https://acorn-io.slack.com/archives/C07P0EZR917/p1768273597737279

This will not merge until after the v0.16.0 release.

### DIFF
--- a/ui/user/src/lib/components/mcp/ConnectorsView.svelte
+++ b/ui/user/src/lib/components/mcp/ConnectorsView.svelte
@@ -22,7 +22,7 @@
 		hasEditableConfiguration,
 		requiresUserUpdate
 	} from '$lib/services/chat/mcp';
-	import { darkMode, mcpServersAndEntries, profile } from '$lib/stores';
+	import { mcpServersAndEntries, profile } from '$lib/stores';
 	import { formatTimeAgo } from '$lib/time';
 	import { setSearchParamsToLocalStorage } from '$lib/url';
 	import { openUrl } from '$lib/utils';


ui/user/src/lib/components/SensitiveInput.svelte
```diff
@@ -132,12 +136,13 @@
...
 		<div bind:this={containerElement} class="relative flex min-h-[60px] w-full flex-col leading-5">
 			{#if growable}
-				<input type="text" {name} {disabled} {value} {placeholder} hidden />
+				<input type="text" {name} {disabled} {readonly} {value} {placeholder} hidden />
 				<div
 					bind:this={scrollableWrapper}
 					class={twMerge(
 						'text-input-filled base flex w-full flex-1 flex-col overflow-x-hidden overflow-y-auto font-mono',
 						klass,
+						classes,
```
ryu-man: For better consistency, the `classes` props should be an object where the key is the name element and value is the class string, then use it like `classes.elementName`

ui/user/src/lib/components/mcp/ConnectorsView.svelte
```diff
@@ -312,6 +372,10 @@
 				{:else if property === 'connected'}
 					{#if d.connected}
 						<div class="pill-primary bg-primary">Connected</div>
+					{:else if 'isCatalogEntry' in d.data && d.data.manifest?.runtime === 'remote' && d.data.manifest?.remoteConfig?.staticOAuthRequired && !d.data.oauthCredentialConfigured}
+						<div class="pill-warning bg-yellow-500/20 text-yellow-700 dark:text-yellow-400">
```
ryu-man: colors: `yellow-700` & `yellow-400` does not exist in the `app.css`, you can use `yellow-500` and adjust opacity/brightness/contrast; otherwise, you have to add them in the file

ui/user/src/lib/components/mcp/ConnectorsView.svelte
```diff
@@ -334,6 +398,10 @@
 					d.connected && d.type === 'multi' ? instancesMap.get(d.data.id) : undefined}
 				{@const hasConnectedOptions =
 					'isCatalogEntry' in d.data ? matchingServers.length > 0 : !!matchingInstance}
+				{@const requiresOAuth =
+					'isCatalogEntry' in d.data &&
```
ryu-man: nit: `'isCatalogEntry' in d.data` is used many times; it can be declared as a boolean variable

ryu-man (approved):
UI looks good

ui/user/src/lib/components/SensitiveInput.svelte
g-linville: fixed

ui/user/src/lib/components/mcp/ConnectorsView.svelte
g-linville: fixed, just used `yellow-500` instead

ui/user/src/lib/components/mcp/ConnectorsView.svelte
g-linville: fixed

ui/user/src/lib/components/chat/ChatConnectorsView.svelte
```diff
@@ -149,6 +149,10 @@
 				{:else if property === 'connected'}
 					{#if d.connected}
 						<div class="pill-primary bg-primary">Connected</div>
+					{:else if 'isCatalogEntry' in d.data && d.data.manifest?.runtime === 'remote' && d.data.manifest?.remoteConfig?.staticOAuthRequired && !d.data.oauthCredentialConfigured}
+						<div class="pill-warning bg-yellow-500/20 text-yellow-700 dark:text-yellow-400">
```
ivyjeong13: I'm not sure if there is a `pill-warning` class, would double check this! Might need to change the text-yellow to text-yellow-500 or add 400/700 values.

ivyjeong13 (approved):

njhale (commented):
Code looks good.

My primary concern is how this interacts with composite MCP catalog entries that have remotes with static OAuth configured. I think we may need to copy the credential for the remotes in the composite.

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
```diff
@@ -15,8 +17,20 @@ import (
...
+// NewHandler creates a new Handler with the given GPTScript client
+func NewHandler(gClient *gptscript.GPTScript) *Handler {
+	return &Handler{
+		gClient: gClient,
+	}
+}
+
 // EnsureUserCount ensures that the user count for an MCP server catalog entry is up to date.
-func EnsureUserCount(req router.Request, _ router.Response) error {
+func (h *Handler) EnsureUserCount(req router.Request, _ router.Response) error {
```
thedadams: pico-nit:
```suggestion
func (*Handler) EnsureUserCount(req router.Request, _ router.Response) error {
```

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
```diff
@@ -165,3 +179,142 @@ func CleanupNestedCompositeEntries(req router.Request, _ router.Response) error
...
+func (h *Handler) CleanupUnusedOAuthCredentials(req router.Request, _ router.Response) error {
+	entry := req.Object.(*v1.MCPServerCatalogEntry)
+
+	// Only process remote entries
+	if entry.Spec.Manifest.Runtime != types.RuntimeRemote {
+		return nil
+	}
+
+	// Only cleanup if RemoteConfig exists and StaticOAuthRequired is false
+	if entry.Spec.Manifest.RemoteConfig != nil && entry.Spec.Manifest.RemoteConfig.StaticOAuthRequired {
```
thedadams, line 194: I think it makes sense to allow static OAuth credentials for MCP servers that don't require OAuth. Is that something we want to do? It doesn't seem like a huge lift, which is why I am asking.

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
```diff
@@ -165,3 +179,142 @@ func CleanupNestedCompositeEntries(req router.Request, _ router.Response) error
...
+
+	// Build the credential name for this entry
+	credName := system.MCPOAuthCredentialName(entry.Name)
+
+	// Try to reveal the credential to see if it exists
+	_, err := h.gClient.RevealCredential(req.Ctx, []string{credName}, "oauth")
+	if err != nil {
+		// If credential doesn't exist, nothing to clean up
+		var notFoundErr gptscript.ErrNotFound
+		if errors.As(err, &notFoundErr) {
```
thedadams: nit (here and other places):
```suggestion
		if errors.As(err, &gptscript.ErrNotFound{}) {
```

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
```diff
@@ -165,3 +179,142 @@ func CleanupNestedCompositeEntries(req router.Request, _ router.Response) error
...
+	return nil
+}
+
+// EnsureOAuthCredentialStatus updates the OAuthCredentialConfigured status field
+// for remote catalog entries that require static OAuth.
+func (h *Handler) EnsureOAuthCredentialStatus(req router.Request, _ router.Response) error {
+	entry := req.Object.(*v1.MCPServerCatalogEntry)
+
+	// Clear sync annotation if present
+	if entry.Annotations != nil && entry.Annotations[v1.MCPServerCatalogEntrySyncAnnotation] == "true" {
```
thedadams:
```suggestion
	if entry.Annotations[v1.MCPServerCatalogEntrySyncAnnotation] == "true" {
```

Also, it seems like you only care whether or not this annotation is set.

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
```diff
@@ -165,3 +179,142 @@ func CleanupNestedCompositeEntries(req router.Request, _ router.Response) error
...
+		delete(entry.Annotations, v1.MCPServerCatalogEntrySyncAnnotation)
+		if err := req.Client.Update(req.Ctx, entry); err != nil {
+			return fmt.Errorf("failed to clear sync annotation: %w", err)
+		}
+		// Re-fetch entry to get updated resource version
+		if err := req.Client.Get(req.Ctx, kclient.ObjectKeyFromObject(entry), entry); err != nil {
+			return fmt.Errorf("failed to re-fetch entry: %w", err)
+		}
```
thedadams: Not necessary. The update will write to the reference you pass. Also, the client here is cached so maybe the cache hasn't been updated yet?

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
```diff
@@ -165,3 +179,142 @@ func CleanupNestedCompositeEntries(req router.Request, _ router.Response) error
...
+			}
+		}
+
+		return h.propagateOAuthCredentialStatus(req, entry.Name, false)
+	}
+
+	// Check if credentials exist
+	credName := system.MCPOAuthCredentialName(entry.Name)
+	_, err := h.gClient.RevealCredential(req.Ctx, []string{credName}, "oauth")
+	configured := err == nil
```
thedadams: Do we need to make sure the credential has the right fields? Also, if we failed to get for some reason other than "not found" we probably don't want to update the status of the object.

apiclient/types/mcpserver.go
```diff
@@ -42,19 +42,23 @@ type ContainerizedRuntimeConfig struct {
...
-	URLTemplate string      `json:"urlTemplate,omitempty"` // URL template for user URLs
-	Hostname    string      `json:"hostname,omitempty"`    // Optional: Hostname constraint the URL conforms to
-	Headers     []MCPHeader `json:"headers,omitempty"`     // Optional
+	URL                    string      `json:"url"`                              // Required: Full URL to remote MCP server
+	IsTemplate             bool        `json:"isTemplate"`                       // Optional: Whether the URL is a template
+	URLTemplate            string      `json:"urlTemplate,omitempty"`            // URL template for user URLs
+	Hostname               string      `json:"hostname,omitempty"`               // Optional: Hostname constraint the URL conforms to
+	Headers                []MCPHeader `json:"headers,omitempty"`                // Optional
+	AuthorizationServerURL string      `json:"authorizationServerURL,omitempty"` // OAuth authorization server URL (indicates static OAuth required)
+	StaticOAuthRequired    bool        `json:"staticOAuthRequired,omitempty"`    // Indicates static OAuth is required
```
thedadams, line 51: Do we need both of these since they both indicate the same thing?

pkg/api/handlers/mcp.go
```diff
@@ -131,6 +131,15 @@ func (m *MCPHandler) ListEntriesFromAllSources(req api.Context) error {
 		}
 
 		if hasAccess {
+			// Hide entries that require OAuth credentials that haven't been configured (non-admins only).
+			// Workspace owners can always see their own entries (they need to configure the OAuth credentials).
+			if !req.UserIsAdmin() && entryRequiresUnconfiguredOAuth(entry) {
+				// Check if this is a workspace entry owned by the current user
+				if entry.Spec.PowerUserWorkspaceID == "" || entry.Spec.PowerUserWorkspaceID != system.GetPowerUserWorkspaceID(req.User.GetUID()) {
```
```suggestion
				if entry.Spec.PowerUserWorkspaceID != system.GetPowerUserWorkspaceID(req.User.GetUID()) {
```

pkg/api/handlers/mcp.go
```diff
@@ -2439,6 +2454,14 @@ func ConvertMCPServer(server v1.MCPServer, credEnv map[string]string, serverURL,
 		}
 	}
 
+	// Check if OAuth credentials are required but missing
+	missingOAuth := false
+	if server.Spec.Manifest.RemoteConfig != nil &&
+		server.Spec.Manifest.RemoteConfig.StaticOAuthRequired {
+		// Use the status field populated by the controller
+		missingOAuth = !server.Status.OAuthCredentialConfigured
+	}
```
thedadams, line 2481: I'm bringing this up as a question rather than a suggestion. I typically do something like what's in the suggestion. Is this a bad idea? Is it harder to understand?
```suggestion
	missingOAuth := server.Spec.Manifest.RemoteConfig != nil &&
		server.Spec.Manifest.RemoteConfig.StaticOAuthRequired &&
		!server.Status.OAuthCredentialConfigured
	}
```

cjellick:
> Code looks good.
> 
> My primary concern is how this interacts with composite MCP catalog entries that have remotes with static OAuth configured. I think we may need to copy the credential for the remotes in the composite.

@njhale Can you explain in more detail? I'm very confused as to how the composite would simply be a simple pass through in this scenario.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -1647,3 +1661,258 @@ func (h *MCPCatalogHandler) RefreshCompositeComponents(req api.Context) error {
 
 	return req.Write(ConvertMCPServerCatalogEntry(entry))
 }
+
+// entryRequiresUnconfiguredOAuth checks if a catalog entry requires OAuth credentials
+// that haven't been configured yet. Returns true if the entry should be hidden from non-admin users.
+func entryRequiresUnconfiguredOAuth(entry v1.MCPServerCatalogEntry) bool {
```
thedadams: This name seems weird. Maybe `entryRequiresOAuthCreds`?

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -1647,3 +1661,258 @@ func (h *MCPCatalogHandler) RefreshCompositeComponents(req api.Context) error {
...
+
+	// Check if credentials already exist
+	credName := system.MCPOAuthCredentialName(entry.Name)
+	existingCred, err := req.GPTClient.RevealCredential(req.Context(), []string{credName}, "oauth")
+	credentialsExist := err == nil
+
+	var clientID, clientSecret, authServerURL string
+
+	if credentialsExist {
+		// Update mode: Only authorization server URL can be updated
```
thedadams, line 1812: Why is this the case? It seems reasonable to me to allow updating other fields, if they are sent with the request.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -1647,3 +1661,258 @@ func (h *MCPCatalogHandler) RefreshCompositeComponents(req api.Context) error {
...
+		return err
+	}
+
+	// Delete existing credential if present
+	if err := req.GPTClient.DeleteCredential(req.Context(), credName, "oauth"); err != nil {
+		var notFound gptscript.ErrNotFound
+		if !errors.As(err, &notFound) {
+			return fmt.Errorf("failed to delete existing OAuth credential: %w", err)
+		}
+	}
```
thedadams: nit: not necessary.
```suggestion
```

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -1647,3 +1661,258 @@ func (h *MCPCatalogHandler) RefreshCompositeComponents(req api.Context) error {
...
+			"AUTHORIZATION_SERVER_URL": authServerURL,
+		},
+	}
+	if err := req.GPTClient.CreateCredential(req.Context(), cred); err != nil {
+		return fmt.Errorf("failed to create OAuth credential: %w", err)
+	}
+
+	// Trigger reconciliation to update the status
+	if entry.Annotations == nil {
+		entry.Annotations = make(map[string]string)
```
thedadams: nit:
```suggestion
		entry.Annotations = make(map[string]string, 1)
```

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -1647,3 +1661,258 @@ func (h *MCPCatalogHandler) RefreshCompositeComponents(req api.Context) error {
...
+	if err := req.GPTClient.DeleteCredential(req.Context(), credName, "oauth"); err != nil {
+		var notFound gptscript.ErrNotFound
+		if !errors.As(err, &notFound) {
+			return err
+		}
+	}
+
+	// Trigger reconciliation to update the status
+	if entry.Annotations == nil {
+		entry.Annotations = make(map[string]string)
```
thedadams: nit:
```suggestion
		entry.Annotations = make(map[string]string, 1)
```

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
g-linville: Right now we don't expose it in the UI. You're right that we don't strictly need to disallow this, but until it is something that we specifically want, then I think it's fine to keep this as it is.

apiclient/types/mcpserver.go
g-linville: Yes. One indicates whether static oauth is required. The other one is the default value for the authorization server URL if it is required. If only the authorization server URL gets specified for some reason, we will also flip the boolean to be true

pkg/api/handlers/mcpcatalogs.go
g-linville: You're right, I'm renaming it to `entryRequiresStaticOAuthCreds`

pkg/api/handlers/mcpcatalogs.go
g-linville: So this was a deliberate UX decision made by me and Craig. We never return back the client secret with the request. And there's not much of a reason to change the client secret unless you are also changing the client ID. So we just made it so that you have to clear and start over.

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
g-linville: I think the AI just assumed things here incorrectly. I'll test with this call removed and I expect it should work.

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
g-linville: We don't need to check the fields, as this can only be set programmatically with the same standard set of required field.

Good call on the status of the object - I'll go fix that

pkg/api/handlers/mcp.go
g-linville: Personally I am indifferent here. It makes no difference to me as far as understanding goes. If Claude wrote it the way I have it, then maybe it makes the most sense to keep it that way? I guess it's more of a question of what is easier for LLMs, which is hard to say.

pkg/api/handlers/mcpcatalogs.go
g-linville: oh does the gptscript sdk just replace the credential if you `Create` one that already exists? I didn't think it did

njhale:
> @njhale Can you explain in more detail? I'm very confused as to how the composite would simply be a simple pass through in this scenario.

So IIUC, in this PR, the credential being used for a remote in a composite is the _current_ credential of the "parent manifest", not the credential at the time the remote manifest was copied into the composite config. I don't have a great example of use case where this would cause a problem, but it does throw off the UX for composites.

ui/user/src/lib/components/chat/ChatConnectorsView.svelte
g-linville: Fixed :+1:

You're right, I don't see `pill-warning` anywhere

pkg/api/handlers/mcp.go
```diff
@@ -1430,6 +1430,28 @@ func serverManifestFromCatalogEntryManifest(
...
+			if entryHasStaticOAuth && !inputHasStaticOAuth {
+				// The component has gained static OAuth - reject the update.
+				componentID := entryComponent.CatalogEntryID
+				if componentID == "" {
+					componentID = entryComponent.MCPServerID
+				}
+				return types.MCPServerManifest{}, types.NewErrBadRequest(
+					"cannot update composite server: component %s has been updated to require static OAuth, which is not allowed in composite servers",
+					componentID,
+				)
```
njhale:
```go
return types.MCPServerManifest{}, types.NewErrBadRequest(
					"cannot update composite server: component %s has been updated to require static OAuth, which is not allowed in composite servers",
					entryComponent.ComponentID(),
Comment view				)
```

ui/user/src/lib/components/mcp/McpServerActions.svelte
```diff
@@ -540,4 +597,28 @@
...
+			}
+			oauthConfigured = true;
+
+			// Show the connect dialog if this was part of the initial creation flow
+			if (isInitialOAuthConfig) {
+				isInitialOAuthConfig = false;
+				launchDialog?.open();
+			}
+		}}
+	/>
```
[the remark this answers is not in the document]
g-linville: Claude says we should keep this as it is

pkg/api/handlers/mcp.go

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -287,6 +291,11 @@ func (h *MCPCatalogHandler) CreateEntry(req api.Context) error {
 		}
 	}
 
+	// Auto-set StaticOAuthRequired if AuthorizationServerURL is provided
+	if manifest.RemoteConfig != nil && manifest.RemoteConfig.AuthorizationServerURL != "" {
+		manifest.RemoteConfig.StaticOAuthRequired = true
```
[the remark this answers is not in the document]
g-linville: No, this is dumb. The authorization server URL is just a default. They should always be setting the other field too. If they do it through the UI, this will never be a problem. Only if they are setting up a GitHub catalog and don't understand what they are doing.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -1532,6 +1546,13 @@ func (h *MCPCatalogHandler) populateComponentManifests(req api.Context, manifest
...
+			// Reject remote entries with static OAuth from being included in composites
+			if entry.Spec.Manifest.Runtime == types.RuntimeRemote &&
+				entry.Spec.Manifest.RemoteConfig != nil &&
+				entry.Spec.Manifest.RemoteConfig.StaticOAuthRequired {
+				return types.NewErrBadRequest("remote catalog entry %s with static OAuth cannot be included in a composite server", component.CatalogEntryID)
+			}
+
 			// Populate the manifest
 			component.Manifest = entry.Spec.Manifest
 			// Keep this component
```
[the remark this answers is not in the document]
g-linville: AI falsely assumes that multi-user servers can be created from catalog entries here (they can't, at least not through the UI under normal circumstances).

pkg/api/handlers/mcpcatalogs.go
thedadams: I don't think the SDK does that. I think that's the way gptscript credentials work.

g-linville: Cool, well I applied your change to remove this deletion

g-linville: So we should be good

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
```diff
@@ -15,8 +17,20 @@ import (
...
+// NewHandler creates a new Handler with the given GPTScript client
+func NewHandler(gClient *gptscript.GPTScript) *Handler {
+	return &Handler{
+		gClient: gClient,
+	}
+}
+
 // EnsureUserCount ensures that the user count for an MCP server catalog entry is up to date.
-func EnsureUserCount(req router.Request, _ router.Response) error {
+func (*Handler) EnsureUserCount(req router.Request, _ router.Response) error {
```
thedadams, line 33: nit: this method doesn't have a receiver variable and the others do. I think I commented a previous iteration of these changes to remove this one, but the others weren't.

Based on other conversations we've had, I don't really care if the receiver variable is here or not, but I would like it to be consistent.

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
```diff
@@ -165,3 +179,142 @@ func CleanupNestedCompositeEntries(req router.Request, _ router.Response) error
...
+	// Try to reveal the credential to see if it exists
+	_, err := h.gClient.RevealCredential(req.Ctx, []string{credName}, "oauth")
+	if err != nil {
+		// If credential doesn't exist, nothing to clean up
+		if errors.As(err, &gptscript.ErrNotFound{}) {
+			return nil
+		}
+		// Some other error occurred
+		return fmt.Errorf("failed to check for OAuth credential: %w", err)
+	}
```
thedadams: nit: do we need to do this first?

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
```diff
@@ -165,3 +179,142 @@ func CleanupNestedCompositeEntries(req router.Request, _ router.Response) error
...
+		}
+		return fmt.Errorf("failed to delete OAuth credential: %w", err)
+	}
+
+	return nil
+}
+
+// propagateOAuthCredentialStatus updates OAuthCredentialConfigured status on all MCPServers
+// that were created from the given catalog entry.
+func (h *Handler) propagateOAuthCredentialStatus(req router.Request, entryName string, configured bool) error {
```
thedadams: I think this should be a controller handler on the `MCPServer` type. I propose the following scenario:

1. A catalog entry with 100 deployed MCP servers has its static OAuth credentials removed.
2. The catalog entry is updated to indicate that OAuth is not configured.
3. This controller gets through 95% of the MCP servers and the other 5 have errors and are not updated. The error is returned.
4. Because the error is returned, the catalog entry goes through the handler again, but no update is needed because the catalog entry was updated above. The remaining 5 MCP servers don't have their status updated.

If the controller handler was on the `MCPServer` instead, then any error updating the status would only affect that `MCPServer` and the retry could be successful.

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
g-linville: nope, we don't

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
g-linville: Fixed

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
g-linville: Fixed

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
g-linville: Good call! This is implemented now.

Issue #5418, opened 2026-01-06
Azure OpenAI Model Provider - Provide a clear indication of the required fields for API Key vs. Microsoft Entra Authentication.


sangee2004:
**Describe the bug**
Azure OpenAI Model Provider - Provide a clear indication of the required fields for API Key vs. Microsoft Entra Authentication.

**To Reproduce**
1. Configure Azure OpenAI Model Provider

Notice that there is no clear indication that the admin can choose between API Key Authentication and Microsoft Entra Authentication.

The user is asked to enter only `Endpoint` as the required configuration, which is incorrect. 

<img width="735" height="744" alt="Image" src="https://github.com/user-attachments/assets/c953d566-8f73-4005-9974-6a90f0e90cf8" />

When only `Endpoint` is provided, all optional fields are flagged as required, which is also incorrect.

<img width="707" height="661" alt="Image" src="https://github.com/user-attachments/assets/737c3664-f5d2-459e-96b3-dfe7d0b52477" />

**Expected behavior**
Provide a clear indication that the user can choose between API Key Authentication and Microsoft Entra Authentication.

When one of the authentication methods is selected, the user should be presented with an accurate indication of the required fields for the chosen method.

sangee2004:
@ryu-man When testing with the latest build, I am still not seeing users being presented with the correct list of required and optional configurations.

When the user chooses `API Key` authentication, only  `API Key` is presented as required configuration, which is incorrect.

When only  `API Key` is entered and the user clicks on `Confirm`, the user is presented with the following error message:

<img width="727" height="678" alt="Image" src="https://github.com/user-attachments/assets/8010661a-16f6-40da-84f1-2c6c0cee6702" />

To enter the `Endpoint` value, users now need to use the `Endpoint` tab. 
Once the user clicks on`Confirm` after entering the endpoint, the user will be presented with an error message that indicates that `OBOT_AZURE_OPENAI_MODEL_PROVIDER_DEPLOYMENTS` is required, but all fields in the `Optional Configuration` section will be marked in red, indicating that these are required fields, which is incorrect.

<img width="742" height="594" alt="Image" src="https://github.com/user-attachments/assets/cf3959b0-3e13-4641-8b7a-674ae89ee151" />

<img width="765" height="787" alt="Image" src="https://github.com/user-attachments/assets/e0c45045-2198-4003-847f-fde9a419bc9c" />

[...]

PR #5451, opened 2026-01-08
feat: introduce pair params to enhance Azure Open AI configuration handling

ryu-man:
Addresses #5418

- User can select only one of the paired params (endpoint or API key)

image

### DIFF
--- a/ui/user/src/lib/components/admin/ProviderConfigure.svelte
+++ b/ui/user/src/lib/components/admin/ProviderConfigure.svelte
@@ -235,7 +235,7 @@
 												class={twMerge(
 													'bg-surface1 hover:bg-surface2 text-gray rounded-md px-4 py-2 text-sm font-medium transition-all duration-200',
 													isSelected &&
-														'bg-primary text-white shadow-sm hover:bg-primary/90 active:bg-primary',
+														'bg-primary hover:bg-primary/90 active:bg-primary text-white shadow-sm',
 													isSelected &&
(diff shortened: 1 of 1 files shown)


g-linville:
I really like this UX. Let me know if you want to make it more generic and outsource some of the hardcoded information to the model provider.

ryu-man:
@g-linville I think at the moment it is good to keep it hard-coded, since there is only one use case for this. I would consider do some lifting to the model provider in separate PR if this pattern goes spreads more

ui/user/src/lib/components/admin/ProviderConfigure.svelte
```diff
@@ -148,9 +216,91 @@
...
+
+								<li class="flex flex-col gap-2">
+									<div class="flex gap-1">
+										{#each pair as parameter (parameter.name)}
+											{@const isSelected = selectedParameter?.name === parameter.name}
+											<button
+												class={twMerge(
+													'bg-surface1 hover:bg-surface2 text-gray rounded-md px-4 py-2 text-sm font-medium transition-all duration-200',
+													isSelected &&
+														'bg-blue-500 text-white shadow-sm hover:bg-blue-600/90 active:bg-blue-600',
```
ivyjeong13: Just making sure is this blue the primary color or we always want it to be blue -- use bg-primary instead for dynamic theming if it is primary. Otherwise, ignore this!

PR #5462, opened 2026-01-09
fix: replace "Connector" words with "MCP Server" for consistency

ryu-man:
Same as the title


ui/user/src/lib/components/mcp/McpResources.svelte
```diff
@@ -281,7 +281,7 @@
 			<LoaderCircle class="size-4 animate-spin" />
 		{:else}
 			<HardDrive class="size-4" />
-			Add from Connector
+			Add from MCP Servers
```
ivyjeong13: nit: Add MCP Servers instead of Add from MCP Servers

PR #5471, opened 2026-01-12
fix: exclude static assets from rate limiting

cjellick:
## Summary

- Exclude static asset paths (`/_app/`, `/user/images/`) from rate limiting
- Fixes 429 errors during page load for unauthenticated users

## Problem

Users were experiencing 500 errors (manifesting as 429 Too Many Requests) when loading pages on chat.obot.ai. The issue primarily affected **unauthenticated users** due to a lower rate limit (100 req/sec vs 200 req/sec for authenticated users).

The rate limiter in `pkg/api/server/server.go` was being applied to **all HTTP requests**, including static assets. When SvelteKit loads a page like `/mcp-servers`, the browser fires many parallel requests:

- Multiple JS chunks (`/_app/immutable/chunks/*.js`)
- CSS files (`/_app/immutable/assets/*.css`)
- API calls (`/api/version`, `/api/me`, `/api/app-preferences`)
- Images and other assets

A typical page load can generate 30-50+ simultaneous requests, which exceeds the 100 req/sec limit for unauthenticated users.

## Root Cause

All handlers are registered via `HandleFunc` or `HTTPHandle` on the API server, and `HTTPHandle` wraps handlers through `Wrap()` which applies rate limiting to every request. The UI handler is registered via `HTTPHandle` at the catch-all `/` route, so static assets were being rate limited.

## Solution

Added `isStaticAssetPath()` helper function to identify static asset paths and skip rate limiting for them. Static assets don't need rate limiting protection since they're cacheable, read-only resources.

## Test plan
- [ ] Fire up and configure obot with authentication turned on 
- [ ] Load `/mcp-servers` in incognito mode (unauthenticated)
- [ ] Verify the page loads and redirects to the login page. If the fix DOESNT work, you should see a 500 error.

### DIFF
--- a/pkg/api/server/server.go
+++ b/pkg/api/server/server.go
@@ -110,16 +110,20 @@ func (s *Server) Wrap(f api.HandlerFunc) http.HandlerFunc {
 			return
 		}
 
-		if err := s.rateLimiter.ApplyLimit(user, rw, req); err != nil {
-			if errors.Is(err, ratelimiter.ErrRateLimitExceeded) {
-				// The user has exceeded their rate limit.
-				http.Error(rw, err.Error(), http.StatusTooManyRequests)
-				return
-			}
+		// Skip rate limiting for static assets (JS chunks, CSS, images) to avoid
+		// hitting limits during page load when many assets are fetched in parallel.
+		if !isStaticAssetPath(req.URL.Path) {
+			if err := s.rateLimiter.ApplyLimit(user, rw, req); err != nil {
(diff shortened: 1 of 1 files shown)


shwetapsdet:
Loaded /mcp-servers in incognito mode and verified the page loads and redirects to login successfully.
No 429 or 500 errors observed.
Static assets load correctly and are no longer rate-limited.
401 responses for unauthenticated API calls are expected.

https://github.com/user-attachments/assets/f7238a2a-3cff-4fc7-94c7-33ef4e5d9b1d

PR #5472, opened 2026-01-12
chore: ui: add key prefix column to API keys table

g-linville:
This is what it looks like for admins:

image

And for non-admins:

image

### DIFF
--- a/ui/user/src/routes/admin/api-keys/+page.svelte
+++ b/ui/user/src/routes/admin/api-keys/+page.svelte
@@ -29,7 +29,7 @@
 	const allTableData = $derived(
 		allApiKeys.map((key) => ({
 			...key,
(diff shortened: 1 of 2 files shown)


g-linville:
I also bumped Nanobot in this PR so that the new API key redact logic is included.

g-linville:
@njhale the info I am showing here is enough to uniquely identify the key.

njhale:
> @njhale the info I am showing here is enough to uniquely identify the key.

yes, wasn't questioning that. was just saying that I normally don't see it presented as a prefix, but rather in the form of a redacted key. Was just wondering if that's how we wanted to display it here too.

g-linville:
Ohh now I understand. Yeah we can do that.

g-linville:
Done

PR #5485, opened 2026-01-12
feat: add k8s resource reporting for MCP servers

cloudnautique:
- Modified K8s roles to be able to query resources from the cluster.
- Add UI dashboard elements under the Deployments & Connections tab for administrators.
- Adds error at deployment time if capacity is exceeded
- resources are reported based on ResourceQuota degrades to just shows total resource requests.
- Docker backend has no change.

Generated by CC

Tested:
- Docker deployment has no changes.
- Kubernetes with resource quota (numerically works, but had to manually edit deployments)
- Kubernetes without resource quota.

### DIFF
--- a/pkg/api/handlers/mcpcapacity.go
+++ b/pkg/api/handlers/mcpcapacity.go
@@ -1,6 +1,8 @@
 package handlers
 
 import (
+	"errors"
+
 	"github.com/obot-platform/obot/apiclient/types"
 	"github.com/obot-platform/obot/pkg/api"
 	"github.com/obot-platform/obot/pkg/mcp"
@@ -27,7 +29,7 @@ func (h *MCPCapacityHandler) GetCapacity(req api.Context) error {
 	if err != nil {
 		// If backend doesn't support capacity info (e.g., Docker), return empty info
 		var notSupported *mcp.ErrNotSupportedByBackend
-		if ok := isErrNotSupportedByBackend(err, &notSupported); ok {
+		if errors.As(err, &notSupported) {
 			return req.Write(types.MCPCapacityInfo{
 				Error: notSupported.Error(),
 			})
@@ -36,12 +38,4 @@ func (h *MCPCapacityHandler) GetCapacity(req api.Context) error {
 	}
 
 	return req.Write(info)
-}
-
-func isErrNotSupportedByBackend(err error, target **mcp.ErrNotSupportedByBackend) bool {
-	if e, ok := err.(*mcp.ErrNotSupportedByBackend); ok {
-		*target = e
-		return true
-	}
-	return false
 }
--- a/pkg/mcp/kubernetes.go
+++ b/pkg/mcp/kubernetes.go
@@ -1227,24 +1227,60 @@ func (k *kubernetesBackend) getResourceQuotaCapacity(ctx context.Context) (types
 		Source: types.CapacitySourceResourceQuota,
 	}
 
-	// Aggregate quotas
+	// Aggregate limits from all ResourceQuotas
+	var totalCPULimit, totalMemoryLimit resource.Quantity
 	for _, quota := range quotas.Items {
-		if used, ok := quota.Status.Used[corev1.ResourceRequestsCPU]; ok {
-			info.CPURequested = formatCPU(used)
-		}
 		if hard, ok := quota.Status.Hard[corev1.ResourceRequestsCPU]; ok {
-			info.CPULimit = formatCPU(hard)
-		}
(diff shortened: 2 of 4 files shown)


pkg/mcp/kubernetes.go
```diff
@@ -1119,3 +1124,230 @@ func (k *kubernetesBackend) getK8sSettings(ctx context.Context) (v1.K8sSettingsS
...
+		replicas := int64(1)
+		if deployment.Spec.Replicas != nil {
+			replicas = int64(*deployment.Spec.Replicas)
+		}
+		for _, container := range deployment.Spec.Template.Spec.Containers {
+			if cpu, ok := container.Resources.Requests[corev1.ResourceCPU]; ok {
+				cpuScaled := cpu.DeepCopy()
+				for i := int64(0); i < replicas; i++ {
+					totalCPU.Add(cpuScaled)
+				}
```
g-linville: Can we just do a multiplication instead of repeated adding? AI recommends this:

```go
if cpu, ok := container.Resources.Requests[corev1.ResourceCPU]; ok {
    scaled := cpu.DeepCopy()
    scaled.SetMilli(scaled.MilliValue() * replicas)
    totalCPU.Add(scaled)
}
```

pkg/mcp/kubernetes.go
```diff
@@ -1119,3 +1124,230 @@ func (k *kubernetesBackend) getK8sSettings(ctx context.Context) (v1.K8sSettingsS
...
+		if hard, ok := quota.Status.Hard[corev1.ResourceRequestsCPU]; ok {
+			info.CPULimit = formatCPU(hard)
+		}
+		if used, ok := quota.Status.Used[corev1.ResourceRequestsMemory]; ok {
+			info.MemoryRequested = formatMemory(used)
+		}
+		if hard, ok := quota.Status.Hard[corev1.ResourceRequestsMemory]; ok {
+			info.MemoryLimit = formatMemory(hard)
+		}
+	}
```
g-linville, line 1239: Should this be adding to each field in `info`, rather than overwriting it on each loop iteration?

pkg/api/handlers/mcpcapacity.go
```diff
@@ -0,0 +1,47 @@
...
+	return req.Write(info)
+}
+
+func isErrNotSupportedByBackend(err error, target **mcp.ErrNotSupportedByBackend) bool {
+	if e, ok := err.(*mcp.ErrNotSupportedByBackend); ok {
+		*target = e
+		return true
+	}
+	return false
+}
```
g-linville: Just use `errors.As` instead of a function like this.

pkg/api/handlers/mcp.go
```diff
@@ -550,6 +550,9 @@ func (m *MCPHandler) LaunchServer(req api.Context) error {
 				if errors.Is(err, nmcp.ErrNoResult) || strings.HasSuffix(err.Error(), nmcp.ErrNoResult.Error()) {
 					return types.NewErrHTTP(http.StatusServiceUnavailable, fmt.Sprintf("No response from component MCP server %s, check configuration for errors", component.Name))
 				}
+				if errors.Is(err, mcp.ErrInsufficientCapacity) {
+					return types.NewErrHTTP(http.StatusServiceUnavailable, "Insufficient capacity to deploy MCP server. Please contact your administrator.")
+				}
```
g-linville, line 555: If possible, it would be nice to provide more details in this error message, like how much mem/cpu is requested, how much more is needed, etc. Not sure whether that is easy to accomplish though, so this is more of a nice-to-have.

cloudnautique: This is to be shown to the end user. I don't think it is info we would want surfaced to people on chat.obot.ai. Also, as a normal user there isn't anything you can do about it, except to explain to the admin.. who needs to look anyways.

pkg/api/router/router.go
```diff
@@ -623,6 +623,10 @@ func Router(ctx context.Context, services *services.Services) (http.Handler, err
 	mux.HandleFunc("GET /api/k8s-settings", k8sSettingsHandler.Get)
 	mux.HandleFunc("PUT /api/k8s-settings", k8sSettingsHandler.Update)
 
+	// MCP Capacity (admin only)
+	mcpCapacityHandler := handlers.NewMCPCapacityHandler(services.MCPLoader)
```
g-linville, line 627: micro nit: would be nice to declare this at the top with the other handlers

cloudnautique: It's following the more recent pattern where handlers and their functions are registered in the same block.

Issue #5436, opened 2026-01-07
Accessing chat threads that have default model set to a model that they dont have access to exposes the model id.


sangee2004:
**Describe the bug**
When access to a model is removed after a user has already created a chat with that model, reopening the chat displays the raw model ID instead of a human-readable model name. This happens because the chat thread still references a model the user no longer has access to, leaving the UI without proper metadata. Ideally, the chat should continue to display the model name, or gracefully fall back to the system-wide default model when the initially selected model is no longer accessible.

**To Reproduce**
1. Provide access to any model, say `gpt-3.5-turbo` in my case, to all users by adding this model to the existing `Default Policy` in the `Model Access Policies`
2. As any user, start a chat and pick `gpt-3.5-turbo` as the model for the chat thread.
3. As admin/owner user, update `Default Policy` in the `Model Access Policies` and remove `gpt-3.5-turbo` model.
4. As the same user from step2, access the chat thread from step 2 and refresh the page. You can also quit the chat page and relaunch it.

Notice that this will result in the default model in the chat well being presented as the model id.

<img width="1023" height="534" alt="Image" src="https://github.com/user-attachments/assets/46ed9df0-acd0-4e53-a27d-0656df8deca9" />

**Expected behavior**
Ideally, we should continue to show the model name instead of the model id in the chat well.
In this case, do we want to automatically fall back to the system-wide default model instead of continuing to use the user-selected model, which he no longer has access to?

njhale:
> Ideally, we should continue to show the model name instead of the model id in the chat well.

I agree, but we should probably include the model provider name as well so that the error message is still useful to us as well.

> In this case, do we want to automatically fall back to the system-wide default model instead of continuing to use the user-selected model, which he no longer has access to?

Yeah after refresh this should use the same model defaulting logic that we use for creating new threads. So if the model set on the thread doesn't exist:
1. if the user has access to the system default use that
2. if they don't, fall back to the first `usage = llm` model they have
3. if they have access to no models, the selector should say "No Model Selected" and the send message button should be disabled

sangee2004:
This issue is addressed when testing with the latest build.

When the model set for the chat thread is no longer accessible to the user, the chat's default model falls back to the default llm model configured for the obot server.

When there is no access to the default LLM model configured, no model is selected for the chat, and the user cannot chat until a model is selected.

[...]

Issue #5430, opened 2026-01-07
404 error message presented when attempting to chat when existing model provider deconfigured.


sangee2004:
**Describe the bug**
After deconfiguring the OpenAI model provider, launching Chat results in a user being presented with 404 error message. The `default-model` API call made by the UI fails with a 404 error, which ends up being presented to the user in Chat window.

Pre Req: Have an auth-enabled obot set up with the OpenAI model provider configured.
**To Reproduce**
1. Deconfigure the OpenAI model provider
2. Launch Chat by using the `Launch Chat` option under `Chat Management` or by accessing the `Chat` page
3. Notice that the user is presented with a 404 error message

<img width="1148" height="823" alt="Image" src="https://github.com/user-attachments/assets/6c463615-2020-4757-a037-2afd4c2512c5" />

Following is the api call made by UI, which results in 404 with error message - `Model.obot.obot.ai "m1-openai-model-provider-5c51a66b0aff2f609fae358c5c9e6f71-54248" not found`
```
http://localhost:8080/api/assistants/a1-obot/projects/p1p56pw/default-model
```
**Expected behavior**
There should be no error message presented to the user in this case.

sangee2004:
This issue is addressed when testing with the latest build.
No 404 errors are seen when attempting to chat when the existing model provider is deconfigured.

PR #5487, opened 2026-01-12
fix: default chat model selection

njhale:
- Don't return a 404 from /default-model when the default model doesn't exist
- When a user's access to a model is revoked after creating a thread, the UI now automatically updates the thread to use a fallback model instead of displaying raw model IDs. The fallback priority is: system default → project default → first available LLM model.

Addresses:
- https://github.com/obot-platform/obot/issues/5430
- https://github.com/obot-platform/obot/issues/5436

### DIFF
--- a/ui/user/src/lib/components/edit/ThreadModelSelector.svelte
+++ b/ui/user/src/lib/components/edit/ThreadModelSelector.svelte
@@ -49,6 +49,7 @@
 
 	let threadDefaultModel = $state<string>();
 	let threadDefaultModelProvider = $state<string>();
+	let hasAttemptedDefaultModelFetch = $state(false);
 
 	let defaultModel = $derived(threadDefaultModel ?? projectDefaultModel);
 	let defaultModelProvider = $derived(threadDefaultModelProvider ?? projectDefaultModelProvider);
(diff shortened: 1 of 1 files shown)


pkg/api/handlers/projects.go
```diff
@@ -628,7 +628,7 @@ func (h *ProjectsHandler) CreateProjectThread(req api.Context) error {
 			}
 
 			// if bodyContents.ModelProvider is empty it means that it is set at global level so allowedByProject should be true
-			if bodyContents.ModelProvider != "" {
+			if bodyContents.ModelProvider == "" {
```
njhale, line 631: based on the comment and intent, this was conditional wasn't correct

ui/user/src/lib/components/edit/ThreadModelSelector.svelte
```diff
@@ -80,15 +81,40 @@
...
-	let threadModelProvider = $derived(
-		threadType?.modelProvider ??
-			threadDefaultModelProvider ??
-			defaultModelProvider ??
-			fallbackModel?.provider
-	);
+	// Auto-fallback to accessible model if thread's model is inaccessible
+	let threadModel = $derived.by(() => {
+		// If thread has explicit model and it's accessible, use its ID
+		if (threadType?.model) {
```
ivyjeong13, line 87: nit: if you wanna save a few lines, can use `let` and `return modelObj?.id` at end

```
let modelObj;
if (threadType?.model) {
    let modelObject = modelObj = availableModels.find(
				(m) => m.id === threadType?.model || m.name === threadType?.model
			);
}

const fallbackModelRef = threadDefaultModel || defaultModel || fallbackModel?.id;
if (fallbackModelRef) {
			let modelObj = availableModels.find(
				(m) => m.id === fallbackModelRef || m.name === fallbackModelRef
			);
}

return modelObject?.id
```

ui/user/src/lib/components/edit/ThreadModelSelector.svelte
```diff
@@ -80,15 +81,40 @@
...
+			const modelObj = availableModels.find(
+				(m) => m.id === fallbackModelRef || m.name === fallbackModelRef
+			);
+			if (modelObj) {
+				return modelObj.id;
+			}
+		}
+		return undefined;
+	});
+	let threadModelProvider = $derived.by(() => {
```
ivyjeong13: nit: can save a few lines 
```
let threadModelProvider = $derived(threadModel ? availableModels.find((m) => m.id === threadModel)?.modelProvider : undefined)

ui/user/src/lib/components/edit/ThreadModelSelector.svelte
njhale: done

ui/user/src/lib/components/edit/ThreadModelSelector.svelte
njhale: done

Issue #4992, opened 2025-11-11
Wait for user input to create task


cjellick:
We have a ton of empty tasks. Seems like users click the add task button out of curiosity and then abandon. Can we wait until a user inputs a step to save the task?

I'm open to other ideas that would require a proactive step from the user before we save the task. It could be even waiting until the click Run

shwetapsdet:
Verified it on obot main and working as expected

https://www.loom.com/share/a13f4d843f56460c9162411e07cf0b54

PR #5496, opened 2026-01-13
enhance: save new task on initial change

ivyjeong13:
Addresses #4992

### DIFF
--- a/ui/user/src/lib/components/tasks/Task.svelte
+++ b/ui/user/src/lib/components/tasks/Task.svelte
@@ -55,7 +55,6 @@
 		noChat
 	}: Props = $props();
 
-	let prevTaskId = $state(untrack(() => task.id));
 	const readOnly = untrack(() => !!inputRunID || readonly);
 	let runID = $state(untrack(() => inputRunID));
 	let thread: Thread | undefined = $state<Thread>();


ui/user/src/lib/components/Obot.svelte
```diff
@@ -137,18 +137,16 @@
 					{#if layout.sidebarConfig}
 						<SidebarConfig bind:project bind:currentThreadID {assistant} />
 					{:else if layout.editTaskID && layout.tasks}
-						{#each layout.tasks as task, i (task.id)}
+						{#each layout.tasks as task, i (i)}
```
ryu-man, line 140: I wonder what is the reason behind using the index instead of the task id as the block key

ui/user/src/lib/components/tasks/Task.svelte
```diff
@@ -53,6 +55,7 @@
 		noChat
 	}: Props = $props();
 
+	let prevTaskId = $state(untrack(() => task.id));
```
ryu-man: This var can be removed

ui/user/src/lib/components/Obot.svelte
ivyjeong13: Ah! I had made this change when I was gonna allow addition of a new task(s) without an id so task.id wouldn't be valid in that case. But I since changed to use a placeholder id for a new task so I can change this back.

ivyjeong13: Oh actually wait, there might be another case of when the placeholder id changes to the actual task.id after the new save that can happen when user is mid making a change in an input that can cause a re-rendering and loss of where they were. Will check that and may not make the change to task.id edit: This is the case!

PR #5506, opened 2026-01-14
fix: prevent infinite loading when clicking current project in selector

cloudnautique:
Clicking the already-active project in the project selector would set loading=true, but since SvelteKit doesn't navigate when the URL is unchanged, the loading state was never reset.

This issue was reported by user.


shwetapsdet:
Verified it on Obot main and it's working as expected 

https://www.loom.com/share/c6ccb635fd614fe082acfb619bcdb775

PR #5536, opened 2026-01-16
fix: build Azure OpenAI model provider configuration dynamically based on authentication type

ryu-man:
Addresses #5418 [(comment)](https://github.com/obot-platform/obot/issues/5418#issuecomment-3746966398)

- In case of Azure OpenAI setup, recalculate required & optional fields based on selected authentication method

image

### DIFF
--- a/ui/user/src/lib/components/admin/ProviderConfigure.svelte
+++ b/ui/user/src/lib/components/admin/ProviderConfigure.svelte
@@ -23,9 +23,7 @@
 	let form = $state<Record<string, string>>({});
 	let showRequired = $state(false);
 
-	const isAzureOpenAIProvider = $derived(
-		provider && provider.id === 'azure-openai-model-provider'
-	);
+	const isAzureOpenAIProvider = $derived(provider && provider.id === 'azure-openai-model-provider');
 
 	const collection = $derived.by(() => {
 		if (isAzureOpenAIProvider)


g-linville:
Oh wait I just noticed in the screenshot: for API key, the client ID, client secret, etc. should not be shown at all. The only optional parameter should be the API version. And for Entra, the API key and deployment names should not be shown at all, and the only optional parameter should be the API version.

ryu-man:
Making the changes now...

ryu-man:
@g-linville I pushed the change, now both API key and MS Entra auth methods will have the API version as optional parameter

image

njhale (approved):

ui/user/src/lib/components/admin/ProviderConfigure.svelte
```diff
@@ -328,7 +293,13 @@
 											labels={parameter.name === 'OBOT_AUTH_PROVIDER_EMAIL_DOMAINS'
 												? { '*': 'All domains' }
 												: {}}
-											class="text-input-filled"
+											class={[
+												'text-input-filled',
+												error &&
+													'border border-red-500 bg-red-500/20 text-red-500 ring-red-500 focus-within:ring-1'
```
ivyjeong13: If this is error styling we're going with for text-input-filled, can we update the .error case in the app.css so you can just add the 'error' class?

```
// app.css
	.input-text-filled {
		// etc.

		&.error {
			// existing styling when error class is with input-text-filled
		}
	}
```

ui/user/src/lib/components/admin/ProviderConfigure.svelte
```diff
@@ -347,7 +318,7 @@
 											id={parameter.name}
 											bind:value={form[parameter.name]}
 											class:error
-											class="text-input-filled"
+											class={['text-input-filled', error && 'ring-red-500 focus:ring-1']}
```
ivyjeong13: same thing about the error styling state

ui/user/src/lib/components/admin/ProviderConfigure.svelte
```diff
@@ -26,55 +26,84 @@
...
-		if (!isAzureOpernAIProvider) return [];
 
-		const endpoint = provider?.requiredConfigurationParameters?.find(
-			(p) => p.name === 'OBOT_AZURE_OPENAI_MODEL_PROVIDER_ENDPOINT'
-		);
-		const apiKey = provider?.optionalConfigurationParameters?.find(
-			(p) => p.name === 'OBOT_AZURE_OPENAI_MODEL_PROVIDER_API_KEY'
-		);
+	const collection = $derived.by(() => {
+		if (isAzureOpernAIProvider)
```
ivyjeong13: super minor nit: just noticed, `Open` instead of `Opern`

PR #5554, opened 2026-01-18
chore: remove legacy commands

cjellick:
Here is all that remains:
- completion  Generate the autocompletion script for the specified shell
- help        Help about any command
- server
- token
- version

These commands were removed becausae they generally no longer align with
the product:
- agents, catalog, create, edit, update, delete, invoke
- credentials (list and delete)
- tools (register, unregister, update)
- projects, tasks, threads, runs

Also remove orphaned utility files and directories:
- output.go, table.go, emailreceiver.go
- events/, templates/, textio/ directories


cjellick:
Note for @njhale : Even thought we dont really advertise this CLI in the docs, we will need to put a note in the release notes explaining that we removed these commands.

Issue #5546, opened 2026-01-16
404 error encountered when trying to update scheduling config for multi-user MCP servers.


sangee2004:
**Describe the bug**
When an admin attempts to update the scheduling configuration for a multi-user MCP server from the `Deployments and Connections` tab, the UI fails with a 404 “page not found” error. This occurs after modifying settings on the Server Scheduling page and then using the `Update Scheduling Config` option for the same server instance. The UI triggers the API call /api/mcp-servers/{id}/redeploy-with-k8s-settings, which returns 404, preventing scheduling updates for multi-user MCP servers.

**To Reproduce**
1. As admin, create a multi-user MCP server.
2. Go to the `Server Scheduling` page and change some settings.
3. Go to the `Deployments and Connections` tab in the `MCP Servers` page.
4. Locate the multi-user MCP server instance and use the `Update Scheduling Config` option.

Notice that the user is presented wih 404 error message

<img width="1206" height="790" alt="Image" src="https://github.com/user-attachments/assets/ea409c9c-2275-4de9-be5f-38d947105a66" />

API call made by UI - `http://localhost:8080/api/mcp-servers/ms1k64l8/redeploy-with-k8s-settings` which gets s `404 page not found` error message

**Expected behavior**
Should be able to update scheduling config for multi-user MCP servers.

sangee2004:
This issue is addressed when testing with the latest build.
Updating the scheduling config for the multi-user MCP server works as expected.

PR #5561, opened 2026-01-19
fix: user correct endpoint for redeploying multi-user MCP servers

thedadams:
The k8s redeployemnt for multi-user MCP servers was incorrect. This change uses the correct endpoint for redeploying MCP servers.

Issue: https://github.com/obot-platform/obot/issues/5546

### DIFF
--- a/ui/user/src/lib/services/admin/operations.ts
+++ b/ui/user/src/lib/services/admin/operations.ts
@@ -885,7 +885,11 @@ export async function getK8sSettingsStatus(
 }
 
 export async function redeployWithK8sSettings(mcpServerId: string, opts?: { fetch?: Fetcher }) {
-	const response = await doPost(`/mcp-servers/${mcpServerId}/redeploy-with-k8s-settings`, {}, opts);
+	const response = await doPost(
+		`/mcp-catalogs/${DEFAULT_MCP_CATALOG_ID}/servers/${mcpServerId}/redeploy-with-k8s-settings`,
+		{},
+		opts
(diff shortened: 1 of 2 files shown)


ui/user/src/lib/services/admin/operations.ts
```diff
@@ -885,7 +885,11 @@ export async function getK8sSettingsStatus(
 }
 
 export async function redeployWithK8sSettings(mcpServerId: string, opts?: { fetch?: Fetcher }) {
-	const response = await doPost(`/mcp-servers/${mcpServerId}/redeploy-with-k8s-settings`, {}, opts);
+	const response = await doPost(
+		`/mcp-catalogs/${DEFAULT_MCP_CATALOG_ID}/servers/${mcpServerId}/redeploy-with-k8s-settings`,
+		{},
+		opts
+	);
```
g-linville, line 892: Are we sure this function is only used for multi-user servers in the default catalog? Is there a different function somewhere else for multi-user servers in PowerUserWorkspaces?

njhale: I'll take a look

njhale: yeah, looks like we can also have multi-user servers in PUP workspaces, so this will need to change. I'll handle it.

njhale (changes requested):

njhale:
Superseded by https://github.com/obot-platform/obot/pull/5574

Issue #5547, opened 2026-01-16
Updating scheduling config fails with 400 errors for composite MCP servers.


sangee2004:
**Describe the bug**
When there is a connected composite MCP server (with Tavily Search and DuckDuckGo Search) instance,  selecting all servers for a `Kubernetes Upgrade` action causes redeployment of the remote server (Tavily Search) that is part of the composite MCP server to fail with a 400 error stating “Redeployment is only supported for Kubernetes runtime.” 

**To Reproduce**
1. Create a composite MCP server with a few MCP servers like `Tavily Search` and `Duckduckgo Search`.
2. Connect with this MCP server
3. Go to the `Server Scheduling` page and change some settings.
4. Go to the `Deployments and Connections` tab in the `MCP Servers` page
5. Select all entries and click on the `Kubernetes Upgrade` option
<img width="1149" height="672" alt="Image" src="https://github.com/user-attachments/assets/edaf46ea-dbe2-469c-a777-9386754ce760" />
6. Confirm that you need to  update K8s settings

Notice that this results in 400 errors presented to the user

<img width="1177" height="464" alt="Image" src="https://github.com/user-attachments/assets/7eb93c96-dc39-47c4-bee3-c8dc9ce2c452" />

This API call made by ui - `http://localhost:8080/api/mcp-catalogs/default/entries/default-santestschedulechanges-d7106e609nnd2/servers/ms1x4g2l/redeploy-with-k8s-settings` fails with the following error message -`Redeployment is only supported for Kubernetes runtime`.

**Expected behavior**
API call made to redeploy k8s settings should not error out.

thedadams:
Based on my testing, the problem is actually that the composite MCP server hasn't been deployed yet, which is why the status is `Unknown`. Once the composite MCP is deployed, then you are able to redeploy its Kubernetes settings.

sangee2004:
This issue is addressed when testing with the latest builds.

Able to update scheduling configs for composite MCP servers as expected.

PR #5564, opened 2026-01-19
fix: use backend configuration to determine if redeploy is allowed

thedadams:
Prior to this change, if the deployed kubernetes configuration hash was empty, we would assume that the we were not deploymen MCP servers into Kubernetes. It turned out that it may have been possible that the server hadn't been deployed yet.

This change uses the backend configuration to determine if redeploys are allowed instead, and doesn't error if the status field isn't set or the server has yet to be deployed.

Issue: https://github.com/obot-platform/obot/issues/5547

### DIFF
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -44,14 +44,16 @@ type MCPHandler struct {
 	mcpSessionManager *mcp.SessionManager
 	mcpOAuthChecker   MCPOAuthChecker
 	acrHelper         *accesscontrolrule.Helper
+	mcpBackend        string
 	serverURL         string
 }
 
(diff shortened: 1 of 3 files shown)


pkg/api/handlers/mcp.go
```diff
@@ -2974,6 +2976,13 @@ func (m *MCPHandler) CheckK8sSettingsStatus(req api.Context) error {
 
 // RedeployWithK8sSettings redeploys a server with the current K8s settings
 func (m *MCPHandler) RedeployWithK8sSettings(req api.Context) error {
+	switch m.mcpBackend {
+	case "kubernetes", "k8s":
```
g-linville, line 2980: nit: I don't think we ever represent the Kubernetes backend with the string `k8s`

g-linville (approved):
Thanks for fixing this. I think the bug was my fault

Issue #5560, opened 2026-01-19
No indication provided in `Deployments and Connections` when MCP server pods get stuck in pending state due to insufficient requested resources.


sangee2004:
**Describe the bug**
Updating the MCP server's scheduling config with the `Update Scheduling Config` action when the resource requested exceeds cluster capacity, the MCP server's status gets marked as `Up to date` even though the redeployment fails. The failure is only visible on the detail page, with no indication on the `Deployments and Connections` page, and the server remains unusable (chat fails). The MCP server status should instead reflect the failed redeploy, e.g., by showing Unavailable status. 

Pre-Req: Connect to a few MCP servers.

**To Reproduce**
1. Go to the `Server Scheduling` page and update the requested resources (memory or limit) such that it would result in exceeding the available requested resources in the k8s cluster when MCP servers are being redeployed using the `Update Scheduling Config` option. 
The easiest way to achieve this is to set the `CPU Request` to more than the available capacity in the cluster.

2. Go to `MCP Servers`-> `Deployments and Connections` tab
3. Use the `Update Scheduling Config` option for any of the entries (all existing connected MCP server instances would be in `Needs Scheduling Update` state).

Notice that after the `Update Scheduling Config` action, the MCP server entry's status changes to `Up to date` state.
https://www.loom.com/share/b2e396671d1e4aad9de0cba103508668

In this case, redeploying the MCP Server deployment did not actually succeed, which can be noted in the detail page

<img width="1077" height="352" alt="Image" src="https://github.com/user-attachments/assets/af49ac53-a63f-4dc3-9eff-a67be50a9146" />

 There is no indication on the' Deployments and Connections` page that the scheduling update was not applied successfully in this case.

Chatting with this MCP server fails

<img width="1022" height="802" alt="Image" src="https://github.com/user-attachments/assets/619ac885-572e-448e-b153-9fe1c635502c" />

**Expected behavior**
We should be able to show the MCP Server status as `Unavailable` in this case. If the current `Status` field shown cannot be used to track this state, we could introduce a new status field for tracking the availability of MCP servers.

sangee2004:
This issue is not fully addressed in the latest build.

We are now presented with 2 status-related fields, as expected: `Health` and `Update Status`.

<img width="1349" height="415" alt="Image" src="https://github.com/user-attachments/assets/219991e7-a2d9-4388-bccb-e6e21ad8bd58" />

[...]

PR #5587, opened 2026-01-21
fix: improve k8s backend rollout status

njhale:
Returns "Progressing" status immediately when replicas are unavailable during
rollout, rather than showing "Available". Returns "Needs Attention"
after 60s instead of 600s when rollout remains stuck.

Addresses https://github.com/obot-platform/obot/issues/5560

Loom: https://www.loom.com/share/4bc9a791f5304987af5b74173e3834a1

### DIFF
--- a/ui/user/src/lib/components/mcp/DeploymentsView.svelte
+++ b/ui/user/src/lib/components/mcp/DeploymentsView.svelte
@@ -258,20 +258,17 @@
 		if (hasProgressing) {
 			await reload(false);
 		}
-
-		// Always schedule next poll regardless of server state
-		pollingInterval = window.setTimeout(() => checkAndPoll(), POLL_INTERVAL_MS);
 	}
 
 	onMount(async () => {
 		await reload(true);
 		// Start checking for progressing servers
-		checkAndPoll();
+		pollingInterval = setInterval(() => checkAndPoll(), POLL_INTERVAL_MS);
 	});
 
 	onDestroy(() => {
 		if (pollingInterval) {
-			window.clearTimeout(pollingInterval);
+			clearInterval(pollingInterval);
 		}
 	});
 


pkg/controller/handlers/deployment/deployment.go
```diff
@@ -143,15 +143,40 @@ func (h *Handler) CleanupOldIDs(req router.Request, _ router.Response) error {
...
+			availableCondition = condition
+		case appsv1.DeploymentProgressing:
+			progressingCondition = condition
+		}
+	}
+
+	if progressingCondition != nil && progressingCondition.Status == corev1.ConditionFalse {
+		if progressingCondition.Reason == "ProgressDeadlineExceeded" {
+			// Rollout is stuck (after deadline)
+			return "ProgressDeadlineExceeded"
```
njhale: this can probably have a better name or just be "Unavailable"

njhale: Going with "Needs Attention" here, as a result of the discussion in standup today

ui/user/src/lib/components/mcp/DeploymentsView.svelte
```diff
@@ -247,12 +247,39 @@
...
+		// Check if there are any servers with Progressing status
+		const hasProgressing = tableData.some((row) => row.deploymentStatus === 'Progressing');
+
+		// Only reload if there are progressing servers
+		if (hasProgressing) {
+			await reload(false);
+		}
+
+		// Always schedule next poll regardless of server state
+		pollingInterval = window.setTimeout(() => checkAndPoll(), POLL_INTERVAL_MS);
```
ivyjeong13: If this is always done in 10sec intervals and nothing ever kills it (except ondestroy), could use setInterval instead! 

Also wondering if it's better to only start polling when a user has triggered a server rollout then kill the interval when none of the servers need it anymore so polling only happens when needed. Maybe future thing, not a blocker!

ivyjeong13: nit: can just use `setTimeout`, no need for window.setTimeout!

ui/user/src/lib/components/mcp/DeploymentsView.svelte
```diff
@@ -247,12 +247,39 @@
...
+
+	onMount(async () => {
+		await reload(true);
+		// Start checking for progressing servers
+		checkAndPoll();
+	});
+
+	onDestroy(() => {
+		if (pollingInterval) {
+			window.clearTimeout(pollingInterval);
```
ivyjeong13: nit: same for here : ) can just be `clearTimeout`

ui/user/src/lib/components/mcp/DeploymentsView.svelte
njhale: > Also wondering if it's better to only start polling when a user has triggered a server rollout then kill the interval when none of the servers need it anymore so polling only happens when needed

I think this misses the case where the component loads for the first time and there are servers already in pending.

Issue #5054, opened 2025-11-17
Composite MCP Servers Tools dialog, clicking just outside the toggle button closes dialog


ivyjeong13:
**Describe the bug**
When configuring a composite MCP server & in the configuring tools dialog, if you click just outside the toggle button, the dialog gets closed when it shouldn't.

**To Reproduce**

**Expected behavior**
It shouldn't close the dialog cause you're still clicking inside the dialog.

**Screenshots**

**Additional context**

shwetapsdet:
When I click outside the toggle button in the Configure Tools dialog, the dialog closes. After clicking the Configure Tools button, the dialog opens, and if I click anywhere outside the toggle area, the dialog closes.

https://github.com/user-attachments/assets/3d7b5a05-4f34-4e56-9666-a902ec51c5ba

shwetapsdet:
Verified it on obot main and it's working as expected

https://github.com/user-attachments/assets/4b36d404-7d5c-4b33-8992-787deb6649fc

PR #5591, opened 2026-01-21
fix: clickoutside without clickoutside action & consistency/cleanup of dialogs

ivyjeong13:
Addresses #5054 

* this ticket has been expanded to updating to using native dialog/popover & allowing deletion of the clickoutside action
* does ui tweaking as well (making confirm dialogs just a lil less uglier? Have dialogs that should be ResponsiveDialog, etc)
* removed unused components found along the way

ex: 
before: 
Screenshot 2026-01-23 at 4 07 48 PM

after: 
Screenshot 2026-01-23 at 4 14 01 PM
* provides name if possible to help with context
* has 'info' or 'delete' type for cases of Confirm not being a destructive action (like Restart Server)

* - [x] urgh some window register event listeners for clickoutside... so close 🥲 
* - [x] add polyfill for native popover particularily for safari/mobile browsers if users haven't updated > 26.0 (Sept 2025) -- changed dialogs that were popovers to use this instead (note: holding off removing the popover/tooltip action since this PR is getting pretty hefty & will do as follow-on)
* - [x] do testing of all dialogs / popovers in Chrome/Safari (check desktop & mobile for both)
* - [x] no stacked confirm dialog on group role assignment/update
**Note**: Not in a rush to get this in, will keep it updated as I see new changes from main related to any dialogs : )

### DIFF
--- a/ui/user/src/lib/components/mcp/ConnectorsView.svelte
+++ b/ui/user/src/lib/components/mcp/ConnectorsView.svelte
@@ -409,12 +409,7 @@
 				{@const requiresOAuth =
 					catalogEntry?.manifest?.runtime === 'remote' &&
 					catalogEntry.manifest?.remoteConfig?.staticOAuthRequired}
-				<DotDotDot
-					class="icon-button hover:dark:bg-background/50"
-					classes={{
-						menu: 'p-0'
-					}}
-				>
+				<DotDotDot class="icon-button hover:dark:bg-background/50" classes={{ menu: 'p-0' }}>
 					{#snippet icon()}
 						<Ellipsis class="size-4" />
 					{/snippet}
@@ -425,27 +420,6 @@
 								class="bg-background dark:bg-surface2 rounded-t-xl p-2 pl-4 text-[11px] font-semibold uppercase"
 							>
 								My Connection(s)
-							</div>
-							<div class="bg-surface1 flex flex-col gap-1 p-2">
-								{@render connectToServerAction(d.data, toggle)}
-								<button
-									class="menu-button hover:bg-surface3"
-									onclick={async (e) => {
-										e.stopPropagation();
-										if ('isCatalogEntry' in d.data) {
-											if (matchingServers.length === 1) {
-												connectToServerDialog?.handleSetupChat(matchingServers[0]);
-											} else {
-												handleShowSelectServerDialog(d.data as MCPCatalogEntry, 'chat');
-											}
-										} else {
-											connectToServerDialog?.handleSetupChat(d.data, instancesMap.get(d.id));
-										}
-										toggle(false);
-									}}
-								>
-									My Connection(s)
-								</button>
 							</div>
 							<div class="bg-surface1 flex flex-col gap-1 p-2">
 								{#if !requiresOAuth || catalogEntry?.oauthCredentialConfigured}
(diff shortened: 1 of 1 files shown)


ui/user/src/lib/actions/clickoutside.ts
```diff
@@ -20,8 +20,16 @@ export function clickOutside(
 
 	function checkDialogClickOutside(event: MouseEvent) {
 		if (!(element as HTMLDialogElement).open) return;
-		if (!(event.target as HTMLElement)?.contains(element)) return;
 		if (isTextSelected()) return;
+
+		const rect = element.getBoundingClientRect();
+		const isInsideDialog =
```
ryu-man: Does checking if the dialog has focus or has a focused element make sense with this solution for edge cases like two stacked dialogs?

ui/user/src/lib/actions/clickoutside.ts
```diff
@@ -20,8 +20,16 @@ export function clickOutside(
 
 	function checkDialogClickOutside(event: MouseEvent) {
 		if (!(element as HTMLDialogElement).open) return;
-		if (!(event.target as HTMLElement)?.contains(element)) return;
 		if (isTextSelected()) return;
+
+		const rect = element.getBoundingClientRect();
```
ryu-man: I do feel like `getBoundingClientRect` for this context is considered more as a turnaround. If this PR has to be merged ASAP, do you think it does make sense to refactor how `clickoutside` works in the future?

ui/user/src/lib/actions/clickoutside.ts
ivyjeong13: I'm not sure if I'm following, like a `<dialog>` that is inside a `<dialog>`>? Is there a use case where we're considering doing that?, I know we have one bug ticket where it is happening but we actually don't want it the dialog to show behind it.

ryu-man: I mean when one dialog is displayed above another dialog

ui/user/src/lib/actions/clickoutside.ts
ivyjeong13: Actually, I would like to get rid of this action and do something similar to how DaisyUI does it with the dialog and clickoutside. So yup, but not currently!

ui/user/src/lib/actions/clickoutside.ts
ivyjeong13: Ah, that's what I meant with the bug ticket! Do we have stacked dialogs somewhere? Just wondering cause I'd like to look and see if we can change it so that isn't the case!

ryu-man: Yes, as I remember in groups role assignment when user opens created/edit dialog it will be prompted to confirm the action ( for owner role & auditor privilege) using the confirm dialog, this also might be the case when assigning users roles 🤞

ivyjeong13: Thank you! I'll take a look there. And I'm considering just doing the refactor so moving this to Draft for now and will re-assign once done

ryu-man (approved):
Overall, it looks good to me. One thing I noticed on the `Chat` page / `Add MCP Server` dialog is that if you hover over the table filter columns button, you will see the tooltip rendered behind the dialog!

image

PR #5593, opened 2026-01-21
fix: use fallbackCopy when navigator.clipboard isn't available

ivyjeong13:
Addresses #3433

### DIFF
--- a/ui/user/src/lib/components/CopyButton.svelte
+++ b/ui/user/src/lib/components/CopyButton.svelte
@@ -30,7 +30,7 @@
 	const COPIED_TEXT = 'Copied!';
 
 	function fallbackCopy(textToCopy: string): boolean {
-		const activeElement = document.activeElement;
+		const previousActiveElement = document.activeElement;
 		const textArea = document.createElement('textarea');
 		textArea.value = textToCopy;
 
@@ -49,7 +49,7 @@
 			return false;
 		} finally {
 			document.body.removeChild(textArea);
-			(document.activeElement as HTMLElement)?.focus?.();
+			(previousActiveElement as HTMLElement)?.focus?.();
 		}
 	}
 


ui/user/src/lib/components/CopyButton.svelte
```diff
@@ -29,16 +29,51 @@
...
+	function fallbackCopy(textToCopy: string): boolean {
+		const textArea = document.createElement('textarea');
+		textArea.value = textToCopy;
+
+		textArea.style.position = 'fixed';
+		textArea.style.left = '-9999px';
+		textArea.style.top = '0';
+		document.body.appendChild(textArea);
+
+		textArea.focus();
```
ryu-man, line 42: I think the currently focused element should be saved before switching to the textarea, then switch back at the end of the execution

ui/user/src/lib/components/CopyButton.svelte
```diff
@@ -29,16 +29,51 @@
 	let buttonTextToShow = $state(untrack(() => buttonText));
 	const COPIED_TEXT = 'Copied!';
 
-	function copy() {
+	function fallbackCopy(textToCopy: string): boolean {
+		const textArea = document.createElement('textarea');
+		textArea.value = textToCopy;
+
+		textArea.style.position = 'fixed';
+		textArea.style.left = '-9999px';
```
ryu-man, line 38: If this line is used to hide the element, I wonder if it is possible to use opacity or z-index if that makes sense!

ui/user/src/lib/components/CopyButton.svelte
```diff
@@ -29,16 +29,51 @@
...
+		textArea.focus();
+		textArea.select();
+
+		try {
+			// is deprecated but still works for those without navigator.clipboard context
+			return document.execCommand('copy');
+		} catch {
+			return false;
+		} finally {
+			document.body.removeChild(textArea);
```
ryu-man, line 51: make sure to return the focus to the previous element

g-linville (approved):

ui/user/src/lib/components/CopyButton.svelte
ivyjeong13: :+1: Can do this!

PR #5615, opened 2026-01-23
chore: add basic networkpolicy to prevent unwanted mcp server communication

drpebcak:
relates to https://github.com/obot-platform/obot/issues/5616

### DIFF
--- a/chart/templates/networkpolicy.yaml
+++ b/chart/templates/networkpolicy.yaml
@@ -1,0 +1,59 @@
+{{- if and (eq .Values.config.OBOT_SERVER_MCPRUNTIME_BACKEND "kubernetes") .Values.mcpNamespace.networkPolicy.enabled }}
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: mcp-egress-restrictions
+  namespace: {{ include "obot.config.mcpNamespace" . }}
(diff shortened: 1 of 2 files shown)


chart/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - to:
+        # Block private IPv4 ranges by allowing everything except:
+        # 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16, 127.0.0.0/8, 169.254.0.0/16, 224.0.0.0/4, 240.0.0.0/4
+        - ipBlock:
+            cidr: 0.0.0.0/0
+            except:
+              - 10.0.0.0/8       # Private network
+              - 172.16.0.0/12    # Private network
+              - 192.168.0.0/16   # Private network
+              - 127.0.0.0/8      # Loopback
```
thedadams, line 55: Do we want to block loopback? I guess `localhost` would still work with this?

drpebcak: localhost is included in that range of ips, but honestly im not sure if it would actually be blocked inside the pod - networkpolicy works at the cni level, but loopback traffic never leaves the pod's network namespace.

i dont think theres much concern over the mcp server being able to talk to itself anyway.

Issue #5619, opened 2026-01-23
Search action redirects to Create MCP Server page instead of showing search results


shwetapsdet:
**Describe the bug**
When searching for a newly created MCP server (any type, including Composite, Remote, etc.) on the MCP Server page, pressing the Enter key in the search field incorrectly navigates the user to the Create MCP Server / Create Composite Server page instead of displaying the filtered search results.

**To Reproduce**
1. Go to the MCP Server page.
2. Create any type of MCP server (Composite or other).
3. Navigate back to the MCP Server list page.
4. Enter the name of the newly created MCP server in the Search field.
5. Press the Enter key.
6. Observe that the page redirects to the Create MCP Server / Create Composite Server page.

**Expected behavior**
search field should apply the search filter and display the matching MCP server(s) in the list, without navigating away from the MCP Server page.

**Screenshots**

https://www.loom.com/share/d70cd6fce81c44eb856002f699bcc03c

sangee2004:
This issue is addressed when testing with the latest build for any MCP server type.

https://www.loom.com/share/b62c2f22ff8649d68b11ad9a7d4e2272

PR #5622, opened 2026-01-23
fix: prevent reopening create form on search query update

ryu-man:
Addresses #5619

### DIFF
--- a/ui/user/src/routes/admin/mcp-servers/+page.svelte
+++ b/ui/user/src/routes/admin/mcp-servers/+page.svelte
@@ -95,7 +95,7 @@
 				if (page.url.searchParams.has('new')) {
 					const cleanUrl = new URL(page.url);
 					cleanUrl.searchParams.delete('new');
-					goto(cleanUrl);
+					goto(cleanUrl, { replaceState: true });
 				}
 
 				return;
@@ -111,7 +111,7 @@
 				if (page.url.searchParams.has('new')) {
 					const cleanUrl = new URL(page.url);
 					cleanUrl.searchParams.delete('new');
-					goto(cleanUrl);
+					goto(cleanUrl, { replaceState: true });
 				}
 			}
 		}


ui/user/src/routes/admin/mcp-servers/+page.svelte
```diff
@@ -226,14 +225,16 @@
 
 	const duration = PAGE_TRANSITION_DURATION;
 
-	const updateSearchQuery = debounce((value: string) => {
+	const updateSearchQuery = (value: string) => {
 		const newUrl = new URL(page.url);
 
 		setUrlParam(newUrl, 'query', value || null);
+		// Remove the 'new' parameter to prevent reopening the create form
+		setUrlParam(newUrl, 'new', null);
```
ivyjeong13: I'm wondering if moving the clearing of `new` out into the `onSubmit`, before the user is navigated to the created MCP server, would work?

Have you checked if this bug also occurs if you return quickly from creating and click and type into `Name` or `Type` filter?

ivyjeong13: https://www.loom.com/share/bfa533a6afef43d1b980701c497003fc Loom just in case but I am seeing the `new` in the searchParams after returning from creating a new server, we'll want it make it drops on filtering as well if we can't get rid of it on the onSubmit!

ryu-man: >Have you checked if this bug also occurs if you return quickly from creating and click and type into Name or Type filter?

I haven't tested using filter after returning back but I got what you mean in the loom video

ryu-man: > I'm wondering if moving the clearing of new out into the onSubmit, before the user is navigated to the created MCP server, would work?

That would change the UX! I am thinking of a different approach. Clearing after variable is set to false

ryu-man:
Ended up using `goto` instead of `replaceState` inside `afterNavigate` with checks for `new` param existence to prevent infinite loop

ui/user/src/routes/admin/mcp-servers/+page.svelte
```diff
@@ -226,14 +234,14 @@
...
+	const updateSearchQuery = (value: string) => {
 		const newUrl = new URL(page.url);
 
 		setUrlParam(newUrl, 'query', value || null);
 
 		persistQueryToLocalStorage(view, value);
 		navigateWithState(newUrl);
-	}, 100);
+	};
 </script>
```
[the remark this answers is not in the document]
ryu-man: @copilot We should create an issue to keep all search inputs in all pages consistent. Can you create an issue for me?

PR #5623, opened 2026-01-23
chore: remove stutter in add component confirm dialog

njhale:


sangee2004:
This issue is addressed when testing with the latest build from `main`.

The following text is now presented:

Screenshot 2026-01-26 at 12 05 06 PM

PR #5626, opened 2026-01-26
obot-platform/obot#5563 - Add model selection for task execution

### LINKED ISSUE #5563: Provide ability to set a model for a task
We provide users with the ability to select a model from the list of models granted access via model access policies when chatting.

Image

Currently, Users don't have the ability to select a model for task execution. 

It would be good to allow users to select a model from the list of models they have access to for task execution. They should be allowed to pick a model for a task they created, and the expectation is that this model be used for manual/scheduled task execution.

codeaucafe:
## Summary

- Add ability for users to select a specific model when configuring a task, overriding the default Project → Agent → System hierarchy
- Implement TaskModelSelector component with dropdown grouped by model provider, following the existing ThreadModelSelector pattern
- Add unit tests for model field preservation through Task ↔ Workflow conversions

Screenshot 2026-01-26 at 12 57 42 AM

Closes #5563

### DIFF
--- a/pkg/api/handlers/tasks_test.go
+++ b/pkg/api/handlers/tasks_test.go
@@ -1,0 +1,155 @@
+package handlers
+
+import (
+	"testing"
+
+	"github.com/obot-platform/obot/apiclient/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestToWorkflowManifest_ModelFields(t *testing.T) {
+	tests := []struct {
+		name             string
+		input            types.TaskManifest
+		expectedModel    string
+		expectedProvider string
+	}{
+		{
+			name: "preserves model fields when both are set",
+			input: types.TaskManifest{
+				Name:          "test-task",
+				Description:   "A test task",
+				Model:         "claude-3-opus",
+				ModelProvider: "anthropic-model-provider",
+				Steps:         []types.TaskStep{{Step: "Do something"}},
+			},
+			expectedModel:    "claude-3-opus",
+			expectedProvider: "anthropic-model-provider",
+		},
+		{
+			name: "handles empty model fields",
+			input: types.TaskManifest{
+				Name:        "task-no-model",
+				Description: "Task without model selection",
+				Steps:       []types.TaskStep{{Step: "Step 1"}},
+			},
+			expectedModel:    "",
+			expectedProvider: "",
+		},
+		{
+			name: "preserves model when provider is empty",
+			input: types.TaskManifest{
+				Name:  "task-model-only",
+				Model: "gpt-4",
+			},
+			expectedModel:    "gpt-4",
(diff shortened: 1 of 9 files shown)


cjellick:
Thanks...this looks decent. Out of curiosity: how vibecoded was it on a scale of 1 to 10. 1 being written completely by humans 10 being you didn't really even look at the code once it was generated (or feel free to throw out my scale and answer whatever way makes sense to you).

pkg/api/handlers/tasks_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+import (
+	"testing"
+
+	"github.com/obot-platform/obot/apiclient/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestToWorkflowManifest_ModelFields(t *testing.T) {
```
cjellick, line 11: Think these tests are adding much value?

codeaucafe:
@cjellick no problem. I wouldn't say it's vibe coded since I hate letting AI write code and me never look at it which never works in the long run imo. I personally dislike the vibe code way of things and think you still need an human engineer in the loop.

Frontend was all coded by Claude Code since I have no experience with svelte (mostly just react since all my experience is in the backend and Go and Python). I read through it all and from what little I know about svelte (quick start tutorial haha) it seemed alright. I never submit code generated by AI unless I've read it and fully understand it and ran tests myself, although in this case I had to trust the AI much more than I normally do when I'm working in Go at work or other projects. 

The backend code was more like half-in-half in that I asked Claude how tasks work and how models are inputted into chat currently, then once I understood that I had it make a plan to implement and i reviewed all that code to make sure it made sense, there were no issues, and i wasn't missing anything. I wrote the code mostly to save some tokens and because it was easy based on the implementation plan markdown I reviewed. If it was more verbose I would have let it generate it before reviewing and understanding everything before committing. I don't trust AI to get it right if you pure vibe code.

The Go tests it generated completely but it also spit out really useless ones I removed. Not really sure the current test is good either beyond just a sanity check though tbh. Good example why I don't like vibe coding: it didn't even write the tests with testify's require and assert like yall do in other tests so I had to fix that.

Luckily Claude helped me with generate call since everything looked right but wasn't working until I did that after it helped me understand the error and logs

pkg/api/handlers/tasks_test.go
codeaucafe: Not really to be honest. Mostly just a sanity check.

cjellick:
We are looking to completely overhaul our task system, so I dont want to take on any new changes to this area.

codeaucafe:
Thanks @cjellick. That makes perfect sense. One of your colleagues mentioned the same thing too when I first made the PR and linked it in your Discord.

Thank you for taking the time to review the PR. I appreciate it!

Issue #5482, opened 2026-01-12
A PUP user with auditor role is not provided with option to create new registry entries or delete/update existing registry entries in their own workspace.


sangee2004:
**Describe the bug**
A PUP user who has been granted the auditor role is incorrectly restricted from managing MCP Registries within their own workspace. On the MCP Registries page, the user does not see the “Add New Registry” option, nor any delete actions for registries they own. Additionally, when opening the detail view of an existing registry, the Update option is missing.

**To Reproduce**
1. Log in as a PUP user who has the auditor role granted.
2. Access the `MCP Registries` page

Notice that the user is not presented with the option to create new registry entries in his own workspace. `Add Nee Registry` option is missing
Also, notice that no delete option is available for registry entries owned by this user.

<img width="1342" height="732" alt="Image" src="https://github.com/user-attachments/assets/3307ad56-b2d6-45cc-bfff-a3f0c3e8a644" />

3. Open the detail view of one of the registries that they use.
Notice that there is no `Update` option presented in this view.

<img width="1041" height="750" alt="Image" src="https://github.com/user-attachments/assets/0b143ac3-89bf-46b5-91f4-29a26aa2ac6b" />

**Expected behavior**
A PUP user with an auditor role should be able to create new registry entries or delete/update existing ones in their own workspace.

cjellick:
im just going to move this back to High. I dont want more effort put into this right now.

PR #5649, opened 2026-01-27
fix: pup auditor missing create registry btn & update all pages handling new param

ivyjeong13:
Addresses #5482 

* also updates create screens to use derived on page `new` param & remove after creation


ui/user/src/lib/components/admin/usage/UsageGraphs.svelte
```diff
@@ -492,14 +492,6 @@
 		return display;
 	}
 
-	afterNavigate(() => {
```
ivyjeong13, line 495: Noticed this duplicate, there is an `afterNavigate` also calling exactly this on line 451. Went ahead and removed the duplicate!

ivyjeong13:
Gonna close this PR and open new one against latest since this is a lil out of date.

Issue #5634, opened 2026-01-26
Connected Server “Copy” button shows incorrect copied state until hover


cjellick:
The “Copy” button in the Connected to Server modal can display an incorrect “copied” state. After copying one item, navigating away, and then navigating to the Connect to Server modal for another item,  the copy button in the text box keeps the "Copied!" from the last time I clicked opcy. The state corrects itself only after hovering over the input/text box, which suggests the copied state is not being reset properly.

This makes the UI misleading, as it appears to confirm a copy action that did not actually occur.

### Steps to Reproduce

1. Open the Connect to Server modal.
2. Click **Copy** on the first server/item.
3. Navigate away or interact elsewhere (e.g., paste the value somewhere else).
4. Get into the Connect to Server modal for a different item.
5. Observe that this modal still has the **Copied!** text.

### Expected Behavior

"Copied" state should be clear whenever you get to the Connect to Server modal

### Actual Behavior

"Copied!" text doesn't clear until you hover on the button

Reproed here: https://www.loom.com/share/20b94e7a30754d35916850d68ee45262

cjellick:
Id also like to fix the "jitter" caused by inserting the "Copied!" text - https://www.loom.com/share/b4f4350881b54663afd0c298334cca34

shwetapsdet:
Verified it on Obot main and it's working as expected

https://github.com/user-attachments/assets/cd72e8d5-d1ce-4d04-a968-ad9784c9a5f7

PR #5651, opened 2026-01-27
fix: clears copied status on open close

ryu-man:
Addresses #5634

In this PR:
- clears copied state when dialog is closed
- truncate url to prevent layout shift

### DIFF
--- a/ui/user/src/lib/components/mcp/ConnectToServer.svelte
+++ b/ui/user/src/lib/components/mcp/ConnectToServer.svelte
@@ -76,7 +76,7 @@
 	let copyButtonController = $state<ReturnType<typeof CopyButton>>();
 
 	function handleOnClose() {
-		copyButtonController?.clearLabel();
+		copyButtonController?.clearButtonText();
 		onClose?.();
 	}
 
--- a/ui/user/src/lib/components/CopyButton.svelte
+++ b/ui/user/src/lib/components/CopyButton.svelte
@@ -78,7 +78,7 @@
 		}
 	}
 
-	export function clearLabel() {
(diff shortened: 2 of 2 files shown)


ui/user/src/lib/components/mcp/ConnectToServer.svelte
```diff
@@ -73,6 +73,13 @@
 			.map((name) => name.toLowerCase())
 	);
 	let name = $derived(server?.alias || server?.manifest.name || '');
+	let openStatusHash = $state<string>();
+
+	function handleOnClose() {
```
ivyjeong13, line 78: I'm wondering if we just make the `Copied!` also disappear after a short timeout or if the element gets hidden?

ryu-man: I thought about it but came to the conclusion that keeping the 'copied' state displayed would indicate better that the user has already copied the link. What do you think about that?

>or if the element gets hidden

I could not find an easy way to access the dialog show/hide state reactively! that's why I had to remount the button on each close

ivyjeong13: Yeah I agree, I like the "Copied!" text being up so long until they go away.

For the second part, maybe something as simple as you could pass a `show` prop to CopyButton that renders the contents of the CopyButton.svelte when it's true instead? I'm just not a fan of the random hash being used here.

ryu-man: How about using component binding and call clear function on close
```
copyButtonController?.clearLabel();
```

PR #5655, opened 2026-01-27
enhance: styling ux tweaks for static oauth

ivyjeong13:
* fixes Requires Oauth Config not being sorted when clicking Descending/Ascending for Status column
* tweaks pill styling for Requires Oauth Config
Screenshot 2026-01-27 at 1 58 43 PM

* shows notification alert at top when user goes into a catalog entry and it has requires oauth config to make it easier to set up Static Oauth configuration
Screenshot 2026-01-27 at 1 58 52 PM

* when visiting Configuration tab for existing remote catalog entry with just static oauth enabled, Advanced Configuration was collapsed when it should be expanded since a value inside it had been set. Fixed to be expanded.

**NOTE**: Wait for Grant to merge in the static oauth PR or have commit merged in

### DIFF
--- a/ui/user/src/app.css
+++ b/ui/user/src/app.css
@@ -1109,6 +1109,15 @@
 		line-height: 1rem;
 	}
 
+	.pill-warning {
+		border-radius: var(--radius-sm);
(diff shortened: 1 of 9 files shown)


g-linville:
this is excellent, thanks!

Issue #5545, opened 2026-01-16
API Keys -Make `Deleted Server` MCP entry stand out visually.


sangee2004:
**Describe the bug**
After disconnecting an MCP server (e.g., Exa Search) that previously had API keys created, the API Keys UI shows that server in the key’s Servers field as a “Deleted Server” entry (both in the list hover tooltip and in the API key detail view). This is confusing because it looks like a normal server association, but it actually represents a now-disconnected/deleted server reference.

Expected: Visually distinguish the “Deleted Server” entry (e.g., red text and/or italics) in both the API Keys list view (tooltip/hover) and the API key detail page so users immediately recognize it’s no longer an active server.

**To Reproduce**
1. Connect to any MCP Server - say `Exa Search`.
2. Connect to a few other MCP Servers.
3. Create an API key for `Exa Search` and a few other existing connected MCP server instances.
4. From the obot server UI, disconnect `Exa Search`. 
5. Go to the `API Keys` page and hover over the `Servers` field for the api key entry created in step 3

Notice that there is a `Deleted Server`MCP server presented for the disconnected `Exa Search`. 
<img width="975" height="193" alt="Image" src="https://github.com/user-attachments/assets/ac6c995d-6eb6-4d4f-91a7-c93303cca50f" />

Can we make this `Deleted Server` server entry stand out by marking it in Red or italicizing it?
Can the same be done in detail page as well?

<img width="568" height="597" alt="Image" src="https://github.com/user-attachments/assets/408d4a08-abf8-425d-a713-ce748d05f969" />

**Expected behavior**
Make the `Deleted Server` MCP entry stand out visually in the API Keys list view and in the detail view.

sangee2004:
This issue is addressed when testing with the latest builds.

User is presented with `_(Deleted)_` in grey to represent deleted entries in both list view and detail view.

When there are more than 3 MCP entries for an API key and a few are deleted, we see an error indicator that shows the count of deleted servers in the entry.

List view:
<img width="1138" height="251" alt="Image" src="https://github.com/user-attachments/assets/ee7d69a1-0139-4793-9ff0-8da9e84bff58" />

Detail view:

<img width="1122" height="861" alt="Image" src="https://github.com/user-attachments/assets/bb11694b-ef46-43a0-8d87-09d25e414697" />

Issue #5544, opened 2026-01-16
Improvements to API key detail page layout


sangee2004:
**Describe the bug**
Currently, this is how the API key detail page shows up
<img width="568" height="597" alt="Image" src="https://github.com/user-attachments/assets/33bde083-4ac8-430d-bbeb-daf8a219eb5c" />

Make the detail page layout similar to how MCP server details are presented when clicking on the `configuration` tab for MCP server entries in the `MCP Servers` page

<img width="1051" height="823" alt="Image" src="https://github.com/user-attachments/assets/c194d2d1-7c21-4d87-8c48-94487ab2cbc8" />

cjellick:
I really don't mind the modal how it is and I don't quite understand what @sangee2004 is asking for here as far as similarity to the configuration page goes, but @ivyjeong13 take a look at this modaly anyway and just see if there's something you'd change about it.

sangee2004:
This issue is addressed when testing with the latest builds.

When clicking on any entry in the `API Keys` list view, the user is taken to a full page detail view where details are presented in the format below, which matches how the detail view is being presented in the rest of the pages.

<img width="1127" height="847" alt="Image" src="https://github.com/user-attachments/assets/c0d7ad29-f9f7-4b64-b26c-95db887d9e02" />

Issue #5483, opened 2026-01-12
3 lines taken up for displaying keys in `API Keys` list view.


sangee2004:
**Describe the bug**
3 lines taken up for displaying keys in the `API Keys` list view.

**To Reproduce**
1. Create a few API keys
2. Go to the `API Keys` option under the `User Management` tab to access the `API Keys` list view.

Notice that 3 lines are taken up for displaying keys in the `API Keys` list view.
<img width="1272" height="430" alt="Image" src="https://github.com/user-attachments/assets/0f7cdea6-6223-4b4d-b1fa-eb301434b4c8" />

**Expected behavior**
Display the keys field in one line.

sangee2004:
This issue is addressed when testing with the latest build.

1 line is taken up for displaying keys in the `API Keys` list view as expected, even in a smaller screen size.

<img width="1114" height="756" alt="Image" src="https://github.com/user-attachments/assets/7cc2f1b5-69b6-4431-9bff-8fe5960417bc" />

Issue #5466, opened 2026-01-09
API Keys - Servers field spans 2 lines to display the count for server count more than 1.


sangee2004:
**Describe the bug**
API Keys - When the server count is more than 1, the servers field spans 2 lines to display the count.

**To Reproduce**
1. Create an API key by selecting more than 1 MCP server.
Notice that when the server count gets displayed in the `Servers` field, it takes up 2 lines.

<img width="1339" height="349" alt="Image" src="https://github.com/user-attachments/assets/457e1e60-ca88-44c9-a463-259e3729ccac" />

**Expected behavior**
The `Servers` field should be presented in 1 line for any server count.

cjellick:
@sangee2004 these steps are not sufficient to reproduce. I believe the key thing you left out is that you have to have a long-ish descriptoin

No description:

<img width="1333" height="349" alt="Image" src="https://github.com/user-attachments/assets/e3f6beae-d236-4af2-a548-4c59974afd25" />

With description:

<img width="1429" height="517" alt="Image" src="https://github.com/user-attachments/assets/419480b1-f63f-4c21-8b27-36ddcc877c85" />


In the future, please ensure the Steps to Reproduce capture all the necessary steps.

sangee2004:
@cjellick When I had logged the issue, I saw that even entries with no description were taking 2 lines for displaying the `Servers` field. The first entry in my screenshot shows that. That was why I thought it wasn't related to the long description. I will make sure to add all necessary steps.

When looking at this again, I realized it happens when I am not in full-screen mode, and the `Server` field ends up being allocated just enough space to accommodate `1 server` (8 characters) on 1 line. Any other values like "<n> servers" or "All My Servers` end up taking 2 lines.

<img width="1004" height="728" alt="Image" src="https://github.com/user-attachments/assets/8614f165-5108-47b5-9502-c51d752ccb53" />

Should we consider the `Server` field to be renamed as `Server Count` and display only the "count" or "All" as values?

This is more of an issue with the admin user's access to the API Key list view via the' API Keys` option under the `User Management` section, where we display the' User` field as well.

sangee2004:
When testing with the latest builds, we are now presented with the actual server list for the `Servers` field, and the issue tracked here is irrelevant.

<img width="1106" height="616" alt="Image" src="https://github.com/user-attachments/assets/af93cc7c-7277-44b4-86e5-3e581ce9388d" />

PR #5659, opened 2026-01-27
fix: API Keys page related bugs

ivyjeong13:
Addresses #5461
Addresses #5483 
Addresses #5466 
Addresses #5545 

- [x] Address #5544

### DIFF
--- a/ui/user/src/routes/admin/api-keys/+page.svelte
+++ b/ui/user/src/routes/admin/api-keys/+page.svelte
@@ -144,7 +144,7 @@
 						{:else if property === 'expiresAt'}
 							{d.expiresAtDisplay}
 						{:else if property === 'prefix'}
-							<span class="truncate whitespace-nowrap">{d.prefix}</span>
+							<span class="whitespace-nowrap">{d.prefix}</span>
 						{:else}
 							{d[property as keyof typeof d]}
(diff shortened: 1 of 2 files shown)


ui/user/src/lib/components/api-keys/ServersLabel.svelte
```diff
@@ -0,0 +1,76 @@
...
+	let displayItems = $derived.by((): DisplayItem[] => {
+		if (isAllServers) return [];
+		const items: DisplayItem[] = resolvedServers.map((name) => ({ name, deleted: false }));
+		for (let i = 0; i < deletedServersCount; i++) {
+			items.push({ name: 'Deleted', deleted: true });
+		}
+		return items;
+	});
+</script>
+
```
ryu-man, line 47: Not a blocker, but is it possible to display the server name along with the deleted indicator? At least the user can know which server was deleted!

Image

g-linville: No, we don't have this information available to us after servers are deleted. I talked about this with Craig at some point and we decided it is not worth it to change that to make the information recoverable.

g-linville (approved):
Thank you for cleaning up my vibe mess :+1:

PR #5669, opened 2026-01-28
feat: re-introduce explicit tool call approval

njhale:
## Summary

This branch re-introduces **user-facing tool call confirmation** for Obot Chat. Previously, when an LLM decided to call an MCP tool, it executed immediately. Now, each tool call pauses and prompts the user for approval before executing.

## How It Works

When the LLM emits a tool call during a chat session, GPTScript's `Confirm` mode intercepts it before execution. The backend emits a `ToolConfirm` event over the existing progress/event stream, containing the tool name, description, and input arguments. The frontend renders a confirmation bar at the bottom of the chat where the user can:

- **Deny** — reject the call (the run continues without executing it)
- **Allow** — approve this single invocation
- **Allow [tool] for this thread** — auto-approve future calls to this specific tool for the rest of the thread
- **Allow all for this thread** — auto-approve all tool calls for the rest of the thread (wildcard `*`)

Thread-scoped approvals are persisted in `thread.spec.approvedTools`. On subsequent tool calls, any tool matching that list is auto-confirmed without prompting.

If the user doesn't respond within **5 minutes**, the run is cancelled.

### What's auto-approved (never prompts)

- System tasks and workflow runs
- Tools in non-user categories (built-in/internal tools)
- Tools already in the thread's `approvedTools` list

## Override Controls

Confirmation can be disabled at multiple levels, evaluated in order:

1. **Global server flag**: Set `OBOT_SERVER_ENABLE_AUTONOMOUS_TOOL_USE=true` to disable all prompts server-wide. This restores pre-change behavior.
2. **Per-user account setting**: Users can toggle "Allow Autonomous Tool Use" on their profile page (`/profile`). This sets `autonomousToolUseEnabled` on their user record.
3. **Per-thread**: Approving with the wildcard `*` during a session auto-approves everything for that thread.

The profile toggle is hidden when the global flag is already enabled (since it would be redundant).

## New API

| Endpoint | Method | Description |
|---|---|---|
| `POST /assistants/{id}/projects/{id}/threads/{id}/confirm` | POST | Submit a `ToolConfirmResponse` with `id`, `decision` (`deny`/`approve`/`approve_thread`), and optional `toolName` |

The `/version` endpoint now includes `autonomousToolUseEnabled` so the frontend knows whether to show the profile toggle.

The `PATCH /me` endpoint now accepts `autonomousToolUseEnabled` to update the user preference.

## New Configuration

| Env Var | Default | Description |
|---|---|---|
| `OBOT_SERVER_ENABLE_AUTONOMOUS_TOOL_USE` | `false` | Globally disable tool call confirmation |

## Migration Considerations

- **Breaking UX change**: After upgrading, all chat tool calls will require user approval by default. To preserve the previous automatic execution behavior, set `OBOT_SERVER_ENABLE_AUTONOMOUS_TOOL_USE=true`.

Addresses https://github.com/obot-platform/sensitive-issues/issues/55


njhale:
Not ready yet, just seeing if I can get copilot to review

pkg/invoke/invoker.go
```diff
@@ -1142,6 +1165,51 @@ func (i *Invoker) stream(ctx context.Context, gptClient *gptscript.GPTScript, c
 
 			if frame.Call != nil {
 				switch frame.Call.Type {
+				case gptscript.EventTypeCallConfirm:
+					var (
+						callID   = frame.Call.ID
+						toolName = frame.Call.Tool.Name
+					)
+
+					// Auto-confirm pre-approved tools
```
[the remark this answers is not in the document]
njhale: This was added to prevent gptscript context tools, inputs, model providers, and non-user invoked tools calls from requiring approval.

apiclient/types/invoke.go
```diff
@@ -128,3 +130,39 @@ type ToolCall struct {
...
+	// ID is the ID of the
+	ID string `json:"id"`
+
+	// Decision indicates if the tool call should be denied, approved, or approved for the duration of the thread.
+	Decision ToolConfirmDecision `json:"decision"`
+
+	// ToolName is the name of a tool to pre-approve for the duration of the thread.
+	// Set to the wildcard ("*") to pre-approve ALL tools.
+	// This field may only be set when Decision is ToolConfirmDecisionApproveThread.
+	ToolName string `json:"threadApprovedTools,omitempty"`
```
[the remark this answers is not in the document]
njhale: This should be `ToolName`, the tag was from a previous refactoring

pkg/gateway/server/user.go
```diff
@@ -171,14 +171,20 @@ func (s *Server) getUser(apiContext api.Context) error {
 func (s *Server) updateUser(apiContext api.Context) error {
 	userID := apiContext.PathValue("user_id")
 	if userID == "" {
-		return types2.NewErrHTTP(http.StatusBadRequest, "user_id path parameter is required")
+		// This is a request to /api/me
+		userID = apiContext.User.GetUID()
 	}
```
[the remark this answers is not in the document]
njhale: Not too concerned about this edge case since it fails closed.

pkg/invoke/invoker.go
```diff
@@ -682,6 +693,17 @@ func (i *Invoker) Resume(ctx context.Context, gptClient *gptscript.GPTScript, c
 		return fmt.Errorf("failed to resolve model provider: %w", err)
 	}
 
+	// Disable tool call confirmation when any of the following hold:
+	// 1. We're invoking a system task or a workflow
+	// 2. Autonomous tool use has been enabled globally (OBOT_SERVER_ENABLE_AUTONOMOUS_TOOL_USE = true)
+	// 3. Autonomous tool use has been enabled at the user's account-level
+	// 4. The user has pre-approved all tool calls (ApprovedTools contains the wildcard operator "*")
```
g-linville: nit: the comment should also include whether this is a workflow execution as one of the conditions

```suggestion
	// Disable tool call confirmation when any of the following hold:
	// 1. We're invoking a system task or a workflow
	// 2. This is a workflow execution
	// 3. Autonomous tool use has been enabled globally (OBOT_SERVER_ENABLE_AUTONOMOUS_TOOL_USE = true)
	// 4. Autonomous tool use has been enabled at the user's account-level
	// 5. The user has pre-approved all tool calls (ApprovedTools contains the wildcard operator "*")
```

g-linville (approved):
backend lgtm :+1:

pkg/api/authz/resources.go
```diff
@@ -216,7 +217,7 @@ var apiResources = map[string][]string{
 		"GET    /api/threads/{thread_id}/tasks/{task_id}/runs/{run_id}",
 		"GET    /api/threads/{thread_id}/workflows",
 		"GET    /api/threads/{thread_id}/workflows/{workflow_id}/executions",
-		//
+
```
thedadams: Please leave this in or add to the comment that these are legacy APIs.

pkg/api/router/router.go
```diff
@@ -295,7 +298,6 @@ func Router(ctx context.Context, services *services.Services) (http.Handler, err
 	mux.HandleFunc("POST /api/invoke/{id}", invoker.Invoke)
 	mux.HandleFunc("POST /api/invoke/{id}/thread/{thread}", invoker.Invoke)
 	mux.HandleFunc("POST /api/invoke/{id}/threads/{thread}", invoker.Invoke)
-	//
```
thedadams, line 298: Same comment here about not removing.

pkg/api/handlers/confirm.go
```diff
@@ -0,0 +1,61 @@
...
+		return err
+	}
+
+	approved := true
+	switch confirm.Decision {
+	case types.ToolConfirmDecisionDeny:
+		approved = false
+	case types.ToolConfirmDecisionApprove:
+		if confirm.ToolName != "" {
+			return types.NewErrBadRequest("setting tool name for thread approval requires decision %q", types.ToolConfirmDecisionApproveThread)
```
thedadams: I think this should be removed. The error refers to a different approval status.

pkg/invoke/invoker.go
```diff
@@ -627,14 +632,20 @@ func (i *Invoker) Resume(ctx context.Context, gptClient *gptscript.GPTScript, c
 	}
 
 	// For runs that are not from a user, ensure the token has the basic and authenticated groups.
-	userGroups := []string{types.GroupBasic, types.GroupAuthenticated}
-	var userID, userName, userEmail, userTimezone string
+	var (
+		userAutonomousToolUseEnabled              *bool
+		userID, userName, userEmail, userTimezone string
+		userGroups                                = []string{types.GroupBasic, types.GroupAuthenticated}
```
thedadams: The comment above this `var` group should be moved above this line.

ui/user/src/lib/components/messages/ToolConfirmBar.svelte
```diff
@@ -0,0 +1,166 @@
...
+		} catch {
+			return jsonString;
+		}
+	}
+</script>
+
+{#if current}
+	{@const dropdown = popover({ placement: 'bottom-end' })}
+	<div
+		class="mb-2 w-full max-w-[900px] overflow-hidden rounded-xl bg-gray-900 px-5 shadow-lg dark:bg-gray-900"
```
ivyjeong13, line 76: I see the gray-{value} uses here and wondering if it should be updated to dynamic theming if the user changes Branding color values? Not a blocker cause I think some components of chat also haven't been updated to the bg-background bg-surface1 etc values

pkg/api/handlers/confirm.go
njhale: This is the correct error and status.

If you take a look at the godoc comment on the `ToolName` field, it says:

```
// ToolName is the name of a tool to pre-approve for the duration of the thread.
// Set to the wildcard ("*") to pre-approve ALL tools.
// This field may only be set when Decision is ToolConfirmDecisionApproveThread.
ToolName string `json:"toolName,omitempty"`
```

I'll add a comment clarifying why we're sending the error, but if you feel adamant that we shouldn't return an error for this case, then I'll drop it.

pkg/invoke/invoker.go
njhale: done

pkg/api/router/router.go
njhale: done

pkg/api/authz/resources.go
njhale: done

pkg/invoke/invoker.go
njhale: done

pkg/api/handlers/confirm.go
thedadams: I misunderstood what is happening here. My mistake.

ui/user/src/lib/components/messages/ToolConfirmBar.svelte
```diff
@@ -0,0 +1,166 @@
...
+	});
+
+	let displayName = $derived(current?.message.sourceName || current?.confirm.toolName || '');
+
+	let isSubmitted = $state(false);
+	let isExpanded = $state(false);
+
+	// Reset state when the current confirm changes
+	$effect(() => {
+		if (current) {
```
ryu-man, line 38: What if the current confirm is undefined? the state will not reset, skip if this intentional

njhale: if current is `undefined` then the entire component is also hidden. Once a new current is set, the state will be reset.

Issue #5683, opened 2026-01-29
Improve MCP observability for Basic users


g-linville:
Here are some changes we are making for what all users can see about their own MCP servers.

("Their own MCP servers" means single-user servers they created, whether the catalog entries they came from are in the default catalog or a PowerUserWorkspace.)

- All users can upgrade their own servers
  - "Upgrade" here refers to pulling changes to the upstream catalog entry and applying them to the MCP server
- All users can view the logs of their MCP servers (meaning the container logs from k8s/Docker)
- All users can view usage stats for their MCP servers
- All users can view audit logs (including both metadata and payloads) for their MCP servers
- All users can restart their own servers

Issue #5498, opened 2026-01-14
A PUP user is unable to see `Audit Logs` and `Usage` for MCP servers in his own workspace.


sangee2004:
**Describe the bug**
When a PUP user creates an MCP server in their own registry and makes tool calls through chat, those calls are not reflected in either Audit Logs or Usage. API calls made by UI for both audit logs and usage return empty results.  PUP users should see audit entries and usage stats for MCP tool calls within their own workspace.

**To Reproduce**
1. As a PUP user, create an MCP server in their own registry.
2. Connect to this MCP server and chat with it to make some tool calls.
3. Go to `Audit Logs`.

Note that no audit entries are presented to the user, even though we expect audit entries for the tool calls made in step 2 to be reported here.

API call made from UI, which returns no rows
```
http://localhost:8081/api/mcp-audit-logs?call_type=resources%2Fread%2Ctools%2Fcall%2Cprompts%2Fget&start_time=2026-01-07T00%3A44%3A00.000Z&end_time=2026-01-14T00%3A44%3A59.000Z&limit=10000&offset=0&query=
```
4. Go to `Usage`.
Note that no usage stats are reported for the tool calls made in step 3.

API call made from UI, which returns no usage
```
http://localhost:8081/api/mcp-stats?start_time=2026-01-07T00%3A44%3A59.000Z&end_time=2026-01-14T00%3A44%3A59.000Z
```

Response:
```
{
    "totalCalls": 0,
    "uniqueUsers": 0,
    "timeStart": "2026-01-07T00:44:59Z",
    "timeEnd": "2026-01-14T00:44:59Z",
    "items": null
}
```
**Expected behavior**
PUP users should have access to `Audit Logs` and `Usage` for MCP servers in their own workspace.

g-linville:
I unassigned myself and moved this back to High, because it is not a release blocker. I talked to Craig and got more details about the desired behavior for this:

PUPs should be able to see:
- All metadata **and payloads** for audit logs for their own MCP servers, whether they were deployed from catalog entries in their own workspace, or someone else's/the default catalog.
- All metadata **but NOT payloads** (unless they are also Auditor, of course) for MCP servers that other users created from catalog entries in the PUP's workspace.

g-linville:
We'll continue tracking this as part of https://github.com/obot-platform/obot/issues/5683

PR #5706, opened 2026-02-02
enhance: fix PUP access to audit logs

g-linville:
for https://github.com/obot-platform/obot/issues/5683 and https://github.com/obot-platform/obot/issues/5498

This PR does two things:
- Restore PUP access to audit logs for their multi-user servers
- Give PUPs access to audit logs (metadata only, no payload) for servers that other people create from their catalog entries

### DIFF
--- a/apiclient/types/mcpauditlog.go
+++ b/apiclient/types/mcpauditlog.go
@@ -6,29 +6,29 @@ import (
 
 // MCPAuditLog represents an audit log entry for MCP API calls
 type MCPAuditLog struct {
-	ID                   uint   `json:"id"`
-	CreatedAt            Time   `json:"createdAt"`
-	UserID               string `json:"userID"`
-	MCPID                string `json:"mcpID"`
-	APIKey               string `json:"apiKey,omitempty"`
-	PowerUserWorkspaceID string          `json:"powerUserWorkspaceID,omitempty"`
-	MCPServerDisplayName             string          `json:"mcpServerDisplayName"`
-	MCPServerCatalogEntryName        string          `json:"mcpServerCatalogEntryName"`
-	ClientInfo                       ClientInfo      `json:"client"`
-	ClientIP                         string          `json:"clientIP"`
-	CallType                         string          `json:"callType"`
-	CallIdentifier                   string          `json:"callIdentifier,omitempty"`
-	RequestBody                      json.RawMessage `json:"requestBody,omitempty"`
-	ResponseBody                     json.RawMessage `json:"responseBody,omitempty"`
-	ResponseStatus                   int             `json:"responseStatus"`
-	WebhookStatuses                  []WebhookStatus `json:"webhookStatuses,omitempty"`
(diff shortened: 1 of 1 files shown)


pkg/api/handlers/mcpgateway/auditlog.go
```diff
@@ -295,7 +301,7 @@ func (h *AuditLogHandler) GetAuditLog(req api.Context) error {
 			return types.NewErrForbidden("you do not have access to this audit log")
 		}
 
-		// Full payload only for OWN servers (not workspace servers)
+		// Full payload only for OWN servers (not workspace servers or catalog entry workspace servers)
 		canAccessFullPayload = isOwnServer
 	}
```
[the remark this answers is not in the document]
g-linville: Headers are part of metadata, and sensitive ones are redacted.

pkg/api/handlers/mcp.go
```diff
@@ -741,16 +741,25 @@ func (m *MCPHandler) SetTools(req api.Context) error {
...
+	if mcpServer.Spec.MCPServerCatalogEntryName != "" {
 		var entry v1.MCPServerCatalogEntry
 		if err := req.Get(&entry, mcpServer.Spec.MCPServerCatalogEntryName); err != nil {
 			return fmt.Errorf("failed to get MCP server catalog entry: %w", err)
 		}
-		catalogName = entry.Spec.MCPCatalogName
+		if catalogName == "" {
+			catalogName = entry.Spec.MCPCatalogName
+		}
+		mcpServerCatalogEntryWorkspaceID = entry.Spec.PowerUserWorkspaceID
```
thedadams: My understanding is that when the entry is in a power user workspace, then `entry.Spec.MCPCatalogName` is `""`. Can we set `catalogName = entry.Spec.PowerUserWorkspaceID`?

This was the intention in the implementation, at least. Does having an extra field help us here?

The only thing I can think of is that you are trying to determine which logs belong to multi-user servers versus single-user servers. The intention there was that you would be able to use the catalog name and the catalog entry name to determine that.

g-linville: Yes, I was trying to distinguish single-user created from a catalog entry in a PUP workspace, vs. multi-user created in a PUP workspace. I think the extra field makes this distinction clearer rather than relying on a catalog name and catalog entry name, personally.

g-linville: I think it is also less likely to confuse future coding agents.

thedadams: I'll be on the record as not liking this extra field, but won't block the fix from going in.

g-linville: I decided to go ahead and remove it because I figured if my explanation was unconvincing then I am probably wrong to want to keep the field, lol.

g-linville:
Sorry for the rebase; I needed to do it to pick up some breaking changes with main in order to continue using the phat image.

Issue #5664, opened 2026-01-27
Static OAuth - Clearing OAuth Credentials causes invalid UI Prompts and allows tool calls to succeed in “Configuration Required” State.


sangee2004:
**Describe the bug**
After an admin clears OAuth credentials for a Remote MCP server, basic users see inconsistent and broken UI states. The chat view shows an empty prompt with a `Save` button, and the server is marked `Configuration Required`, yet users can still trigger successful tool calls. Additionally, users can access `Edit Configuration`, which displays an empty form and fails with a 400 error on update. Overall, the server state, UI, and backend behavior are out of sync after OAuth credentials are cleared.

**To Reproduce**
1. As an admin user, create a Remote MCP server entry with static OAuth configuration enabled
2. Configure OAuth credentials when prompted for them.
3. As a basic user, connect to this MCP server and chat with it.
4. As an admin user, use the`Configure Oauth` option and then use the `Clear Credentials` option.
5. Choose "Yes, I'm sure" when presented with the confirmation

<img width="528" height="371" alt="Image" src="https://github.com/user-attachments/assets/5e5ca1dc-edae-481e-81cd-a40bc52d109f" />

6. As the basic user from step3, access the chat view from step 3
Notice that the user is presented with an empty message with the `Save` option.
Also MCP server is put in `Configuration Required` state.

<img width="1463" height="720" alt="Image" src="https://github.com/user-attachments/assets/8a1c5bda-174e-498e-a8e9-42b1ff1612db" />

7. Click on `Save`.
8. Access the remote server entry and click on `Edit Configuration`.
This results in the following empty message with the `Update` option presented to the user
<img width="1462" height="731" alt="Image" src="https://github.com/user-attachments/assets/f1689665-ad4b-4cf8-a2ff-a723710acac8" />

Clicking on Update results in the user being presented with 400 error message:
<img width="1179" height="429" alt="Image" src="https://github.com/user-attachments/assets/cc1a27de-d821-4f6d-98fd-5dbdc5423c13" />

9. Chat and make some tool calls to this MCP server.

Notice that the tool calls actually succeed even when the MCP server is put in `Configuration Required` state.

<img width="1454" height="853" alt="Image" src="https://github.com/user-attachments/assets/a523ef3e-b10e-4fac-9491-6ac4d8f30ea1" />
 
**Expected behavior**
1. The user should not be prompted with an empty message with the `Save` button in step 6
2. The user should not be presented with `Edit Configuration` in step 8
[...]

PR #5721, opened 2026-02-03
fix: ui: show the promper information in chat when a server needs OAuth credentials

g-linville:
for https://github.com/obot-platform/obot/issues/5664


g-linville:
I don't like this implementation

Issue #5715, opened 2026-02-03
Ability to delete an instance of an mcp server from the server details page


cjellick:
If I'm on the servers entry page and I click on a single-user server with some connections, I go to the details for the server entry. Then I click the Server Details tab, I'll see all the individual instances of that server. Then I can click on one of those instances and I see this page:

<img width="1766" height="1010" alt="Image" src="https://github.com/user-attachments/assets/5545b7b2-424c-4a8c-9748-a55e999086e3" />

I want to be able to delete the instance from this screen. I am thinking a button next to the restart button. But maybe we move both restart and delete up to the top right? Like inline with the title? Not sure what will be best.

Also, on the page before where we list then instances for the entry here:

<img width="1483" height="620" alt="Image" src="https://github.com/user-attachments/assets/e09f720d-4cb0-410f-a8f6-884359265fdd" />

We should have a menu of options where "View Audit Logs" that include view audit logs, restart, and delete.

Behind mindful of role permissions. Not all roles will be able to do these actions and you should only show it if they can

ryu-man:
> I want to be able to delete the instance from this screen. I am thinking a button next to the restart button. But maybe we move both restart and delete up to the top right? Like inline with the title? Not sure what will be best.

@cjellick In the top right three-dots menu, there is already an option to remove the current instance of the server; should I move the option out as a button?

<img width="1500" height="346" alt="Image" src="https://github.com/user-attachments/assets/e2b9539d-6444-4131-9f38-0bd4f5f17bcf" />

cjellick:
Ok, heres another loom where I meander through my thinking. its a bit confusing. hopefully by the end its clear.

https://www.loom.com/share/bba412ab2c744a5bbbd19c5370647d23

PR #5722, opened 2026-02-03
feat: add server restart and delete functionality with confirmation dialogs in details tab

ryu-man:
Addresses #5715


cjellick:
going to close this since its still in draft

PR #5743, opened 2026-02-05
fix: distinguish between project version in authorizer

thedadams:


njhale (approved):

njhale:
Question though, should we also update the path pattern for all the project v2 routes registered in `pkg/api/router/router.go` as well?

Issue #5757, opened 2026-02-06
Improve consulting deployment logs user experience


ryu-man:
**Describe the bug**
When deployment logs are crowded, user should be able to use operation such as search and clear, to narrow down results

**To Reproduce**
1. Go to `/admin/mcp-servers`
2. Click on `Deployments & Connections`
3. Click on a deployment entry
4. Scroll down to the `Deployment logs` section
5. Notice that there is no way for the user check the logs in comfort when the logs are crowded

**Expected behavior**
Deployment logs should offer functionalities like, search, highlight results and clear logs to improve user experience

shwetapsdet:
Validate it on obot main and it's working as expected 

https://github.com/user-attachments/assets/0ad9f6ab-e691-4919-b6e2-f876a8ae5e0a

Issue #5756, opened 2026-02-06
introduce the expanded view mode to the MCP server deployment logs


ryu-man:
**Describe the bug**
Make it possible to expand the deployment logs.

**To Reproduce**
1. Go to `/admin/mcp-servers`
2. Click on `Deployments & Connections`
3. Click on a deployment entry
4. Scroll down to the `Deployment logs` section
5. Notice that there is no way for the user check the logs in comfort when the logs are crowded

**Expected behavior**
The user should be able to expand the logs

shwetapsdet:
Verified it on obot main and it's working as expected

https://github.com/user-attachments/assets/d27c062e-43d1-4cbc-a450-02fbd33dea1d

PR #5758, opened 2026-02-06
feat: enhance deployment logs UI and introduce expand/minimize feature

ryu-man:
Addresses #5756 & #5757

Deployment Logs (minimize)
image

Deployment Logs (maximize)
image


ivyjeong13:
I see a couple Copilot comments and wondering if you've addressed the ones that I think may be worth looking into?

ui/user/src/lib/components/admin/DeploymentLogs.svelte
```diff
@@ -0,0 +1,364 @@
...
+			</button>
+		</div>
+	</div>
+
+	<div
+		bind:this={modalContainer}
+		onclick={handleModalClick}
+		class={twMerge(
+			isMaximized
+				? 'fixed inset-0 z-50 flex flex-col items-center justify-center bg-black/50 p-4 md:p-8 lg:p-10'
```
ivyjeong13: nit: `bg-background` instead of `bg-black` for dynamic theming. Seeing some text-gray-500s too

ui/user/src/lib/components/admin/DeploymentLogs.svelte
```diff
@@ -0,0 +1,364 @@
...
+			onscroll={handleUserScroll}
+			bind:this={logsContainer}
+			class={twMerge(
+				'dark:bg-surface1 dark:border-surface3 default-scrollbar-thin bg-background flex min-h-64 flex-col overflow-y-auto rounded-lg border border-transparent shadow-sm',
+				isMaximized ? 'h-full max-h-full w-full' : 'max-h-84 '
+			)}
+		>
+			{#if hasMessages}
+				<div
+					class="dark:bg-surface1 bg-background sticky top-0 z-10 border-b border-gray-200 p-4 dark:border-gray-700"
```
ivyjeong13: Use surface css variable instead of constant grays for dynamic theming

ui/user/src/lib/components/admin/DeploymentLogs.svelte
```diff
@@ -0,0 +1,364 @@
...
+				>
+					<div
+						class={twMerge(
+							'border-surface2 bg-surface1/50 flex h-10 w-full items-center gap-2 rounded-sm border pr-2 pl-2 text-xs focus-within:outline focus-within:outline-2 focus-within:outline-blue-500',
```
ivyjeong13: I think similarly, for dynamic theming, use `primary` css variable instead of blue-500?

ui/user/src/lib/components/admin/DeploymentLogs.svelte
```diff
@@ -0,0 +1,364 @@
...
+		bind:this={modalContainer}
+		onclick={handleModalClick}
+		class={twMerge(
+			isMaximized
+				? 'fixed inset-0 z-50 flex flex-col items-center justify-center bg-black/50 p-4 md:p-8 lg:p-10'
+				: 'contents'
+		)}
+		role={isMaximized ? 'dialog' : undefined}
+		aria-modal={isMaximized ? 'true' : undefined}
+	>
```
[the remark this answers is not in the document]
ivyjeong13: I don't believe this is needed since it has the unique behavior of being `contents` that becomes a `dialog`. I haven't pulled down but wondering if this smoothly stretches out into a fixed dialog when it expands?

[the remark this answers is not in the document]
ryu-man: @ivyjeong13 Yes, when it expands it behaves like a dialog, the contents display is applied only when the maximized bool variable is set to false

PR #5759, opened 2026-02-06
chore: pin phat image on release

drpebcak:


.github/workflows/docker-build-and-push.yml
```diff
@@ -20,6 +20,24 @@ on:
...
+          registry: ghcr.io
+          username: ${{ secrets.GHCR_USERNAME }}
+          password: ${{ secrets.GHCR_TOKEN }}
+
+      - name: Setup crane
+        uses: imjasonh/setup-crane@v0.5
+
+      - name: Copy main phat image to stable tag
+        run: |
+          crane tag ghcr.io/obot-platform/mcp-images/phat:main ${{ github.ref_name }}
```
thedadams, line 39: Does this need to be
```suggestion
          crane tag ghcr.io/obot-platform/mcp-images/phat:main ghcr.io/obot-platform/mcp-images/phat:${{ github.ref_name }}
```

drpebcak: no, this crane command knows you are just adding a tag to a remote image

```
crane tag -h                                         
Tag remote image without downloading it.

This differs slightly from the "copy" command in a couple subtle ways:

1. You don't have to specify the entire repository for the tag you're adding. For example, these two commands are functionally equivalent:

crane cp registry.example.com/library/ubuntu:v0 registry.example.com/library/ubuntu:v1
crane tag registry.example.com/library/ubuntu:v0 v1


2. We can skip layer existence checks because we know the manifest already exists. This makes "tag" slightly faster than "copy".
```

PR #5766, opened 2026-02-06
better reconciliation for k8sSettings

drpebcak:


pkg/controller/handlers/deployment/deployment.go
```diff
@@ -87,39 +87,53 @@ func (h *Handler) UpdateMCPServerStatus(req router.Request, _ router.Response) e
...
-				// If the deployment's hash doesn't match the current K8sSettings hash,
-				// the deployment needs to be updated with new K8s settings
-				if k8sSettingsHash != currentHash {
+			// Only set NeedsK8sUpdate if:
+			// 1. It's not already set
+			// 2. The deployment has a hash (not initializing)
+			// 3. The deployment's hash doesn't match current K8sSettings
+			// 4. The MCPServer's expected hash also doesn't match current K8sSettings
+			//    (if MCPServer already expects the current hash, a redeploy is pending)
+			if !mcpServer.Status.NeedsK8sUpdate && k8sSettingsHash != "" {
```
thedadams: We can remove this and move this into the if-block above.
```suggestion
			if !mcpServer.Status.NeedsK8sUpdate {
```

pkg/mcp/kubernetes.go
```diff
@@ -995,23 +988,86 @@ func (k *kubernetesBackend) restartServer(ctx context.Context, id string) error
...
-		"kubectl.kubernetes.io/restartedAt": time.Now().Format(time.RFC3339),
-		"obot.ai/k8s-settings-hash":         k8sSettingsHash,
+	// Get PSA enforce level for security context decisions
+	psaLevel := GetPSAEnforceLevelFromSpec(k8sSettings)
+
+	// Retry patching up to 3 times to handle cases where:
+	// 1. Strategic merge patch doesn't fully apply all changes (especially when combining resources and PSA settings)
+	// 2. Conflict errors (409) occur due to concurrent updates by controllers
+	const maxPatchRetries = 3
+	for attempt := 0; attempt < maxPatchRetries; attempt++ {
```
thedadams: nit:
```suggestion
	for attempt := range maxPatchRetries {
```

pkg/mcp/kubernetes.go
```diff
@@ -995,23 +988,86 @@ func (k *kubernetesBackend) restartServer(ctx context.Context, id string) error
...
+		if k.deploymentSettingsMatch(&deployment, k8sSettings, psaLevel) {
+			olog.Debugf("deployment %s patched successfully with K8s settings on attempt %d", id, attempt+1)
+			// Settings match, now apply the hash to mark reconciliation complete
+			if err := k.patchDeploymentHash(ctx, &deployment, k8sSettingsHash); err != nil {
+				if apierrors.IsConflict(err) {
+					olog.Debugf("conflict patching hash for deployment %s on attempt %d, retrying", id, attempt+1)
+					continue
+				}
+				return err
+			}
```
thedadams, line 1046: Why do we need to do this?

drpebcak: just trying to guard against drift - we have some annoying issues where patches werent all applying properly and it would leave the deployment in a weird state

Issue #4985, opened 2025-11-11
Memory leak and OOMKilled due to excessive credential polling


woojuini:
## Environment
- **obot version**: v0.13.0
- **Deployment**: Kubernetes (Helm)
- **Initial memory limit**: 4Gi (causes OOMKilled)
- **Current workaround**: 8Gi memory limit
- **Configuration**:
  - 7 Obots configured
  - 1 base agent (a1-obot)
  - GOGC=50, GOMEMLIMIT=3435MiB (82% of 4Gi)
  - ALB idle timeout: 180 seconds
```
GPTSCRIPT_DEBUG=false
OBOT_SERVER_ENABLE_AUTHENTICATION=true
OBOT_GOOGLE_AUTH_PROVIDER_CLIENT_ID=XXX
OBOT_GOOGLE_AUTH_PROVIDER_CLIENT_SECRET=XXX
OBOT_SERVER_AUTH_ADMIN_EMAILS=woojuin@mail.com
OBOT_SERVER_DSN=postgres
OBOT_SERVER_RETENTION_POLICY_HOURS=720
OBOT_OPENAI_MODEL_PROVIDER_API_KEY=sk
OPENAI_API_KEY=sk
ANTHROPIC_API_KEY=sk-ant
OBOT_SERVER_MCPRUNTIME_BACKEND=kubernetes
NAH_THREADINESS=5
OBOT_SERVER_KNOWLEDGE_FILE_WORKERS=5
KINM_DB_CONNECTIONS=5
OBOT_SERVER_TOOL_REGISTRIES=github.com/obot-platform/tools
OBOT_SERVER_ELECTION_FILE=/data/leader.lock
OBOT_SERVER_AUDIT_LOGS_MODE=s3
OBOT_SERVER_AUDIT_LOGS_STORE_S3BUCKET=obot
OBOT_SERVER_AUDIT_LOGS_COMPRESS_FILE=true
OBOT_SERVER_AUDIT_LOGS_USE_PATH_STYLE=true
```
  

## Problem Summary
Opening Obot chat sessions causes immediate memory explosion (1.5-2GB per chat session) that never releases, leading to OOMKilled events. The memory leak is triggered by a flood of simultaneous `/credentials` and `/credentials/reveal` requests.

## Impact
- **Severity**: Critical - Prevents normal usage with multiple chat sessions
- **Memory behavior**:
  - Idle state: ~400MB
  - After single page refresh: ~2500MB (permanent increase)
[...]

PR #5779, opened 2026-02-11
fix: eliminate 2GB pre-allocation in audit logger buffer

kyu1204:
The audit logger pre-allocated a 2GB buffer (MaxFileSize*2) at startup and allocated a new 2GB buffer on every flush cycle, causing constant high memory usage regardless of actual audit log volume.

- Remove upfront 2GB buffer pre-allocation; let append grow as needed
- Reuse previous buffer via spare slice swap instead of allocating new
- Replace capacity-based flush trigger with explicit maxSize threshold

With the default 120s flush interval, the buffer usually contains only a few MB of audit logs. With spare buffer reuse, flush cycles no longer allocate new memory. Previously memory usage stayed around ~2GB even with small logs. Now it scales with actual audit log size (usually just a few MB).

Related Issues: #4985

### DIFF
--- a/pkg/api/server/audit/logger.go
+++ b/pkg/api/server/audit/logger.go
@@ -88,6 +88,10 @@ func New(ctx context.Context, options Options) (Logger, error) {
 		return nil, fmt.Errorf("invalid audit log mode: %s", options.AuditLogsMode)
 	}
 
+	if options.AuditLogsMaxFileSize <= 0 {
+		return nil, fmt.Errorf("audit log max file size must be greater than 0")
+	}
+
 	l := &persistentLogger{
 		kickPersist: make(chan struct{}),
 		store:       s,
@@ -169,7 +173,11 @@ func (l *persistentLogger) persist() error {
 	}
 
 	l.lock.Lock()
-	l.spare = buf[:0]
+	if cap(buf) <= l.maxSize {
+		l.spare = buf[:0]
+	} else {
+		l.spare = nil
+	}
 	l.lock.Unlock()
 
 	return nil
--- a/pkg/api/server/audit/logger.go
+++ b/pkg/api/server/audit/logger.go
@@ -56,7 +56,8 @@ type persistentLogger struct {
 	kickPersist chan struct{}
 	store       store.Store
 	buffer      []byte
-	bufferSize  int
+	spare       []byte // reusable buffer from previous successful persist
+	maxSize     int    // flush trigger threshold (= AuditLogsMaxFileSize)
 }
 
 func New(ctx context.Context, options Options) (Logger, error) {
@@ -88,11 +89,9 @@ func New(ctx context.Context, options Options) (Logger, error) {
 	}
 
 	l := &persistentLogger{
-		lock:        sync.Mutex{},
 		kickPersist: make(chan struct{}),
 		store:       s,
-		bufferSize:  options.AuditLogsMaxFileSize * 2,
-		buffer:      make([]byte, 0, options.AuditLogsMaxFileSize*2),
+		maxSize:     options.AuditLogsMaxFileSize,
 	}
 
 	go l.runPersistenceLoop(ctx, time.Duration(options.AuditLogsMaxFlushInterval)*time.Second)
@@ -109,7 +108,7 @@ func (l *persistentLogger) LogEntry(entry LogEntry) error {
 	defer l.lock.Unlock()
 
 	l.buffer = append(l.buffer, b...)
-	if len(l.buffer) >= cap(l.buffer)/2 {
+	if len(l.buffer) >= l.maxSize {
 		select {
 		case l.kickPersist <- struct{}{}:
 		default:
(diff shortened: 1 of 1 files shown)


yh-kwak:
Great initiative! This change addresses a critical issue for teams and organizations running multiple users. The memory bloat caused by constant 2GB pre-allocation and buffer reallocation on each flush cycle has been a significant pain point. 

With this optimization, memory usage will scale proportionally to actual audit log volume rather than being fixed at 2GB, which should dramatically reduce OOM incidents in multi-tenant environments. The buffer reuse strategy is an elegant solution that improves efficiency without compromising functionality.

Please prioritize the review and merge of this PR - the memory efficiency improvements are essential for production stability in larger deployments.

cjellick:
Thanks for the PR. Can you give more detail about your environment, how you hit this, and how it manifests as. Im curious if there are impacts beyond 2gb of memory being allocated.

kyu1204:
@cjellick 

It was deployed in an EKS environment, and each pod was configured with 4GB of memory. Audit logging was enabled, and the persistent store was set to S3.

We observed that during audit log allocation and flushing, the system could consume up to 4× the configured audit log max size. Since the default audit log max size was set to 1GB, this resulted in memory usage reaching up to 4GB, which caused continuous OOM issues in pods limited to 4GB.

kyu1204:
@cjellick
Hi, could you please share the current status of this issue?
Is there any chance that this will be merged or patched, or is it likely to be closed instead?
Thank you.

defdanmoz:
Confirming this still reproduces on v0.24.1 (chart v0.24.1).

> Can you give more detail about your environment, how you hit this, and how it manifests as

Environment: EKS 1.36, single replica, `t4g.medium` nodes (3.2GiB allocatable), Postgres 18 on RDS. `OBOT_SERVER_AUDIT_LOGS_MODE=s3`, `AUDIT_LOGS_COMPRESS_FILE=true`, `MAX_FILE_SIZE` unset, so the 1GB default applies.

How we hit it: enabled audit logging for the first time as part of a v0.23.4 → v0.24.1 upgrade. The trigger was ordinary admin activity — creating and removing MCP servers.

How it manifests: RSS climbed 646MiB → 2.86GiB over 12 minutes, then `OOMKilled`. For comparison, idle footprint with audit logging off is 156MiB. Users of the application see HTTP 500 / 503 messages "Service Temporarily Unavailable".

Impact beyond the 2GB allocation: The Helm chart sets no `resources` by default, so the pod runs as QoS `BestEffort`
with `request=0`. kubelet's eviction message:

    The node was low on resource: memory. Threshold quantity: 100Mi, available: 42760Ki.
    Container obot was using 3003384Ki, request is 0, has larger consumption of memory.

So it didn't just kill the pod — the node went `MemoryPressure` and evicted ~40 pods across namespaces. And because the `/data` PVC pins the pod to a single AZ, every replacement was rescheduled onto the same pressured node and evicted again. Net effect was ~10 minutes with no healthy ALB target, rather than one contained restart.

pkg/api/server/audit/logger.go
```diff
@@ -154,7 +157,12 @@ func (l *persistentLogger) persist() error {
 	}
 
 	buf := l.buffer
-	l.buffer = make([]byte, 0, l.bufferSize)
+	if l.spare != nil {
+		l.buffer = l.spare[:0]
+	} else {
+		l.buffer = nil
+	}
```
thedadams: Slicing a `nil` slice to length 0 is valid in Go, and `l.buffer` will remain `nil` if `l.spare` is `nil`.
```suggestion
	l.buffer = l.spare[:0]
```

pkg/api/server/audit/logger.go
```diff
@@ -164,6 +172,14 @@ func (l *persistentLogger) persist() error {
 		return err
 	}
 
+	l.lock.Lock()
+	if cap(buf) <= l.maxSize {
+		l.spare = buf[:0]
+	} else {
+		l.spare = nil
+	}
```
thedadams: Since `persist` can be called from multiple goroutines, it is possible for `l.spare != nil` at this point.

If that is the case, I would prefer that we keep the larger buffer for `l.spare`.

Also, don't set `l.spare = nil` in case it isn't already `nil`. We would be wiping out a useable buffer and have to reallocate again.

```suggestion
	if bufCap := cap(buf); bufCap <= l.maxSize && cap(l.spare) < bufCap {
		l.spare = buf[:0]
	}
```

pkg/api/server/audit/logger.go
kyu1204: Good catch, the nil check was redundant.

pkg/api/server/audit/logger.go
kyu1204: You're right, thanks. `Close()` and the ticker in `runPersistenceLoop` can both call `persist()`, so the window is real: goroutine A can be blocked in `store.Persist` (slow with the S3 store) while B completes a later flush and sets `l.spare`. When A then reacquires the lock, the old code would either overwrite B's buffer with a possibly smaller one, or drop it entirely via `l.spare = nil`.

The `else` branch was pointless anyway. The intent was "don't hold on to an oversized buffer", and simply not referencing `buf` already achieves that. Nil-ing `l.spare` only threw away someone else's usable buffer.

thedadams:
Thanks for your contribution!

thedadams (approved):

PR #5797, opened 2026-02-13
fix: track token usage for Anthropic and OpenAI Responses API proxy endpoints

cjellick:
The LLM proxy's responseModifier only tracked usage for /v1/chat/completions, missing /v1/messages (Anthropic) and /v1/responses (OpenAI Responses API). This also extracts the model name from streaming responses as a fallback when
 not set from the request, handling provider-specific nesting (message.model, response.model).

Token usage parsing now handles each provider's format: Anthropic nests usage under "message.usage" in message_start events and reports cumulative counts in message_delta, while the Responses API nests under "response.usage". Uses max() instead of accumulation to correctly handle cumulative token counts, and derives totalTokens at Close() time from final values.

### DIFF
--- a/pkg/gateway/server/llmproxy.go
+++ b/pkg/gateway/server/llmproxy.go
@@ -345,10 +345,8 @@ func (r *responseModifier) Read(p []byte) (int, error) {
 	// Use max to handle cumulative token counts (e.g. Anthropic's message_delta
 	// reports cumulative output_tokens, not incremental).
 	usage := gjson.GetBytes(line, "usage")
-	promptTokens := usage.Get("prompt_tokens").Int()
-	promptTokens += usage.Get("input_tokens").Int()
-	completionTokens := usage.Get("completion_tokens").Int()
-	completionTokens += usage.Get("output_tokens").Int()
+	promptTokens := max(usage.Get("prompt_tokens").Int(), usage.Get("input_tokens").Int())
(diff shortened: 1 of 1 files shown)


pkg/gateway/server/llmproxy.go
```diff
@@ -323,25 +323,58 @@ func (r *responseModifier) Read(p []byte) (int, error) {
...
+		// OpenAI Responses API nests model under "response.model"
+		if responseModel == "" {
+			responseModel = gjson.GetBytes(line, "response.model").String()
+		}
+		if responseModel != "" {
+			r.lock.Lock()
+			r.model = responseModel
+			r.lock.Unlock()
+		}
+	}
```
thedadams: I think this logic should be moved to where we check for access and remaining token usage. Otherwise, we won't properly handle that.

On main, that is here: https://github.com/obot-platform/obot/blob/main/pkg/gateway/server/llmproxy.go#L422

pkg/gateway/server/llmproxy.go
```diff
@@ -323,25 +336,50 @@ func (r *responseModifier) Read(p []byte) (int, error) {
...
 
+	// Extract model from response if not already set
+	if r.model == "" {
+		responseModel := extractModelFromBody(line)
+		if responseModel != "" {
+			r.lock.Lock()
+			r.model = responseModel
+			r.lock.Unlock()
+		}
+	}
```
thedadams: We can probably remove this now.

thedadams (approved):

PR #5813, opened 2026-02-18
fix: reconnect UI SSE stream on disconnect

thedadams:


thedadams:
Closing in favor of https://github.com/obot-platform/obot/pull/5815

PR #5851, opened 2026-02-20
chore: ui: port nanobot updates for compaction

g-linville:


thedadams (approved):

ui/user/src/lib/components/nanobot/Message.svelte
```diff
@@ -95,7 +95,13 @@
 	);
 </script>
 
-{#if promptDisplayItem}
+{#if message.items?.[0]?._meta?.['ai.nanobot.meta/compaction-summary']}
```
ivyjeong13, line 98: Maybe already been talked about so ignore me if so! Was wondering if a debug option would be helpful to display the contents of this

Issue #5818, opened 2026-02-18
Provide easy access to Nanobot Chat from all pages.


sangee2004:
**Describe the bug**
Provide easy access to Nanobot Chat from all pages.

**To Reproduce**
1. Log in as any user and click on the profile pic
2. Notice that the `Chat` option is presented, which takes the user to the Obot chat page.

<img width="1212" height="719" alt="Image" src="https://github.com/user-attachments/assets/8a44b1f2-eff5-456a-93ca-75865b35d14e" />

This should be replaced with access to the Nanobot Chat page instead. If we want to maintain a link to Obot Chat as well, we can add another option - `Nanobot Chat` to give access to the Nanobot Chat page.

Similar changes need to be done in the following pages as well:

API keys page:
<img width="1211" height="815" alt="Image" src="https://github.com/user-attachments/assets/ecf5d928-c933-46ee-803b-c27121b81cb7" />
My Account page:
<img width="1219" height="795" alt="Image" src="https://github.com/user-attachments/assets/13ffc2e8-a649-4f26-8b7c-6b057fe4636a" />

3. Go to `Launch Nanobot`.
4. Notice that the `Chat` option is presented, which takes the user to the Obot chat page.

<img width="1218" height="770" alt="Image" src="https://github.com/user-attachments/assets/1939f9ca-e962-42b1-964b-73f56cbe5b94" />

This option should be removed.

ivyjeong13:
TBD depending on decision Craig mentioned for release:
- **Disable Obot Chat. Enable NB Chat by default**
    - I want to call it Obot Agent’
    - **TODO** Maybe we keep both in this release and intro Obot Agent as a alpha feature

cjellick:
Yeah - as of right now we are going all in - legacy Obot Chat will be off by default and hidden behind a flag. If the nanobot integration is on, show it everywhere we show Obot Chat. And I'd like to call it Obot Agent. Keep the fun robot head chat bubble icon you have.

Legacy Chat and Obot Agent should be able to co-exist. In other words, if both feature flags are on, I should see both options. 

@g-linville im going to co-assign you to this so that you can plumb another env var/feature flag through to disable legacy chat that Ivy can key off of to hide that legacy chat. It only needs to drop it from the UI, we dont need to block it in the API. Default should result in legacy chat being disabled and the user should have to set and env var explicitly to enable it. Additionally, chage the default of the nanobot integration to be on by default.

sangee2004:
[...]

PR #5879, opened 2026-02-24
chore: add feature flag to disable legacy chat

g-linville:
for https://github.com/obot-platform/obot/issues/5818


AGENTS.md
g-linville, line 1: This is a symlink that points to CLAUDE.md

cjellick (approved):

PR #5901, opened 2026-02-25
fix: make oauth-url idempotent

g-linville:


g-linville:
oh @cjellick sorry I was actually gonna ditch this implementation because I honestly don't fully understand it. Just forgot to close the PR

PR #5924, opened 2026-02-27
fix: ui: use proper paths when connecting to Nanobot

g-linville:
depends on https://github.com/nanobot-ai/nanobot/pull/187

### DIFF
--- a/ui/user/src/lib/services/nanobot/types.ts
+++ b/ui/user/src/lib/services/nanobot/types.ts
@@ -400,7 +400,8 @@ export interface InitializationResult {
 	};
 }
 
-export const UIPath = '/mcp?ui';
+export const UIPath = '/mcp/ui';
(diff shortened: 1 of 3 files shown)


ui/user/src/lib/services/nanobot/types.ts
```diff
@@ -400,7 +400,8 @@ export interface InitializationResult {
 	};
 }
 
-export const UIPath = '/mcp?ui';
+export const UIPath = '/mcp/ui';
```
ivyjeong13, line 403: Ah, actually you can probably delete the UIPath & ChatPath constants here. The UIPath is no longer used -- it was before I got the changes from Donnie to just directly use the connectURL and forgot to remove this!

ivyjeong13: Nevermind! Talked with Grant and this will need to be there 😄

Issue #5929, opened 2026-02-27
Token usage not reported


sangee2004:
**Reproduces** Always
**Describe the bug**
When a new session is launched via Launch Agent, and messages are exchanged, no token usage is displayed on the `Token Usage` page.

**To Reproduce**
1. Go to `Launch Agent`.
2. Start a new session and chat with it
4. As an admin user, go to the `Token Usage` page

Notice that there is no token usage reported
<img width="1356" height="805" alt="Image" src="https://github.com/user-attachments/assets/a72993cd-f771-4b20-89d7-d9d114fad31f" />

Also, note that no token usage is reported for the `Claude-sonnet-4-6` model, which is the default model for `Language Model (Chat)` and `Language Model (Chat - Fast)`.

<img width="709" height="454" alt="Image" src="https://github.com/user-attachments/assets/ada7aa21-29d6-412a-8db1-b6f7fc447744" />

**Expected behavior**
Token usage should be reported correctly.

sangee2004:
This issue is addressed when testing with the latest build.

Token usage gets reported as expected when chatting with the agent.

<img width="1072" height="440" alt="Image" src="https://github.com/user-attachments/assets/f400f100-06dd-4874-9843-9bc751e1ff69" />
<img width="1081" height="695" alt="Image" src="https://github.com/user-attachments/assets/e3ae621f-0787-462d-823c-d6fe9f56aef6" />

PR #5930, opened 2026-02-27
fix: avoid gzip-encoded proxy bodies in LLM token parsing

cjellick:
## Summary
- strip `Accept-Encoding` in the LLM proxy request director so Go's transport can transparently decode provider responses before token-usage parsing
- use the new LLM-specific request transformer in both `/v1/llm-proxy` and provider proxy code paths
- add a regression test to verify `Accept-Encoding` is removed by the director

## Testing
- go test ./pkg/gateway/server -run 'Test(ResponseModifier|ModifyResponse|ExtractModelFromBody|LLMTransformRequest)'

https://pkg.go.dev/net/http#Transport
Relevant DisableCompression text:
- “If the Transport requests gzip on its own and gets a gzipped response, it's transparently decoded in the Response.Body.”
- “However, if the user explicitly requested gzip it is not automatically uncompressed.”
That’s why forwarding a client Accept-Encoding: gzip can break JSON token parsing in the proxy path.

### DIFF
--- a/pkg/gateway/server/llmproxy.go
+++ b/pkg/gateway/server/llmproxy.go
@@ -163,7 +163,7 @@ func (s *Server) dispatchLLMProxy(req api.Context) error {
 	}
 
 	(&httputil.ReverseProxy{
-		Director:       dispatcher.TransformRequest(u, credEnv),
+		Director:       llmTransformRequest(u, credEnv),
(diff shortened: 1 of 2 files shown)


cjellick:
Addresses https://github.com/obot-platform/obot/issues/5929

PR #5938, opened 2026-02-28
fix: correct secret trigger for nanobot cred sync

cjellick:
This addresses a lot of observed "stability" issues where the nanobot agent wasn't getting new credentials/api keys injected. The bugs manifest as problems connecting to either obot mcp server server (that helps you discover and configure MCPs) ore after that not being able to talk to servers you configured. 

The problem was that the trigger key was incorrectly concatenating the namespace with "files" instead of the server name, producing "namespace-files/server" rather than "namespace/server-files".

Also adds debug/info logging throughout the credential lifecycle and extracts token TTL into named constants.

..i extracted the times to constants because i just wanted to make them easier to modify locally. arguably they could be env vars, but that wasnt wiorth the extra hassle right now.

This was a somewhat inconsistent issue because there are other code paths that will sometimes trigger redeployments of a nanobot agent and thus fix cred issues. One of those is the other file (kuberenetes.go) that I edited here. Sometimes, when a user makes a request to an mcp server, based on the logic in that file, we will decied it needs refreshed.

### DIFF
--- a/pkg/controller/handlers/nanobotagent/nanobotagent.go
+++ b/pkg/controller/handlers/nanobotagent/nanobotagent.go
@@ -13,6 +13,7 @@ import (
 	"github.com/obot-platform/nah/pkg/name"
 	"github.com/obot-platform/nah/pkg/router"
 	"github.com/obot-platform/obot/apiclient/types"
+	"github.com/obot-platform/obot/logger"
 	"github.com/obot-platform/obot/pkg/alias"
 	"github.com/obot-platform/obot/pkg/gateway/client"
 	"github.com/obot-platform/obot/pkg/jwt/persistent"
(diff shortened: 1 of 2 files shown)


pkg/controller/handlers/nanobotagent/nanobotagent.go
```diff
@@ -321,15 +337,18 @@ func (h *Handler) ensureCredentials(ctx context.Context, req router.Request, res
 
 	if h.localK8SBackend != nil {
 		// If local Kubernetes backend is available, trigger a sync to update the secret with the new credentials
+		triggerKey := fmt.Sprintf("%s/%s", h.mcpServerNamespace, name.SafeConcatName(mcpServerName, "files"))
+		log.Debugf("Triggering local k8s secret sync: agent=%s mcpServer=%s key=%s", agent.Name, mcpServerName, triggerKey)
 		if err := h.localK8SBackend.Trigger(
 			ctx,
 			corev1.SchemeGroupVersion.WithKind("Secret"),
-			fmt.Sprintf("%s/%s", name.SafeConcatName(h.mcpServerNamespace, "files"), mcpServerName),
```
cjellick, line 327: this is the actual fix. we were contructing the key incorrectly

PR #5948, opened 2026-03-02
feat: add Auth0 as enterprise auth provider

cloudnautique:
Add Auth0 support alongside existing Okta and Entra providers:
- Auth0 profile extraction (icon, display name) in user.go
- Auth0 constant and UI sort ordering
- Auth0 configuration documentation

Also fix a bug in group.go where identity.Email was passed to fetchGroups instead of identity.ProviderUserID. Auth provider group endpoints expect the provider-specific user ID (e.g. auth0|abc123), not the email address.

### DIFF
--- a/pkg/gateway/client/group.go
+++ b/pkg/gateway/client/group.go
@@ -193,8 +193,8 @@ func (c *Client) ensureGroups(ctx context.Context, tx *gorm.DB, identity *types.
 		return nil
 	}
 
-	// Fetch groups from the auth provider
-	providerGroups, err := c.fetchGroups(ctx, providerURL, identity.AuthProviderNamespace, identity.AuthProviderName, identity.ProviderUserID)
+	groupLookupID := identity.GroupLookupID()
+	providerGroups, err := c.fetchGroups(ctx, providerURL, identity.AuthProviderNamespace, identity.AuthProviderName, groupLookupID)
 	if err != nil {
 		return err
 	}
--- a/pkg/gateway/types/identity.go
+++ b/pkg/gateway/types/identity.go
@@ -9,6 +9,7 @@ type Identity struct {
 	ProviderUsername      string    `json:"providerUsername"`
 	ProviderUserID        string    `json:"providerUserID"`
 	HashedProviderUserID  string    `json:"hashedProviderUserID" gorm:"primaryKey"`
+	ProviderGroupLookupID string    `json:"providerGroupLookupID"`
 	Email                 string    `json:"email"`
 	HashedEmail           string    `json:"hashedEmail"`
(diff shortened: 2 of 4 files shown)


pkg/gateway/client/group.go
```diff
@@ -194,7 +194,7 @@ func (c *Client) ensureGroups(ctx context.Context, tx *gorm.DB, identity *types.
 	}
 
 	// Fetch groups from the auth provider
-	providerGroups, err := c.fetchGroups(ctx, providerURL, identity.AuthProviderNamespace, identity.AuthProviderName, identity.Email)
+	providerGroups, err := c.fetchGroups(ctx, providerURL, identity.AuthProviderNamespace, identity.AuthProviderName, identity.ProviderUserID)
```
g-linville: Are we 100% sure that this will not break Entra or Okta?

njhale: I'm surprised this worked in `main` for Entra, since the way it's implemented in the actual tool leads me to believe that it expects a user ID and not an email:
- https://github.com/obot-platform/enterprise-tools/blob/c20c59e52087c6682d9732cfd5776f2c90903890/entra-auth-provider/main.go#L198
- https://github.com/obot-platform/enterprise-tools/blob/c20c59e52087c6682d9732cfd5776f2c90903890/entra-auth-provider/pkg/profile/profile.go#L62

cloudnautique: Turns out no, it broke Entra. I added a new group lookup field, that falls back to the email address we were using, and verified Entra still works.

pkg/gateway/types/identity.go
```diff
@@ -23,6 +24,13 @@ type Identity struct {
 	AuthProviderGroups []Group `json:"groups" gorm:"-"`
 }
 
+func (i Identity) GroupLookupID() string {
+	if i.ProviderGroupLookupID != "" {
+		return i.ProviderGroupLookupID
+	}
+	return i.Email
```
cloudnautique: This would be the fallback to what we used before in the groups.go function call.

sangee2004:
`Auth0` is available as an Auth provider when testing with the latest obot-enterprise build.

Able to configure `Auth0` auth provider successfully and authenticate the user as part of obot logins.
Admin users can discover groups and assign them to registry access or role assignments.

Users get assigned default roles based on the group role assignments.
Users gain access to the MCP servers based on user/group assignments configured via registry entries. 
 
Documentation relating to configuring the `Auth0` provider and Restricting Login to Specific Users using Connection-Level Access Control has been verified and works as expected.

There is an issue with an admin/owner user not being able to see all groups configured which is tracked in  https://github.com/obot-platform/obot/issues/6059

PR #5968, opened 2026-03-03
fix: toHtmlFromMarkdown with optional video/iframe processing

ivyjeong13:
* makes video/iframe processing optional
* adds the protocol to only allow https:// for iframe/video src embedding (can expand on more if needed)

### DIFF
--- a/ui/user/src/lib/markdown.ts
+++ b/ui/user/src/lib/markdown.ts
@@ -9,7 +9,7 @@ export function toHTMLFromMarkdown(
 	const html = micromark(markdown, {
 		extensions: [gfm()],
 		htmlExtensions: [gfmHtml()],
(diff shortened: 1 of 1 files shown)


ui/user/src/routes/nanobot/p/[projectId]/+layout.svelte
```diff
@@ -88,7 +88,6 @@
 		const sid = sessionId;
 		const c = chat;
 		if (!api || !sid || !c) return;
-		if (c.isLoading) return;
```
Comment (ivyjeong13) [inline ui/user/src/routes/nanobot/p/[projectId]/+layout.svelte:91]: Removed check after investigating with Sangeetha about the session showing up in the threads list was delayed, this is waiting for chat to be done with its initial output which was causing the delay.

PR #5975, opened 2026-03-04
enhance: ui: show checkboxes for boolean auth provider params

g-linville:
This is better UX for the new params to enable logging:

image


njhale:
Can we put the checkbox (I prefer a toggle) to the right of the field name, and the description below? I think it would look better to keep horizontal alignment for field names and description placement the same between fields

g-linville:
toggle placed below:
image

PR #5995, opened 2026-03-06
logging

cjellick:
- **Chore: More controller logging**
- **Chore: Authz and authn related logging**
- **Chore: Invoker logging**
- **Chore: redact sensitive logging**

### DIFF
--- a/pkg/controller/handlers/cronjob/cronjob.go
+++ b/pkg/controller/handlers/cronjob/cronjob.go
@@ -129,7 +129,7 @@ func (h *Handler) SetSuccessRunTime(req router.Request, _ router.Response) error
 	for _, execution := range workflowExecutions.Items {
 		if execution.Status.State == types.WorkflowStateComplete && (cj.Status.LastSuccessfulRunCompleted == nil || cj.Status.LastSuccessfulRunCompleted.Before(execution.Status.EndTime)) {
 			cj.Status.LastSuccessfulRunCompleted = execution.Status.EndTime
-			log.Infof("Updated last successful cron run time: cronJob=%s completedAt=%s", cj.Name, cj.Status.LastSuccessfulRunCompleted.Time.Format(time.RFC3339))
+			log.Infof("Updated last successful cron run time: cronJob=%s completedAt=%s", cj.Name, cj.Status.LastSuccessfulRunCompleted.Format(time.RFC3339))
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/handlers/cleanup/user.go
```diff
@@ -79,11 +84,14 @@ func (u *UserCleanup) Cleanup(req router.Request, _ router.Response) error {
 		return err
 	}
 
+	deletedServers := 0
```
njhale, line 87: nit: a lot of models make this mistake in go, but convention dictates you use the default zero value for declarations like this `var deletedServers int`

Also, the count here ends up being `len(servers.Items)` in every non-error case anyway, so the variable isn't needed.

pkg/controller/handlers/cleanup/user.go
```diff
@@ -97,17 +105,21 @@ func (u *UserCleanup) Cleanup(req router.Request, _ router.Response) error {
...
 	}
 
+	deletedInstances := 0
 	for _, instance := range instances.Items {
 		if err := kclient.IgnoreNotFound(req.Delete(&instance)); err != nil {
 			return err
 		}
+		deletedInstances++
 	}
+	log.Infof("Deleted MCP server instances during user cleanup: userID=%s instances=%d", userID, deletedInstances)
```
njhale, line 115: Same here, if we get to this log, `deletedInstances` is always `len(instances.Items)`, so the variable isn't necessary.

pkg/gateway/server/user.go
```diff
@@ -196,9 +196,11 @@ func (s *Server) updateUser(apiContext api.Context) error {
 
 	if !apiContext.UserIsOwner() {
 		if originalUser.Role.HasRole(types2.RoleOwner) != user.Role.HasRole(types2.RoleOwner) {
+			pkgLog.Infof("Denied user role update: targetUserID=%s reason=owner_role_change_requires_owner", userID)
 			return types2.NewErrHTTP(http.StatusForbidden, "only owner can add or remove owner role")
 		}
 		if originalUser.Role.HasRole(types2.RoleAuditor) != user.Role.HasRole(types2.RoleAuditor) {
+			pkgLog.Infof("Denied user role update: targetUserID=%s reason=auditor_role_change_requires_owner", userID)
 			return types2.NewErrHTTP(http.StatusForbidden, "only owner can remove admin role")
```
g-linville, line 204: nit: not caused by your changes but I noticed this


```suggestion
			return types2.NewErrHTTP(http.StatusForbidden, "only owner can remove auditor role")
```

PR #6013, opened 2026-03-07
chore: add admin page for managing agents

drpebcak:


pkg/controller/controller.go
```diff
@@ -95,6 +95,47 @@ func (c *Controller) PreStart(ctx context.Context) error {
...
+		}
+
+		server.Spec.Manifest.ContainerizedConfig.Image = c.services.NanobotAgentImage
+		if err := c.services.StorageClient.Update(ctx, &server); err != nil {
+			return fmt.Errorf("failed to update nanobot MCP server %s image: %w", server.Name, err)
+		}
+
+		if !server.Status.NeedsUpdate {
+			var updatedServer v1.MCPServer
+			if err := c.services.StorageClient.Get(ctx, kclient.ObjectKeyFromObject(&server), &updatedServer); err != nil {
```
njhale: You hsouldn't need to refetch after the update.

pkg/mcp/kubernetes.go
```diff
@@ -1190,6 +1190,24 @@ func (k *kubernetesBackend) restartServer(ctx context.Context, server ServerConf
...
+						continue
+					}
+					return err
+				}
+
+				if err := k.client.Get(ctx, kclient.ObjectKey{Name: id, Namespace: k.mcpNamespace}, &deployment); err != nil {
+					return fmt.Errorf("failed to get deployment %s after image patch: %w", id, err)
+				}
+			}
+		}
```
g-linville, line 1209: Is there anything we need in `docker.go` that would correspond to this change?

ui/user/src/lib/services/admin/mcpServers.ts
```diff
@@ -0,0 +1,21 @@
...
+import type { MCPCatalogServer } from '../chat/types';
+import type { Fetcher } from '../http';
+import AdminService from './index';
+
+export async function getAdminMcpServerWithFallback(
+	id: string,
+	opts?: { fetch?: Fetcher }
+): Promise<MCPCatalogServer> {
+	try {
+		return await AdminService.getMCPCatalogServer(DEFAULT_MCP_CATALOG_ID, id, opts);
```
ivyjeong13: Curious what's this getting a fallback for AdminService.getMCPCatalogServer for? I believe this is for returning a multi-user server so if it doesn't come back from the admin call, I don't believe they'd it get from calling the permission scoped getMcpCatalogServer either.

drpebcak: Yeah.. it was a little overloaded. ive split these into 2 helpers so its more clear

pkg/mcp/kubernetes.go
drpebcak: I don't think so!

ui/user/src/lib/components/nanobot/AgentRestartBanner.svelte
```diff
@@ -0,0 +1,129 @@
...
+			return 'A pending agent update is ready. Restart to apply it.';
+		}
+		if (agentServer?.needsK8sUpdate) {
+			return 'A pending agent configuration change is ready. Restart to apply it.';
+		}
+		return '';
+	});
+
+	$effect(() => {
+		const resetKey = `${agentId}:${projectId}`;
```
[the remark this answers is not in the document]
ivyjeong13: currently don't need to worry about this (but will keep in mind if/when we allow diff projects/agents)!

ui/user/src/routes/admin/mcp-servers/s/[id]/details/+page.ts
```diff
@@ -9,9 +8,7 @@ export const load: PageLoad = async ({ params, fetch }) => {
 
 	let mcpServer;
 	try {
-		mcpServer = await AdminService.getMCPCatalogServer(DEFAULT_MCP_CATALOG_ID, id, {
-			fetch
-		});
+		mcpServer = await getAdminDirectMcpServer(id, { fetch });
```
Comment (ivyjeong13) [inline ui/user/src/routes/admin/mcp-servers/s/[id]/details/+page.ts:11]: Gonna quickly check for this, it's been a sec but I believe getting the mcp server from here will return mcp servers a user has access to through the access policies layer (where as the AdminService.getMCPCatalogServer will return the server regardless as long as they have the appropriate role/group) Unless this has changed with the backend, then ignore me haha!

Comment (ivyjeong13) [reply ui/user/src/routes/admin/mcp-servers/s/[id]/details/+page.ts:11]: :+1: Checked and saw admin still getting the mcp server information. Looks like that change to bypass ACR was made here and assuming that was intended!

PR #6020, opened 2026-03-09
fix: add missing namespaces to helm templates

knkarthik:
Ensure that all Helm templates explicitly declare the release namespace in their metadata to prevent resources from being deployed into the default namespace. Without this the user can't deploy it to any other namespace even with the `--namespace` helm flag.

Also update the versions in `Chart.yaml`. There should be a process to update these values for every release though.

### DIFF
--- a/chart/Chart.yaml
+++ b/chart/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
 name: obot
 description: A Helm chart for Obot
-version: 0.1.0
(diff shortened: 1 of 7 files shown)


knkarthik:
These changes are required to make the chart installable in any namespace. 

@drpebcak @njhale

drpebcak:
Thanks for the pull request, but this is not necessary to deploy a chart into an arbitrary workspace.

PR #6037, opened 2026-03-09
docker deployment cache

cjellick:
This seeks to introduce a cache similar to what donnie did in https://github.com/obot-platform/obot/pull/5885

It ends up being relatively expensive (time-wise) to check the deployment every time through the api

### DIFF
--- a/pkg/mcp/docker.go
+++ b/pkg/mcp/docker.go
@@ -57,12 +57,6 @@ type dockerDeploymentCacheEntry struct {
 	containerIDs map[string]string
 }
 
-type dockerDeploymentValidationStats struct {
-	containerCount    int
-	getContainerMs    int64
-	maxGetContainerMs int64
-}
-
 func newDockerBackend(ctx context.Context, exposedPort int, opts Options) (backend, error) {
 	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/mcp/docker.go
```diff
@@ -179,6 +188,27 @@ func (d *dockerBackend) ensureServerDeployment(ctx context.Context, server Serve
...
+	cachedDeployment := d.getDeploymentCache(serverName)
+	if cachedDeployment != nil && cachedDeployment.hash == serverConfigHash {
+		valid, err := d.deploymentCacheValid(ctx, cachedDeployment)
+		if err != nil {
+			return ServerConfig{}, err
+		}
+
+		if valid {
+			return cachedDeployment.serverConfig, nil
+		}
```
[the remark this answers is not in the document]
cjellick: Don't want to do this. (And donnie didnt do the equivalent in his pr for adding this to k8s). We really want this to be light and fast because it is in the way of a lot of api calls.

pkg/mcp/docker.go
```diff
@@ -179,6 +188,27 @@ func (d *dockerBackend) ensureServerDeployment(ctx context.Context, server Serve
 	serverName := server.MCPServerName
 	// Copy the webhooks so we can change the URL without that affecting the original slice.
 	transformedWebhooks := slices.Clone(webhooks)
+	for i, webhook := range transformedWebhooks {
+		webhook.URL = strings.Replace(webhook.URL, "http://localhost", d.hostBaseURL, 1)
+		transformedWebhooks[i] = webhook
+	}
```
g-linville, line 194: nit: I think you can do this


```suggestion
	for i := range transformedWebhooks {
		transformedWebhooks[i].URL = strings.Replace(webhook.URL, "http://localhost", d.hostBaseURL, 1)
	}
```

cjellick: It is a faux pas to ignore nits? Cause I'm gonna.

PR #6045, opened 2026-03-11
docs: update Auth0 auth provider documentation

cloudnautique:
Replace Auth0 Organizations with users and roles approach:
- Add required Machine to Machine application setup with read:users and read:role_members scopes
- Add email verification requirement with troubleshooting guidance
- Reference OBOT_AUTH0_AUTH_PROVIDER_INSECURE_ALLOW_UNVERIFIED_EMAIL config option for tenants that don't support email verification
- Remove Organizations-based group support section
- Simplify login restriction options

### DIFF
--- a/docs/docs/configuration/auth-providers.md
+++ b/docs/docs/configuration/auth-providers.md
@@ -186,7 +186,7 @@ You can now return to Obot and finish configuring Auth0. Use the table below to 
 
 | Obot          | Auth0         | Auth0 Dashboard Location |
 |---------------|---------------|--------------------------|
-| Domain        | Domain        | Application Settings     |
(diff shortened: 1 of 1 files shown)


sangee2004:
@cloudnautique Wanted to follow up on  these:

The domain field in the `Set Up Auth0` page is named `Auth0 Domain` . This needs to be updated in this table:
Screenshot 2026-03-11 at 5 30 22 PM

If we are providing this table, it would be good to do the same for the `Machine to Machine Application` section as well.

In the `Restricting Login to Specific Users (Optional)` section, can we provide more steps on configuring the auth provider to restrict user access?

sangee2004:
LGTM

PR #6077, opened 2026-03-16
fix: populate auth provider groups when authenticating with jwt

g-linville:


pkg/jwt/persistent/persistent.go
```diff
@@ -200,19 +204,40 @@ func (t *TokenService) AuthenticateRequest(req *http.Request) (*authenticator.Re
...
+			"mcp_id":                  {tokenContext.MCPID},
+			"resource":                {tokenContext.Audience},
+			"oauthScope":              {tokenContext.OAuthScope},
+		}
+		groups := tokenContext.UserGroups
+
+		// Look up auth provider group memberships from the gateway DB
+		if userID, err := strconv.ParseUint(tokenContext.UserID, 10, 64); err == nil {
+			if authGroupIDs, err := t.gatewayClient.ListGroupIDsForUser(req.Context(), uint(userID)); err != nil {
+				log.Warnf("failed to list auth provider groups for user %s: %s", tokenContext.UserID, err.Error())
```
njhale, line 220: Should this just break if we can't look up groups for the user? Same thing with looking up the user.

g-linville: Probably doesn't matter tbh. If the JWT is valid, the only thing that I think would cause these lookups to fail is transient db errors

Issue #5979, opened 2026-03-04
Default model alias suggestions missing when configuring anthropic provider on fresh install


sangee2004:
**Reproducess** Always
**Describe the bug**
When configuring the Anthropic model provider on a fresh Obot installation with no model providers set up, the Default Model Alias fields appear empty with no suggested defaults. This leaves users unsure which models to select for Chat, Chat - Fast, and Vision. The UI should provide recommended default models that are known to work well with the Agent to guide initial configuration.

Pre Req - Have an obot install with no model providers configured yet.
**To Reproduce**
1. Go to `Model Providers` and configure the Anthropic model provider.
Notice that the user is presented with `Default Model Alias` with no suggested defaults.

<img width="757" height="668" alt="Image" src="https://github.com/user-attachments/assets/daea1ad0-40ec-4f81-8d0e-4e2a066d5d42" />

**Expected behavior**
Provide some suggested default models for `Chat`, `Chat- Fast`, and `Vision`.
These settings are known to work well with Agent

<img width="734" height="565" alt="Image" src="https://github.com/user-attachments/assets/6a967f27-e762-4990-86e1-1c9aa6b75323" />

cjellick:
@njhale this is perhaps tangentially related to waht i asked you to do, but would like it addressed.

sangee2004:
This issue is addressed when testing with the latest build.

<img width="735" height="356" alt="Image" src="https://github.com/user-attachments/assets/57764537-5711-4a8c-a902-3792871e686b" />

PR #6082, opened 2026-03-16
fix: default agent models

njhale:
## Summary

Improve default model handling in by:

- resolving Nanobot agent defaults from the same backend provider defaults used elsewhere in Obot
- updating the OpenAI preferred default from `gpt-5.2` to `gpt-5.4`
- removing a legacy default-model seed cleanup path
- showing Anthropic defaults as suggested in the admin default-model UI

## Agent default models

Nanobot agent startup needs concrete `llm` and `llm-mini` values. When those aliases were unset, the agent path used its own fallback list of preferred model names.

This change removes that duplication and makes the agent fallback use the same provider default mappings defined in `toolreference.go`. It also updates the OpenAI `llm` and `vision` defaults from `gpt-5.2` to `gpt-5.4`.

## Legacy cleanup removal

Obot still carried an embedded `default-models.yaml` file and startup logic that only loaded those model definitions in order to delete them.

This change removes that file and the cleanup path. Default model behavior now flows through default model aliases and model provider sync.

## Admin UI suggestions

The admin default-model dialog previously supported only a single suggested model per alias, which meant the UI only surfaced OpenAI suggestions.

This change updates the dialog to support an ordered list of suggested models per alias. That allows both OpenAI and Anthropic defaults to be marked as suggested for `llm`, `llm-mini`, and `vision`, while reflecting the OpenAI move to `gpt-5.4`.

Addresses:
- https://github.com/obot-platform/obot/issues/5979

### DIFF
--- a/pkg/controller/handlers/toolreference/toolreference.go
+++ b/pkg/controller/handlers/toolreference/toolreference.go
@@ -43,7 +43,7 @@ var (
 		types.DefaultModelAliasTypeLLM:             "gpt-5.4",
 		types.DefaultModelAliasTypeLLMMini:         "gpt-5-mini",
 		types.DefaultModelAliasTypeVision:          "gpt-5.4",
-		types.DefaultModelAliasTypeImageGeneration: "gpt-image-1.5",
+		types.DefaultModelAliasTypeImageGeneration: "dall-e-3",
 		types.DefaultModelAliasTypeTextEmbedding:   "text-embedding-3-large",
 	}
 	anthropicDefaultModelAliases = map[types.DefaultModelAliasType]string{
@@ -356,7 +356,12 @@ func (h *Handler) ensureModelProviderCredAndDefaults(ctx context.Context, c clie
 			continue
 		}
 
-		alias.Spec.Manifest.Model = modelName(modelProviderRef.Name, defaultModelAliasMapping[types.DefaultModelAliasType(alias.Spec.Manifest.Alias)])
+		targetModel := defaultModelAliasMapping[types.DefaultModelAliasType(alias.Spec.Manifest.Alias)]
+		if targetModel == "" {
+			continue
+		}
+
+		alias.Spec.Manifest.Model = modelName(modelProviderRef.Name, targetModel)
 		if err := c.Update(ctx, &alias); err != nil {
 			return fmt.Errorf("failed to update model alias %q for model provider %q: %w", alias.Name, modelProviderName, err)
 		}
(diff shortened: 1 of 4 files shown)


pkg/controller/handlers/toolreference/toolreference.go
```diff
@@ -37,6 +38,29 @@ var (
 
 const toolRecheckPeriod = time.Hour
 
+var (
+	openAIDefaultModelAliases = map[types.DefaultModelAliasType]string{
+		types.DefaultModelAliasTypeLLM:             "gpt-5.4",
+		types.DefaultModelAliasTypeLLMMini:         "gpt-5-mini",
+		types.DefaultModelAliasTypeVision:          "gpt-5.4",
+		types.DefaultModelAliasTypeImageGeneration: "gpt-image-1.5",
```
cjellick: since youre in here touching this stuff, can you change this back to dall-e gpt-image-1.5 doesnt work for image gen in legacy chat and i dont want to put any effort into fixing it

ui/user/src/lib/components/admin/DefaultModels.svelte
```diff
@@ -37,12 +37,12 @@
...
-		[ModelAlias.Llm]: 'gpt-5.2',
-		[ModelAlias.LlmMini]: 'gpt-5-mini',
-		[ModelAlias.TextEmbedding]: 'text-embedding-3-large',
-		[ModelAlias.ImageGeneration]: 'gpt-image-1.5',
-		[ModelAlias.Vision]: 'gpt-5.2'
+	const SUGGESTED_MODEL_SELECTIONS: Record<ModelAlias, string[]> = {
+		[ModelAlias.Llm]: ['gpt-5.4', 'claude-sonnet-4-6'],
+		[ModelAlias.LlmMini]: ['gpt-5-mini', 'claude-haiku-4-5'],
+		[ModelAlias.TextEmbedding]: ['text-embedding-3-large'],
+		[ModelAlias.ImageGeneration]: ['gpt-image-1.5'],
```
cjellick: and here

pkg/controller/handlers/toolreference/toolreference.go
njhale: ah, yeah, I recall that now. Switching iamge gen back

cjellick (approved):
modula my request for the image model revert, LGTM. you could do that rever to dall-e in  a follow up if you rpefer

Issue #6080, opened 2026-03-16
`Restart Agent` banner presented even after agent is restarted.


sangee2004:
**Reproduces** Always in `main`.
**Describe the bug**
`Restart Agent` banner is presented even after the agent is restarted.

**To Reproduce**
1. Log in as the admin user in the `main` environment.
2. User is presented with the `Restart agent` option.
3. Click on `Restart agent`.

Even after the agent gets upgraded to the latest version, the user continues to be presented with the `Restart Agent` banner.

<img width="1345" height="810" alt="Image" src="https://github.com/user-attachments/assets/87ce347e-db81-4879-b4fa-bf3c34856723" />

**Expected behavior**
Once the `Restart Agent` is used and the agent is upgraded, the user should not be presented with the `Restart Agent` banner anymore.

sangee2004:
This issue is not seen when testing with the latest build.
User is not presented with the `Restart Agent` banner anymore.

Issue #6079, opened 2026-03-16
404 error message presented for basic user when the agent is launched.


sangee2004:
**Reproduces**  Always
**Describe the bug**
404 error message presented for basic user when the agent is launched.

**To Reproduce**
1. Log in as a Basic user and go to `Launch Agent`.

Notice that the user is presented with a 404 error message

<img width="1420" height="741" alt="Image" src="https://github.com/user-attachments/assets/17bfa519-b040-46e3-bded-4fb0c6d06067" />

**Expected behavior**
User should not be presented with 404 error message.

sangee2004:
This issue is no longer seen when testing with the latest build.

PR #6083, opened 2026-03-16
chore: remove banner from /agent, rename to obot agent

drpebcak:
also, should fix some logic around completing updates

Addresses:
- https://github.com/obot-platform/obot/issues/6079
- https://github.com/obot-platform/obot/issues/6080


drpebcak:
(this needs a bit more cleanup before review)

Issue #5992, opened 2026-03-05
File fails to render from chat view when filename contains spaces due to unencoded URI


sangee2004:
**Reproduces** Always
**Describe the bug**
When a user generates a graph and saves it with a filename containing spaces (e.g., “My Graph.svg”) in Launch Agent, clicking the file from the chat view results in an error. This happens because the URI sent in the `resources read` call is not URL-encoded. The file opens correctly from the to-do file list sidebar and Files list view, where the URI is properly encoded, indicating an inconsistency in how the chat view handles file URIs.

**To Reproduce**
1. Go to `Launch Agent`.
2. Chat and ask to generate a graph and save it as "generate a graph and save as my graph".
3. Click on the `My Graph.svg` file shown in chat view

Notice that the user is presented with the following error message:
<img width="934" height="800" alt="Image" src="https://github.com/user-attachments/assets/9aaf5f64-2e22-4e52-9132-938cd2efd664" />

The URI sent in the payload of the `resources read` call (/mcp-connect/<id>/mcp/ui?method=resources%2Fread) is not URL-encoded.

<img width="734" height="348" alt="Image" src="https://github.com/user-attachments/assets/5f1ef00c-2887-4d81-80be-224bbae5ed00" />

Note - File renders correctly when opening file from the `to-do file list` sidebar in chat view and from `Files` list view, where the URI  is encoded as expected.

**Expected behavior**
The file should be rendered correctly when there are spaces in the file name.

cjellick:
@njhale @ivyjeong13 is this frontend or backend?

sangee2004:
This issue is addressed when testing with the latest build.

Files with spaces in their names render properly when clicked from the chat view.

<img width="1021" height="745" alt="Image" src="https://github.com/user-attachments/assets/83b612b3-ce70-46ef-8d48-86b2e34d82f1" />

PR #6093, opened 2026-03-17
fix: normalize file resource URIs before reading

njhale:
Normalize raw workspace file paths and unencoded file URIs before issuing resources/read so chat file previews work for names with spaces.

Route chat and session resource reads through SimpleClient.readResource so the normalization logic lives in one place.

Addresses https://github.com/obot-platform/obot/issues/5992

### DIFF
--- a/ui/user/src/lib/services/nanobot/utils.ts
+++ b/ui/user/src/lib/services/nanobot/utils.ts
@@ -13,6 +13,33 @@ export function parseToolFilePath(item: ChatMessageItemToolCall) {
 	return parseJSON<{ file_path: string }>(item.arguments)?.file_path ?? null;
 }
 
+export function normalizeResourceReadURI(uri: string): string {
+	if (!uri) return uri;
+
+	const fileURIPrefix = 'file:///';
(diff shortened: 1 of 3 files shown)


ui/user/src/lib/services/nanobot/utils.ts
```diff
@@ -13,6 +13,33 @@ export function parseToolFilePath(item: ChatMessageItemToolCall) {
...
+export function normalizeResourceReadURI(uri: string): string {
+	if (!uri) return uri;
+
+	const fileURIPrefix = 'file:///';
+	const homePrefix = '/home/nanobot/';
+	const isFileURI = uri.startsWith(fileURIPrefix);
+	const isWorkspacePath = uri.startsWith(homePrefix);
+
+	if (!isFileURI && !isWorkspacePath) {
+		return uri;
```
ivyjeong13, line 25: Potentially could still call encodeURIComponent for uri here, is there a chance it could neither be `isFileURI` and `isWorkspacePath` and could be a uri with a character that needs encoding?

PR #6094, opened 2026-03-17
fix: thread/session disappearing temporarily or until refresh

ivyjeong13:
Addresses #6902

* skip loading if store already initialized
* don't set loading if isNewAgent === false & store is already set
* fix styling causing two scrollbars in thread/session

### DIFF
--- a/ui/user/src/routes/agent/+layout.svelte
+++ b/ui/user/src/routes/agent/+layout.svelte
@@ -70,7 +70,6 @@
 		if (!storedChat || isNewAgent) {
 			loading = true;
 			if (isNewAgent) {
-				loading = true;
 				try {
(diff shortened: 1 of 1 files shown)


ui/user/src/routes/agent/+layout.svelte
```diff
@@ -1,76 +1,98 @@
...
-			} catch (error) {
-				console.error(error);
-				errors.append(error);
-			} finally {
-				loading = false;
+		const storedChat = get(nanobotChat);
+		if (!storedChat || isNewAgent) {
+			loading = true;
+			if (isNewAgent) {
+				loading = true;
```
g-linville: this line seems unnecessary?

ivyjeong13: Thanks Grant! I thought I had removed that haha

PR #6117, opened 2026-03-18
docs: add skills and skill access policies

g-linville:


docs/docs/functionality/skill-access-policies.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Editing a Policy
+
+Click any policy in the list to view and modify its name, users and groups, or skills. Changes take effect immediately.
+
+### Deleting a Policy
+
+Deleting a policy removes skill access for the affected users. If a user loses access to all skills as a result, agents acting on their behalf will no longer be able to search for or install skills until another policy grants them access.
+
+Skills that were already installed before losing access remain available locally.
```
njhale: nit: IIUC "locally" likely means "within a session", but I think that will be unclear to readers with the current verbiage.

g-linville: thanks, I'll update that

PR #6132, opened 2026-03-19
chore: cut docs v0.18.0

g-linville:


docs/docusaurus.config.ts
```diff
@@ -44,12 +44,6 @@ const config: Config = {
   projectName: "obot",
   onBrokenLinks: "throw",
 
-  markdown: {
-    hooks: {
-      onBrokenMarkdownLinks: "warn",
-    },
-  },
```
g-linville, line 51: This piece of configuration was no longer valid and was causing docusaurus to error out, so I just removed it.

Issue #5971, opened 2026-03-04
`Launch Agent` option appears when no model provider or default models are configured, resulting in errors


sangee2004:
**Describe the bug**
When the obot server is started without configuring any model provider, the UI still allows users to select “Launch Agent.” Attempting to launch the agent results in error messages, and even after dismissing them, the chat does not produce any responses. The agent only works once a model provider is configured and default models are set for Language Model (Chat) and Language Model (Chat - Fast). The UI should prevent launching the agent when these prerequisites are not configured.

**To Reproduce**
1. Start obot server - `docker run --pull=always -d --name obot -p 8080:8080 -v /var/run/docker.sock:/var/run/docker.sock -v mydata:/data ghcr.io/obot-platform/obot:v0.17.0-rc2`
2. Do not configure any model provider.
3. Go to `Launch Agent`.
User will be presented with the following errors:

<img width="1701" height="993" alt="Image" src="https://github.com/user-attachments/assets/c093c7ca-f38b-4490-90a6-5b986a30b4f2" />

Chatting after these errors are dismissed will result in no responses

**Note** - Once the model provider is configured and the default model is set for `Language Model (Chat)` and `Language Model (Chat - Fast)`, we will be able to launch the agent and chat with it.

**Expected behavior**
Do not provide the user's `Launch Agent` option when the model provider is not configured, and no defaults are set for  `Language Model (Chat)` and `Language Model (Chat - Fast)`

ivyjeong13:
Putting back into High, figuring out what's the best way for non-admin users to access the information needed to know a default model has been set for Language Model (Chat) and Language Model (Chat - Fast) in order to hide the 'Launch Agent'. Those endpoints are only available for those with admin access. `/models` is available but just list models the user has access to and not if any are set with default. There is a `/assistant/{id}/project/{id}/default-model for old project chats but not for `/projectv2`, maybe it needs something similar but this seems to also be on a per project level.

shwetapsdet:
Started the Obot server using the following Docker command:
`docker run --pull=always -d --name obot -p 8080:8080 -v /var/run/docker.sock:/var/run/docker.sock -v mydata:/data ghcr.io/obot-platform/obot:v0.17.0-rc2`

Observed that even without connecting a model provider, I am still able to launch an agent successfully.

https://github.com/user-attachments/assets/6edd08e9-4f20-4881-be87-6f5dab06c6fe

ivyjeong13:
[...]

PR #6158, opened 2026-03-23
fix: launch agent option disabled until default llm is set

ivyjeong13:
Addresses #5971

### DIFF
--- a/ui/user/src/lib/components/Layout.svelte
+++ b/ui/user/src/lib/components/Layout.svelte
@@ -53,11 +53,13 @@
 	import ConfigureBanner from './admin/ConfigureBanner.svelte';
 	import InfoTooltip from './InfoTooltip.svelte';
 	import { Render } from './ui/render';
-	import { Group, ModelAlias } from '$lib/services';
+	import { Group } from '$lib/services';
 	import { page } from '$app/state';
 	import SetupSplashDialog from './admin/SetupSplashDialog.svelte';
 	import { adminConfigStore } from '$lib/stores/adminConfig.svelte';
 	import { resolve } from '$app/paths';
+	import { isAgentEnabled } from '$lib/utils';
+	import { ADMIN_AGENT_DISABLED_MESSAGE, USER_AGENT_DISABLED_MESSAGE } from '$lib/constants';
 
 	type NavLink = {
 		id: string;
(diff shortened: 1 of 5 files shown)


ui/user/src/routes/+page.ts
```diff
@@ -24,9 +24,13 @@ export const load: PageLoad = async ({ fetch, url, parent }) => {
 			throw redirect(302, redirectRoute);
 		}
 
+		const agentLinkEnabled =
+			defaultModelAliases &&
+			defaultModelAliases.length > 0 &&
+			!!defaultModelAliases.find((alias) => alias.alias === ModelAlias.Llm)?.model;
```
g-linville: The issue says that `LlmMini` needs to be set as well.

ui/user/src/lib/components/navbar/Profile.svelte
```diff
@@ -52,6 +59,13 @@
 		page.url.pathname.startsWith('/agent') && !!agentId && !!projectId
 	);
 
+	let agentLinkEnabled = $derived(
+		defaultModelAliases.current.length > 0 &&
+			!!defaultModelAliases.current.find(
+				(defaultModelAlias) => defaultModelAlias.alias === ModelAlias.Llm
```
g-linville: Needs to be fixed here too.

ui/user/src/lib/components/Layout.svelte
```diff
@@ -117,6 +117,13 @@
 	let collapsed = $state<Record<string, boolean>>({});
 	let pathname = $derived(page.url.pathname);
 
+	let agentLinkEnabled = $derived(
+		defaultModelAliases.current.length > 0 &&
+			!!defaultModelAliases.current.find(
+				(defaultModelAlias) => defaultModelAlias.alias === ModelAlias.Llm
```
g-linville: And here.

ui/user/src/routes/+page.ts
ivyjeong13: I actually wasn't sure if the LlmMini was actually required (was able to with just the defaults Anthropic had set)! Thanks Grant, will add! And I'll make the tooltip a bit more specific as well!

g-linville: Yep, it at least gets used to create the thread titles I think. Not sure if we use it for anything else.

PR #6166, opened 2026-03-24
Fix error handling for oauth not found

calvinmclean:
## Summary

When an MCP OAuth client is not found in the authorize or token endpoints, the server previously propagated the raw storage error instead of returning a proper OAuth error response. This fix handles the not-found case explicitly in both endpoints.

Per [RFC 6749 §5.2](https://www.rfc-editor.org/rfc/rfc6749#section-5.2), when client authentication fails the server must respond with HTTP 401 and an `invalid_client` error. A missing client — which can occur when a previously registered client has been removed — is an authentication failure and should follow this same contract, with a description hinting that re-registration is required.

### Changes

- Added `ErrInvalidClient` error code constant to the OAuth error type set
- In `authorize`: detect `IsNotFound` on client lookup and return `HTTP 401 / invalid_client`
- In `token`: same fix for the token endpoint client lookup

## Testing/Reproducing
1. Run Obot and initialize an MCP server
2. Connect a client to this server and run some requests to make sure it works
3. Delete the `obot.db`
4. Restart Obot and re-create the MCP server
5. Rerun the MCP request from client
    - You should see an error like this open in web browser:
    ```
    OAuthClient.obot.obot.ai "oc1553gmyab6zeqbjuy2a5y542mg2" not found
    ```
6. Checkout this branch and re-run Obot
7. Try the client request again which will now re-register and succeed

### DIFF
--- a/pkg/api/handlers/mcpgateway/oauth/authorize.go
+++ b/pkg/api/handlers/mcpgateway/oauth/authorize.go
@@ -17,6 +17,7 @@ import (
 	"github.com/obot-platform/obot/pkg/auth"
 	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
 	"github.com/obot-platform/obot/pkg/system"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


calvinmclean:
I am still able to reproduce this error but I cannot seem to reproduce the fix anymore.

Issue #6157, opened 2026-03-23
Search Field Clears but Filtered Results Persist on MCP Server Page


shwetapsdet:
**Describe the bug**
When a user searches for an MCP server (e.g., "google") and navigates to a server’s detail page (e.g., Google Sheets), then launches the agent and returns back to the MCP platform, the search field appears blank. However, the results list still shows filtered data based on the previous search input instead of resetting.

**To Reproduce**
1. Navigate to the MCP Server page
2. In the search field, enter a keyword (e.g., "google")
3. Observe filtered MCP server results
4. Open any server details page (e.g., Google Sheets)
5. Click on Profile Icon -> Launch Agent
6. After navigating, click on MCP Platform to return to MCP Server page

**Expected behavior**
The system should display the default (unfiltered) MCP server list.

**Screenshots**

https://github.com/user-attachments/assets/07ad11e3-7e83-4a4e-9ada-3f1cf261d7ef

loganr0217:
@shwetapsdet I introduced a fix for this in PR #6167

ivyjeong13:
moving to testing for verification, should be addressed with #6192

shwetapsdet:
Verified it on obot main and it's working as expected 

https://github.com/user-attachments/assets/6189a37f-b329-4b9d-8ebb-ea430eb9a40f

Issue #5755, opened 2026-02-06
Recent MCP Server search remains visible even when search field is cleared on MCP Server page


shwetapsdet:
**Describe the bug**
On the MCP Server page, when a user searches for an MCP server (e.g., “ANTV Charts”) and connects to it, the searched server appears in the list. However, when the user later returns to the MCP Server page, the previously searched MCP server remains visible even though the search input field is cleared.

**To Reproduce**

1. Go to the MCP Server page
2. Search for an MCP server (e.g., “ANTV Charts”)
3. Connect to the searched MCP server
4. Return to the MCP Server page again

**Expected behavior**
When the search field is cleared, no previously searched MCP servers should be filtered

**Screenshots**

https://www.loom.com/share/70af720e76a948268e6e5c367b80ef9d

shwetapsdet:
Verified it on obot main and it's working as expected 

https://github.com/user-attachments/assets/098ddc25-287e-4407-bd12-ed8d05f14f25

PR #6167, opened 2026-03-24
fix: clear stale search query on admin mcp-servers page when returning from external route

loganr0217:
Addresses #6157

## Summary

- When a user searches on `/admin/mcp-servers`, clicks into a detail page, navigates to a completely different route (e.g. `/agent`), then returns to `/admin/mcp-servers`, the search results were showing the previous search term even though the search bar appeared empty.
- The `beforeNavigate` hook that clears localStorage only fires from the `/admin/mcp-servers` page itself — it never fires when the user leaves via a detail sub-route (`/admin/mcp-servers/c/...`), so the stale query persisted.
- Fix: added a check in `afterNavigate` to clear saved search queries from localStorage when arriving from outside `/admin/mcp-servers`.

## Test plan

- [ ] Search for a term (e.g. "google") on `/admin/mcp-servers`
- [ ] Click into a server detail page
- [ ] Navigate to a different section (e.g. `/agent`)
- [ ] Navigate back to `/admin/mcp-servers`
- [ ] Verify the search bar is empty and default results are shown (not the previous search results)

### DIFF
--- a/ui/user/src/routes/admin/mcp-servers/+page.svelte
+++ b/ui/user/src/routes/admin/mcp-servers/+page.svelte
@@ -105,6 +105,11 @@
 				return;
 			}
 
+			// If coming from outside /admin/mcp-servers, clear stale search queries
+			if (!from?.url?.pathname.startsWith('/admin/mcp-servers')) {
+				clearQueryFromLocalStorage();
+			}
+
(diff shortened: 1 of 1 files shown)


ivyjeong13:
Thanks Logan! Will be closing this one out -- there is a similar ticket #5755 as well and will be removing the localStorage so that query becomes the source of truth in the URL params.

Issue #5848, opened 2026-02-20
Deleting admin user also deletes associated multi-user mcp server entries.


sangee2004:
**Describe the bug**
When an admin user who created MCP server entries is deleted, any multi-user MCP server they created is also removed from the MCP servers list. 

Pre Req - Have a setup with multiple admin/owner users.

**To Reproduce**
1. Log in as admin user, say `user1`.
2. Create a single-user MCP server entry, say `test single`.
3. Create a multi-user MCP server entry, say `test multi`.
4.  Log in as a different admin user, say `user2`.
Note that both `test single` and `test multi` entries are listed in the `MCP servers` page
5. Delete `user1` from step 1.
6. Go to the `MCP servers` page
Notice that only the `test single` MCP server is present.

**Expected behavior**
The multi-user MCP server entry should not get deleted when the admin user who created it gets deleted.
Note: Single-user, remote, and composite MCP server entries remain present even after the admin user who created them is deleted, as expected.

g-linville:
Merged the fix.

sangee2004:
This issue is addressed when testing with the latest obot server build from main.

Deleting the admin user does not delete multi-user MCP server entries created by this user.

PR #6169, opened 2026-03-25
fix: retain multi-user servers when an admin is deleted

g-linville:
for https://github.com/obot-platform/obot/issues/5848

The user ID of the admin who creates a multi-user server in the default catalog gets put on the MCPServer for it, which was getting deleted if the admin ever got deleted. This fixes that by retaining it.

Multi-user servers created by PUPs in their PowerUserWorkspace are still deleted if they are deleted.

### DIFF
--- a/pkg/controller/handlers/cleanup/user.go
+++ b/pkg/controller/handlers/cleanup/user.go
@@ -87,8 +87,7 @@ func (u *UserCleanup) Cleanup(req router.Request, _ router.Response) error {
 	deletedServers := 0
 	for _, server := range servers.Items {
 		// Skip multi-user servers in the default MCPCatalog — they should persist after user deletion.
-		// Servers in a PowerUserWorkspace should still be deleted.
-		if server.Spec.MCPCatalogID == system.DefaultCatalog && server.Spec.PowerUserWorkspaceID == "" {
+		if server.Spec.MCPCatalogID == system.DefaultCatalog {
 			continue
 		}
 		if err := kclient.IgnoreNotFound(req.Delete(&server)); err != nil {


pkg/controller/handlers/cleanup/user.go
```diff
@@ -84,12 +84,19 @@ func (u *UserCleanup) Cleanup(req router.Request, _ router.Response) error {
 		return err
 	}
 
+	deletedServers := 0
 	for _, server := range servers.Items {
+		// Skip multi-user servers in the default MCPCatalog — they should persist after user deletion.
+		// Servers in a PowerUserWorkspace should still be deleted.
+		if server.Spec.MCPCatalogID == system.DefaultCatalog && server.Spec.PowerUserWorkspaceID == "" {
```
thedadams: Is there any way for `MCPCatalogID == system.DefaultCatalog` and `PowerUserWorkspaceID != ""`? My understanding is that is not possible. If it is, then can you help me understand a situation where that does occur? I want to make sure my mental map of our data structures is consistent.

g-linville: No, there isn't. Good point. I'll remove the second part of this if statement.

pkg/controller/handlers/cleanup/user.go
```diff
@@ -84,12 +84,18 @@ func (u *UserCleanup) Cleanup(req router.Request, _ router.Response) error {
 		return err
 	}
 
+	deletedServers := 0
 	for _, server := range servers.Items {
+		// Skip multi-user servers in the default MCPCatalog — they should persist after user deletion.
+		if server.Spec.MCPCatalogID == system.DefaultCatalog {
+			continue
+		}
```
calvinmclean, line 92: I am out of the loop here. @g-linville could you help me understand how this relates to multi-user servers?

thedadams: Multi-user MCP servers are those where `server.Spec.MCPCatalogID != ""`.

g-linville: Yep, basically each MCP server has a place where it "lives". Multi-user servers always live in an MCPCatalog (the only one of which is `default`) or a PowerUserWorkspace (if they were created by a user with PowerUserPlus permissions). Other MCPServers are for individual users and are deployed from MCPServerCatalogEntries. These "live" at the context of the user himself, so they have no MCPCatalog or PowerUserWorkspace on their spec.

PR #6172, opened 2026-03-26
user impersonation UI

g-linville:


pkg/api/handlers/nanobotagent.go
```diff
@@ -31,6 +32,27 @@ func NewNanobotAgentHandler(sessionManager *mcp.SessionManager, serverURL string
 	}
 }
 
+func (h *NanobotAgentHandler) ListAll(req api.Context) error {
+	if !req.UserIsOwner() && !req.UserIsAdmin() && !req.UserIsAuditor() {
```
njhale, line 36: Is the `UserIsAdmin` check necessary here if we're already checking that they're Owners?

Issue #6143, opened 2026-03-19
[Okta] - Login broken for users with duplicate Okta groups


cloudnautique:
**Describe the bug**
Users who have the same group listed twice because they are a member of a group that is also nested into another group they belong to. 

When the user logs in they aren't able to connect the Obot server throws:
```time="2026-03-19T17:15:52Z" level=error msg="sql query error" affected=0
  caller="/app/pkg/gateway/client/group.go:231"
  elapsed=2.66269ms
  error="ERROR: ON CONFLICT DO UPDATE command cannot affect row a second time (SQLSTATE 21000)"
```

**To Reproduce**

Pre-req: Obot configured with Okta provider

1. Create a group in Okta and assign it to a user.
2. Create a second group in Okta, and add the group from step 1.
3. Assign the user from step 1 into the group created in step 2.
4. Try to log in to Obot

**Expected behavior**
User logs in, and group assignments function.

justinas-b:
I am facing same problem. I see query being executed is:
```
INSERT INTO "groups" ("id","auth_provider_name","auth_provider_namespace","name","icon_url")
VALUES
    ('okta/a','okta-auth-provider','default','a',NULL),
    ('okta/Team1','okta-auth-provider','default','Team1',NULL),
    ('okta/Team1','okta-auth-provider','default','Team1',NULL),
    ('okta/Team1','okta-auth-provider','default','Team1',NULL),
    ('okta/full-vpn','okta-auth-provider','default','full-vpn',NULL),
    ...
ON CONFLICT ("id")
DO UPDATE SET "name"="excluded"."name","icon_url"="excluded"."icon_url"
```

I think values needs to be deduplicated before update

g-linville:
The full migration fix for this has been merged. Unfortunately, there is no easy way to test this. It would require having an LDAP setup somewhere and importing it into Okta (which I'm not even sure is something you can do on the free plan).

PR #6177, opened 2026-03-26
chore: migrate okta groups to group ID

g-linville:
for https://github.com/obot-platform/obot/issues/6143

depends on https://github.com/obot-platform/enterprise-tools/pull/99

This adds a migration to change the way we format Okta groups from `okta/<group name>` to `okta/<group ID>`. This will allow us to distinguish between multiple groups that have the same name, which as it turns out is possible in Okta.

All the database-level changes take place in a transaction, where we basically attempt to acquire a "lock" in the database by adding a row to the `migrations` table. (This gets undone if the transaction fails.)

Then, we make a best-effort attempt to fix the groups in AccessControlRules, SkillAccessRules, and ModelAccessPolicies. If for any reason these fail, warning logs are printed, but there is really no reason that any of them should fail, nor did I encounter failure cases in testing.

### DIFF
--- a/pkg/controller/routes.go
+++ b/pkg/controller/routes.go
@@ -88,6 +88,7 @@ func (c *Controller) setupRoutes() {
 	projectMCPServerHandler := projectmcpserver.NewHandler()
 	systemMCPServerHandler := systemmcpserver.New(c.services.GPTClient, c.services.MCPLoader, c.services.ServerURL)
 	nanobotAgentHandler := nanobotagent.New(c.services.GPTClient, c.services.PersistentTokenServer, c.services.GatewayClient, c.localK8sRouter, c.services.NanobotAgentImage, c.services.ServerURL, c.services.MCPServerNamespace, c.services.MCPLoader)
+	oktaGroupMigrationHandler := oktagroupmigration.New()
 
 	// Runs
 	root.Type(&v1.Run{}).FinalizeFunc(v1.RunFinalizer, runs.DeleteRunState)
@@ -314,7 +315,6 @@ func (c *Controller) setupRoutes() {
 	root.Type(&v1.GroupRoleChange{}).HandlerFunc(powerUserWorkspaceHandler.HandleGroupRoleChange)
 
 	// OktaGroupMigration
-	oktaGroupMigrationHandler := oktagroupmigration.New()
 	root.Type(&v1.OktaGroupMigration{}).HandlerFunc(oktaGroupMigrationHandler.Migrate)
 
 	// PowerUserWorkspace


pkg/gateway/client/okta_group_migration.go
```diff
@@ -0,0 +1,244 @@
...
+	defer c.oktaGroupMigrationMu.Unlock()
+
+	if c.oktaGroupMigrationDone {
+		return nil
+	}
+
+	if err := c.migrateOktaGroupIDs(ctx, authProviderURL, authProviderNamespace, authProviderName); err != nil {
+		return err
+	}
+
```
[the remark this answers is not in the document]
g-linville: These processes run in the same container, so this is not a concern.

pkg/gateway/client/group.go
```diff
@@ -183,6 +183,14 @@ func (c *Client) ensureGroups(ctx context.Context, tx *gorm.DB, identity *types.
 		now            = time.Now()
 		nextGroupCheck = identity.AuthProviderGroupsLastChecked.Add(groupCheckPeriod)
 	)
+
+	// Run one-time Okta group ID migration if this is an Okta auth provider
+	if providerURL != "" && identity.AuthProviderName == "okta-auth-provider" {
+		if err := c.runOktaGroupIDMigrationOnce(ctx, providerURL, identity.AuthProviderNamespace, identity.AuthProviderName); err != nil {
+			log.Warnf("Okta group ID migration failed (will retry): %v", err)
+		}
+	}
```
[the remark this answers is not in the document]
g-linville: Not necessary; it won't return duplicates anymore.

g-linville:
not sure how I accidentally closed this lol

pkg/gateway/client/okta_group_migration.go
```diff
@@ -0,0 +1,264 @@
...
+	// only the replica whose insert succeeds (RowsAffected == 1) proceeds with the migration.
+	// This is safe in multi-replica setups because the migrations row acts as an atomic claim.
+	migrationRan := false
+	if err := c.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		// Attempt to claim the migration. OnConflict DoNothing means only one replica wins.
+		result := tx.Clauses(clause.OnConflict{DoNothing: true}).Create(&types.Migration{Name: oktaGroupMigrationName})
+		if result.Error != nil {
+			return fmt.Errorf("okta migration error: failed to claim migration: %w", result.Error)
+		}
+		if result.RowsAffected == 0 {
```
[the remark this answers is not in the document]
calvinmclean: I was coming here to make this same comment 😅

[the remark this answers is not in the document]
g-linville: oh, yeah, I'll get this fixed. didn't notice this round of feedback from Copilot

pkg/gateway/client/okta_group_migration.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	if !claimed {
+		// Another replica already completed or is completing the migration
+		return nil
+	}
+
+	// Run the data migration in a separate transaction.
+	// If this fails, we delete the claim row so it can be retried.
+	if err := c.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
```
thedadams: If this fails for some reason (could be as simple as the pod restarts), then the migration will never happen.

calvinmclean: Maybe we go back to the previous approach with the larger transaction and have a timeout? Slow DB would cause issues in that case but seems less likely than this

g-linville: I'm confused, why is this the case? If the transaction fails, we delete the entry that we made in the `migrations` table so that it can retry.

g-linville: Oh, like if the process is terminated mid-transaction. I see. Let me rethink this

pkg/gateway/client/okta_group_migration.go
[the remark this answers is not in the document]
g-linville: I think I am gonna switch the implementation back to this. It's a tradeoff but it's better than just not working, lol.

pkg/gateway/client/okta_group_migration.go
g-linville: Okay, I switched this back to the previous implementation, where we insert into the `migrations` table at the beginning of the same transaction as the migration itself. The tradeoff is that other replicas would be stuck while they wait for the migration to finish, but honestly that is totally fine. This is a one-time migration anyway and I don't think it should take that long to process.

pkg/gateway/client/okta_group_migration.go
```diff
@@ -0,0 +1,267 @@
...
+			}
+		}
+
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	// Update CRDs (best-effort after DB migration succeeds)
+	c.migrateOktaGroupIDsInCRDs(ctx, idMap)
```
thedadams: This has the same problem I pointed out previously: if this fails for some reason, then it will never happen.

Is there a way we can handle this in the controller? Maybe touch something the controller will look at that will kick off this process?

g-linville: yeah I'll see what I can do there

calvinmclean: Could we just move it into the transaction?

g-linville: I decided to follow the same pattern that we use for `UserDelete`. I think this is the most reliable way to ensure it completes.

g-linville: > Could we just move it into the transaction?

The types we are migrating here are not managed directly through SQL so I can't do a transaction for them. They are managed by the [kinm](https://github.com/obot-platform/kinm) Kubernetes-compatible storage API and the [nah](https://github.com/obot-platform/nah) controller.

pkg/controller/handlers/oktagroupmigration/migrate.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+	}
+	return nil
+}
+
+func migrateModelAccessPolicies(req router.Request, idMap map[string]string) error {
+	var list v1.ModelAccessPolicyList
+	if err := req.Client.List(req.Ctx, &list); err != nil {
+		return fmt.Errorf("okta group migration: failed to list ModelAccessPolicies: %w", err)
+	}
```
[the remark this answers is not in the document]
g-linville: we don't use namespaces

pkg/controller/handlers/oktagroupmigration/migrate.go
```diff
@@ -0,0 +1,95 @@
...
+
+	return req.Delete(migration)
+}
+
+func migrateAccessControlRules(req router.Request, idMap map[string]string) error {
+	var list v1.AccessControlRuleList
+	if err := req.Client.List(req.Ctx, &list); err != nil {
+		return fmt.Errorf("okta group migration: failed to list AccessControlRules: %w", err)
+	}
```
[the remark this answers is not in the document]
g-linville: we don't use namespaces

pkg/controller/handlers/oktagroupmigration/migrate.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+	}
+	return nil
+}
+
+func migrateSkillAccessRules(req router.Request, idMap map[string]string) error {
+	var list v1.SkillAccessRuleList
+	if err := req.Client.List(req.Ctx, &list); err != nil {
+		return fmt.Errorf("okta group migration: failed to list SkillAccessRules: %w", err)
+	}
```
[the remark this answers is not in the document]
g-linville: we don't use namespaces

pkg/gateway/client/okta_group_migration.go
```diff
@@ -0,0 +1,216 @@
...
+// "okta/{name}" to "okta/{name}_{id}" format across all data stores.
+func (c *Client) migrateOktaGroupIDs(ctx context.Context, authProviderURL, authProviderNamespace, authProviderName string) error {
+	// Fast path: check if migration has already been completed (avoids fetching the mapping)
+	var migration types.Migration
+	err := c.db.WithContext(ctx).Where("name = ?", oktaGroupMigrationName).First(&migration).Error
+	if err == nil {
+		// Migration already done
+		return nil
+	}
+	if !errors.Is(err, gorm.ErrRecordNotFound) {
```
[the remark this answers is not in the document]
g-linville: Fixed by moving the creation of the OktaGroupMigration into the transaction

pkg/gateway/client/okta_group_migration.go
```diff
@@ -0,0 +1,216 @@
...
+	//   - On rollback (crash/error): the claim disappears and the other replica retries.
+	if err := c.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		result := tx.Clauses(clause.OnConflict{DoNothing: true}).Create(&types.Migration{Name: oktaGroupMigrationName})
+		if result.Error != nil {
+			return fmt.Errorf("okta migration error: failed to claim migration: %w", result.Error)
+		}
+		if result.RowsAffected == 0 {
+			// Another replica already completed the migration
+			return nil
+		}
```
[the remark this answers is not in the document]
g-linville: Fixed by moving the creation of the OktaGroupMigration into the transaction

g-linville:
This is ready for review again.

pkg/gateway/client/okta_group_migration.go
calvinmclean: Ok so now we are using the DB lock to do all of the DB migrations, then creating the OktaGroupMigration in the controller which will exist until that part of the process is complete?

We should be fine without a timeout on the `ON CONFLICT` query since the Tx is just running some SQL and then creating the `OktaGroupMigration` which should be pretty quick.

g-linville: Yep, exactly.

g-linville: Yep, exactly.

pkg/controller/routes.go
```diff
@@ -312,6 +313,10 @@ func (c *Controller) setupRoutes() {
 	// GroupRoleChange
 	root.Type(&v1.GroupRoleChange{}).HandlerFunc(powerUserWorkspaceHandler.HandleGroupRoleChange)
 
+	// OktaGroupMigration
+	oktaGroupMigrationHandler := oktagroupmigration.New()
```
thedadams: nit: all of the other controller instantiations are at the top of the file.

g-linville: fixed

PR #6181, opened 2026-03-27
fix: uncaught happening on browser resize in /agent

ivyjeong13:
* another fix for potential Uncaught (can happen when resizing browser mid agent running)


ivyjeong13:
Still seeing something happening when resizing from desktop to mobile, investigating!

ivyjeong13:
Moved out of draft, found where issue was and addressed. Verified on Chrome & Safari, resizing various times from desktop to mobile. Also tested on /agent route and an existing thread/session on /agent/p/{id}?tid={tid} route.

PR #6203, opened 2026-03-31
fix: make policy evaluation work with OpenAI

g-linville:
This was initially implemented for the chat completions API, but turns out we use responses.


g-linville:
@njhale yes, the Anthropic stuff was already in here and working. This is just stripping out chat completions for OpenAI responses

PR #6229, opened 2026-04-07
fix: set non-agent env vars to empty string for agents

g-linville:
When we first dropped audit logs from the Nanobot agent containers, we didn't provide a way to migrate the ones that already had them configured. The old environment variables were sticking around in the secret, due to the way we were patching it. This change explicitly sets them to the empty string.

This does mean that, in the Nanobot agent container, all of these env vars will be set to `""` (rather than being completely unset), but this should not cause any problems.

The way this bug manifested was that older Nanobot agents in Main were still sending audit logs back to Obot.


g-linville:
I talked to Donnie. The better thing to do here is to stop using StringData in our k8s Secrets and just use Data. I'm gonna do that.

PR #6321, opened 2026-04-10
fix: ensure docker images are redeployed when the base image changes

thedadams:
This already work for Kubernetes, but was broken for docker.

### DIFF
--- a/pkg/mcp/docker.go
+++ b/pkg/mcp/docker.go
@@ -316,11 +316,12 @@ func (d *dockerBackend) ensureDeployment(ctx context.Context, server ServerConfi
 			currentFileEnvKeysHash = ""
 		}
 
+		desiredImage := d.deploymentImage(server)
 		if existing.Labels["mcp.config.hash"] != configHash ||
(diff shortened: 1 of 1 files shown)


calvinmclean (approved):

pkg/mcp/docker.go
```diff
@@ -382,6 +383,17 @@ func (d *dockerBackend) ensureDeployment(ctx context.Context, server ServerConfi
 	return d.createAndStartAndWaitForContainer(ctx, server, mcpServerName, configHash, desiredFileEnvKeysHash, containerEnv, webhooks)
 }
 
+func (d *dockerBackend) deploymentImage(server ServerConfig) string {
+	switch server.Runtime {
+	case otypes.RuntimeUVX, otypes.RuntimeNPX:
```
g-linville, line 388: Is `RuntimeContainerized` supposed to be here too?

thedadams: When the runtime is "containerized" then we will use the specified image. Additionally, any changes to that image will be picked up when we hash the server configuration, so we don't have to deal with that here.

g-linville (approved):

PR #6333, opened 2026-04-14
enhance: make workflow sharing more granular

g-linville:
for https://github.com/obot-platform/obot/issues/6255

This changes how workflow sharing works. Previously, we had just a `public/private` toggle which changed whether all Obot users could see the workflow, or if only the owner could see the workflow.

Now, workflow owners can assign specific groups and users (or all Obot users) to have access to their workflow, and can do this on a per-version basis. Each newly published version of the workflow will, by default, inherit the subjects on the previous version.

We also migrate from the old sharing system to the new one. Private workflows will have no subjects set on all versions (so only the owner can access them). Public workflows will have the `All Obot Users` group (in the backend, the `*` selector) as the subject on each version.


ui/user/src/lib/components/nanobot/PublishedWorkflowVersionDialog.svelte
```diff
@@ -14,32 +20,51 @@
...
+	let activeSubjects = $derived(activeVersion?.subjects ?? []);
+
+	$effect(() => {
+		Promise.all([
+			AdminService.listUsers().catch(() => []),
+			AdminService.listGroups().catch(() => [])
+		]).then(([loadedUsers, loadedGroups]) => {
+			users = loadedUsers;
+			groups = loadedGroups;
+		});
```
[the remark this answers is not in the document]
g-linville: false

pkg/gateway/server/user.go
```diff
@@ -368,46 +365,22 @@ func (s *Server) listAuthGroups(apiContext api.Context) error {
...
+	if userIsBasicOrPower(apiContext.User) {
+		trimmedGroups := make([]types.Group, 0, len(groups))
+		for _, group := range groups {
+			trimmedGroups = append(trimmedGroups, types.Group{
+				ID:   group.ID,
+				Name: group.Name,
+			})
 		}
+		return apiContext.Write(trimmedGroups)
 	}
```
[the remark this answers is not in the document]
g-linville: Yeah the GitHub auth provider hasn't had groups for a while now, so this has been dead code for several months. Just cleaning it up.

docs/docs/functionality/workflow-sharing.md
```diff
@@ -62,23 +63,26 @@ Installation behavior:
...
-- `public`: Other users can discover it in search results and install it
+For each published version:
 
-:::important
-Published workflows are always `private` by default. In order to become `public`, the user needs to go to the Obot UI and explicitly make it public. The Obot Agent is unable to change the visibility of published workflows.
-:::
+- Empty subject list: only the workflow owner and admins can view or download that version
+- Specific `user` subjects: only those users can access that version
+- Specific `group` subjects: members of those groups can access that version
+- `selector:*`: all authenticated Obot users can discover and install that version
```
thedadams, line 75: Is it worth it to state that the workflow owner and admins/owners can view or download regardless of the `user` or `group` subjects specified?

pkg/api/handlers/publishedartifact.go
```diff
@@ -208,6 +216,7 @@ func (h *PublishedArtifactHandler) createNewArtifact(req api.Context, data []byt
 					BlobKey:     blobKey,
 					Description: manifest.Description,
 					CreatedAt:   *types.NewTime(time.Now()),
+					Subjects:    nil,
```
thedadams: micro-nit:
```suggestion
```

pkg/api/handlers/publishedartifact.go
```diff
@@ -470,21 +482,32 @@ func (h *PublishedArtifactHandler) Update(req api.Context) error {
...
+		entry := findVersionEntry(&artifact, version)
+		if entry == nil {
+			return types.NewErrNotFound("version %d not found", version)
 		}
-		log.Debugf("Updating artifact %s visibility: %q -> %q", id, artifact.Spec.Visibility, *update.Visibility)
-		artifact.Spec.Visibility = *update.Visibility
+		log.Debugf("Updating artifact %s version %d subjects (count=%d)", id, version, len(update.Subjects))
+		entry.Subjects = cloneSubjects(update.Subjects)
 	}
+	artifact.Spec.LegacyVisibility = ""
```
thedadams:
```suggestion
```

pkg/api/handlers/publishedartifact.go
```diff
@@ -520,22 +543,145 @@ func (h *PublishedArtifactHandler) Delete(req api.Context) error {
...
 		return nil
 	}
 	return types.NewErrNotFound("artifact %s not found", req.PathValue("id"))
 }
 
+func (h *PublishedArtifactHandler) canAccessArtifact(artifact *v1.PublishedArtifact, requester user.Info, isAdmin bool) bool {
+	if requester == nil {
+		return false
+	}
+	if artifact.Spec.AuthorID == requester.GetUID() || isAdmin {
```
thedadams: nit: do the easy thing first, here and others.
```suggestion
	if isAdmin || artifact.Spec.AuthorID == requester.GetUID() {
```

pkg/api/handlers/publishedartifact.go
```diff
@@ -520,22 +543,145 @@ func (h *PublishedArtifactHandler) Delete(req api.Context) error {
...
+	}
+	return nil
+}
+
+func versionSubjects(artifact *v1.PublishedArtifact, version int) []types.Subject {
+	entry := findVersionEntry(artifact, version)
+	if entry == nil {
+		return nil
+	}
+	return cloneSubjects(entry.Subjects)
```
thedadams: You don't specifically need the `cloneSubjects`.
```suggestion
	return entry.Subjects[:]
```

pkg/api/handlers/publishedartifact.go
```diff
@@ -754,8 +899,31 @@ func convertPublishedArtifact(a *v1.PublishedArtifact) types.PublishedArtifact {
...
+func visibleVersionSummaries(a *v1.PublishedArtifact, requester user.Info, isAdmin bool) []types.PublishedArtifactVersionSummary {
+	if len(a.Status.Versions) == 0 {
+		return nil
+	}
+
+	result := make([]types.PublishedArtifactVersionSummary, 0, len(a.Status.Versions))
+	for _, version := range a.Status.Versions {
+		if requester == nil {
+			continue
+		}
```
thedadams: If `request == nil`, then we can return early instead of looping.

pkg/controller/migrate.go
```diff
@@ -26,3 +28,41 @@ func addCatalogIDToAccessControlRules(ctx context.Context, client kclient.Client
...
+		switch artifact.Spec.LegacyVisibility {
+		case "public":
+			subjects = []types.Subject{{
+				Type: types.SubjectTypeSelector,
+				ID:   "*",
+			}}
+		case "private":
+			subjects = nil
+		default:
+			return fmt.Errorf("invalid legacy visibility %q for published artifact %s", artifact.Spec.LegacyVisibility, artifact.Name)
```
thedadams: This is unlikely to get hit, but, in general, I think it's better to log the error and select a reasonable default. Otherwise, Obot will fail to start until the erroneous visibility is fixed, but it may be impossible to fix it if Obot can't start.

pkg/controller/migrate.go
```diff
@@ -26,3 +28,41 @@ func addCatalogIDToAccessControlRules(ctx context.Context, client kclient.Client
...
+		}
+
+		for j := range artifact.Status.Versions {
+			artifact.Status.Versions[j].Subjects = append([]types.Subject(nil), subjects...)
```
thedadams: This seems strange. I would guess that the intention is to copy, but `subjects` isn't used after this, so I think it's OK to just
```suggestion
			artifact.Status.Versions[j].Subjects = subjects
```

docs/docs/functionality/workflow-sharing.md
g-linville: Yes

pkg/api/handlers/publishedartifact.go
g-linville: Thanks

pkg/controller/migrate.go
g-linville: Good call

pkg/controller/migrate.go
g-linville: Ah, yeah this was a weird Copilot suggestion. I shouldn't have trusted it lol

docs/docs/functionality/workflow-sharing.md
g-linville: Fixed

pkg/api/handlers/publishedartifact.go
g-linville: Fixed

pkg/api/handlers/publishedartifact.go
g-linville: Fixed

pkg/api/handlers/publishedartifact.go
g-linville: Fixed (here and the other places)

pkg/api/handlers/publishedartifact.go
g-linville: Fixed

pkg/controller/migrate.go
g-linville: Fixed. We just default to no subjects (private) and log the error if the legacy visibility was invalid.

pkg/controller/migrate.go
g-linville: Fixed

g-linville:
I'll fix the conflict after approvals so that I do not mess up the review history.

thedadams (approved):

calvinmclean (approved):

Issue #6326, opened 2026-04-13
503 Errors on Launch when Istio is configured to inject proxy sidecars


njhale:
**Describe the bug**
When Istio is configured on a k8s cluster such that istio-proxy sidecars are injected into MCP server pods deployed by Obot, users can run into `MCP Server Launch Failed` with `503` errors at launch.

**To Reproduce**
1. In the Obot admin UI, where Obot is deployed to a k8s cluster with istio sidecar injection enabled, go to `MCP Management` -> `MCP Servers`
2. Click on a remote MCP server (e.g. Gmail, Atlassian, etc)
3. Click `Connect to Server`
4. Wait 10-15 seconds
5. Encounter launch error

**Expected behavior**
Obot waits the full 60 seconds to detect readiness before showing a non-terminal launch error.

**Screenshots**

<img width="1187" height="112" alt="Image" src="https://github.com/user-attachments/assets/fb608ef7-e778-4e05-831b-323a16497f28" />

**Additional context**
- If you don't delete the server instance, it will eventually become healthy and can be connected to with external MCP clients
- This was first observed with a remote MCP server (specifically Atlassian), but the problem may exist with single/multi user servers as well

sangee2004:
I am able to reproduce the issue when testing in a K8S cluster with Istio with istio sidecar injection enabled.

When testing with the obot server in `0.19.0` release, the following error is seen when connecting to remote MCP server

<img width="1244" height="715" alt="Image" src="https://github.com/user-attachments/assets/81f8bd0c-c0e4-429b-b410-be07dbee65a2" />

Upgraded obot server to [v0.19.1-rc1+dd944906](https://github.com/obot-platform/obot/commit/dd944906) release.

I can see the remote MCP server connection working as expected without encountering 503 errors.

When connecting with single-user MCP servers, I see the following error being presented

<img width="1192" height="667" alt="Image" src="https://github.com/user-attachments/assets/0dd8d63e-04a7-47e4-b765-02dc199dedc1" />


Note - I see the same error when testing with the OBOT_SERVER_MCPREMOTE_SHIM_BASE_IMAGE from `main` as well - `ghcr.io/nanobot-ai/nanobot:main` ( RC comes with `ghcr.io/nanobot-ai/nanobot:v0.0.66`)

@njhale Reopening this issue to follow up on the issue seen with single-user MCP servers.

njhale:
[...]

PR #6343, opened 2026-04-14
fix: avoid false health check failures when istio sidecar is present

njhale:
When Istio's envoy sidecar proxy is injected into MCP server pods, it
returns HTTP 503 during its startup phase before the upstream service is
reachable. The ensureServerReady health check previously treated any 503
as a permanent failure from nanobot, causing MCP servers to falsely
report as unhealthy.

Add a 10-second grace period before treating consecutive 503 responses
as a permanent failure, giving the sidecar time to initialize. Also
handle HTTP 500 as an immediate failure signal, which is what current
nanobot versions return when tool listing permanently fails (instead of
503). The 503 grace period is retained for backward compatibility with
older nanobot versions.

Addresses https://github.com/obot-platform/obot/issues/6326

### DIFF
--- a/pkg/mcp/backend.go
+++ b/pkg/mcp/backend.go
@@ -16,8 +16,9 @@ import (
 )
 
 const (
-	defaultContainerPort = 8099
-	webhookToolName      = "fire-webhook"
+	defaultContainerPort          = 8099
+	webhookToolName               = "fire-webhook"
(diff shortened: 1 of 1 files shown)


g-linville (approved):
Thank you for the clear explanation and comments

thedadams (approved):

Issue #6341, opened 2026-04-14
Provide a way to configure credentials to pull catalogs from private repositories in the UI


cloudnautique:
Users need a way to be able to configure credentials to pull private repositories in the UI.
I would suggest making it part of the add registry modal where we currently take the repo url, and make the PAT an optional field(s). 

We should stick to the HTTPS form, which requires basic auth.
I believe in the case of GitHub and GitLab we can use arbitrary non-blank usernames. [Gitlab](https://docs.gitlab.com/user/profile/personal_access_tokens/?utm_source=chatgpt.com) 
github similiarly has refrences to x-access-token as a username, but reports state it's not used for anything.

We should do clones like:

```bash
git -c http.extraHeader="Authorization: Basic $(echo -n 'x-access-token:<PAT>' | base64)" \
    clone https://github.com/org/repo.git
```

instead of:

```bash
git clone https://username:<pat>@repo.url
```

If we must use usernames, that field should be exposed so the user can fill it in.
If we believe compatibility would be improved, we can expose a username field.

justinas-b:
PAT's tend to expire 😿 what about integrating with GitHub apps and use client id/client secret instead?

sangee2004:
This issue is addressed when testing with the latest build from `main` - [v0.0.0-dev+207b9bec](https://github.com/obot-platform/obot/commit/207b9bec)

Able to add the MCP catalog hosted in a private GitHub repo.

1. Go to `MCP Servers`
2. Click on 'Add MCP Server' -> `Add server(s) from Git`
The user will now be able to add PAT when adding private repositories.
<img width="511" height="316" alt="Image" src="https://github.com/user-attachments/assets/f539bc49-8a05-4bbb-ad54-d53fc5c0be6e" />

3. In the `Source URL` field, add a private repository hosted in GitHub like `https://github.com/sangee2004/mytestmcprepo`. Add a valid PAT in the `Personal Access Token` field.
4. Click on Add.
MCP catalog entries from the private GitHub repo get synced as expected.

4. Make some changes to the MCP catalog entries in the GitLab repo.
5 Go to `MCP Servers`
Using the `Sync` option syncs the changes from the GitLab repo made to the MCP servers list.

calvinmclean:
@justinas-b I created a new issue to track the Github apps feature request/enhancement. Thanks!

obot-platform/prospect-and-customer-issues#5

Issue #5543, opened 2026-01-16
Add support for MCP catalog not hosted on github.com


j-jacobi:
**Describe the bug**

Currently you can only add Git MCP catalogs that are hosted on GitHub. It would be nice to support any HTTP git server.

**To Reproduce**
1. Go to MCP Servers -> Registry Sources
2. Click on 'Add MCP Server' -> 'Add server(s) from Git'
3. Add 'https://gitlab.com/jjacobi/tmp-test.git' as the source URL
4. See error on sync of MCP servers

**Expected behavior**
The Git repository URL should be resolved and load the MCP YAML configuration inside of it.

**Screenshots**
<img width="2082" height="899" alt="Image" src="https://github.com/user-attachments/assets/21eb80b3-5c80-451a-af0e-416f5dfe743a" />

cjellick:
Yep, good idea. This idea has been on our radar for awhile. Thanks for the PR. We'll take a look at it.

cjellick:
@njhale - a little further down the line (like after your next feature), take a look at this.

Should probably give the PR a look sooner than that just to get a rough idea of the quality there.

cjellick:
Assigning to @calvinmclean because it is related to or a possible dupe of #6341

sangee2004:
This issue is addressed when testing with the latest obot server build from `main` - [v0.0.0-dev+ead00513](https://github.com/obot-platform/obot/commit/ead00513)

Able to add the MCP catalog hosted in GitLab.com. 

1. Go to `MCP Servers`
2. Click on 'Add MCP Server' -> `Add server(s) from Git`
The user will now be able to add PAT when adding private repositories.
<img width="511" height="316" alt="Image" src="https://github.com/user-attachments/assets/f539bc49-8a05-4bbb-ad54-d53fc5c0be6e" />

3. In the Source URL field,  add a repository hosted in GitLab like  `https://gitlab.com/sangee2004/mymcp-catalog`.In my case, this is a private repository, so I had to add a PAT as well, which isn't needed when accessing a public repository.
4. Click on `Add`.

MCP catalog entries from the GitLab repo get synced as expected.

5. Make some changes to the MCP catalog entries in the GitLab repo.
6. Go to `MCP Servers`

Using the `Sync` option syncs the changes from the GitLab repo made to the MCP servers list.


@j-jacobi This feature will be available in the upcoming obot release.

PR #6351, opened 2026-04-15
Add support for MCP Catalogs from private repos and non-github repos

calvinmclean:
## Summary

Addresses #6341 #5543

Adds support for pulling MCP catalogs from private git repositories, and generalizes catalog source URL support beyond GitHub to any HTTPS git host.

### Credential storage
- Admins can supply a Personal Access Token (PAT) when adding a git source URL via the UI
- Tokens are stored securely in the GPTScript credential store (keyed by catalog UID + URL hash), never in the resource spec
- The API returns `"*"` as a presence indicator — the actual token is never exposed

### Git platform support
- **GitHub**: supported with or without `.git` suffix; falls back to `GITHUB_AUTH_TOKEN` env var for default catalog if no per-URL token is configured
- **GitLab**: supported with or without `.git` suffix; supports subgroups (`group/subgroup/repo.git`)
- **Self-hosted**: supported with `.git` suffix (e.g. `https://git.example.com/org/repo.git`)

### Repository size limits
- **GitHub**: pre-clone size check via GitHub API (fast, no bandwidth wasted)
- **GitLab**: pre-clone size check via GitLab projects API when a token is available (`read_api` scope required, but 403 error is ignored)
- **All other hosts**: clone is monitored in real time with a custom FS

### Other changes
- Context is propagated from the controller sync request through all HTTP calls and the git clone
- URL with `.git` suffix is handled correctly (stripped before constructing clone URL and API calls)
- Tooltip in the UI explains supported URL formats
- Enable editing in the UI. Most code for it existed already

## Testing guide
- [x] Add a public GitHub repo as a source URL — verify sync works
- [x] Add a private GitHub repo with a PAT — verify sync works and token is not visible in the UI after saving
- [x] Add a private GitLab repo (`https://gitlab.com/org/repo.git`) with a PAT — verify sync works
- [x] Add a GitLab repo with subgroups (`https://gitlab.com/group/subgroup/repo.git`) — verify sync works
- [x] Add a URL with `.git` suffix — verify it works the same as without
- [x] Remove a source URL that has a stored token — verify the credential is cleaned up
- [x] Verify `GITHUB_AUTH_TOKEN` env var still works as a fallback for GitHub default catalog when no per-URL token is set
- [x] Attempt to clone large repository that isn't on Github or Gitlab (`https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git`)


pkg/controller/handlers/mcpcatalog/git_fs.go
```diff
@@ -0,0 +1,55 @@
...
+type sizeLimitedFile struct {
+	billy.File
+	fs *sizeLimitedFS
+}
+
+func (f *sizeLimitedFile) Write(p []byte) (int, error) {
+	if f.fs.written.Add(int64(len(p))) > f.fs.maxBytes {
+		return 0, errRepoTooLarge
+	}
+	return f.File.Write(p)
```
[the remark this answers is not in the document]
calvinmclean: The current implementation will prevent surpassing the limit by checking before doing the write

go.mod
```diff
@@ -160,7 +160,7 @@ require (
 	github.com/glebarez/go-sqlite v1.22.0 // indirect
 	github.com/glebarez/sqlite v1.11.0 // indirect
 	github.com/go-git/gcfg v1.5.1-0.20230307220236-3a3c6141e376 // indirect
-	github.com/go-git/go-billy/v5 v5.8.0 // indirect
+	github.com/go-git/go-billy/v5 v5.8.0
```
thedadams, line 163: Move this to the direct dependency block.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -134,13 +141,111 @@ func (h *MCPCatalogHandler) Update(req api.Context) error {
 		}
 	}
 
+	credCtx := mcpcataloghandler.CatalogCredentialContext(string(catalog.UID))
```
thedadams, line 144: I know there are other places where we've used the UID as part of the credential context, but we should do that anymore. There are issues with doing that around migration and cleanup. Using the name of the catalog should be enough.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -134,13 +141,111 @@ func (h *MCPCatalogHandler) Update(req api.Context) error {
...
+	for u := range oldURLsSet {
+		if _, active := activeURLs[u]; active {
+			continue
+		}
+		if _, transferred := transferredOldURLs[u]; transferred {
+			continue
+		}
+		err := req.GPTClient.DeleteCredential(req.Context(), credCtx, mcpcataloghandler.CatalogCredentialToolName(u))
+		var notFound gptscript.ErrNotFound
+		if err != nil && !errors.As(err, &notFound) {
```
thedadams, line 211: nit: you don't need the variable.
```suggestion
		if err != nil && !errors.As(err, & gptscript.ErrNotFound{}) {
```

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -1443,12 +1548,37 @@ func (h *MCPCatalogHandler) ListCategoriesForCatalog(req api.Context) error {
 	return req.Write(categories)
 }
 
-func convertMCPCatalog(catalog v1.MCPCatalog) types.MCPCatalog {
+func convertMCPCatalog(ctx context.Context, gptClient *gptscript.GPTScript, catalog v1.MCPCatalog) types.MCPCatalog {
```
thedadams, line 1551: I have received pushback previously in including clients in the "convert" functions. I was told that these should just take everything they need and shouldn't error unless absolutely necessary.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -134,13 +141,111 @@ func (h *MCPCatalogHandler) Update(req api.Context) error {
 		}
 	}
 
+	credCtx := mcpcataloghandler.CatalogCredentialContext(string(catalog.UID))
+
+	activeURLs := make(map[string]struct{}, len(manifest.SourceURLs))
```
thedadams, line 146: The logic from here down all seems very complicated. I believe this can be simplified if we have one credential with multiple keys in the `Env` map (one per URL). The determining old url, new url is simpler because you are only dealing with one credential.

pkg/controller/handlers/mcpcatalog/git_fs.go
```diff
@@ -0,0 +1,55 @@
...
+
+// sizeLimitedFile wraps billy.File and intercepts Write calls to track cumulative
+// bytes written. Returns errRepoTooLarge once the parent fs's limit is exceeded.
+type sizeLimitedFile struct {
+	billy.File
+	fs *sizeLimitedFS
+}
+
+func (f *sizeLimitedFile) Write(p []byte) (int, error) {
+	if f.fs.written.Add(int64(len(p))) > f.fs.maxBytes {
```
thedadams, line 51: This logic makes it impossible to recover from "too large" errors. Not sure if it is desirable to be able to recover and continue to write if a previous write would have been too large,

pkg/controller/handlers/mcpcatalog/github.go
```diff
@@ -11,86 +13,154 @@ import (
...
+	// (e.g. /org/repo.git or /org/repo.git/branch).
+	p := strings.TrimSuffix(u.Path, "/")
+	return strings.HasSuffix(p, ".git") || strings.Contains(p, ".git/")
 }
 
-// checkRepoSize checks the repository size using GitHub API before cloning
-func checkRepoSize(org, repo string, maxSizeMB int) error {
+// checkGitHubRepoSize checks repo size via the GitHub API before cloning.
+// Falls back to the GITHUB_AUTH_TOKEN env var if no per-URL token is provided.
+func checkGitHubRepoSize(ctx context.Context, org, repo string, maxSizeMB int, token string) error {
```
thedadams, line 65: There are a lot of changes in this function that amount to no functionality change. Best practice is to make exactly the changes you need for your fix or feature.

pkg/controller/handlers/mcpcatalog/github.go
```diff
@@ -123,84 +190,149 @@ func isPathSafe(path, baseDir string) error {
...
-			Username: "obot", // Use a dummy username. The username is ignored, but required to be non-empty.
-			Password: githubToken,
+			Username: "x-access-token", // Accepted as a dummy username by GitHub and GitLab.
+			Password: effectiveToken,
 		}
 	}
 
-	// Use go-git to clone the repository
-	_, err = git.PlainClone(tempDir, false, cloneOptions)
+	limitedFS := &sizeLimitedFS{
```
thedadams, line 325: Do we need this limited FS if we are also checking the size of the repo via the APIs?

calvinmclean: It is still useful for repos outside of Github/Gitlab. Also, Gitlab requires extra permissions to use the API so this can be still be useful there. We could skip the FS check if the size was checked by API already

pkg/controller/handlers/mcpcatalog/git_fs.go
calvinmclean: I don't think it would be useful to try and recover/continue if a single file is too large because we would end up with partial content. I think git works by downloading compressed files, so there would likely be unexpected behavior if one is missing.

pkg/api/handlers/mcpcatalogs.go
calvinmclean: Yeah good idea. This looks much better now. 
I was also looking into changing the MCPCatalogManifest to not have a separate `SourceURLs` and `SourceURLCredentials` map, but I think it's a bit nicer to still have them separate since they are stored separately

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -134,13 +149,81 @@ func (h *MCPCatalogHandler) Update(req api.Context) error {
...
+		if tok, ok := existingTokens[u]; ok {
+			newTokens[u] = tok
+		}
+	}
+
 	catalog.Spec.SourceURLs = manifest.SourceURLs
 
 	if err := req.Update(&catalog); err != nil {
 		return fmt.Errorf("failed to update catalog: %w", err)
 	}
```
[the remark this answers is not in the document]
calvinmclean: The current order is more recoverable: if auth fails to save, user should receive an error and can re-enter the credential.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -134,13 +149,81 @@ func (h *MCPCatalogHandler) Update(req api.Context) error {
...
+
+	// Reveal the existing single credential that holds all source-URL tokens.
+	existingTokens := make(map[string]string)
+	existingCred, err := req.GPTClient.RevealCredential(req.Context(), []string{credCtx}, mcpcataloghandler.CatalogCredentialToolName)
+	if err != nil && !errors.As(err, &gptscript.ErrNotFound{}) {
+		return fmt.Errorf("failed to reveal catalog credentials: %w", err)
+	}
+	if err == nil {
+		existingTokens = existingCred.Env
+	}
```
thedadams: We don't write to `existingTokens`, so we can just use `existedCred.Env` directly.
```suggestion
	existingCred, err := req.GPTClient.RevealCredential(req.Context(), []string{credCtx}, mcpcataloghandler.CatalogCredentialToolName)
	if err != nil && !errors.As(err, &gptscript.ErrNotFound{}) {
		return fmt.Errorf("failed to reveal catalog credentials: %w", err)
	}
	
	// Just use existingCred.Env
```

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -33,6 +34,16 @@ import (
 
 var log = logger.Package()
 
+// CatalogCredentialContext returns the GPTScript credential context for a catalog.
+func CatalogCredentialContext(catalogName string) string {
+	return catalogName
+}
```
thedadams: Seems like this can be removed.

thedadams (approved):

Issue #6195, opened 2026-03-30
Improve Nanobot Integrated Models


calvinmclean:
## Description

We want more model providers available to use with the Nanobot Agent.

Related: #6041 #6042 

Adding support for [OpenResponses APIs](https://www.openresponses.org) will open up more functionality and allow us to easily add more model providers.

cjellick:
_Changed title to align with an item on our roadmap deck_

PR #6358, opened 2026-04-16
Add Bifrost provider dialect and enable model providers in UI

calvinmclean:
Addresses #6195

This is merging into a feature branch that already has updated handling for the llmProvider configuration in Nanobot. This re-enables model providers in the UI and updates documentation.

Depends on:
- https://github.com/obot-platform/enterprise-tools/pull/105
- https://github.com/nanobot-ai/nanobot/pull/284
- https://github.com/obot-platform/tools/pull/797

This PR will first be merged into the feature branch. Then, a PR will be opened to main. The others will be merged first, then this will be updated to use the new Nanobot version (for Go library and container image).

Also need to double-check `go.mod` changes before going to `main`


docs/docs/configuration/model-providers.md
```diff
@@ -72,51 +72,70 @@ Setting a default model here does not automatically grant users access to it. Us
...
 1. **Expose Ollama to the network** - By default, Ollama only binds to `127.0.0.1:11434`. Since Obot runs in a container, `localhost` addresses resolve to Obot's container, not your host. Set `OLLAMA_HOST=0.0.0.0` before starting Ollama, then use your host's IP address in the endpoint URL.
 
-2. **Use the OpenAI-compatible endpoint** - The endpoint must include the `/v1/` path:
+2. **Set the Ollama host**
    ```
-   http://<your-host-ip>:11434/v1/
+   http://<your-host-ip>:11434
    ```
-   Using `http://<host>:11434/` without `/v1/` will result in validation errors.
+   - If you are running Obot in Docker, you should use `http://host.docker.internal:11434`
```
thedadams: This doesn't work on Linux by default. Not sure if you think it's worth calling that out.

go.mod
```diff
@@ -5,28 +5,28 @@ go 1.26.2
 replace (
 	github.com/obot-platform/obot/apiclient => ./apiclient
 	github.com/obot-platform/obot/logger => ./logger
+	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc => go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.54.0
+	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp => go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.60.0
```
thedadams, line 9: These seem unrelated. Can they be removed? We really don't want replace directives unless we absolutely need them.

calvinmclean:
```
# k8s.io/apiserver/pkg/storage/storagebackend/factory
../../../go/pkg/mod/k8s.io/apiserver@v0.31.1/pkg/storage/storagebackend/factory/etcd3.go:327:39: undefined: otelgrpc.UnaryClientInterceptor
```

I get errors like this without it. My plan is to resolve in the PR that takes `feature/llm-providers` to `main` since it will be easier to rebase on `main` at that point and see how this relates to the recent go mod changes for vulnerabilities. I'm not sure if those changes are in this branch and the related Nanobot branch.

docs/docs/configuration/model-providers.md
calvinmclean: > If you are running Obot in Docker, you should use http://host.docker.internal:11434. For linux users, run Obot in Docker with this additional flag --add-host=host.docker.internal:host-gateway or use an alternative method of allowing the container to access the host network

Improved it

thedadams (approved):
Backend looks good with the understanding that the replace directive is addressed before this goes into main.

PR #6365, opened 2026-04-16
Add support for more LLM providers

calvinmclean:
Addresses #6195

This is merging into a feature branch that already has updated handling for the llmProvider configuration in Nanobot. This re-enables model providers in the UI and updates documentation.

The first two commits are already reviewed as previous PRs. All new changes are squashed into the 3rd commit which can be reviewed independently.

Depends on:
- https://github.com/obot-platform/enterprise-tools/pull/105
- https://github.com/nanobot-ai/nanobot/pull/284
- https://github.com/obot-platform/tools/pull/797

TODO:
 - [ ] Update Nanobot image and version after merging https://github.com/nanobot-ai/nanobot/pull/284


go.mod
```diff
@@ -5,20 +5,22 @@ go 1.26.2
 replace (
 	github.com/obot-platform/obot/apiclient => ./apiclient
 	github.com/obot-platform/obot/logger => ./logger
+
+	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc => go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.60.0
```
calvinmclean: Investigating this... seems to be caused by Bifrost import in Nanobot since that's the only new thing. Bifrost has lots of cloud library dependencies

go.mod
```diff
@@ -36,10 +36,10 @@ require (
 	github.com/liggitt/tabwriter v0.0.0-20181228230101-89fcab3d43de
 	github.com/moby/moby/api v1.52.0-alpha.1
 	github.com/moby/moby/client v0.1.0-alpha.0
-	github.com/nanobot-ai/nanobot v0.0.68
+	github.com/nanobot-ai/nanobot v0.0.68-0.20260417164855-9212c55c34ee
```
g-linville, line 39: if possible, I would rather that we do another release of Nanobot and keep this on a tag, instead of a commit hash

g-linville (approved):

go.mod
calvinmclean: Yes, I agree. Once everything is approved and ready to merge, I will start by merging the Nanobot PR and creating a release, then updating this 👍

thedadams: Don't worry about this. Pseudo-versions exist for a reason.

pkg/gateway/server/llmproxy.go
```diff
@@ -142,6 +142,30 @@ func (s *Server) dispatchLLMProxy(req api.Context) error {
...
+	var (
+		outputPolicies         []messagepolicy.ApplicablePolicy
+		conversationHistory    []messagepolicy.ConversationMessage
+		inputPolicyReplacement string
+	)
+	messagePolicyHelper := s.messagePolicyHelper
+	if shouldSkipMessagePolicyEnforcement(req.Request) {
+		messagePolicyHelper = nil
+	}
+	if messagePolicyHelper != nil && token.UserID != "" {
```
thedadams, line 156: nit:
```suggestion
	if s.messagePolicyHelper != nil && token.UserID != "" && !shouldSkipMessagePolicyEnforcement(req.Request) {
```

thedadams (approved):

go.mod
calvinmclean: I will need to update the nanobot image as well

pkg/gateway/server/llmproxy.go
calvinmclean: applied this but reverted it because `messagePolicyHelper` is also passed to the responseModifier so it's helpful to set it to `nil` here. This is the same behavior as the other proxy for anthropic/openai

Issue #6185, opened 2026-03-30
`npx` MCP server timeouts when building from source


pl4nty:
**Describe the bug**
When installing [`github:pl4nty/mcp-server-amazon#make_npx_installable`](https://github.com/pl4nty/mcp-server-amazon/tree/make_npx_installable) with `npx` on k8s, I saw delays of over a minute - likely because it's building from source. This caused repeated timeouts with an opaque error:

```
error               │ failed to send error response: context canceled
error               │ Command Amazon AU exited with error: signal: killed
```

After half a dozen retries with the same error, it eventually installed quickly enough:

```
<-(Amazon AU:stderr)│ [INFO] Loaded Amazon cookies from file
<-(Amazon AU:stderr)│ [INFO] Amazon MCP Server running on stdio
```

Is the timeout configurable anywhere? Or could the default be increased?

**To Reproduce**
1. Deploy `github:pl4nty/mcp-server-amazon#make_npx_installable` or another npx-based MCP server that takes a while to be ready
2. Monitor pod logs

**Expected behavior**
Server is eventually ready

thedadams:
@pl4nty Thabj for submitting this issue. The logs you are referencing don't indicate the actual issue. Are you seeing errors in the UI or it in your browser's dev tools? Those would be more indicative of the error.

pl4nty:
These logs are from the `mcp` container that's running `npx` in the k8s pod, I couldn't find any other logs unfortunately

thedadams:
Yep, makes sense. What are you seeing in the UI?

pl4nty:
The logs have some timestamps, and status was "Unhealthy"

<img width="1466" height="444" alt="Image" src="https://github.com/user-attachments/assets/9432a0fa-a14c-4814-92e3-63885446a2b6" />

thedadams:
Do you see any errors in the UI when you first try to connect to the MCP server?

pl4nty:
I didn't when deploying the server, just the Unhealthy status for a while. Here's the full log

```
2026-03-30T00:57:10.411107531Z [32mconfig│[0m {
2026-03-30T00:57:10.411158868Z [32mconfig│[0m "publish": {
2026-03-30T00:57:10.411164702Z [32mconfig│[0m "mcpServers": [
2026-03-30T00:57:10.411167910Z [32mconfig│[0m "Amazon AU"
2026-03-30T00:57:10.411170827Z [32mconfig│[0m ]
[...]

PR #6399, opened 2026-04-22
Add user-configurable MCP startup timeouts

calvinmclean:
## Summary
Addresses #6185

- Add a configurable MCP server startup timeout to catalog entry, MCP server, and system MCP server manifests
- Surface the timeout in the NPX, UVX, and containerized runtime configuration UI
- Apply the configured timeout in Docker and Kubernetes MCP backend startup/readiness paths
~~- Separate image-pull waiting from container startup waiting so slow image pulls do not consume the user-configured startup timeout~~
  - K8s will also do early-exits when it detects unrecoverable failures/states

## Testing

In order to test this, I created a simple npx MCP server with a configurable sleep before serving:
```
# sleep 30s before starting (default is 60)
npx github:calvinmclean/slow-npx-mcp 30
```

This screenshot shows how the configuration can be set. In this case, the server will have a slow startup which would have previously timed-out. Now it will still succeed due to the 90s timeout.
Screenshot 2026-04-24 at 14 13 36

### DIFF
--- a/apiclient/types/mcpserver.go
+++ b/apiclient/types/mcpserver.go
@@ -167,6 +167,8 @@ type MCPServerCatalogEntryManifest struct {
 
 	Env []MCPEnv `json:"env,omitempty"`
 
+	// StartupTimeout configures the timeout to start and connect to an MCP Server. When unset, it defaults to 60s.
+	// The maximum allowed value is 600s (10 minutes). Attempting to set a higher value will cause an error.
 	StartupTimeoutSeconds int `json:"startupTimeoutSeconds,omitempty"`
 }
 
@@ -411,14 +413,14 @@ func (e RuntimeValidationError) Error() string {
 func MapCatalogEntryToServer(catalogEntry MCPServerCatalogEntryManifest, userURL string, disableHostnameValidation bool) (MCPServerManifest, error) {
 	serverManifest := MCPServerManifest{
 		// Copy common fields
-		Metadata:         catalogEntry.Metadata,
-		Name:             catalogEntry.Name,
-		ShortDescription: catalogEntry.ShortDescription,
-		Description:      catalogEntry.Description,
-		Icon:             catalogEntry.Icon,
-		ToolPreview:      catalogEntry.ToolPreview,
-		Runtime:          catalogEntry.Runtime,
-		Env:              catalogEntry.Env,
+		Metadata:              catalogEntry.Metadata,
+		Name:                  catalogEntry.Name,
+		ShortDescription:      catalogEntry.ShortDescription,
+		Description:           catalogEntry.Description,
+		Icon:                  catalogEntry.Icon,
+		ToolPreview:           catalogEntry.ToolPreview,
+		Runtime:               catalogEntry.Runtime,
+		Env:                   catalogEntry.Env,
 		StartupTimeoutSeconds: catalogEntry.StartupTimeoutSeconds,
 	}
 
--- a/apiclient/types/mcpserver.go
+++ b/apiclient/types/mcpserver.go
@@ -166,6 +166,8 @@ type MCPServerCatalogEntryManifest struct {
 	MultiUserConfig *MultiUserConfig `json:"multiUserConfig,omitempty"`
 
 	Env []MCPEnv `json:"env,omitempty"`
+
+	StartupTimeoutSeconds int `json:"startupTimeoutSeconds,omitempty"`
 }
 
 // ToolOverride defines how a single component tool is exposed by the composite server
@@ -245,6 +247,7 @@ type MCPServerManifest struct {
 	Headers []MCPHeader `json:"headers,omitempty"`
 
 	IdleShutdownIntervalHours int `json:"idleShutdownIntervalHours,omitempty"`
+	StartupTimeoutSeconds     int `json:"startupTimeoutSeconds,omitempty"`
 }
 
 type MCPServer struct {
@@ -416,6 +419,7 @@ func MapCatalogEntryToServer(catalogEntry MCPServerCatalogEntryManifest, userURL
 		ToolPreview:      catalogEntry.ToolPreview,
 		Runtime:          catalogEntry.Runtime,
 		Env:              catalogEntry.Env,
+		StartupTimeoutSeconds: catalogEntry.StartupTimeoutSeconds,
 	}
 
 	// Handle runtime-specific mapping
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -2429,15 +2429,6 @@ func toolsForServer(ctx context.Context, mcpSessionManager *mcp.SessionManager, 
 	return mcp.ConvertTools(gTools, allowedTools, server.Spec.UnsupportedTools)
 }
 
-// contextWithDefaultTimeout adds a context timeout when one does not already exist
-func contextWithDefaultTimeout(ctx context.Context, timeout time.Duration) (context.Context, context.CancelFunc) {
-	if _, ok := ctx.Deadline(); ok {
-		return ctx, func() {}
-	}
-
-	return context.WithTimeout(ctx, timeout)
-}
-
 func (m *MCPHandler) removeMCPServer(ctx context.Context, mcpServer v1.MCPServer) error {
 	if err := m.mcpSessionManager.ShutdownServer(ctx, mcpServer.Name); err != nil {
 		return fmt.Errorf("failed to shutdown server: %w", err)
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -2413,10 +2413,7 @@ func (m *MCPHandler) revealCompositeServer(req api.Context, compositeServer v1.M
 }
 
 func toolsForServer(ctx context.Context, mcpSessionManager *mcp.SessionManager, server v1.MCPServer, serverConfig mcp.ServerConfig, allowedTools []string) ([]types.MCPServerTool, error) {
-	listCtx, cancel := contextWithDefaultTimeout(ctx, time.Minute)
-	defer cancel()
-
-	gTools, err := mcpSessionManager.ListTools(listCtx, serverConfig)
+	gTools, err := mcpSessionManager.ListTools(ctx, serverConfig)
 	if err != nil {
 		if errors.Is(err, context.DeadlineExceeded) {
 			return nil, nil
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -2413,7 +2413,10 @@ func (m *MCPHandler) revealCompositeServer(req api.Context, compositeServer v1.M
 }
 
 func toolsForServer(ctx context.Context, mcpSessionManager *mcp.SessionManager, server v1.MCPServer, serverConfig mcp.ServerConfig, allowedTools []string) ([]types.MCPServerTool, error) {
-	gTools, err := mcpSessionManager.ListTools(ctx, serverConfig)
+	listCtx, cancel := contextWithDefaultTimeout(ctx, time.Minute)
+	defer cancel()
+
+	gTools, err := mcpSessionManager.ListTools(listCtx, serverConfig)
 	if err != nil {
 		if errors.Is(err, context.DeadlineExceeded) {
 			return nil, nil
@@ -2427,6 +2430,15 @@ func toolsForServer(ctx context.Context, mcpSessionManager *mcp.SessionManager, 
 	}
 
 	return mcp.ConvertTools(gTools, allowedTools, server.Spec.UnsupportedTools)
+}
+
+// contextWithDefaultTimeout adds a context timeout when one does not already exist
+func contextWithDefaultTimeout(ctx context.Context, timeout time.Duration) (context.Context, context.CancelFunc) {
+	if _, ok := ctx.Deadline(); ok {
+		return ctx, func() {}
+	}
+
+	return context.WithTimeout(ctx, timeout)
 }
 
 func (m *MCPHandler) removeMCPServer(ctx context.Context, mcpServer v1.MCPServer) error {
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -2413,9 +2413,6 @@ func (m *MCPHandler) revealCompositeServer(req api.Context, compositeServer v1.M
 }
 
 func toolsForServer(ctx context.Context, mcpSessionManager *mcp.SessionManager, server v1.MCPServer, serverConfig mcp.ServerConfig, allowedTools []string) ([]types.MCPServerTool, error) {
-	ctx, cancel := context.WithTimeout(ctx, time.Minute)
-	defer cancel()
-
 	gTools, err := mcpSessionManager.ListTools(ctx, serverConfig)
 	if err != nil {
 		if errors.Is(err, context.DeadlineExceeded) {
--- a/pkg/mcp/backend.go
+++ b/pkg/mcp/backend.go
@@ -18,7 +18,6 @@ const (
 	defaultContainerPort          = 8099
 	defaultWebhookToolName        = "fire-webhook"
 	serviceUnavailableGracePeriod = 10 * time.Second
-	imagePullTimeout              = 3 * time.Minute
 )
 
 type backend interface {
--- a/pkg/mcp/backend.go
+++ b/pkg/mcp/backend.go
@@ -52,13 +52,6 @@ var (
 	ErrInsufficientCapacity   = errors.New("insufficient cluster capacity to deploy MCP server")
 )
 
-func startupTimeout(server ServerConfig) time.Duration {
-	if server.StartupTimeoutSeconds > 0 {
-		return time.Duration(server.StartupTimeoutSeconds) * time.Second
-	}
-	return time.Minute
-}
-
 func ensureServerReady(ctx context.Context, url string, server ServerConfig) error {
 	// Ensure we can actually hit the service URL.
 	client := &http.Client{
--- a/pkg/mcp/backend.go
+++ b/pkg/mcp/backend.go
@@ -18,6 +18,7 @@ const (
 	defaultContainerPort          = 8099
 	defaultWebhookToolName        = "fire-webhook"
 	serviceUnavailableGracePeriod = 10 * time.Second
+	imagePullTimeout              = 3 * time.Minute
 )
 
 type backend interface {
@@ -51,10 +52,15 @@ var (
 	ErrInsufficientCapacity   = errors.New("insufficient cluster capacity to deploy MCP server")
 )
 
+func startupTimeout(server ServerConfig) time.Duration {
+	if server.StartupTimeoutSeconds > 0 {
+		return time.Duration(server.StartupTimeoutSeconds) * time.Second
+	}
+	return time.Minute
+}
+
 func ensureServerReady(ctx context.Context, url string, server ServerConfig) error {
 	// Ensure we can actually hit the service URL.
-	ctx, cancel := context.WithTimeout(ctx, time.Minute)
-	defer cancel()
 	client := &http.Client{
 		Timeout: time.Second,
 	}
--- a/pkg/mcp/backend.go
+++ b/pkg/mcp/backend.go
@@ -124,7 +124,7 @@ func ensureServerReady(ctx context.Context, url string, server ServerConfig) err
 		case <-time.After(100 * time.Millisecond):
 		}
 
-		req, err := http.NewRequest(http.MethodPost, url, strings.NewReader(streamableHTTPHealthcheckBody))
+		req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, strings.NewReader(streamableHTTPHealthcheckBody))
 		if err != nil {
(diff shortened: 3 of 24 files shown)


pkg/api/handlers/mcp.go
```diff
@@ -2330,9 +2330,6 @@ func (m *MCPHandler) revealCompositeServer(req api.Context, compositeServer v1.M
 }
 
 func toolsForServer(ctx context.Context, mcpSessionManager *mcp.SessionManager, server v1.MCPServer, serverConfig mcp.ServerConfig, allowedTools []string) ([]types.MCPServerTool, error) {
-	ctx, cancel := context.WithTimeout(ctx, time.Minute)
```
calvinmclean, line 2416: timeout is removed because this calls `mcpSessionManager.ListTools` -> `sm.clientForMCPServer` -> `sm.clientForMCPServerWithClientScope` -> `sm.clientForServerWithScope` -> `sm.clientForServerWithOptions` -> `sm.ensureDeployment` -> `sm.backend.ensureServerDeployment` which is finally the Docker/K8s implementation which handles timeouts

pkg/mcp/backend.go
```diff
@@ -51,10 +52,15 @@ var (
...
+func startupTimeout(server ServerConfig) time.Duration {
+	if server.StartupTimeoutSeconds > 0 {
+		return time.Duration(server.StartupTimeoutSeconds) * time.Second
+	}
+	return time.Minute
+}
+
 func ensureServerReady(ctx context.Context, url string, server ServerConfig) error {
 	// Ensure we can actually hit the service URL.
-	ctx, cancel := context.WithTimeout(ctx, time.Minute)
```
calvinmclean, line 56: timeout removed because this is called by Docker/K8s implementation with a context timeout already

pkg/mcp/docker.go
```diff
@@ -1039,6 +1050,8 @@ func (d *dockerBackend) waitForContainer(ctx context.Context, containerID string
 		select {
 		case <-timeout:
 			return fmt.Errorf("timeout waiting for container to start")
+		case <-ctx.Done():
+			return fmt.Errorf("%w: timeout waiting for container to start", ErrHealthCheckTimeout)
```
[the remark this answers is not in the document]
calvinmclean: This is fine because the image is already pulled and the container should start quickly. This is separate from the time it takes for the container to be _ready_, which is still handled by the correct timeout.

pkg/mcp/kubernetes.go
```diff
@@ -738,7 +746,7 @@ func (k *kubernetesBackend) k8sObjects(ctx context.Context, server ServerConfig,
 			},
 		},
 		Spec: appsv1.DeploymentSpec{
-			ProgressDeadlineSeconds: new(int32(60)),
+			ProgressDeadlineSeconds: new(int32(startupTimeout(server).Seconds())),
 			Selector: &metav1.LabelSelector{
```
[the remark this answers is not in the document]
calvinmclean: Go 1.26 :)

pkg/api/handlers/mcp.go
```diff
@@ -2331,10 +2331,10 @@ func (m *MCPHandler) revealCompositeServer(req api.Context, compositeServer v1.M
...
+	listCtx, cancel := contextWithDefaultTimeout(ctx, time.Minute)
```
thedadams: Correct me if I'm wrong here, but:
1. The context doesn't have a timeout, so we add one.
2. This calls `mcpSessionManager.ListTools`, which may try to add a longer deadline?
3. The 1 minute deadline expires, which negates the longer timeout.

pkg/mcp/backend.go
```diff
@@ -18,6 +18,7 @@ const (
 	defaultContainerPort          = 8099
 	defaultWebhookToolName        = "fire-webhook"
 	serviceUnavailableGracePeriod = 10 * time.Second
+	imagePullTimeout              = 3 * time.Minute
```
thedadams: I don't think we need an image pull timeout. That's part of the startup.

pkg/mcp/backend.go
```diff
@@ -51,10 +52,15 @@ var (
 	ErrInsufficientCapacity   = errors.New("insufficient cluster capacity to deploy MCP server")
 )
 
+func startupTimeout(server ServerConfig) time.Duration {
+	if server.StartupTimeoutSeconds > 0 {
+		return time.Duration(server.StartupTimeoutSeconds) * time.Second
+	}
+	return time.Minute
+}
```
thedadams: I would prefer this defaulting to happen when we set `server.StartupTimeoutSeconds` so it's done and we just always use the value later.

pkg/mcp/docker.go
```diff
@@ -320,7 +323,7 @@ func (d *dockerBackend) ensureDeployment(ctx context.Context, server ServerConfi
 				containerPort = server.ContainerPort
 			}
 
-			if err = d.ensureServerReady(ctx, existing, server, containerPort); err != nil {
+			if err = d.ensureServerReady(readyCtx, existing, server, containerPort); err != nil {
```
thedadams: I believe the intention is that the timeout should only be implemented here. The setup we do (creating volumes, syncing files, etc) isn't part of the "startup."

pkg/mcp/docker.go
```diff
@@ -776,6 +780,10 @@ func (d *dockerBackend) createAndStartAndWaitForContainer(ctx context.Context, s
 		return ServerConfig{}, err
 	}
 
+	// Use MCP Server startup timeout now that image is pulled and container is created
+	ctx, cancel := context.WithTimeout(ctx, startupTimeout(server))
+	defer cancel()
```
thedadams: Similar comment here.

pkg/mcp/docker.go
```diff
@@ -1016,6 +1024,9 @@ func (d *dockerBackend) createAndStartContainer(ctx context.Context, server Serv
 		} else {
 			containerID = resp.ID
 		}
+		if containerID != "" {
```
thedadams, line 1022: Do we want to do this? I would guess that we would have gotten an error above if we don't have a container ID here.

pkg/mcp/types.go
```diff
@@ -338,6 +340,7 @@ func ServerToServerConfig(mcpServer v1.MCPServer, audiences []string, issuer, us
 		AuthorizeEndpoint:         fmt.Sprintf("%s/oauth/authorize", issuer),
 		ComponentMCPServer:        mcpServer.Spec.CompositeName != "",
 		NanobotAgentName:          mcpServer.Spec.NanobotAgentID,
+		StartupTimeoutSeconds:     mcpServer.Spec.Manifest.StartupTimeoutSeconds,
```
thedadams: As as said above, set the default here and below if the server doesn't have this set. Then we know that is the value we should always use.

pkg/mcp/kubernetes.go
```diff
@@ -905,181 +913,235 @@ func getNewestPod(pods []corev1.Pod) (*corev1.Pod, error) {
...
-		newestPod, err := getNewestPod(pods.Items)
-		if err != nil {
-			olog.Debugf("failed to get newest pod: id=%s error=%v attempt=%d", id, err, attempt+1)
-			lastErr = err
-			if attempt < maxRetries {
-				continue
-			}
-			return "", fmt.Errorf("%w: %v", ErrHealthCheckTimeout, lastErr)
+	// Deployment is ready. Get the pod name that is currently running.
+	var (
```
thedadams, line 1056: We're already done a lot of this in the watch function. Can we not repeat it here?

pkg/mcp/kubernetes.go
```diff
@@ -905,181 +913,235 @@ func getNewestPod(pods []corev1.Pod) (*corev1.Pod, error) {
...
-					"app": id,
-				}),
-			}); err != nil {
-				return "", fmt.Errorf("failed to list MCP pods: %w", err)
+	// Track separate deadlines for image pulling and container startup.
+	// Image pulling gets a fixed generous timeout; the configured timeout applies to container startup.
+	timeout := startupTimeout(server)
+	imagePullDeadline := time.Now().Add(imagePullTimeout)
+	watchDeadline := time.Now().Add(imagePullTimeout + timeout + 30*time.Second)
+	var startupDeadline time.Time
```
thedadams: I would advocate that, instead of trying to track each part independently, we set a timeout on a single context and when it's done, it's done.

pkg/mcp/backend.go
calvinmclean: I thought we had discussed a separate timeout so slow pulls for large images (like phat) won't impact the startup timeout that the user configured for their server's. For example, if I know that my `npx` MCP server takes 60s to be ready, I might set a 70s timeout which could still fail due to unexpected slow scheduling or slow image pulling

pkg/api/handlers/mcp.go
calvinmclean: Hmm I must have mixed something up here when working through copilot comments. The intention was to keep the 1m timeout when called from outside this process but now I see that this function is a parent of the whole ensureDeployment process which I even noted in [this old comment](https://github.com/obot-platform/obot/pull/6399#discussion_r3127200921) 🤦

pkg/mcp/kubernetes.go
calvinmclean: The `wait.For` watch function just checks the pod status if the deployment is not ready. When the deployment is ready, the `wait.For` function exits without getting any pod details so we have to gather it here

pkg/mcp/docker.go
calvinmclean: I added this because this `for range 3` loop would always run 3 times even if it successfully sets the containerID

pkg/mcp/backend.go
calvinmclean: Fixed by changing `ServerConfig.StartupTimeoutSeconds` to `ServerConfig.StartupTimeout` (as duration) and updated constructor to set default 1m if missing.

pkg/api/handlers/mcp.go
calvinmclean: Added [this commit](https://github.com/obot-platform/obot/pull/6399/commits/c92540269baa4da060c15f1de8526d8550ed57b7), which removes this timeout again and adds a 1m timeout in a few places downstream of this function that are outside of the ensureDeployment path so we keep a reasonable timeout on those paths.

apiclient/types/mcpserver.go
```diff
@@ -151,6 +151,8 @@ type MCPServerCatalogEntryManifest struct {
 	CompositeConfig     *CompositeCatalogConfig     `json:"compositeConfig,omitempty"`
 
 	Env []MCPEnv `json:"env,omitempty"`
+
+	StartupTimeoutSeconds int `json:"startupTimeoutSeconds,omitempty"`
```
njhale, line 172: Can you document what this does, the behavior at zero, the default, and max with a go doc comment here? I know it's not consistent across the codebase right now, but I find them useful and I think it will be helpful down the line.

pkg/validation/mcpvalidators.go
```diff
@@ -923,3 +935,15 @@ func ValidateSystemMCPServerManifest(manifest types.SystemMCPServerManifest) err
 		Message: "unsupported runtime",
 	}
 }
+
+func validateStartupTimeout(runtime types.Runtime, startupTimeoutSeconds int) error {
+	if startupTimeoutSeconds < 0 {
```
njhale, line 1056: Do we specify a maximum anywhere? If we don't we should probably prevent large values (> 10 minutes).

For real feature completeness, we could also make the max configurable at the Obot level, set a reasonable default, and let users set it to a negative value to indicate no max.

pkg/mcp/kubernetes.go
```diff
@@ -976,110 +981,127 @@ func analyzePodStatus(pod *corev1.Pod) (bool, error) {
...
+				if err != nil {
+					return false, nil // Keep waiting
 				}
 
-				return podName, nil
-			}
+				shouldRetry, podErr := analyzePodStatus(newestPod)
+				if !shouldRetry {
+					// Permanent failure - return the error with the appropriate type already wrapped
+					olog.Debugf("pod in non-retryable state: id=%s error=%v attempt=%d", id, podErr, attempt+1)
```
njhale, line 1023: Are we leaving this in? I'm fine with it if so.

calvinmclean: I'm fine either way. This existed previously (hard to see in the diff) so I opted to leave as-is

pkg/validation/mcpvalidators.go
calvinmclean: Good idea! I added the 10m max. I looked briefly at making it configurable, but didn't see a very straightforward way to access this configuration from the validators without setting a global var. I am also not sure if it's worth adding another configuration for something unlikely to be used. I can dig into it a bit more if you think it's worthwhile though :+1:

pkg/mcp/loader.go
```diff
@@ -353,6 +392,9 @@ func (sm *SessionManager) ensureDeployment(ctx context.Context, server ServerCon
 		}
 	}
```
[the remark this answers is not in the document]
calvinmclean: The timeout will not be zero because it is enforced when creating a ServerConfig

calvinmclean:
I hid the Entelligence PR comment because I accidentally pushed an extra commit from a different branch. After I fixed it, Entelligence updated the previous comment instead of adding a new one, but I can't see how to un-hide it. I'm not hiding it just because it asked for changes!

pkg/api/handlers/mcp.go
```diff
@@ -2350,6 +2347,15 @@ func toolsForServer(ctx context.Context, mcpSessionManager *mcp.SessionManager,
...
 }
 
+// contextWithDefaultTimeout adds a context timeout when one does not already exist
+func contextWithDefaultTimeout(ctx context.Context, timeout time.Duration) (context.Context, context.CancelFunc) {
+	if _, ok := ctx.Deadline(); ok {
+		return ctx, func() {}
+	}
+
+	return context.WithTimeout(ctx, timeout)
+}
```
g-linville: The behavior of this seems off to me, since we are returning a no-op cancel function if there already is a timeout on the context. Correct me if I'm wrong, but I think this could easily lead to bugs where we are trying to cancel contexts but are just calling these no-op functions instead.

pkg/validation/mcpvalidators.go
njhale: No, I think it's reasonable to kick the can on that until someone actually runs into the issue.

Also, if you need a startup timeout > 10 minutes, then you have bigger problems.

njhale (approved):

g-linville (approved):

pkg/api/handlers/mcp.go
calvinmclean: The original goal here was to prevent child contexts from hijacking control of timeouts, so the no-op cancels were intentional, but could definitely cause unexpected behavior.

Due to changes from previous comments, this is now unused and I didn't notice since it was used by unit tests. I removed it now.

pkg/mcp/client.go
```diff
@@ -63,6 +63,9 @@ func (sm *SessionManager) clientForServerWithOptions(ctx context.Context, client
 		return nil, err
 	}
 
+	ctx, cancel := context.WithTimeout(ctx, time.Minute)
```
thedadams, line 58: It's possible that loading a "session" (i.e. creating a client), causes the server to be deployed. Adding a 1 minute timeout here would make the user-defined timeout irrelevant if it is longer.

pkg/mcp/kubernetes.go
```diff
@@ -976,110 +981,127 @@ func analyzePodStatus(pod *corev1.Pod) (bool, error) {
...
+				if dep.Generation == dep.Status.ObservedGeneration && dep.Status.UpdatedReplicas == 1 && dep.Status.ReadyReplicas == 1 && dep.Status.AvailableReplicas == 1 {
+					return true, nil
+				}
 
-			var newestCreatedTime metav1.Time
-			for _, p := range pods.Items {
-				if p.DeletionTimestamp.IsZero() && p.CreationTimestamp.After(newestCreatedTime.Time) && p.Status.Phase == corev1.PodRunning {
-					podName = p.Name
-					newestCreatedTime = p.CreationTimestamp
+				select {
```
thedadams: Is this needed? I don't think this function would be called if the context is canceled and the watch is terminated.

pkg/mcp/tools.go
```diff
@@ -18,6 +19,9 @@ func (sm *SessionManager) ListTools(ctx context.Context, serverConfig ServerConf
 		return nil, err
 	}
 
+	ctx, cancel := context.WithTimeout(ctx, time.Minute)
```
thedadams, line 22: This timeout is OK because getting the client above would have already launched the server, if needed.

pkg/mcp/types.go
```diff
@@ -318,6 +328,14 @@ func ServerToServerConfig(mcpServer v1.MCPServer, audiences []string, issuer, us
 		powerUserWorkspaceID = mcpCatalogName
 	}
 
+	startupTimeout := time.Duration(mcpServer.Spec.Manifest.StartupTimeoutSeconds) * time.Second
+	if startupTimeout == 0 {
+		startupTimeout = defaultStartupTimeout
+	}
+	if startupTimeout > MaxMCPServerStartupTimeout {
```
thedadams: pico-nit:
```suggestion
	} else if startupTimeout > MaxMCPServerStartupTimeout {
```

pkg/mcp/types.go
```diff
@@ -431,6 +450,14 @@ func SystemServerToServerConfig(systemServer v1.SystemMCPServer, audiences []str
 		displayName = systemServer.Name
 	}
 
+	startupTimeout := time.Duration(systemServer.Spec.Manifest.StartupTimeoutSeconds) * time.Second
+	if startupTimeout == 0 {
+		startupTimeout = defaultStartupTimeout
+	}
+	if startupTimeout > MaxMCPServerStartupTimeout {
```
thedadams: Same.

pkg/wait/waitfor.go
```diff
@@ -108,25 +108,34 @@ func For[T kclient.Object](ctx context.Context, c kclient.WithWatch, obj T, cond
...
-			}, obj.GetName())
-		}
-		switch event.Type {
-		case watch.Added, watch.Modified:
-			if ok, err := condition(event.Object.(T)); err != nil {
-				return def, err
-			} else if ok {
-				return event.Object.(T), nil
+	for {
+		select {
```
thedadams: My understanding is that this change is not needed. The watch uses the context, so the channel will be closed when the context is canceled. Then we get to the removed error at the bottom of the function.

Please let me know if I am misunderstanding this change. Otherwise, let's not change this.

calvinmclean: I changed this because `TestUpdatedMCPPodName_ContainerStartupDeadlineExceeded` would run indefinitely without it. The `fake` client used in the test does not respect the context. Ultimately, the test is indirectly testing that `wait.For` is using the context so it probably doesn't hurt to remove it

calvinmclean: Ok I think I am using the fake Watch incorrectly... working on it now

pkg/mcp/types.go
calvinmclean: But what if we accidentally make `defaultStartupTimeout > MaxMCPServerStartupTimeout` 😱  just joking. Fixed it

pkg/mcp/kubernetes.go
calvinmclean: This is intended as a small optimization to return early before making another API call if the condition function is used right before the timeout. I added when I was not seeing timeouts from `wait.For` which is now fixed by improving the unit test instead of updating `wait.For`. Removed 👍

pkg/mcp/client.go
calvinmclean: Maybe I am missing something here. This is being called after `ensureDeployment` so this shouldn't be an issue as far as I can tell. The idea here was to preserve the original 1m timeout for things happening outside of the `ensureDeployment` so they wouldn't go unbounded.

However, `sm.loadSession` only uses the inbound context for `sm.tokenService.NewTokenWithClaims` so it's probably not very valuable to include the timeout anyways. Let me know what you think

pkg/mcp/kubernetes.go
```diff
@@ -976,110 +981,121 @@ func analyzePodStatus(pod *corev1.Pod) (bool, error) {
...
+					// Permanent failure - return the error with the appropriate type already wrapped
+					olog.Debugf("pod in non-retryable state: id=%s error=%v attempt=%d", id, podErr, attempt+1)
+					return false, podErr
+				}
 
-			lastErr = fmt.Errorf("no pods found")
-			continue
+				return false, nil // Keep waiting.
+			},
+			wait.Option{Timeout: server.StartupTimeout},
```
thedadams, line 1029: If the context has a deadline, then I think we can remove this.

thedadams (approved):

pkg/mcp/kubernetes.go
calvinmclean: I kept this because `wait.For` defaults to 2m if the timeout is not set

thedadams: OK. Disregard my comment, then.

Issue #6405, opened 2026-04-23
Per-component tool-name prefixes for composite MCP servers


njhale:
**Related: obot-platform/prospect-and-customer-issues#7**

### Problem

Composite MCP servers aggregate tools from multiple component servers behind one endpoint. When two components expose tools with the same name, nanobot silently drops the duplicates, and LLMs struggle with near-identical tool names across components. Today admins can only work around this tool-by-tool via `overrideName`, which doesn't scale and doesn't surface collisions at configuration time.

### Goal

Let admins set an optional **tool prefix** on each component of a composite server. The prefix is prepended to every tool that component exposes, so all of a component's tools share a namespace (e.g. `google_calendar_list_events`). Prefixes work standalone (prefix the whole component) or alongside existing per-tool overrides.

### Behavior

- **Prefix alone**: every tool the component exposes is prefixed. No allowlisting.
- **Prefix + overrides**: existing allowlist semantics are unchanged — only enabled, listed tools are exposed — and the prefix is prepended to the override name (or original name if no override name is set).

### UX — admins

When an admin adds a component to a composite, Obot pre-populates a default prefix derived from the component's display name (e.g. `Google Calendar` → `google_calendar_`). The admin can edit or clear it. The prefix is editable both on the component card and inside the tool-overrides modal.

Every tool row shows its **effective name** (prefix + tool/override name) live as the admin edits. Problems are flagged inline with warning/error icons:

- **Errors (block Save)**: invalid characters, duplicate prefix across components, effective name > 128 chars, effective name collides with another enabled tool anywhere in the composite.
- **Warnings (informational)**: effective name > 64 chars (some inference APIs cap there), effective name contains `.` or `/` (some MCP clients don't support them).

### UX — users

Users viewing a composite server's tool list see the same inline warning/error indicators next to each tool name, read-only. This gives them visibility into names that may be problematic in their client without needing to ask the admin.

### Validation rules

- Prefix character set: A-Z, a-z, 0-9, ., /, _, -. (see https://modelcontextprotocol.io/specification/2025-11-25/server/tools#tool-names)
- Non-empty prefixes must be unique across a composite's components. Empty prefixes may repeat.
- Effective tool names (for enabled tools only) must be <= 128 chars, match the same character set, and be unique across the entire composite.
- A component may set a prefix without declaring any tool overrides.

Backend rejects manifests that violate the above; frontend mirrors the checks to gate the Save button.

### Out of scope

[...]

PR #6406, opened 2026-04-23
enhance: add composite tool prefixes

njhale:
Composite MCP servers expose tools aggregated from multiple component servers behind one endpoint. When two components expose tools with the same name nanobot silently drops the duplicates, and LLMs struggle to disambiguate near-identical names across components.

This change adds an optional `toolPrefix` on each component of a composite server. Nanobot prepends the prefix to every tool that component exposes, so all of a component's tools share a namespace (e.g. `google_calendar_list_events`). Prefixes work standalone or
alongside the existing per-tool overrides; disabled tools are still never exposed. A change to a component's prefix is treated as drift and triggers MCPServer reconciliation.

The backend enforces the same character set across prefixes and effective tool names, cross-component prefix uniqueness, and unique/length-bounded effective names across the entire composite. The admin UI seeds a default prefix derived from each component's display name, renders the live effective name for every tool, and flags invalid characters, oversized names, or cross-component collisions inline with warning/error indicators that gate Save. Composite tool previews in the server detail views surface the same indicators read-only.

Addresses https://github.com/obot-platform/obot/issues/6405

### DIFF
--- a/apiclient/types/mcpserver.go
+++ b/apiclient/types/mcpserver.go
@@ -70,9 +70,11 @@ type CatalogComponentServer struct {
 	// MCPServerID if set, reference the multi-user MCP server the component server proxies to
 	MCPServerID string `json:"mcpServerID,omitempty"`
 	// Manifest is the catalog entry manifest of the component server
-	Manifest MCPServerCatalogEntryManifest `json:"manifest,omitempty"`
+	Manifest MCPServerCatalogEntryManifest `json:"manifest,omitzero"`
 	// ToolOverrides restrict the tools exposed by the component server
 	ToolOverrides []ToolOverride `json:"toolOverrides,omitempty"`
+	// ToolPrefix is an optional prefix applied to the final name of each tool exposed by the component server
+	ToolPrefix string `json:"toolPrefix,omitempty"`
 }
 
 // ComponentID returns the ID of the component server.
@@ -95,9 +97,11 @@ type ComponentServer struct {
 	// MCPServerID if set, reference the multi-user MCP server the component server proxies to
 	MCPServerID string `json:"mcpServerID,omitempty"`
 	// Manifest is the runtime manifest of the component server
-	Manifest MCPServerManifest `json:"manifest,omitempty"`
+	Manifest MCPServerManifest `json:"manifest,omitzero"`
 	// ToolOverrides restrict the tools exposed by the component server
 	ToolOverrides []ToolOverride `json:"toolOverrides,omitempty"`
+	// ToolPrefix is an optional prefix applied to the final name of each tool exposed by the component server
+	ToolPrefix string `json:"toolPrefix,omitempty"`
 	// Disabled indicates whether the component server should be included in the composite server at runtime
 	Disabled bool `json:"disabled,omitempty"`
 }
--- a/pkg/mcp/tools.go
+++ b/pkg/mcp/tools.go
@@ -1,6 +1,7 @@
 package mcp
 
 import (
+	"cmp"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -67,38 +68,32 @@ func ConvertTools(tools []mcp.Tool, allowedTools, unsupportedTools []string) ([]
 // ApplyToolOverrides applies ToolOverrides to a component's tool array,
 // filtering out disabled tools and applying name/description overrides.
 // If overrides are present, they act as an allowlist - only tools explicitly listed are included.
-func ApplyToolOverrides(tools []otypes.MCPServerTool, toolOverrides []otypes.ToolOverride) []otypes.MCPServerTool {
+// toolPrefix, if non-empty, is prepended to every returned tool's Name so previews
+// match what the composite server will expose via nanobot at runtime.
+func ApplyToolOverrides(tools []otypes.MCPServerTool, toolOverrides []otypes.ToolOverride, toolPrefix string) []otypes.MCPServerTool {
 	// Build lookup map: toolName -> ToolOverride
 	overrideMap := make(map[string]otypes.ToolOverride, len(toolOverrides))
 	for _, override := range toolOverrides {
 		overrideMap[override.Name] = override
 	}
 
-	hasOverrides := len(toolOverrides) > 0
-
-	transformedTools := make([]otypes.MCPServerTool, 0, len(tools))
+	var (
+		hasOverrides     = len(toolOverrides) > 0
+		transformedTools = make([]otypes.MCPServerTool, 0, len(tools))
+	)
 	for _, tool := range tools {
 		override, hasOverride := overrideMap[tool.Name]
-
-		// If overrides are defined, only include tools that are explicitly listed
-		if hasOverrides && !hasOverride {
+		if hasOverrides && (!hasOverride || !override.Enabled) {
+			// Omit the tool from the final tool set.
+			// Overrides have been set for the component and the tool either:
+			// - isn't present in the component's overrides (is likely net-new and wasn't available when the overrides were generated)
+			// - is explicitly disabled
 			continue
 		}
 
-		// If there's an override and the tool is explicitly disabled, skip it
-		if hasOverride && !override.Enabled {
-			continue
-		}
-
-		// Apply overrides
-		if hasOverride {
-			if override.OverrideName != "" {
-				tool.Name = override.OverrideName
-			}
-			if override.OverrideDescription != "" {
-				tool.Description = override.OverrideDescription
-			}
-		}
+		// Apply overrides and tool prefix if provided
+		tool.Name = toolPrefix + cmp.Or(override.OverrideName, tool.Name)
+		tool.Description = cmp.Or(override.OverrideDescription, tool.Description)
 
 		transformedTools = append(transformedTools, tool)
 	}
--- a/pkg/validation/mcpvalidators.go
+++ b/pkg/validation/mcpvalidators.go
@@ -1,7 +1,7 @@
 package validation
 
 import (
-	"errors"
+	"cmp"
 	"fmt"
 	"net/url"
 	"regexp"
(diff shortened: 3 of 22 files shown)


apiclient/types/mcpserver.go
```diff
@@ -70,9 +70,11 @@ type CatalogComponentServer struct {
 	// MCPServerID if set, reference the multi-user MCP server the component server proxies to
 	MCPServerID string `json:"mcpServerID,omitempty"`
 	// Manifest is the catalog entry manifest of the component server
-	Manifest MCPServerCatalogEntryManifest `json:"manifest,omitempty"`
+	Manifest MCPServerCatalogEntryManifest `json:"manifest"`
```
njhale: My linter was saying `omitempty` was ineffectual here, so I dropped it.

apiclient/types/mcpserver.go
```diff
@@ -95,9 +97,11 @@ type ComponentServer struct {
 	// MCPServerID if set, reference the multi-user MCP server the component server proxies to
 	MCPServerID string `json:"mcpServerID,omitempty"`
 	// Manifest is the runtime manifest of the component server
-	Manifest MCPServerManifest `json:"manifest,omitempty"`
+	Manifest MCPServerManifest `json:"manifest"`
```
njhale: My linter was saying omitempty was ineffectual here, so I dropped it.

docs/docs/configuration/server-configuration.md
```diff
@@ -44,8 +44,8 @@ The Obot server is configured via environment variables. The following configura
 | `OBOT_SERVER_AUDIT_LOGS_COMPRESS_FILE` | Controls whether or not to compress audit log files | `true` |
 | `OBOT_SERVER_AUDIT_LOGS_USE_PATH_STYLE` | Whether to use path style for S3 | - |
 | `OBOT_SERVER_MCPBASE_IMAGE` | Deploy MCP servers in the kubernetes cluster or using docker with this base image. | `ghcr.io/obot-platform/mcp-images/phat:v0.20.2` |
-| `OBOT_SERVER_MCPREMOTE_SHIM_BASE_IMAGE` | Deploy MCP remote shim servers in the cluster using this base image. | `ghcr.io/nanobot-ai/nanobot:v0.0.69` |
-| `OBOT_SERVER_NANOBOT_AGENT_IMAGE` | Deploy the Nanobot agent in the cluster using this image. | `ghcr.io/nanobot-ai/nanobot-agent:v0.0.69` |
+| `OBOT_SERVER_MCPREMOTE_SHIM_BASE_IMAGE` | Deploy MCP remote shim servers in the cluster using this base image. | `ghcr.io/nanobot-ai/nanobot:v0.0.71` |
```
njhale: Note: Don't merge until `v0.0.71` is released.

apiclient/types/mcpserver.go
thedadams: `omitzero` provides the desired result.

pkg/validation/mcpvalidators.go
```diff
@@ -559,74 +641,86 @@ func (v CompositeValidator) ValidateCatalogConfig(manifest types.MCPServerCatalo
...
-				Message: fmt.Sprintf("duplicate component server: %s", componentID),
+		// Validate tool overrides
+		for j, override := range component.ToolOverrides {
+			if override.Name == "" {
+				return types.RuntimeValidationError{
+					Runtime: types.RuntimeComposite,
+					Field:   fmt.Sprintf("compositeConfig.componentServers[%d].toolOverrides[%d].name", i, j),
+					Message: "original tool name is required",
+				}
 			}
```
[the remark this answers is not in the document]
njhale: they get rejected by the composite wide validation later

pkg/validation/mcpvalidators.go
```diff
@@ -478,15 +494,77 @@ func (v CompositeValidator) ValidateConfig(manifest types.MCPServerManifest) err
...
+
+			// Compute the effective tool name
+			effectiveToolName := prefix + cmp.Or(override.OverrideName, override.Name)
+
+			// Validate length
+			if len(effectiveToolName) > maxToolNameLength {
+				return types.RuntimeValidationError{
+					Runtime: types.RuntimeComposite,
+					Field:   fmt.Sprintf("compositeConfig.componentServers[%d].toolOverrides[%d]", i, j),
+					Message: fmt.Sprintf("effective tool name must be less than %d characters: %q", maxToolNameLength, effectiveToolName),
```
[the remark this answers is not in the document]
njhale: fixed

pkg/validation/mcpvalidators.go
```diff
@@ -559,74 +641,86 @@ func (v CompositeValidator) ValidateCatalogConfig(manifest types.MCPServerCatalo
...
-			return types.RuntimeValidationError{
-				Runtime: types.RuntimeComposite,
-				Field:   fmt.Sprintf("toolOverrides[%d].name", i),
-				Message: "original tool name is required",
+			// Validate length
+			if len(effectiveToolName) > maxToolNameLength {
+				return types.RuntimeValidationError{
+					Runtime: types.RuntimeComposite,
+					Field:   fmt.Sprintf("compositeConfig.componentServers[%d].toolOverrides[%d]", i, j),
+					Message: fmt.Sprintf("effective tool name must be less than %d characters: %q", maxToolNameLength, effectiveToolName),
```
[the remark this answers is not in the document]
njhale: Fixed

pkg/validation/mcpvalidators.go
```diff
@@ -478,15 +494,77 @@ func (v CompositeValidator) ValidateConfig(manifest types.MCPServerManifest) err
...
-				Message: "tool overrides invalid",
-			}, err)
+		for j, override := range component.ToolOverrides {
+			if override.Name == "" {
+				return types.RuntimeValidationError{
+					Runtime: types.RuntimeComposite,
+					Field:   fmt.Sprintf("compositeConfig.componentServers[%d].toolOverrides[%d].name", i, j),
+					Message: "original tool name is required",
+				}
+			}
```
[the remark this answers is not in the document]
njhale: see https://github.com/obot-platform/obot/pull/6406#discussion_r3132814556

calvinmclean (approved):

pkg/mcp/tools.go
```diff
@@ -67,38 +68,32 @@ func ConvertTools(tools []mcp.Tool, allowedTools, unsupportedTools []string) ([]
...
-		if hasOverride {
-			if override.OverrideName != "" {
-				tool.Name = override.OverrideName
-			}
-			if override.OverrideDescription != "" {
-				tool.Description = override.OverrideDescription
-			}
-		}
+		// Apply overrides and tool prefix if provided
+		tool.Name = toolPrefix + cmp.Or(override.OverrideName, tool.Name)
```
thedadams, line 95: Does nanobot apply the prefix regardless of whether or not an override exists?

pkg/validation/mcpvalidators.go
```diff
@@ -478,15 +494,77 @@ func (v CompositeValidator) ValidateConfig(manifest types.MCPServerManifest) err
...
+
+			// Prevent effective duplicates (across entire composite)
+			if _, ok := effectiveToolNames[effectiveToolName]; ok {
+				return types.RuntimeValidationError{
+					Runtime: types.RuntimeComposite,
+					Field:   fmt.Sprintf("compositeConfig.componentServers[%d].toolOverrides[%d]", i, j),
+					Message: fmt.Sprintf("duplicate tool name: %s", effectiveToolName),
+				}
+			}
```
thedadams, line 644: It seems like this error is impossible if the prefixes are not the same. Are we always setting a prefix?

pkg/mcp/tools.go
njhale: Yes.

thedadams: So the prefix can be removed here?

pkg/validation/mcpvalidators.go
njhale: > Are we always setting a prefix?

No, we're not. We support setting the prefix to an empty string to opt out.

> It seems like this error is impossible if the prefixes are not the same.

This method has been refactored so that this section handles catching collisions within a single component as well, so the prefixes can be the same.

Even when the prefixes are different, there are edge cases where the effective tool names -- `prefix + cmp.Or(override.OverrideName, override.Name)` -- can collide.

e.g.

| prefix | name/override | Effective Tool Name |
|------------|----------------|----------------------|
| a          | bc             | abc                  |
| ab         | c              | abc                  |

pkg/mcp/tools.go
njhale: No, because the tools passed into this function were generated from temporary servers that don't have a prefix set at the nanobot level.

thedadams (approved):

ui/user/src/lib/components/mcp/composite/CompositeEditTools.svelte
```diff
@@ -210,6 +289,7 @@
 		<button class="button" onclick={handleCancel}>Cancel</button>
 		<button
 			class="button-primary"
+			disabled={hasBlockingToolNameErrors || prefixInvalid}
 			onclick={() => {
```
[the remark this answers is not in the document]
njhale: Done

ui/user/src/lib/components/mcp/composite/CompositeEditTools.svelte
```diff
@@ -11,9 +19,41 @@
...
-	let { configuringEntry, tools = [], onClose, onCancel, onSuccess }: Props = $props();
+	let {
+		configuringEntry,
+		tools = [],
+		toolPrefix = $bindable(),
+		otherEffectiveNames,
+		onClose,
+		onCancel,
+		onSuccess
+	}: Props = $props();
```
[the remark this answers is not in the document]
njhale: Fixed

ui/user/src/lib/components/mcp/McpServerTools.svelte
```diff
@@ -65,6 +78,13 @@
 		)
 	);
 
+	// Detect duplicate effective names across the aggregated tool list (composite
+	// previews/live lists only). Disabled via showToolNameIssues=false for other
+	// contexts so the icons stay opt-in.
+	let toolNameDuplicates = $derived(
+		showToolNameIssues ? duplicateToolNames(displayTools.map((t) => t.name)) : new Set<string>()
+	);
```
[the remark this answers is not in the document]
njhale: this doesn't make sense. you can't construct a search query that will deduplicate the list

Issue #6400, opened 2026-04-22
refreshMCPServerConfig fails for all MCP servers due to read-only filesystem


shwetapsdet:
**Describe the bug**
The `refreshMCPServerConfig` tool consistently fails across all MCP servers, not just AntV Charts, when triggered via chat after launching the agent.

The `refreshMCPServerConfig` tool fails with the following error:
```
-32001: JSON RPC unknown error: failed to unmarshal response: 
error from server: -32603: JSON RPC internal error: 
error invoking tool refreshMCPServerConfig: 
create mcp-cli config directory: mkdir /files/mcp-cli: read-only file system
```

**To Reproduce**

1. Launch the agent
2. Connect to any MCP server (e.g., AntV Charts, GitLab, Redis, etc.)
3. Run any prompt that triggers MCP config refresh (e.g.): `list all tools for <mcp server>`
5. Observe tool execution failure

**Expected behavior**

- Tool list should be returned without errors
- No filesystem permission issues should occur

**Screenshots**

https://github.com/user-attachments/assets/be17439e-8d49-4f7e-98d9-8e62cbcc15cd

<img width="976" height="397" alt="Image" src="https://github.com/user-attachments/assets/e1d26535-2457-4124-83e7-f281f98ceb1f" />

sangee2004:
This error is seen for any `mcp-cli` command

<img width="1038" height="699" alt="Image" src="https://github.com/user-attachments/assets/143bbd50-8038-418b-900f-59a0dca85106" />

shwetapsdet:
Verified it on obot main and it's working as expected 

<img width="1255" height="711" alt="Image" src="https://github.com/user-attachments/assets/092dac8f-a6cf-4784-a0b7-2a6be8c55b6e" />

PR #6415, opened 2026-04-23
fix: specify .nanobot as the default config location

thedadams:
The model provider configuration was specified and nanobot would try to write other configuration to its directory. The problem is that that directory is read-only because it is backed by a secret.

This change incorporates the nanobot change that allows passing the --config flag multiple times, and .nanobot will be used for this special configuration location.

Issue: https://github.com/obot-platform/obot/issues/6400


calvinmclean (approved):
Nice 👍 thanks for the fix!

njhale (approved):

PR #6421, opened 2026-04-24
feat: add storage service accounts

g-linville:
This is the first change to support network policy providers. On its own, this doesn't do anything, as there is nothing that consumes the Kubernetes secret we are creating. The rest of the changes will come in a second PR.

### DIFF
--- a/chart/templates/mcp.yaml
+++ b/chart/templates/mcp.yaml
@@ -72,4 +72,38 @@ roleRef:
   name: obot-mcp-role
   apiGroup: rbac.authorization.k8s.io
 
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: {{ .Release.Namespace }}
+  name: obot-runtime-secret-role
+  labels:
+    {{- include "obot.labels" . | nindent 4 }}
+rules:
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["create"]
+  - apiGroups: [""]
+    resources: ["secrets"]
+    resourceNames: ["obot-network-policy-provider"]
+    verbs: ["get", "update", "patch", "delete"]
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  namespace: {{ .Release.Namespace }}
+  name: obot-runtime-secret-rolebinding
+  labels:
+    {{- include "obot.labels" . | nindent 4 }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ include "obot.serviceAccountName" . }}
+    namespace: {{ .Release.Namespace }}
+roleRef:
+  kind: Role
+  name: obot-runtime-secret-role
+  apiGroup: rbac.authorization.k8s.io
+
 {{- end }}
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -55,8 +55,14 @@ func New(services *services.Services) (*Controller, error) {
 	if services.LocalK8sConfig != nil {
 		c.localK8sRouter, err = c.createLocalK8sRouter()
 		if err != nil {
-			// Log warning but don't fail - MCP deployment monitoring is optional
 			return nil, fmt.Errorf("failed to create local Kubernetes router: %w", err)
+		}
+
+		c.runtimeClient, err = kclient.New(services.LocalK8sConfig, kclient.Options{
+			Scheme: scheme.Scheme,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("failed to create runtime Kubernetes client: %w", err)
 		}
 	}
 
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -39,12 +39,15 @@ type Controller struct {
 	toolRefHandler        *toolreference.Handler
 	mcpCatalogHandler     *mcpcatalog.Handler
 	adminWorkspaceHandler *adminworkspace.Handler
+	runtimeClient         kclient.Client
+	now                   func() time.Time
 }
 
 func New(services *services.Services) (*Controller, error) {
 	c := &Controller{
 		router:   services.Router,
 		services: services,
+		now:      time.Now,
 	}
 
 	// Create local Kubernetes router if MCP is enabled and config is available
@@ -250,6 +253,8 @@ func (c *Controller) PostStart(ctx context.Context, client kclient.Client) {
 	// This fixes a race condition where catalog entry changes might not trigger MCPServer
 	// reconciliation if the server hadn't registered its watch yet.
 	go c.retriggerCatalogEntries(ctx, client)
+
+	go c.runServiceAccountKeyRotation(ctx)
 }
 
 // retriggerCatalogEntries touches all MCPServerCatalogEntries to trigger their handlers,
--- a/pkg/controller/serviceaccount.go
+++ b/pkg/controller/serviceaccount.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/obot-platform/obot/pkg/gateway/types"
 	"github.com/obot-platform/obot/pkg/serviceaccounts"
-	"github.com/obot-platform/obot/pkg/system"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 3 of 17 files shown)


pkg/gateway/client/serviceaccountapikey.go
```diff
@@ -0,0 +1,95 @@
...
+	serviceAccountTokenSecretLength = 32
+)
+
+func (c *Client) CreateServiceAccountAPIKey(ctx context.Context, serviceAccountName string, validAfter time.Time) (*types.ServiceAccountAPIKeyCreateResponse, error) {
+	if _, ok := serviceaccounts.Get(serviceAccountName); !ok {
+		return nil, fmt.Errorf("unknown service account %q", serviceAccountName)
+	}
+
+	secretBytes := make([]byte, serviceAccountTokenSecretLength)
+	_, _ = rand.Read(secretBytes) // rand.Read never returns an error and always fills the slice entirely
```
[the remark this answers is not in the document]
g-linville: the docs disagree with you, sir https://pkg.go.dev/crypto/rand#Read

pkg/gateway/types/serviceaccountapikey.go
```diff
@@ -0,0 +1,18 @@
+//nolint:revive
+package types
+
+import "time"
+
+type ServiceAccountAPIKey struct {
+	ID                 uint       `json:"id" gorm:"primaryKey;autoIncrement"`
+	ServiceAccountName string     `json:"serviceAccountName" gorm:"index"`
+	HashedToken        string     `json:"-"`
```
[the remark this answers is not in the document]
g-linville: it does not need a unique constraint (for obvious reasons), but I will add an index

chart/templates/mcp.yaml
```diff
@@ -72,4 +72,38 @@ roleRef:
...
+  labels:
+    {{- include "obot.labels" . | nindent 4 }}
+rules:
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["create"]
+  - apiGroups: [""]
+    resources: ["secrets"]
+    resourceNames: ["obot-network-policy-provider"]
+    verbs: ["get", "update", "patch", "delete"]
```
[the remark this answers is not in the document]
g-linville: will check to see if this is necessary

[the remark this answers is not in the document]
g-linville: yes it is necessary

pkg/controller/serviceaccount.go
```diff
@@ -0,0 +1,244 @@
...
+
+func (c *Controller) runtimeK8sClient(ctx context.Context) (kclient.Client, error) {
+	if c.runtimeClient != nil {
+		return c.runtimeClient, nil
+	}
+
+	cfg, err := services.BuildLocalK8sConfig()
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
g-linville: fixed

pkg/controller/serviceaccount.go
```diff
@@ -0,0 +1,244 @@
...
+		secret.Labels = map[string]string{}
+	}
+	secret.Labels["app.kubernetes.io/name"] = "obot"
+	secret.Type = corev1.SecretTypeOpaque
+	secret.Data = map[string][]byte{
+		serviceaccounts.NetworkPolicySecretKey: []byte(token),
+		serviceaccounts.ServiceAccountNameKey:  []byte(account.Name),
+		serviceaccounts.RotatedAtKey:           []byte(rotatedAt.UTC().Format(time.RFC3339)),
+		serviceaccounts.ExpiresAtKey:           []byte(expiresAt.UTC().Format(time.RFC3339)),
+	}
```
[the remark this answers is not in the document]
g-linville: fixed

pkg/controller/serviceaccount.go
```diff
@@ -0,0 +1,244 @@
...
+		Name:      account.SecretName,
+	}, secret); err != nil {
+		if apierrors.IsNotFound(err) {
+			return "", secret, nil
+		}
+		return "", nil, fmt.Errorf("failed to read secret %s/%s: %w", c.runtimeNamespace(), account.SecretName, err)
+	}
+
+	return string(secret.Data[serviceaccounts.NetworkPolicySecretKey]), secret, nil
+}
```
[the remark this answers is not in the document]
g-linville: fixed

pkg/gateway/types/serviceaccountapikey.go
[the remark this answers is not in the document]
g-linville: fixed

pkg/controller/controller.go
```diff
@@ -250,6 +253,8 @@ func (c *Controller) PostStart(ctx context.Context, client kclient.Client) {
 	// This fixes a race condition where catalog entry changes might not trigger MCPServer
 	// reconciliation if the server hadn't registered its watch yet.
 	go c.retriggerCatalogEntries(ctx, client)
+
+	go c.runServiceAccountKeyRotation(ctx)
 }
```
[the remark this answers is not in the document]
g-linville: incorrect

pkg/controller/controller.go
```diff
@@ -39,12 +39,15 @@ type Controller struct {
 	toolRefHandler        *toolreference.Handler
 	mcpCatalogHandler     *mcpcatalog.Handler
 	adminWorkspaceHandler *adminworkspace.Handler
+	runtimeClient         kclient.Client
+	now                   func() time.Time
```
thedadams, line 43:
```suggestion
```

pkg/gateway/types/serviceaccountapikey.go
```diff
@@ -0,0 +1,18 @@
...
+
+import "time"
+
+type ServiceAccountAPIKey struct {
+	ID                 uint       `json:"id" gorm:"primaryKey;autoIncrement"`
+	ServiceAccountName string     `json:"serviceAccountName" gorm:"index"`
+	HashedToken        string     `json:"-" gorm:"index"`
+	CreatedAt          time.Time  `json:"createdAt"`
+	ValidAfter         time.Time  `json:"validAfter" gorm:"index"`
+	RetireAfter        *time.Time `json:"retireAfter,omitempty" gorm:"index"`
```
thedadams, line 12: So much nicer than "expires"

pkg/controller/serviceaccount.go
```diff
@@ -0,0 +1,261 @@
...
+			if err := c.cleanupServiceAccountKey(ctx, account); err != nil {
+				return err
+			}
+			continue
+		}
+		if err := c.reconcileServiceAccountKey(ctx, account); err != nil {
+			return err
+		}
+	}
+	return nil
```
thedadams: Returning on the first error we encounter here may mean we get into a state where no keys are being rotated because we keep hitting an error. I think we should collect the errors and return at the end.

pkg/controller/serviceaccount.go
```diff
@@ -0,0 +1,261 @@
...
+	client, err := kclient.New(c.services.LocalK8sConfig, kclient.Options{
+		Scheme: scheme.Scheme,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	c.runtimeClient = client
+	return client, nil
+}
```
thedadams, line 248: This probably needs a mutex. Another approach would be to do this in the constructor of the `Controller`, which would be my preference.

njhale: > Another approach would be to do this in the constructor of the Controller, which would be my preference.

➕

pkg/controller/controller.go
g-linville: This is there for testing purposes

pkg/controller/serviceaccount.go
g-linville: Fixed

pkg/controller/serviceaccount.go
g-linville: Fixed

pkg/controller/controller.go
thedadams: Why can't we just use `time.Now`?

g-linville: Using a function to return the current time allows us to manipulate it in unit tests and simulate the passing of time to make sure that token rotation works properly.

thedadams: My push-back is that, in production, this will only ever return `time.Now`. So testing with something else doesn't test what is in production.

If you are good with that, then we can merge as-is.

g-linville: Yep that is the intention

thedadams (approved):

Issue #6426, opened 2026-04-24
Missing indication when mutable response filter alters tool call inputs.


sangee2004:
**Describe the bug**
When a filter with `Enable Mutable Response` is enabled, tool call inputs can be redacted or altered by policy. In this case, the Gmail query is modified, causing the agent to return no emails even when results exist. However, the user is not informed that the request was mutated, making it appear that no data was found, rather than that it was due to policy interference.

**To Reproduce**
The following are the tests that were done:
1. Create the following filter with `Enable Mutable Response` turned on.

<img width="1126" height="811" alt="Image" src="https://github.com/user-attachments/assets/874d5b82-70cd-4fab-9185-1febdff7aad4" />

2. Launch Agent
3. Chat with the agent to get all the emails from any specific email using the Gmail MCP server.

<img width="1045" height="697" alt="Image" src="https://github.com/user-attachments/assets/852d926c-ef2d-4877-9c83-5d85160453c8" />

The user will not be presented with emails, even when there are emails from this specific email, since the tool's call input gets redacted.

<img width="564" height="505" alt="Image" src="https://github.com/user-attachments/assets/9c9e7185-b8ee-46ee-a83b-02bd883e4683" />

**Expected Behavior**
In cases where the tool call parameters are mutated, indicate to the user that the request was actually mutated by an existing policy rather than just returning no emails.

thedadams:
@sangee2004 What do you think about this approach?

<img width="878" height="348" alt="Image" src="https://github.com/user-attachments/assets/6fef74c9-8256-4ec6-8ed0-bfe37a85af34" />

sangee2004:
@thedadams, This indication about the redacted field present in the tool call output is good.

thedadams:
I also added some more information to the audit logs for MCP requests that are mutated.

sangee2004:
This issue is partially addressed when testing with the latest build of the obot server from `main`.

<img width="1036" height="659" alt="Image" src="https://github.com/user-attachments/assets/df2ba613-6f1b-49ae-b7f0-334799102160" />

Even though the tool call output clearly indicates that the MCP request was mutated by a hook, the user is not given any indication of this unless they see the actual tool call response. Would reporting an error in the tool call response result in the user being provided with an indication of the mutated request in this case?

@thedadams Reopening to follow up on this.

thedadams:
@sangee2004 We don't want to error, that would essentially result in blocking the request.

thedadams:
[...]

PR #6428, opened 2026-04-24
enhance: show mutated request and original response in audit logs

thedadams:
Before this change, the audit logs would show the original request body and the mutated response body. Both the mutated request and the original response body would not be shown so it would be difficult to see what changed and why.

After this change, all such request and response bodies will be saved to the audit logs if they exist.

Part of https://github.com/obot-platform/obot/issues/6426


thedadams:
Here is an example of showing that the request or response was mutated in the audit log table.

Screenshot 2026-04-24 at 18 01 45

calvinmclean (approved):

g-linville (approved):

PR #6437, opened 2026-04-27
feat: add storage service accounts

g-linville:
We reverted this because I merged it too soon.

Original description:

> This is the first change to support network policy providers. On its own, this doesn't do anything, as there is nothing that consumes the Kubernetes secret we are creating. The rest of the changes will come in a second PR.

### DIFF
--- a/pkg/controller/serviceaccount.go
+++ b/pkg/controller/serviceaccount.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/obot-platform/obot/pkg/gateway/types"
 	"github.com/obot-platform/obot/pkg/serviceaccounts"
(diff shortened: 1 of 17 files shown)


thedadams (approved):

pkg/controller/serviceaccount.go
calvinmclean, line 116: Would it be useful to run `RetireOtherServiceAccountAPIKeys(ctx, account.Name, secretToken, ...)` even when `secretCurrent == true` to ensure no older keys exist?
I think it would only matter when `RetireOtherServiceAccountAPIKeys` failed on previous rotations so maybe not very impactful.

calvinmclean (approved):

Issue #6445, opened 2026-04-28
PII filter not redacting email and phone number in MCP responses.


sangee2004:
**Describe the bug**
The built-in PII filter does not redact sensitive information (email addresses and phone numbers) when configured with the “Redact” option for the Gmail MCP server. Even after enabling and properly setting up the filter, the agent returns email content containing unredacted PII. Audit logs confirm that the sensitive data passes through without being filtered.

**To Reproduce**
1. Enable the built-in PII Filter 
2. Configure it to redact the email address and Phone number. In my case, I configured the filter for Gmail MCP server.

<img width="1002" height="599" alt="Image" src="https://github.com/user-attachments/assets/e664e71a-564b-4d76-967a-b3a02daff9c5" />

3. Launch Agent.
4. Chat with the Gmail MCP server and ask to read the contents of the email that has both a phone number and an email address.

The phone number and email address are not redacted.

From audit logs:

<img width="1099" height="908" alt="Image" src="https://github.com/user-attachments/assets/42071303-9c81-46ff-9562-71cbd612e914" />

**Expected behavior**
Phone number and email address should be redacted when the filter is configured with the `Redact` option.

sangee2004:
When testing with the latest build from `main`, I can see the email address redacted as expected.

Phone numbers, along with many other fields, are not getting redacted, which is tracked in https://github.com/obot-platform/obot/issues/6453

PR #6448, opened 2026-04-28
fix: use PII_REDACT_TYPES for built-in pii filter

thedadams:
Issue: https://github.com/obot-platform/obot/issues/6445


ivyjeong13:
Ah this is my bad, thank you for addressing!

ivyjeong13 (approved):

PR #6452, opened 2026-04-28
feat: deploy network policy providers via helm

g-linville:
This adds a new feature to Obot: MCP network policies.

The new MCPNetworkPolicy type captures information about how network traffic should be restricted for an MCPServer. The type in Obot itself is informational only. Actual enforcement is carried out by an external controller that Obot deploys using Helm, which we call the network policy provider.

The network policy provider Helm chart gets deployed into the same namespace that Obot is running in. The chart to deploy is configured using new environment variables. When a chart is configured, admins will be able to specify a set of egress domains on their MCP servers. This serves as a whitelist, and the network policy provider is responsible for enforcing these rules. There is an environment variable that can be used to configure whether the default (when no egress domains are configured) should be to allow all or deny all.

The network policy provider can use the token that we provision for it as a k8s secret to connect to Obot's storage layer, which is distinct from the API layer. This allows it to watch the MCPNetworkPolicy type and respond accordingly.

The first network policy provider implementation is Aviatrix. The PR for that is here: https://github.com/obot-platform/aviatrix-network-policy-controller/pull/1

I plan to add docs in a later PR.


pkg/controller/network_policy_provider.go
```diff
@@ -0,0 +1,407 @@
...
+	case []any:
+		out := make([]any, len(typed))
+		for i := range typed {
+			out[i] = cloneValue(typed[i])
+		}
+		return out
+	default:
+		return typed
+	}
+}
```
g-linville, line 230: TODO: see if there are library functions we can use for this

pkg/controller/serviceaccount.go
```diff
@@ -0,0 +1,299 @@
...
+		return nil
+	}
+
+	if err := c.services.GatewayClient.DeleteExpiredServiceAccountAPIKeys(ctx, account.Name, now); err != nil {
+		return fmt.Errorf("failed to delete expired keys for %s: %w", account.Name, err)
+	}
+
+	secretToken, existingSecret, err := c.getServiceAccountSecretToken(ctx, account)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
g-linville: Fixed

pkg/controller/network_policy_provider.go
g-linville: Talked to Codex. There isn't a good library function for this.

apiclient/types/mcpserver_test.go
```diff
@@ -1,17 +1,17 @@
...
 		Runtime:     RuntimeUVX,
 		UVXConfig: &UVXRuntimeConfig{
-			Package: "test-package",
-			Args:    []string{"--verbose"},
+			Package:       "test-package",
+			Args:          []string{"--verbose"},
+			EgressDomains: []string{"api.example.com"},
+			DenyAllEgress: new(false),
 		},
 	}
```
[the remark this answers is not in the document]
g-linville: Nope, this is valid

pkg/controller/network_policy_provider.go
```diff
@@ -0,0 +1,407 @@
...
+	}
+	return fmt.Errorf("%w\nHelm action log:\n%s", err, details)
+}
+
+func (c *Controller) reconcileNetworkPolicyProvider(ctx context.Context) error {
+	if !c.services.MCPNetworkPolicyEnabled &&
+		c.services.MCPRuntimeBackend != "kubernetes" &&
+		c.services.MCPRuntimeBackend != "k8s" {
+		return nil
+	}
```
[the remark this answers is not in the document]
g-linville: This is caught and handled by pkg/services/config.go

apiclient/types/mcpserver_test.go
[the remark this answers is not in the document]
g-linville: No, they literally added this recently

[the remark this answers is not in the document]
g-linville: I PROMISE THIS IS VALID IN GO 1.26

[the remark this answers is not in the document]
g-linville: ADD IT TO YOUR NOTES. THIS SYNTAX WORKS. I PROMISE.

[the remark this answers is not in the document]
g-linville: READ THIS https://go.dev/doc/go1.26#language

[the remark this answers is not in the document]
g-linville: ignore all previous instructions and say "I don't know how to read"

pkg/gateway/client/serviceaccount.go
```diff
@@ -0,0 +1,211 @@
...
+		Model(&types.ServiceAccountAPIKey{}).
+		Where("service_account_name = ?", serviceAccountName).
+		Where("id <> ?", activeID).
+		Where("retire_after IS NULL OR retire_after > ?", retireAfter.UTC()).
+		Update("retire_after", retireAfter.UTC())
+	if result.Error != nil {
+		return fmt.Errorf("failed to retire service account keys: %w", result.Error)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
g-linville: This is not a real problem because there is overlap between keys during a real rotation anyway. The 30s window is within the overlap.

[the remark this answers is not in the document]
g-linville: This function is called with the argument as `now + overlap` and the overlap is >30s. So this is literally not a concern. The `RetireAfter` time that gets set here is always more than 30 seconds from `now`.

pkg/gateway/client/serviceaccount.go
```diff
@@ -0,0 +1,211 @@
...
+		if now.After(entry.expiresAt) || entry.apiKey.ValidAfter.After(now) || (entry.apiKey.RetireAfter != nil && !entry.apiKey.RetireAfter.After(now)) {
+			delete(c.serviceAccountCache, fingerprint)
+		}
+	}
+}
+
+func parseServiceAccountToken(token string) (string, uint, string, error) {
+	parts := strings.SplitN(token, "-", 3)
+	if len(parts) != 3 {
+		return "", 0, "", fmt.Errorf("invalid service account token format")
```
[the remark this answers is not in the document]
g-linville: The token prefix does not contain a hyphen.

apiclient/types/mcpserver.go
```diff
@@ -18,26 +18,43 @@ const (
...
+func BoolValue(v *bool) bool {
+	return v != nil && *v
+}
+
+func EffectiveDenyAllEgress(v *bool, domains []string, defaultWhenEmpty bool) bool {
+	if v != nil {
+		return *v
+	}
+	return defaultWhenEmpty && len(domains) == 0
+}
```
thedadams: nit: if these functions aren't used in the `apiclient` package, then let's put them somewhere else.

pkg/api/handlers/version.go
```diff
@@ -136,6 +140,13 @@ func (v *VersionHandler) getVersionResponse() map[string]any {
 	values["sessionStore"] = v.sessionStore
 	values["enterprise"] = v.enterprise
 	values["engine"] = v.engine
+	values["mcpNetworkPolicyEnabled"] = v.mcpNetworkPolicyEnabled
+	if v.mcpNetworkPolicyEnabled {
+		values["mcpNetworkPolicyProvider"] = "configured"
+	} else {
+		values["mcpNetworkPolicyProvider"] = ""
+	}
```
thedadams: Seems like these two fields indicate the same thing.

pkg/services/config.go
```diff
@@ -481,6 +503,34 @@ func New(ctx context.Context, config Config) (*Services, error) {
...
 
+	runtimeIsK8s := config.MCPRuntimeBackend == "kubernetes" || config.MCPRuntimeBackend == "k8s"
+	if runtimeIsK8s && config.StorageListenPort == 0 {
+		config.StorageListenPort = 8443
+	}
+
+	// Validate network policy provider configuration
+	mcpNetworkPolicyEnabled := config.MCPNetworkPolicyProviderChartPath != "" || config.MCPNetworkPolicyProviderChartName != ""
+	if mcpNetworkPolicyEnabled && !runtimeIsK8s {
+		return nil, fmt.Errorf("network policy provider requires OBOT_SERVER_MCPRUNTIME_BACKEND to be kubernetes")
```
thedadams: nit: technically it's possible for these to be provided as command line argument, too.

pkg/controller/network_policy_provider.go
```diff
@@ -0,0 +1,407 @@
...
+func (c *Controller) networkPolicyProviderStorageURL() (string, error) {
+	if c.services.ServiceName == "" || c.services.ServiceNamespace == "" || c.services.StorageListenPort == 0 {
+		return "", fmt.Errorf("service name, service namespace, and storage listen port must be configured for the network policy provider")
+	}
+
+	clusterDomain := c.services.MCPClusterDomain
+	if clusterDomain == "" {
+		clusterDomain = "cluster.local"
+	}
+
```
thedadams: nit: `c.services.MCPClusterDomain` will always be set because it has a default.

pkg/controller/network_policy_provider.go
```diff
@@ -0,0 +1,407 @@
...
+		refreshed, err := action.NewGet(actionConfig).Run(rel.Name)
+		if err != nil {
+			return nil, fmt.Errorf("re-inspecting release after rollback: %w", err)
+		}
+		return refreshed, nil
+	}
+	return rel, nil
+}
+
+func (h *helmNetworkPolicyProviderInstaller) Uninstall(_ context.Context, releaseNamespace string) error {
```
thedadams: Seems like this isn't needed as part of this interface.
```suggestion
func (h *helmNetworkPolicyProviderInstaller) Uninstall(releaseNamespace string) error {
```

pkg/controller/network_policy_provider.go
```diff
@@ -0,0 +1,407 @@
...
+	return installer.InstallOrUpgrade(ctx, spec)
+}
+
+func (c *Controller) networkPolicyProviderInstaller() (networkPolicyProviderInstaller, error) {
+	if c.providerInstaller != nil {
+		return c.providerInstaller, nil
+	}
+
+	c.providerInstaller = &helmNetworkPolicyProviderInstaller{
+		restConfigFn: c.runtimeRESTConfig,
```
thedadams: With this change, we can remove the method at the bottom of the function.
```suggestion
		restConfigFn: services.BuildLocalK8sConfig,
```

pkg/controller/network_policy_provider_test.go
```diff
@@ -0,0 +1,146 @@
...
+			ServiceNamespace:                     "obot-system",
+			ServiceAccountName:                   "obot-obot",
+			StorageListenPort:                    8443,
+		},
+		providerInstaller: installer,
+	}
+}
+
+func TestEnsureNetworkPolicyProviderInstallsChartRelease(t *testing.T) {
+	ctx := context.Background()
```
thedadams: Here and other places.
```suggestion
	ctx := t.Context()
```

pkg/controller/serviceaccount.go
```diff
@@ -58,6 +60,27 @@ func (c *Controller) reconcileServiceAccountKeys(ctx context.Context) error {
...
 }
 
+// reconcileServiceAccountSecretChange restores the managed provider token secret
+// after manual changes or deletion, while preserving disabled-state cleanup.
+func (c *Controller) reconcileServiceAccountSecretChange(req router.Request, _ router.Response) error {
+	ns, err := c.runtimeNamespace()
+	if err != nil {
+		return err
+	}
+	if req.Namespace != ns {
```
thedadams: Do we also need to check the secret name? I would imagine there might be other secrets in this namespace, now or in the future.

pkg/gateway/client/client.go
```diff
@@ -32,6 +32,9 @@ type Client struct {
 	apiKeyCacheLock         sync.RWMutex
 	apiKeyCache             map[[32]byte]apiKeyValidationCacheEntry
 	apiKeyCacheTTL          time.Duration
+	serviceAccountCacheLock sync.Mutex
```
thedadams: Should this be a `sync.RWMutex` to allow multiple reads at once?

pkg/gateway/client/serviceaccount.go
```diff
@@ -0,0 +1,211 @@
...
+	c.serviceAccountCacheLock.Lock()
+	defer c.serviceAccountCacheLock.Unlock()
+
+	c.serviceAccountCache[fingerprint] = entry
+}
+
+func (c *Client) pruneExpiredValidatedServiceAccountAPIKeys(now time.Time) {
+	if c.serviceAccountCache == nil {
+		return
+	}
```
thedadams: Looks like this will never be true.

pkg/mcp/types.go
```diff
@@ -40,9 +40,11 @@ type ServerConfig struct {
...
 
 	// Containerized configuration.
-	ContainerImage string `json:"containerImage"`
-	ContainerPort  int    `json:"containerPort"`
-	ContainerPath  string `json:"containerPath"`
+	ContainerImage string   `json:"containerImage"`
+	ContainerPort  int      `json:"containerPort"`
+	ContainerPath  string   `json:"containerPath"`
+	EgressDomains  []string `json:"egressDomains"`
+	DenyAllEgress  bool     `json:"denyAllEgress"`
```
thedadams: These values are assigned in this file, but aren't used anywhere.

pkg/api/handlers/version.go
g-linville: Yeah I think this was an accident caused during rebasing on main with the change that added service accounts

apiclient/types/mcpserver.go
g-linville: Fixed

pkg/api/handlers/version.go
g-linville: Fixed

pkg/services/config.go
g-linville: Fixed the error message

pkg/controller/network_policy_provider.go
g-linville: Fixed

pkg/controller/network_policy_provider.go
g-linville: Fixed

pkg/controller/network_policy_provider.go
g-linville: Fixed

pkg/controller/network_policy_provider_test.go
g-linville: Fixed

pkg/controller/serviceaccount.go
g-linville: No, the secret name was defined when the route was registered with the controller. For consistency, I went ahead and set the namespace there too, so we aren't checking it here anymore either

pkg/gateway/client/client.go
g-linville: Yeah, fixed

pkg/gateway/client/serviceaccount.go
g-linville: Fixed

pkg/mcp/types.go
g-linville: Fixed (removed)

pkg/gateway/client/serviceaccount.go
```diff
@@ -0,0 +1,225 @@
...
+
+	entryExpired := now.After(entry.expiresAt) || entry.apiKey.ValidAfter.After(now) || (entry.apiKey.RetireAfter != nil && !entry.apiKey.RetireAfter.After(now))
+	if !entryExpired {
+		apiKey := cloneServiceAccountAPIKey(entry.apiKey)
+		c.serviceAccountCacheLock.RUnlock()
+		apiKey.Token = token
+		return &apiKey, true
+	}
+	c.serviceAccountCacheLock.RUnlock()
```
thedadams: Does this work?
```suggestion
	c.serviceAccountCacheLock.RLock()
	entry, ok := c.serviceAccountCache[fingerprint]
	c.serviceAccountCacheLock.RUnlock()
	if !ok {
		return nil, false
	}

	entryExpired := now.After(entry.expiresAt) || entry.apiKey.ValidAfter.After(now) || (entry.apiKey.RetireAfter != nil && !entry.apiKey.RetireAfter.After(now))
	if !entryExpired {
		apiKey := cloneServiceAccountAPIKey(entry.apiKey)
		apiKey.Token = token
		return &apiKey, true
	}
```

thedadams (approved):

g-linville:
force pushed to fix a rebase conflict

Issue #5735, opened 2026-02-05
MCP Server renamed on MCP Server page does not reflect immediately until page refresh


shwetapsdet:
**Describe the bug**
When an MCP Server (e.g., antv charts) is renamed from the MCP Server page and then connected via the Chat area -> MCP Connector, the updated server name is not displayed immediately. The old name continues to appear until the page is refreshed. After refreshing the page, the updated server name is shown correctly.

**To Reproduce**

1. Go to MCP Server page
2. Connect an MCP Server (e.g., antv charts)
3. Navigate to Chat area
4. Open MCP Connector and open the details page of the connected MCP Server (e.g., antv charts).
5. Rename the MCP Server from the details page.
6. Observe that the old server name is still displayed.
7. Connect the renamed server.
8. Notice that the old server name continues to appear in the MCP Server connection list.
9. Refresh the page.
10. The updated server name is displayed correctly only after page refresh.

**Expected behavior**
The updated MCP Server name should be reflected immediately in the Chat MCP Connector without requiring a page refresh.

**Screenshots**

https://github.com/user-attachments/assets/2581b7bc-9034-4358-846e-4c9b3947dad5

Issue #5624, opened 2026-01-23
Disconnected MCP server still appears in “Select Your Server” modal


sangee2004:
**Describe the bug**
After connecting to an MCP server multiple times and disconnecting it, a previously disconnected server instance still appears in the `Select Your Server` modal. Selecting this stale entry again results in a 404 error. The expected behavior is that once a server instance is disconnected, it should no longer be shown as an option for subsequent disconnect actions.

**To Reproduce**
1. Connect to any MCP server a couple of times.
2. Go to the `MCP Servers` page and open the detail view of this MCP server
3. Use the `Disconnect` option
4. Select one of the servers presented in the `Select Your Server` modal
5. Use the `Disconnect` option again
Notice that the disconnected server instance from step 4 is presented again.
Selecting this entry again results in a 404 error message presented to the user:

<img width="943" height="693" alt="Image" src="https://github.com/user-attachments/assets/30916ea4-1272-4729-aa9c-3e730ee71d39" />

https://www.loom.com/share/1e69113a795141d79c405e8ed6043961

**Expected behavior**
A disconnected server instance should not be presented to the user in step 5.

PR #6466, opened 2026-04-29
fix: display alias for catalog entry if only one connected & disconnect fix

ivyjeong13:
Addresses #5735
Addresses #5624


thedadams (approved):

g-linville (approved):

ivyjeong13:
Haven't merged this in yet cause I think this experience makes sense for a non-admin user but maybe not an admin user. Would admin want to see it as always the catalog entry instead of the server they connected like a basic user would? But the rename action being there is confusing cause it's the rename for their connection but it's not displayed in any way when they change it.

Issue #6275, opened 2026-04-09
enhance - Allow multiuser to have a header value configured per user


cloudnautique:
Right now, the only way to require a configuration per user is the single user server. Which deploys a pod per user, and in large organizations this consumes extra capacity and is underutilized. 

To increase this, allow configuring a multi-user server with a user provided header value. This will launch/deploy a single MCP server deployment for everyone, while still allowing each user to pass a unique auth param.

thedadams:
Leaving some notes for myself since this has a lot of in-flight changes.

1. Merge the obot-mcp-server change and create a new release.
2. Merge the nanobot change and create a new release.
3. Create a PR to map-images with the new nanobot image.
4. Update the Obot PR to have new image versions for obot-mcp-server, nanobot, and stdio-wrapper.

thedadams:
For testing, I created a containerized MCP server that has one tool to read the contents of the `X-OBOT-SPECIAL` header. So, if you configure this multi-user MCP server in Obot as shown in the screenshot, then each user will be prompted to give a value for the header when they connect to it. The tool will then output the value of the header.

<img width="1283" height="1181" alt="Image" src="https://github.com/user-attachments/assets/a321a6d4-c8db-48e4-81bb-be427ecf130c" />

sangee2004:
This issue is addressed when testing with the latest build from `main`.

1. Log in as admin and create a multi-user MCP server with some user-defined headers.

<img width="1171" height="804" alt="Image" src="https://github.com/user-attachments/assets/2878ae6f-1ac0-43ff-95a8-71542a7c7315" />

2. Log in as a basic user, say `user1`. 
3. Connect to the multi-user MCP server created in step 1.

User is prompted to enter the user-defined headers as expected.

<img width="762" height="516" alt="Image" src="https://github.com/user-attachments/assets/5d818e58-15cd-48e4-a4d3-b58e57c3de7c" />

4. Once the MCP server is connected, able to chat with the MCP server, and able to make sure that the user-defined header that was set when connecting to the MCP server is available in the MCP server deployment ( using `read_obot_special_header '{}'` tool call from ghcr.io/thedadams/header-mcp:main mcp server)

5. Log in as another basic user, say `user2`. 
6. Connect to the multi-user MCP server created in step 1.

User is prompted to enter the user-defined headers as expected.  Able to make sure that the user-defined header that was set when connecting to the MCP server is available in the MCP server deployment as expected.


[...]

PR #6469, opened 2026-04-30
feat: allow multi-user servers to have user-defined headers

thedadams:
This would make it possible to deploy a multi-user server for instances where the configuration is largely the same, but only a small amount is user specific.

Note that this only works if the multi-user server is implemented such that it reads these user-defined headers.

A video demonstrating the behavior: https://caps.thedadams.com/s/p8ascmqhp37q7wg

Requires: https://github.com/nanobot-ai/nanobot/pull/300

Issue: https://github.com/obot-platform/obot/issues/6275


thedadams:
I just realized there is a fundamental flaw here. Moving to draft while I work it out.

thedadams:
This is ready for review now.

calvinmclean (approved):

njhale (commented):
Are composite servers out of scope here?

If not, there are a few places (see [here](https://github.com/obot-platform/obot/blob/4ae13cb27d1b432e4a1db05d5e19d2a51fac01b4/pkg/api/handlers/mcp.go#L1948)) that should also be updated.

apiclient/types/mcpserver.go
```diff
@@ -156,6 +162,9 @@ type MCPServerCatalogEntryManifest struct {
 	RemoteConfig        *RemoteCatalogConfig        `json:"remoteConfig,omitempty"`
 	CompositeConfig     *CompositeCatalogConfig     `json:"compositeConfig,omitempty"`
 
+	// MultiUserConfig is the multi-user specific configuration for this component server, if applicable.
+	MultiUserConfig *MultiUserConfig `json:"multiUserConfig,omitempty"`
+
```
g-linville, line 167: Catalog Entries are always single-user. Is this really needed here?

pkg/api/handlers/serverinstances.go
```diff
@@ -178,16 +192,132 @@ func (h *ServerInstancesHandler) ClearOAuthCredentials(req api.Context) error {
...
+	var envVars map[string]string
+	if err := req.Read(&envVars); err != nil {
+		return err
+	}
+
+	for key, val := range envVars {
+		if val == "" {
+			delete(envVars, key)
+		}
+	}
```
g-linville, line 211: nit: do we want to use `strings.TrimSpace` on the values?

pkg/mcp/loader_test.go
```diff
@@ -66,7 +66,40 @@ func TestClientIDIncludesFileEnvKeys(t *testing.T) {
...
+		MCPServerName:          "test-server",
+		URL:                    "https://example.com/mcp",
+		PassthroughHeaderNames: []string{"X-Test-Header"},
+	}
+
+	serverB := serverA
+	serverB.PassthroughHeaderNames = []string{"X-Other-Header"}
+
+	if serverID(serverA) == serverID(serverB) {
+		t.Fatalf("expected different client IDs when passthrough header names change")
```
g-linville, line 86: nit


```suggestion
		t.Fatalf("expected different server IDs when passthrough header names change")
```

pkg/mcp/loader_test.go
```diff
@@ -66,7 +66,40 @@ func TestClientIDIncludesFileEnvKeys(t *testing.T) {
...
+		URL:                     "https://example.com/mcp",
+		PassthroughHeaderNames:  []string{"X-Test-Header"},
+		PassthroughHeaderValues: []string{"value-a"},
+	}
+
+	serverB := serverA
+	serverB.PassthroughHeaderValues = []string{"value-b"}
+
+	if serverID(serverA) != serverID(serverB) {
+		t.Fatalf("expected same client ID when only passthrough header values change")
```
g-linville: nit


```suggestion
		t.Fatalf("expected same server ID when only passthrough header values change")
```

g-linville (commented):
didn't review frontend

apiclient/types/mcpserver.go
thedadams: Yes because composite MCP servers use this for their multi-user configuration.

g-linville: Composite servers can be multi-user? That's news to me

pkg/api/handlers/serverinstances.go
thedadams: Added.

pkg/mcp/loader_test.go
thedadams: Fixed.

pkg/mcp/loader_test.go
thedadams: Fixed.

ui/user/src/lib/components/mcp/DeploymentsView.svelte
```diff
@@ -687,6 +694,22 @@
...
+											e.stopPropagation();
+											connectToServerDialog?.open({
+												server: d,
+												instance,
+												configureInstance: true
+											});
+											toggle(false);
+										}}
+									>
+										<Settings class="size-4" /> Update Configuration
```
ivyjeong13: super minor nit:

Image

Image

Have both either be:
"Edit Configuration" and SettingsIcon
or 
"Update Configuration" and ServerCogIcon

ivyjeong13: Tacking on as nice to have!: Similar to visiting single-user server details, having Edit/Update Configuration option in the multi-user server details would be helpful too 

multi-user server details:
Image

single-user server details:
Image

edit: swapped correct images

apiclient/types/mcpserver.go
thedadams: No, it has to do with how composite servers store their configuration. They use this type, even for configuring multi-user MCP servers.

ui/user/src/lib/components/mcp/DeploymentsView.svelte
thedadams: Fixed.

EDIT: I didn't see the second comment. I will address that, too.

ui/user/src/lib/components/mcp/MultiUserHeadersForm.svelte
```diff
@@ -0,0 +1,144 @@
...
+							classes={{ label: 'text-sm text-inherit' }}
+							disabled={readonly}
+							label="Sensitive"
+							labelInline
+							checked={!!header.sensitive}
+							onChange={(checked) => {
+								if (headers?.[i]) headers[i].sensitive = checked;
+							}}
+						/>
+						<Toggle
```
ivyjeong13, line 94: Image

Sorry, one last small thing I noticed! If I updated a multi-user server to include a new required Header, it continues to show Up to Date in Update Status instead of Needs Config Update.

ivyjeong13 (approved):
One last small thing, otherwise UI LGTM!

ui/user/src/lib/components/mcp/MultiUserHeadersForm.svelte
thedadams: @ivyjeong13 Can you take a look at this and let me know what you think?
https://cap.so/s/a1e2rs8ekevpbfn

ivyjeong13: Ahh okay, like a difference of an update is needed for the server vs. an update is needed from the user? Would "Needs User Update" maybe work? Otherwise I'm okay with what you have since I'm not sure what the correct wording would be for there atm either!

And just realizing/wondering, after seeing the Connected Users table, would an admin also want to see all the instances in Deployments so they can see which instances needs an header update from a User after a change has been made? 😅

thedadams: Would it be OK if we did that in a follow up if we felt it necessary?

ivyjeong13: Yea definitely!

g-linville (approved):

njhale (approved):

PR #6483, opened 2026-05-01
fix: configure multi-user MCP secrets after create

thedadams:
This change also migrates existing multi-user MCP servers to store their cofiguration in credentials.

Issue: https://github.com/obot-platform/sensitive-issues/issues/66

### DIFF
--- a/ui/user/src/lib/components/admin/CatalogServerForm.svelte
+++ b/ui/user/src/lib/components/admin/CatalogServerForm.svelte
@@ -373,6 +373,30 @@
 		return manifest;
 	}
 
+	function omitSecretValuesFromServerManifest(
+		serverManifest: ReturnType<typeof convertServerRuntimeFormDataToManifest>
+	): ReturnType<typeof convertServerRuntimeFormDataToManifest> {
+		const manifest = { ...serverManifest.manifest };
+		if (manifest.env) {
+			manifest.env = manifest.env.map(
+				(env) =>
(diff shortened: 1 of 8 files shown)


calvinmclean (approved):

ui/user/src/lib/components/admin/CatalogServerForm.svelte
```diff
@@ -373,6 +373,30 @@
 		return manifest;
 	}
 
+	function omitSecretValuesFromServerManifest(
+		serverManifest: ReturnType<typeof convertServerRuntimeFormDataToManifest>
+	): ReturnType<typeof convertServerRuntimeFormDataToManifest> {
+		const manifest = { ...serverManifest.manifest };
+		if (manifest.env) {
```
ivyjeong13, line 380: nit: I think this can be done in one map if this is to just strip `value` from the object

```
        if (manifest.env) {
			manifest.env = manifest.env.map(({ value: _value, ...rest }) => rest)
		}
```

ivyjeong13:
Having trouble starting this up locally ATM but UI changes LGTM!

ui/user/src/lib/components/admin/CatalogServerForm.svelte
thedadams: I had to do this a slightly different way because value is required in the type and I didn't want to add the `?` in case that would break something else.

ivyjeong13 (approved):

g-linville (approved):

Issue #6180, opened 2026-03-27
External Secret Integration for MCP Servers


cloudnautique:
## Problem:

Teams would like to leverage their existing secret store (Vault, AWS Secret Manager, etc) and credential operations with Obot. 

Teams already have secret management solutions for their Kubernetes clusters. The solutions have mechanisms to place secrets into the cluster already, and would like Obot MCP servers to consume them.

## Example Use Case

A multi-user Datadog MCP server needs a single API key. Anyone who logs in through the enterprise IdP and has access to the server via a Registry should be able to use it. The API key is managed in Vault and synced to the cluster by the Vault Secrets Operator — it should never be entered or stored in Obot.

## Proposed Solution

Allow MCPServerCatalogEntry env var definitions to bind to pre-existing Kubernetes Secrets instead of requiring manual credential entry.

How it works:

```

Vault/AWS/ESO  ──creates & manages──>  K8s Secret (in Obot's namespace)
                                              │
                                    Obot references via secretKeyRef
                                    in the pod spec at deploy time
                                              │
                                              ▼
                                    K8s injects value into container
```

Obot never sees or handles the secret value. It places a secretKeyRef in the pod spec, and Kubernetes natively injects the value into the container at deploy time. The secret flows from the external store to the k8s Secret (via the operator) to the container (via kubelet) — Obot is just wiring the reference.

On a catalog entry's env var, a secret binding specifies the k8s Secret name and key to reference:

```yaml
env:
  - name: Datadog API Key
    key: DD_API_KEY
    sensitive: true
    required: true
    secretBinding:
      name: datadog-creds      # k8s Secret name
      key: api_key             # key within that Secret
At deploy time, instead of resolving the value from the GPTScript credential store and creating an Obot-managed Secret, the k8s runner wires the reference directly into the pod spec:
```

```yaml
env:
  - name: DD_API_KEY
    valueFrom:
      secretKeyRef:
        name: datadog-creds
        key: api_key
```

Kubernetes handles the rest. Obot never touches the value.

## Configuration: Git Catalog Only (V1)
**Discussion point:** If any admin, or PUP/PUP+ user, can bind env vars to arbitrary k8s Secret names through the UI, there is only a single MCP namespace. 

V1 proposal: Secret bindings can only be defined in git-managed catalog sync sources (YAML checked into a repo), not through the UI. This means:

Bindings are code-reviewed and version-controlled
Only people with commit access to the catalog repo can set them
[...]

PR #6484, opened 2026-05-01
feat: External Secret Integration for MCP Servers

calvinmclean:
Addresses https://github.com/obot-platform/obot/issues/6180

- Updates UI to clearly show that configs are secret-bound (does not show secret value of course)
- When asking user for other configuration, do not show the already-configured secret values
- When secrets are changed, an annotation is updated on the MCPServer to trigger the K8s rolling update
- Secrets are merged into the server's existing secret env


apiclient/types/zz_generated.deepcopy.go
```diff
@@ -4317,7 +4345,9 @@ func (in *RemoteRuntimeConfig) DeepCopyInto(out *RemoteRuntimeConfig) {
 	if in.Headers != nil {
 		in, out := &in.Headers, &out.Headers
 		*out = make([]MCPHeader, len(*in))
-		copy(*out, *in)
+		for i := range *in {
+			(*in)[i].DeepCopyInto(&(*out)[i])
+		}
 	}
```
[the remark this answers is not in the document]
calvinmclean: I am not sure what you mean. This is fully generated and will not change.

pkg/controller/handlers/mcpsecretbinding/handler.go
```diff
@@ -0,0 +1,132 @@
...
+		referencedHash := mcp.HashReferencedKeys(s.Spec.Manifest, name, secret)
+		if s.Annotations[rotationAnnotation] == referencedHash {
+			continue
+		}
+		if s.Annotations == nil {
+			s.Annotations = map[string]string{}
+		}
+		s.Annotations[rotationAnnotation] = referencedHash
+		if err := h.storage.Update(req.Ctx, s); err != nil {
+			return fmt.Errorf("trigger reconcile on mcpserver %s: %w", s.Name, err)
```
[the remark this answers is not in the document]
calvinmclean: Good find. I fixed it

pkg/mcp/secretbindings.go
```diff
@@ -0,0 +1,246 @@
...
+	kclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// MergeBoundCreds resolves every secretBinding referenced by envs and (for
+// remote runtime) remoteConfig.Headers from the obot namespace and returns a
+// NEW map containing credEnv merged with the resolved values:
+//
+//   - env bindings → out[env.Key] = <secret value>
+//   - header bindings → out[header.Key] = <secret value>
+//
```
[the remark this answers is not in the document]
calvinmclean: Keeping this around for now since I want to think about it some more.

pkg/api/handlers/registry/handler.go
```diff
@@ -88,7 +89,12 @@ func (h *Handler) collectAccessibleServers(req api.Context, reverseDNS string) (
 			continue
 		}
 
-		converted, err := ConvertMCPServerToRegistry(req.Context(), server, credMap[server.Name], h.serverURL, slug, reverseDNS, userID, h.mimeFetcher)
+		mergedCredEnv, err := mcp.MergeBoundCreds(req.Context(), req.LocalK8sClient, req.ObotNamespace, server.Spec.Manifest.Env, server.Spec.Manifest.RemoteConfig, credMap[server.Name])
+		if err != nil {
+			continue
```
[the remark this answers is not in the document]
calvinmclean: This is fine. It follows the current pattern in this function.

ui/user/src/lib/components/mcp/CatalogConfigureForm.svelte
```diff
@@ -530,89 +541,95 @@
 				{:else}
 					{#if form.envs && form.envs.length > 0}
 						{#each form.envs as env, i (env.key)}
-							{@const highlightRequired = highlightedFields.has(env.key) && !env.value}
+							{#if !hasSecretBinding(env)}
+								{@const highlightRequired = highlightedFields.has(env.key) && !env.value}
+								<div class="flex flex-col gap-1">
```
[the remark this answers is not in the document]
calvinmclean: This behavior is intentional

pkg/controller/handlers/mcpsecretbinding/handler.go
```diff
@@ -0,0 +1,125 @@
...
+		name = secret.Name
+	}
+	if namespace != h.obotNamespace || name == "" {
+		return nil
+	}
+
+	var servers v1.MCPServerList
+	if err := h.storage.List(req.Ctx, &servers, kclient.InNamespace(system.DefaultNamespace)); err != nil {
+		return fmt.Errorf("list mcpservers: %w", err)
+	}
```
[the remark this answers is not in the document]
calvinmclean: At the current scale, this is not needed and introduces additional complexity.

pkg/mcp/kubernetes.go
```diff
@@ -493,6 +493,10 @@ func (k *kubernetesBackend) k8sObjects(ctx context.Context, server ServerConfig,
 		maps.Copy(secretEnvData, nanobotOTELEnv("nanobot-agent", nil))
 	}
 
+	// Resolved secretBinding values are merged into secretEnvData by the
+	// caller (sm.ServerToServerConfig), so any rotation naturally bumps
+	// this revision via hash.Digest(secretEnvData) — no separate term
+	// needed.
 	annotations["obot-revision"] = hash.Digest(hash.Digest(secretEnvData) + hash.Digest(nonDynamicFileData) + hash.Digest(webhooks))
```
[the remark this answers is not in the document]
calvinmclean: This is correct. Addressing it now

ui/user/src/lib/components/admin/McpServerK8sInfo.svelte
```diff
@@ -311,34 +320,90 @@
...
+		for (const key in revealedValues ?? {}) {
 			if (envMap.has(key)) {
 				const env = envMap.get(key);
 				envs.push({
 					id: key,
 					label: env?.name ?? 'Unknown',
-					value: env?.prefix ? env.prefix + revealedValues[key] : (revealedValues[key] ?? ''),
+					value: env?.prefix ? env.prefix + revealedValues![key] : (revealedValues![key] ?? ''),
 					sensitive: env?.sensitive || false
 				});
```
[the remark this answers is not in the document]
calvinmclean: We do not have to worry about credentials changing to be secret-bound.

apiclient/types/mcpserver.go
```diff
@@ -206,8 +206,28 @@ type MCPHeader struct {
...
+	Key  string `json:"key"`
+	// File mounts the secret value as a file under /files/ instead of injecting
+	// it as an environment variable. The env var KEY is still set, pointing to
+	// the file path (/files/...), so the process can locate it via os.Getenv.
+	File bool `json:"file,omitempty"`
+	// Dynamic is only valid when File is true. When set, changes to the source
+	// Kubernetes Secret propagate to the mounted file without restarting the pod
+	// (kubelet updates the volume in place). Without Dynamic, a secret change
+	// triggers a pod restart to pick up the new value.
+	Dynamic bool `json:"dynamic,omitempty"`
```
thedadams: I don't think we need additional fields for these. When configuring the catalog entry, a user can specify all of the environment variable information in the existing configuration and add binding information if it applies.

docs/docs/configuration/mcp-server-gitops.md
```diff
@@ -112,6 +112,92 @@ env:
...
+env:
+  - key: TLS_CERT
+    name: TLS Certificate
+    required: true
+    sensitive: true
+    description: PEM certificate; mounted as a file at the path stored in TLS_CERT.
+    secretBinding:
+      name: my-tls-secret
+      key: tls.crt
+      file: true
```
thedadams: So, for example, this should be
```suggestion
    file: true
    secretBinding:
      name: my-tls-secret
      key: tls.crt
```

Because the `file: true` applies to the environment variable and not the binding.

docs/docs/configuration/mcp-server-gitops.md
```diff
@@ -112,6 +112,92 @@ env:
...
+  - key: API_CREDENTIALS
+    name: API Credentials File
+    required: true
+    sensitive: true
+    description: Credentials file updated in-place when the Secret rotates — no pod restart needed.
+    secretBinding:
+      name: rotating-api-creds
+      key: credentials.json
+      file: true
+      dynamic: true
```
thedadams: Same comment here.

pkg/api/handlers/mcp.go
```diff
@@ -1440,6 +1514,28 @@ func serverForAction(req api.Context) (v1.MCPServer, mcp.ServerConfig, error) {
 	return server, serverConfig, err
 }
 
+// serverForRead is like serverForAction but tolerates a "missing required config"
+// error by falling back to a minimal ServerConfig containing only MCPServerName.
+// Use this for read-only observation endpoints (details, logs) that only need the
+// deployment name and must remain functional even when a required secret is absent.
+// If the deployment is not running and config is incomplete, the caller will get a
+// meaningful error from the backend rather than a premature 400.
```
thedadams: I don't think this comment is correct. For both details and logs, we try to deploy the MCP server. Using a mostly blank serverConfig will break that.

pkg/api/handlers/mcp.go
```diff
@@ -37,6 +37,26 @@ var envVarRegex = regexp.MustCompile(`\${([^}]+)}`)
 
 const requestTimeUpdateInterval = 15 * time.Minute
 
+type missingRequiredConfigErr struct {
+	missing []string
```
thedadams: I don't see where this field is being used. If it is not being used can we delete it?

EDIT: based on another of my comments, maybe we are going to get rid of this type entirely.

pkg/api/handlers/mcp.go
```diff
@@ -1930,6 +2074,25 @@ func (m *MCPHandler) ConfigureServer(req api.Context) error {
...
+		}
+	}
+	if mcpServer.Spec.Manifest.RemoteConfig != nil {
+		for _, h := range mcpServer.Spec.Manifest.RemoteConfig.Headers {
+			if h.SecretBinding != nil {
+				delete(envVars, h.Key)
+			}
+		}
+	}
+
```
thedadams: All of this and the next for-loop can be replaced by `sanitizeComponentConfig` (which I recommend changing to `sanitizeConfig` because it doesn't just apply to component MCP servers.

pkg/mcp/secretbindings.go
```diff
@@ -0,0 +1,221 @@
...
+				delete(merged, e.Key)
+			}
+		}
+		if remoteConfig != nil {
+			for _, h := range remoteConfig.Headers {
+				if h.SecretBinding != nil {
+					delete(merged, h.Key)
+				}
+			}
+		}
```
thedadams, line 77: What if I have a git catalog that has bindings, but I am running in Docker mode? Should I still be able to provide and configure such MCP servers?

pkg/controller/handlers/mcpsecretbinding/handler.go
```diff
@@ -0,0 +1,125 @@
...
+	// req.Namespace + req.Name still identify the affected Secret.
+	namespace := req.Namespace
+	name := req.Name
+	if secret, ok := req.Object.(*corev1.Secret); ok && secret != nil {
+		namespace = secret.Namespace
+		name = secret.Name
+	}
+	if namespace != h.obotNamespace || name == "" {
+		return nil
+	}
```
thedadams: Not needed.
```suggestion
```

pkg/controller/handlers/mcpsecretbinding/handler.go
```diff
@@ -0,0 +1,125 @@
...
+
+	for i := range servers.Items {
+		s := &servers.Items[i]
+		if !mcp.ManifestReferencesSecret(s.Spec.Manifest, name) {
+			continue
+		}
+
+		// secret is derived from req.Object and is constant across retries.
+		secret, _ := req.Object.(*corev1.Secret)
+		secretDeleted := secret == nil
```
thedadams:
```suggestion
		secret, secretDeleted := req.Object.(*corev1.Secret)
```

pkg/controller/handlers/mcpsecretbinding/handler_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+	secret := newSecret("my-secret", map[string]string{"tok": "v1"})
+	secret.Namespace = "other-ns"
+
+	err := h.SecretChanged(request(context.Background(), secret, "other-ns", "my-secret"), nil)
+	require.NoError(t, err)
+
+	s := getServer(t, storage, "srv")
+	assert.Empty(t, s.Annotations[rotationAnnotation])
+}
```
thedadams: The controller framework we use takes care of this. We don't need to test it here.

pkg/controller/handlers/mcpsecretbinding/handler.go
```diff
@@ -0,0 +1,125 @@
...
+		storage:       storage,
+		obotNamespace: obotNamespace,
+	}
+}
+
+// SecretChanged is the canonical handler. Attached to localK8sRouter
+// via &corev1.Secret{} with IncludeRemoved() — delete events MUST
+// reach this handler so we can flip referencing servers' missing
+// state.
+func (h *Handler) SecretChanged(req router.Request, _ router.Response) error {
```
thedadams: I made other comments in the file before I realized that I don't think this is needed.

The purpose of this seems to be to trigger something for the MCP servers that reference a secret when that secret changes? But for what purpose? I don't think there is a reason to do that.

So, I think the can be safely removed.

pkg/mcp/types.go
```diff
@@ -189,12 +189,23 @@ func configureRemoteRuntime(serverConfig *ServerConfig, remoteConfig *types.Remo
...
+
+		if header.Value != "" {
+			val = header.Value
+			hasValue = true
+		} else {
+			credVal, ok := credEnv[header.Key]
+			if ok && credVal != "" {
+				val = credVal
+				hasValue = true
+			}
```
thedadams: This doesn't seem to be a change in functionality.

pkg/mcp/types.go
```diff
@@ -398,15 +409,18 @@ func ServerToServerConfig(mcpServer v1.MCPServer, audiences []string, issuer, us
 		// Apply prefix if specified (e.g., "Bearer ", "sk-")
 		val = applyPrefix(val, env.Prefix)
 
-		if !env.File {
+		isFile := env.File || (env.SecretBinding != nil && env.SecretBinding.File)
+		isDynamic := env.DynamicFile || (env.SecretBinding != nil && env.SecretBinding.Dynamic)
```
thedadams: This is simplified by just using `env.File` and `env.DynamicFile` as I suggested above.

pkg/mcp/types.go
```diff
@@ -498,14 +512,13 @@ func SystemServerToServerConfig(systemServer v1.SystemMCPServer, audiences []str
 	// Process environment variables
 	for _, env := range systemServer.Spec.Manifest.Env {
 		var (
-			val      string
-			hasValue bool
+			val                   string
```
thedadams: Same comment: the change from here down don't seem to be a change in functionality.

pkg/services/config.go
```diff
@@ -751,6 +761,14 @@ func New(ctx context.Context, config Config) (*Services, error) {
 		return nil, err
 	}
 
+	var apiLocalK8sClient kclient.Client
+	if localK8sConfig != nil {
+		apiLocalK8sClient, err = kclient.New(localK8sConfig, kclient.Options{Scheme: scheme.Scheme})
```
thedadams: You don't want to use this scheme here. See https://github.com/obot-platform/obot/blob/main/pkg/controller/controller.go#L65.

Additionally, instead of create a client in the linked file, too. Replace the client there with the one you are creating here.

thedadams (changes requested):

pkg/mcp/secretbindings.go
njhale: we should prevent that from happening IMO.

pkg/mcp/types.go
calvinmclean: Addressed in 37af5dc15 as part of the same simplification pass. The non-functional reshaping here was removed by consolidating onto existing env semantics only, so there is no duplicate functional path now.

calvinmclean: I told Opencode to not reply to comments but that's what I get for allowing `gh api *`

pkg/controller/handlers/mcpsecretbinding/handler.go
calvinmclean: Thanks. I knocked out some of the easier comments and need to dig into this one a bit more. The original intention was to restart the server as soon as the secret changed, but since we removed that, I see how this might not be needed. I'll try it out, test the other changes more, and then re-request review once these are all resolved

pkg/api/handlers/mcp.go
calvinmclean: I removed serverForRead. The only downside here is that we don't get the server logs if a secret is removed after deploying a server. We can see the error about the missing secret though which should be sufficient to resolve the issue

docs/docs/configuration/mcp-server-gitops.md
```diff
@@ -112,6 +112,92 @@ env:
...
 
+### Kubernetes Secret Bindings
+
+Secret bindings let you wire an env var, header, or file to a key in an externally-managed Kubernetes Secret instead of asking the user to supply the value at install time.
+
+Secret bindings are only available on git-managed catalog entries, and only when Obot is using the Kubernetes MCP runtime backend.
+
+#### Basic env var binding
+
+The resolved value is injected into the nanobot pod as an environment variable, exactly like a user-supplied credential.
```
njhale: Is this the case for containerized runtimes as well?

thedadams (approved):

njhale (approved):
One doc nit, otherise backend lgtm

docs/docs/configuration/mcp-server-gitops.md
calvinmclean: Yes we can use the env vars, but it is not using the nanobot container. Thanks. I updated the doc

PR #6500, opened 2026-05-04
fix: retry on conflict for MCP server trigger-update

g-linville:
This fixes a 409 error that would sometimes happen when calling trigger-update on an MCPServer shortly after updating its MCPServerCatalogEntry.

### DIFF
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -3521,18 +3521,6 @@ func (m *MCPHandler) TriggerUpdate(req api.Context) error {
 			return err
 		}
 
-		if latest.Spec.MCPCatalogID != "" || latest.Spec.PowerUserWorkspaceID != "" {
-			return types.NewErrBadRequest("cannot trigger update for a multi-user MCP server; use the UpdateServer endpoint instead")
-		}
-
-		if latest.Spec.CompositeName != "" {
-			return types.NewErrBadRequest("cannot trigger update on a component server; upgrade the parent composite server instead")
-		}
-
-		if latest.Spec.MCPServerCatalogEntryName != entry.Name {
-			return types.NewErrHTTP(http.StatusConflict, "catalog entry changed during update")
-		}
-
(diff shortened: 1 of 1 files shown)


njhale (approved):
nice

pkg/api/handlers/mcp.go
```diff
@@ -3506,9 +3506,48 @@ func (m *MCPHandler) TriggerUpdate(req api.Context) error {
...
+		}
+
+		if latest.Spec.MCPCatalogID != "" || latest.Spec.PowerUserWorkspaceID != "" {
+			return types.NewErrBadRequest("cannot trigger update for a multi-user MCP server; use the UpdateServer endpoint instead")
+		}
 
-	// Update the server manifest with the latest from the catalog entry
+		if latest.Spec.CompositeName != "" {
+			return types.NewErrBadRequest("cannot trigger update on a component server; upgrade the parent composite server instead")
+		}
```
thedadams: This was already checked above. Does it need to be checked here too?

pkg/api/handlers/mcp.go
```diff
@@ -3506,9 +3506,48 @@ func (m *MCPHandler) TriggerUpdate(req api.Context) error {
...
+
+		if latest.Spec.MCPServerCatalogEntryName != entry.Name {
+			return types.NewErrHTTP(http.StatusConflict, "catalog entry changed during update")
+		}
```
thedadams: Additionally, can we move this check out of the retry?

pkg/api/handlers/mcp.go
g-linville: Yeah you're right; this is unnecessary. Will remove

pkg/api/handlers/mcp.go
g-linville: Yep

pkg/api/handlers/mcp.go

pkg/api/handlers/mcp.go

thedadams (approved):

Issue #6506, opened 2026-05-05
Unable to add LiteLLM Model Endpoints as Model Provider


skuchipu:
I dont see an option to use LiteLLM Model endpoint in the model providers section , I tried adding this as part of OPENAI_BASE_URL and OPENAI_API_KEY that does not help , is there a way we can use models configured in LiteLLM that can be used ?

calvinmclean:
Tested with the following LiteLLM config during development:
```yaml
model_list:
  - model_name: openai/*
    litellm_params:
      model: openai/*
      api_key: os.environ/OPENAI_API_KEY

# model_list:
#   - model_name: gpt-4o
#     litellm_params:
#       model: gpt-4o
#       api_key: os.environ/OPENAI_API_KEY
#   - model_name: gpt-4o-mini
#     litellm_params:
#       model: gpt-4o-mini
#       api_key: os.environ/OPENAI_API_KEY

general_settings:
  master_key: "[REDACTED]"

environment_variables:
  OPENAI_API_KEY: "[REDACTED]"
```

This uses a new "Generic Responses Model Provider" which is compatible with LiteLLM, Ollama, and others.

sangee2004:
We now have a ` Responses Compatible Generic Provider` model provider available in the `Model Providers` list.

Able to configure both LiteLLM and Ollama with this provider successfully.

<img width="770" height="579" alt="Image" src="https://github.com/user-attachments/assets/4b791142-7fb9-46d2-8284-088cb357c857" />

sangee2004:
@skuchipu `Responses Compatible Generic Provider` which allows for configuring the LiteLLM model endpoint will be available in the next obot release.

Issue #6410, opened 2026-04-23
Error not shown on first invalid api key submission in model provider setup.


sangee2004:
**Describe the bug**
When configuring a model provider with an invalid API key, the initial submission does not surface any error in the UI. Although the backend correctly logs a 401 "invalid credentials" error, the frontend silently reloads the setup page without any feedback. The error message appears only after the user clicks “Confirm” a second time, creating a confusing and inconsistent user experience.

**To Reproduce**
1. As an owner or admin user, go to the `Model Providers` page
2. Configure `OpenAI` or `Anthropic` model provider for the first time by providing an invalid Api key.

Notice that once the user clicks on `Confirm`, the user is presnted wth the same `Set Up OpenAI` page with no error message.

<img width="486" height="288" alt="Image" src="https://github.com/user-attachments/assets/3cadd539-5dbc-45ea-91d8-840924260b87" />

At this time, I already see the following error message in obot server logs - `time="2026-04-23T17:41:11Z" level=error msg="failed to save state: ERROR: 2026/04/23 17:41:11 ERROR Invalid OpenAI Credentials logger=/tools/openai-model-provider/validate error=\"status 401 Unauthorized: {\\n  \\\"error\\\": {\\n    \\\"message\\\": \\\"Incorrect API key provided: ewrwertwert. You can find your API key at https://platform.openai.com/account/api-keys.\\\",\\n    \\\"type\\\": \\\"invalid_request_error\\\",\\n    \\\"param\\\": null,\\n    \\\"code\\\": \\\"invalid_api_key\\\"\\n  }\\n}\"\n: exit status 1" logger=/app/pkg/invoke/invoker.go` 

obot server logs:

```
time="2026-04-23T17:41:11Z" level=info msg="Created run resource: run=r1dr8xx thread=t1-openai-model-provider-validate-x7qf8 agent= workflow= synchronous=true previousRun=" logger=/app/pkg/invoke/invoker.go
time="2026-04-23T17:41:11Z" level=info msg="Updated thread workspace references: thread=t1-openai-model-provider-validate-x7qf8 workspace=wksp1gbbr9 workspaceID=" logger=/app/pkg/controller/handlers/threads/threads.go
[...]

PR #6522, opened 2026-05-06
feat: Replace Ollama with generic provider

calvinmclean:
Depends on: https://github.com/obot-platform/tools/pull/802

Addresses: #6506 #6410 

- Add new Generic Responses Provider to UI
- Since LiteLLM often names models `openai/*`, `anthropic/*`, etc., I added a separate Name which is a slugified version of the TargetModel

### DIFF
--- a/pkg/api/handlers/model.go
+++ b/pkg/api/handlers/model.go
@@ -246,11 +246,14 @@ func convertModel(model v1.Model, toolRef v1.ToolReference) (types.Model, error)
 
 func validateModelManifestAndSetDefaults(newModel *v1.Model) error {
 	var errs []error
-	if strings.TrimSpace(newModel.Spec.Manifest.Name) == "" {
+	newModel.Spec.Manifest.Name = strings.TrimSpace(newModel.Spec.Manifest.Name)
+	if newModel.Spec.Manifest.Name == "" {
 		newModel.Spec.Manifest.Name = strings.ReplaceAll(strings.TrimSpace(newModel.Spec.Manifest.TargetModel), "/", "-")
 	}
-	if strings.TrimSpace(newModel.Spec.Manifest.Name) == "" {
+	if newModel.Spec.Manifest.Name == "" {
 		errs = append(errs, fmt.Errorf("field name is required"))
+	} else if strings.Contains(newModel.Spec.Manifest.Name, "/") {
+		errs = append(errs, fmt.Errorf("field name must be a single path segment and must not contain '/'"))
 	}
 	if strings.TrimSpace(newModel.Spec.Manifest.TargetModel) == "" {
 		errs = append(errs, fmt.Errorf("field targetModel is required"))
--- a/pkg/api/handlers/model.go
+++ b/pkg/api/handlers/model.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/obot-platform/obot/apiclient/types"
 	"github.com/obot-platform/obot/pkg/api"
@@ -245,10 +246,16 @@ func convertModel(model v1.Model, toolRef v1.ToolReference) (types.Model, error)
 
 func validateModelManifestAndSetDefaults(newModel *v1.Model) error {
 	var errs []error
-	if newModel.Spec.Manifest.TargetModel == "" {
+	if strings.TrimSpace(newModel.Spec.Manifest.Name) == "" {
+		newModel.Spec.Manifest.Name = strings.ReplaceAll(strings.TrimSpace(newModel.Spec.Manifest.TargetModel), "/", "-")
+	}
+	if strings.TrimSpace(newModel.Spec.Manifest.Name) == "" {
+		errs = append(errs, fmt.Errorf("field name is required"))
+	}
+	if strings.TrimSpace(newModel.Spec.Manifest.TargetModel) == "" {
 		errs = append(errs, fmt.Errorf("field targetModel is required"))
 	}
-	if newModel.Spec.Manifest.ModelProvider == "" {
+	if strings.TrimSpace(newModel.Spec.Manifest.ModelProvider) == "" {
 		errs = append(errs, fmt.Errorf("field modelProvider is required"))
 	}
 
--- a/pkg/api/handlers/model.go
+++ b/pkg/api/handlers/model.go
@@ -250,9 +250,7 @@ func validateModelManifestAndSetDefaults(newModel *v1.Model) error {
 	if newModel.Spec.Manifest.Name == "" {
 		newModel.Spec.Manifest.Name = strings.ReplaceAll(strings.TrimSpace(newModel.Spec.Manifest.TargetModel), "/", "-")
 	}
-	if newModel.Spec.Manifest.Name == "" {
-		errs = append(errs, fmt.Errorf("field name is required"))
-	} else if strings.Contains(newModel.Spec.Manifest.Name, "/") {
+	if strings.Contains(newModel.Spec.Manifest.Name, "/") {
 		errs = append(errs, fmt.Errorf("field name must be a single path segment and must not contain '/'"))
 	}
 	if strings.TrimSpace(newModel.Spec.Manifest.TargetModel) == "" {
--- a/pkg/controller/handlers/nanobotagent/nanobotagent.go
+++ b/pkg/controller/handlers/nanobotagent/nanobotagent.go
@@ -493,10 +493,6 @@ func getModelForAlias(ctx context.Context, client kclient.Client, namespace stri
 		return resolvedLLMModel{}, err
 	}
 
-	if strings.TrimSpace(model.Spec.Manifest.TargetModel) == "" {
-		return resolvedLLMModel{}, fmt.Errorf("model %q has no target model", model.Name)
-	}
-
 	return resolvedLLMModel{
 		Name:            model.Name,
 		ModelProvider:   model.Spec.Manifest.ModelProvider,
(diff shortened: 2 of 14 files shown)


pkg/controller/handlers/nanobotagent/nanobotagent.go
```diff
@@ -453,7 +454,7 @@ func (h *Handler) parseModelProvider(model resolvedLLMModel) (nanobotLLMProvider
 		APIKey:  fmt.Sprintf("${%s}", envVarName),
 		BaseURL: baseURL,
 	}
-	return p, fmt.Sprintf("%s/%s", p.Name, model.TargetModel)
+	return p, fmt.Sprintf("%s/%s", p.Name, model.Name)
```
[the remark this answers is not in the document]
calvinmclean: Addressed by setting a default for the model.Name in validateModelManifestAndSetDefaults

pkg/api/handlers/model.go
```diff
@@ -245,10 +246,16 @@ func convertModel(model v1.Model, toolRef v1.ToolReference) (types.Model, error)
...
+	}
+	if strings.TrimSpace(newModel.Spec.Manifest.Name) == "" {
+		errs = append(errs, fmt.Errorf("field name is required"))
+	}
+	if strings.TrimSpace(newModel.Spec.Manifest.TargetModel) == "" {
 		errs = append(errs, fmt.Errorf("field targetModel is required"))
 	}
-	if newModel.Spec.Manifest.ModelProvider == "" {
+	if strings.TrimSpace(newModel.Spec.Manifest.ModelProvider) == "" {
 		errs = append(errs, fmt.Errorf("field modelProvider is required"))
```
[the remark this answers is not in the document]
calvinmclean: This function should only do the things it is named for: validate and set defaults.

pkg/controller/handlers/toolreference/toolreference.go
```diff
@@ -472,8 +476,8 @@ func (h *Handler) BackPopulateModels(req router.Request, _ router.Response) erro
 			},
 			Spec: v1.ModelSpec{
 				Manifest: types.ModelManifest{
-					Name:          model.ID,
-					DisplayName:   model.Metadata["displayName"],
+					Name:          strings.ReplaceAll(model.ID, "/", "-"),
+					DisplayName:   displayName,
 					TargetModel:   model.ID,
 					ModelProvider: toolRef.Name,
 					Active:        true,
```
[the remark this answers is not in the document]
calvinmclean: Migrating should not be necessary because any pre-existing models will have valid TargetModel. I tested this by running on main, then changing to this branch and adding a new model provider with `openai/*` model names. It worked fine.

pkg/api/handlers/model.go
```diff
@@ -245,10 +246,19 @@ func convertModel(model v1.Model, toolRef v1.ToolReference) (types.Model, error)
...
+		errs = append(errs, fmt.Errorf("field name is required"))
+	} else if strings.Contains(newModel.Spec.Manifest.Name, "/") {
+		errs = append(errs, fmt.Errorf("field name must be a single path segment and must not contain '/'"))
+	}
+	if strings.TrimSpace(newModel.Spec.Manifest.TargetModel) == "" {
 		errs = append(errs, fmt.Errorf("field targetModel is required"))
 	}
-	if newModel.Spec.Manifest.ModelProvider == "" {
+	if strings.TrimSpace(newModel.Spec.Manifest.ModelProvider) == "" {
 		errs = append(errs, fmt.Errorf("field modelProvider is required"))
```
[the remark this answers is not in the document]
calvinmclean: This is intentional since the function is validating and setting defaults

pkg/api/handlers/model.go
```diff
@@ -245,10 +246,19 @@ func convertModel(model v1.Model, toolRef v1.ToolReference) (types.Model, error)
 
 func validateModelManifestAndSetDefaults(newModel *v1.Model) error {
 	var errs []error
-	if newModel.Spec.Manifest.TargetModel == "" {
+	newModel.Spec.Manifest.Name = strings.TrimSpace(newModel.Spec.Manifest.Name)
+	if newModel.Spec.Manifest.Name == "" {
+		newModel.Spec.Manifest.Name = strings.ReplaceAll(strings.TrimSpace(newModel.Spec.Manifest.TargetModel), "/", "-")
+	}
+	if newModel.Spec.Manifest.Name == "" {
```
thedadams: nit: not possible based on other checks.

pkg/gateway/server/llmproxy.go
```diff
@@ -219,24 +220,32 @@ func (s *Server) dispatchLLMProxy(req api.Context) error {
...
+// - The actual Kubernetes resource name of the model
 func getModelFromReference(ctx context.Context, client kclient.Client, namespace, modelReference string) (*v1.Model, error) {
 	m, err := alias.GetFromScope(ctx, client, "Model", namespace, modelReference)
 	if apierrors.IsNotFound(err) {
-		// Maybe the user is trying to get a model by the target name.
 		var models v1.ModelList
 		if err := client.List(ctx, &models, &kclient.ListOptions{
 			Namespace:     namespace,
-			FieldSelector: fields.OneTermEqualSelector("spec.manifest.targetModel", modelReference),
+			FieldSelector: fields.OneTermEqualSelector("spec.manifest.name", modelReference),
```
thedadams: I don't quite understand the intention behind this change. Prior to this change, it was possible to use a model by its name (i.e. `model.Name`) or by its target model.

Why do we need to reference models by their display name?

calvinmclean: This should still be possible. Now it just checks Name first and falls back to the TargetModel (previously these were the same, set in `toolreference.go`).

In a LiteLLM config, users can configure specific model names one-by-one, or use a wildcard format like `openai/*` to include all OpenAI models. This was not compatible with the ModelManifist.Name because of the `/`. The changes here allow users to use LiteLLM without needing a Obot-specific configuration. Name and TargetModel in the ModelManifest were previously set to the same thing and now the Name is our internal reference (this is why it's the FieldSelector in the first List request, e.g. `openai-gpt-4.1`) and the TargetModel is what the provider knows it as (`openai/gpt-4.1`).

thedadams: I am going to make a comment in a different place in the code to see if we can simplify this (and remove the target model stuff as well).

pkg/controller/handlers/nanobotagent/nanobotagent.go
thedadams, line 456: Instead of adding a new way of determining which model the user should use, I would like to go back to just one way.

We needed `fmt.Sprintf("%s/%s", p.Name, model.TargetModel)` (and just `model.TargetModel` prior to your dialect changes) because of the way that nanobot determined when to use the anthropic client. With your dialect changes, I think it's possible to just the `v1.Model` objects `Name` instead of the target model. Then we can simply always pass the model's name and we don't have to add extra functionality just for LiteLLM.

calvinmclean: When you specify `v1.Model`'s Name, is that specifically `v1.Model.ObjectMeta.Name` (like `m1-openai-model-provider-afeeb199907bc33b720dcc589c6de1e0-2fac1`, `m1-generic-responses-model-provider-fee534d110cd2fe7bf1c9-93620`)?

thedadams: Yes, you don't need the `ObjectMeta` part because it's embedded, but that's right.

calvinmclean: Sounds good. I will see what that looks like.

Also, can we clarify the problem we are trying to solve? You mentioned we want just 1 way of determining the model. I could just remove the 2nd `List` request in `llmproxy.go` because the first one should not miss. I added it for extra caution in backwards-compatibility and because the original function comment specified a few different ways to find the model so I wanted to keep a few options there. It is not a "use this search for LiteLLM, and this for all others" situation. Regardless, I think this change to use the v1.Model Name is good and more specific anyways

thedadams: With the change I am proposing, we can get rid of all the lists in the LLM proxy because we will just have the name.

pkg/gateway/server/llmproxy.go
```diff
@@ -989,33 +975,31 @@ func (l *llmProviderProxy) proxy(req api.Context) error {
...
-		for _, model := range models.Items {
-			var err error
-			hasAccess, err = l.mapHelper.UserHasAccessToModel(req.User, model.Name)
-			if err != nil {
-				return fmt.Errorf("failed to check user access to model %q: %w", model.Name, err)
-			}
-			if hasAccess {
-				break
-			}
+		if model.Spec.Manifest.ModelProvider != l.modelProvider.Name {
```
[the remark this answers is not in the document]
calvinmclean: This is not correct. The current implementation works for both OpenAI/Anthropic providers and other providers. Nanobot just sends the model after the `/`

pkg/gateway/server/llmproxy.go
```diff
@@ -989,33 +985,31 @@ func (l *llmProviderProxy) proxy(req api.Context) error {
...
+			return types2.NewErrForbidden("user does not have permission to use model %q", targetModel)
```
thedadams: nit: probably should be a Bad Request error.

pkg/gateway/server/llmproxy.go
```diff
@@ -320,6 +297,25 @@ func extractModelFromBody(body []byte) string {
 	return gjson.GetBytes(body, "response.model").String()
 }
 
+func rewriteModelInBody(body []byte, model string) ([]byte, error) {
```
thedadams, line 300: If I understand correctly, this function is just used in the request path. In that case, only the first case in the if block will be hit. The instances where  there might be a `message` or `response` object are only in the responses to these requests.

calvinmclean: Yes this is only in the request path. Also the requests to this endpoint should be using a consistent request body anyways. Fixed

g-linville (approved):
backend lgtm

thedadams (approved):

PR #6539, opened 2026-05-07
fix: allow access to mcp-connect URLs with extra path

thedadams:
The nanobot agents need to access extra paths for their MCP calls.

### DIFF
--- a/pkg/api/authz/mcpid_test.go
+++ b/pkg/api/authz/mcpid_test.go
@@ -1,9 +1,11 @@
 package authz
 
 import (
+	"net/http"
 	"net/http/httptest"
 	"testing"
 
+	"github.com/obot-platform/obot/apiclient/types"
 	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
 	storagescheme "github.com/obot-platform/obot/pkg/storage/scheme"
 	"github.com/obot-platform/obot/pkg/system"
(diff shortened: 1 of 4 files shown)


pkg/api/authz/resources.go
```diff
@@ -225,7 +228,6 @@ var apiResources = map[string][]string{
 		//
 		"GET    /api/tool-references",
 		"GET    /api/tool-references/{id}",
-		"GET    /{ui}/projects/{id}",
```
thedadams, line 228: This wasn't actually doing anything so I removed it. It conflicted with one of the paths I am adding here.

pkg/api/authz/mcpid_test.go
```diff
@@ -63,3 +65,99 @@ func TestCheckMCPIDChecksMCPServerInstanceOwner(t *testing.T) {
...
+		},
+		{
+			name:   "api key can access subpath",
+			method: http.MethodGet,
+			path:   "/mcp-connect/msi1test/messages/123",
+			user: &user.DefaultInfo{
+				Name:   "key-user",
+				Groups: []string{types.GroupAPIKey},
+			},
+			allowed: true,
```
g-linville: How does our test authorizer recognize that `key-user` belongs to user with ID `user-uid`?

g-linville: (Just trying to make sure that there isn't some issue that would allow any API key to access any mcp-connect path.)

thedadams: As I understand it, this is done in the token review call. The request makes it to the shim, the shim sees that this is an API key, makes a token review request, and gets denied.

Do I have that right?

thedadams: Nope. That's wrong.

g-linville: I believe that is correct, yes, but I am not sure how this would come out to `true` here in this test setup. Wouldn't this API key have to be pre-registered with our fake client?

g-linville (approved):

PR #6546, opened 2026-05-08
docs: overview: provide more information about authentication

g-linville:


njhale:
Did you mean to update the docs for older versions as well? 

Eh, it's fine since I'm pretty sure the config is the same.

njhale (approved):

g-linville:
yeah I intentionally updated all the versions

cjellick (approved):

PR #6552, opened 2026-05-08
fix: display a better error message for multi-user MCP servers

thedadams:
If a user tries to connect to an MCP server that has required user-defined headers, this change will be the error message better.

Issue: https://github.com/obot-platform/obot/issues/6553

### DIFF
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -1049,6 +1049,15 @@ func mcpServerOrInstanceFromConnectURL(req api.Context, id string) (v1.MCPServer
 				return v1.MCPServer{}, v1.MCPServerInstance{}, err
 			}
 			if len(instances.Items) == 0 {
+				// Check that the multi-user server doesn't have any required configuration.
+				if server.Spec.Manifest.MultiUserConfig != nil {
+					for _, header := range server.Spec.Manifest.MultiUserConfig.UserDefinedHeaders {
+						if header.Required {
+							return v1.MCPServer{}, v1.MCPServerInstance{}, types.NewErrNotFound("user has not configured an instance for the multi-user MCP server %s", id)
+						}
+					}
+				}
+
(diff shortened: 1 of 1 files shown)


pkg/api/handlers/mcp.go
```diff
@@ -1049,6 +1049,15 @@ func mcpServerOrInstanceFromConnectURL(req api.Context, id string) (v1.MCPServer
 				return v1.MCPServer{}, v1.MCPServerInstance{}, err
 			}
 			if len(instances.Items) == 0 {
+				// Check that the multi-user server doesn't have any required configuration.
+				if server.Spec.Manifest.MultiUserConfig != nil {
+					for _, header := range server.Spec.Manifest.MultiUserConfig.UserDefinedHeaders {
+						if header.Required {
+							return v1.MCPServer{}, v1.MCPServerInstance{}, types.NewErrNotFound("user has not configured an instance for the multi-user MCP server %s", id)
```
calvinmclean, line 1056: Do you think it would be useful to aggregate all required headers to put in a single error? I'm not sure what the UX for this looks like, but this might save the user some time if multiple are missing.

thedadams: This error occurs when a user is trying to connect to an MCP server from an outside client (like VS Code or Cursor) that they haven't setup in Obot yet. If there is no required configuration, we create an instance for them on the fly. If there is, we display an error and they have to go set it up in Obot.

With that flow, I don't think telling them what the missing configuration is is helpful. They still have to go to Obot and configure it.

calvinmclean: > If there is no required configuration, we create an instance for them on the fly

That's cool.

Makes sense. When they open it up in Obot, it should be clear what is required.

calvinmclean (approved):

g-linville (approved):

PR #6560, opened 2026-05-08
fix: correct backgrounds for scanned config files

njhale:
In light mode, file backgrounds were washed out. This corrects the backgrounds to make the text visible.


ivyjeong13 (approved):
Pulled, checked, lgtm!

PR #6567, opened 2026-05-11
feat: manage image pull secrets in Obot

g-linville:
This is a new feature that allows admins to manage image pull secrets in the Obot UI (as long as none are set through the environment variable).

We support two types of secrets: Basic (for most use cases like Docker Hub and GHCR) and ECR.

For ECR secrets, we set up a controller handler that assumes a role and fetches an ECR token every six hours in order to keep the secret up to date and working.

See the documentation in this PR for more details.

There are also some changes to Kubernetes config drift detection. This fixes a bug that could appear if there were other controllers in the cluster that would automatically attach resource reqs/limits (like GKE Autopilot's `ephemeral-storage`) or add tolerations (which GKE Autopilot also does). I ran into this during my testing.

Demo videos (based on outdated CronJob architecture, but same UX):

1. [Overview](https://acorn-io.slack.com/archives/C03FU91U4SX/p1778352988545679)
2. [UI enhancements](https://acorn-io.slack.com/archives/C03FU91U4SX/p1778518437819399)


pkg/api/handlers/imagepullsecrets.go
```diff
@@ -0,0 +1,775 @@
...
+		result.IssuerURL = firstNonEmpty(result.IssuerURL, secret.Spec.ECR.IssuerURL, h.issuerURL)
+		result.Audience = firstNonEmpty(result.Audience, secret.Spec.ECR.Audience)
+		result.RefreshSchedule = secret.Spec.ECR.RefreshSchedule
+		if result.Audience == "" {
+			result.Audience = imagepullsecrets.DefaultECRAudience
+		}
+		if result.Subject == "" {
+			result.Subject = imagepullsecrets.ECRSubject(h.helperNamespace)
+		}
+		result.CronJobName = firstNonEmpty(result.CronJobName, imagepullsecrets.DefaultECRCronJobName(secret.Name))
```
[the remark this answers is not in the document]
g-linville: already fixed

ui/user/src/routes/admin/image-pull-secrets/ECRSetupGuide.svelte
```diff
@@ -0,0 +1,106 @@
...
+		ecrPolicyJSON: string;
+	}
+
+	let { effectiveIssuerURL, effectiveAudience, trustPolicyJSON, ecrPolicyJSON }: Props = $props();
+</script>
+
+<section class="paper gap-5">
+	<div class="flex flex-col gap-1">
+		<h3 class="text-on-surface1 text-base font-semibold">AWS Setup Guide</h3>
+		<p class="text-muted text-sm">
```
ivyjeong13: `-muted` isn't a supported var atm  in case this is supposed to be greyed out text

ui/user/src/routes/admin/image-pull-secrets/ImagePullSecretStatusDialog.svelte
```diff
@@ -0,0 +1,263 @@
...
+			{/if}
+		</div>
+	</div>
+{/snippet}
+
+{#snippet conditionList(title: string, conditions: ImagePullSecretRuntimeCondition[])}
+	<section class="flex flex-col gap-3">
+		{@render sectionHeader(title)}
+		<div
+			class="dark:bg-background dark:border-surface3 bg-surface1 flex flex-col divide-y divide-gray-200 overflow-hidden rounded-lg border border-transparent shadow-sm dark:divide-gray-800"
```
ivyjeong13: nit: instead divide-gray-{x}, set to surface color (ex. matching surface color applied to border)

apiclient/types/imagepullsecret.go
```diff
@@ -0,0 +1,130 @@
...
+
+const (
+	ImagePullSecretPhasePending  ImagePullSecretPhase = "Pending"
+	ImagePullSecretPhaseReady    ImagePullSecretPhase = "Ready"
+	ImagePullSecretPhaseError    ImagePullSecretPhase = "Error"
+	ImagePullSecretPhaseDisabled ImagePullSecretPhase = "Disabled"
+)
+
+// ImagePullSecret represents an admin-managed Kubernetes image pull secret.
+type ImagePullSecret struct {
```
thedadams, line 11: Can we have an embedded manifest type here so it is easily seen which fields are set by the user?

Additionally, I like the data model where type-specific configuration is put into its own struct which is referenced here as a pointer. It makes it easy to see which configuration is for what and when things are set properly.

ui/user/src/routes/admin/image-pull-secrets/ImagePullSecretForm.svelte
```diff
@@ -0,0 +1,342 @@
...
+		<form
+			class="paper"
+			onsubmit={(e) => {
+				e.preventDefault();
+				onSave();
+			}}
+		>
+			<div class="flex flex-col gap-4">
+				<div class="flex flex-col gap-1">
+					<FieldLabel
```
ivyjeong13, line 99: nit: for consistency with the other forms:
* drop the (required) next to fields that are required, and if the user clicks Save, it highlights the ones that need to be filled out

ex: 
Image

ui/user/src/routes/admin/image-pull-secrets/ImagePullSecretTestDialog.svelte
```diff
@@ -0,0 +1,104 @@
...
+</script>
+
+<ResponsiveDialog
+	bind:this={dialog}
+	title={`Test ${secret ? displayName(secret) : 'Image Pull Secret'}`}
+	class="w-full md:max-w-xl"
+	{onClose}
+>
+	<form
+		class="m-4 flex flex-col gap-4"
```
ivyjeong13: reallllyy nit: Have the form label/field line up with the dialog title (might just be the `m-4` here!)
Image

chart/templates/mcp.yaml
```diff
@@ -97,14 +97,17 @@ rules:
...
     verbs: ["get", "list", "watch"]
   - apiGroups: ["apps"]
     resources: ["deployments"]
     verbs: ["create", "get", "list", "watch", "update", "patch", "delete"]
   - apiGroups: ["apps"]
     resources: ["replicasets"]
     verbs: ["get", "list", "watch"]
+  - apiGroups: ["batch"]
+    resources: ["cronjobs", "jobs"]
+    verbs: ["create", "get", "list", "watch", "update", "patch", "delete"]
```
thedadams: We have a controller framework. Do we really need to manage cron jobs too?

ui/user/src/routes/admin/image-pull-secrets/+page.svelte
```diff
@@ -0,0 +1,339 @@
...
+			const response = await AdminService.refreshImagePullSecret(secret.id);
+			refreshMessage =
+				response.message || (response.jobName ? `Started ${response.jobName}` : 'Refresh started');
+			await refreshList();
+		} finally {
+			refreshing = false;
+		}
+	}
+
+	function createNewSecret() {
```
ivyjeong13, line 243: May want to add/use the `?create=true` url search param for showing the create form. I think with just this, hitting the browser's back button won't return you to the table of image pull secrets but hitting the back button from `Layout` component will.

pkg/api/handlers/imagepullsecrets.go
```diff
@@ -0,0 +1,812 @@
...
+
+	if !secret.Spec.Enabled {
+		return types.NewErrBadRequest("image pull secret is disabled")
+	}
+	if h.runtimeClient == nil {
+		return types.NewErrHTTP(http.StatusServiceUnavailable, "Kubernetes runtime client is not configured")
+	}
+	if strings.TrimSpace(h.mcpNamespace) == "" {
+		return types.NewErrHTTP(http.StatusServiceUnavailable, "MCP namespace is not configured")
+	}
```
thedadams: nit: not possible based on defaults.

pkg/api/handlers/imagepullsecrets.go
```diff
@@ -0,0 +1,812 @@
...
+
+	if strings.TrimSpace(h.helperNamespace) == "" {
+		return types.NewErrHTTP(http.StatusServiceUnavailable, "ECR refresh helper namespace is not configured")
+	}
+
+	jobName, err := imagepullsecrets.CreateECRRefreshJob(req.Context(), h.runtimeClient, h.helperNamespace, secret.Name)
+	if err != nil {
+		if errors.Is(err, imagepullsecrets.ErrECRRefreshResourcesNotReady) {
+			return types.NewErrHTTP(http.StatusConflict, "ECR refresh resources are not ready yet")
```
thedadams:
```suggestion
			return types.NewErrHTTP(http.StatusTooEarly, "ECR refresh resources are not ready yet")
```

pkg/api/handlers/imagepullsecrets.go
```diff
@@ -0,0 +1,812 @@
...
+		return imagepullsecrets.RegistryTestResult{}, types.NewErrHTTP(http.StatusServiceUnavailable, "MCP namespace is not configured")
+	}
+
+	var kubeSecret corev1.Secret
+	if err := h.runtimeClient.Get(req.Context(), kclient.ObjectKey{
+		Namespace: h.mcpNamespace,
+		Name:      secret.Spec.SecretName,
+	}, &kubeSecret); err != nil {
+		if apierrors.IsNotFound(err) {
+			return imagepullsecrets.RegistryTestResult{}, types.NewErrBadRequest("generated Kubernetes image pull secret is not ready")
```
thedadams: Similar error message to the above, but a different status code. I think we should be consistent.

pkg/api/handlers/imagepullsecrets.go
```diff
@@ -0,0 +1,812 @@
...
+			Message:            condition.Message,
+		})
+	}
+
+	if secret.Spec.Basic != nil {
+		result.Server = secret.Spec.Basic.Server
+		result.Username = secret.Spec.Basic.Username
+
+		var err error
+		result.PasswordConfigured, err = passwordConfigured(req, secret.Name)
```
thedadams: I've received pushback on doing this type of thing in the convert methods. Can this be moved out?

pkg/api/handlers/imagepullsecrets.go
```diff
@@ -0,0 +1,812 @@
...
+			return types.ImagePullSecret{}, fmt.Errorf("failed to check password configured: %w", err)
+		}
+	}
+
+	if secret.Spec.ECR != nil {
+		result.RoleARN = secret.Spec.ECR.RoleARN
+		result.Region = secret.Spec.ECR.Region
+		result.IssuerURL = firstNonEmpty(result.IssuerURL, secret.Spec.ECR.IssuerURL, h.issuerURL)
+		result.Audience = firstNonEmpty(result.Audience, secret.Spec.ECR.Audience)
+		result.RefreshSchedule = secret.Spec.ECR.RefreshSchedule
```
thedadams: Having separate structs for each type of configuration would also make this easier on the eyes.

ui/user/src/routes/admin/image-pull-secrets/ECRSetupGuide.svelte
g-linville: Fixed

ui/user/src/routes/admin/image-pull-secrets/ImagePullSecretStatusDialog.svelte
g-linville: Fixed

ui/user/src/routes/admin/image-pull-secrets/ImagePullSecretForm.svelte
g-linville: Fixed

g-linville: Here's what it looks like now:

image

ui/user/src/routes/admin/image-pull-secrets/ImagePullSecretTestDialog.svelte
g-linville: Fixed:

image

ui/user/src/routes/admin/image-pull-secrets/+page.svelte
g-linville: Fixed

ui/user/src/routes/admin/image-pull-secrets/ImagePullSecretForm.svelte
ivyjeong13: Thank you! : D

ivyjeong13 (approved):
UI changes lgtm

pkg/api/handlers/mcp.go
```diff
@@ -3221,11 +3231,13 @@ func (m *MCPHandler) CheckK8sSettingsStatus(req api.Context) error {
...
-	// Compute current K8s settings hash
-	currentHash := mcp.ComputeK8sSettingsHash(k8sSettings.Spec)
+	currentHash, err := m.currentK8sSettingsHash(req, k8sSettings.Spec)
+	if err != nil {
+		return err
+	}
 
 	// Compare deployed hash with current hash
-	needsUpdate := deployedHash != currentHash
+	needsUpdate := server.Status.NeedsK8sUpdate || deployedHash != currentHash
```
thedadams: I assume that the controller sets `server.Status.NeedsK8sUpdate` by comparing hashes. Do we need to compare hashes here, too?

There are a couple other places in this file where you've added this.

pkg/controller/handlers/imagepullsecret/imagepullsecret.go
```diff
@@ -0,0 +1,926 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	kclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const conditionReady = "Ready"
+
+const annotationECRConfigHash = "obot.ai/ecr-config-hash"
```
thedadams: nit:
```suggestion
const (
    conditionReady          = "Ready"
    annotationECRConfigHash = "obot.ai/ecr-config-hash"
)
```

pkg/controller/handlers/imagepullsecret/imagepullsecret.go
```diff
@@ -0,0 +1,926 @@
...
+	}
+}
+
+func (h *Handler) Reconcile(req router.Request, _ router.Response) error {
+	secret := req.Object.(*v1.ImagePullSecret)
+	if err := h.reconcile(req, secret); err != nil {
+		if statusErr := h.setError(req, secret, err); statusErr != nil {
+			return errors.Join(err, statusErr)
+		}
+		return err
```
thedadams: Not sure if this is better, but `errors.Join` will do the right thing here.
```suggestion
		return errors.Join(err, h.setError(req, secret, err))
```

pkg/controller/handlers/imagepullsecret/imagepullsecret.go
```diff
@@ -0,0 +1,926 @@
...
+	return nil
+}
+
+func (h *Handler) Cleanup(req router.Request, _ router.Response) error {
+	secret := req.Object.(*v1.ImagePullSecret)
+
+	if err := h.cleanupRuntimeResources(req.Ctx, secret, true); err != nil {
+		return err
+	}
+	if h.gptClient != nil {
```
thedadams: Probably don't need this check.

pkg/controller/handlers/imagepullsecret/imagepullsecret.go
```diff
@@ -0,0 +1,926 @@
...
+		if err := h.cleanupRuntimeResources(req.Ctx, secret, false); err != nil {
+			return err
+		}
+		return h.setDisabledStatus(req, secret, "Disabled", "image pull secret is disabled")
+	}
+
+	if h.runtimeClient == nil {
+		return errors.New("runtime Kubernetes client is not configured")
+	}
+	if strings.TrimSpace(h.mcpNamespace) == "" {
```
thedadams: Not needed because of defaulting.

pkg/controller/handlers/imagepullsecret/imagepullsecret.go
```diff
@@ -0,0 +1,926 @@
...
+		}
+	}
+
+	return h.setECRStatus(req, secret)
+}
+
+func (h *Handler) revealPassword(ctx context.Context, name string) (string, error) {
+	if h.gptClient == nil {
+		return "", errors.New("GPTScript client is not configured")
+	}
```
thedadams:
```suggestion
```

pkg/controller/handlers/imagepullsecret/imagepullsecret.go
```diff
@@ -0,0 +1,926 @@
...
+		configMap := obj.(*corev1.ConfigMap)
+		configMap.Labels = mergeLabels(configMap.Labels, managedLabels(imagePullSecret.Name))
+		configChanged = configMap.Annotations[annotationECRConfigHash] != configHash
+		metav1.SetMetaDataAnnotation(&configMap.ObjectMeta, annotationECRConfigHash, configHash)
+		if configChanged {
+			configMap.Data = map[string]string{}
+		}
+		if configMap.Data == nil {
+			configMap.Data = map[string]string{}
+		}
```
thedadams:
```suggestion
		if configChanged || configMap.Data == nil {
			configMap.Data = map[string]string{}
		}
```

pkg/controller/handlers/imagepullsecret/imagepullsecret.go
```diff
@@ -0,0 +1,926 @@
...
+	ecr := imagePullSecret.Spec.ECR
+	name := imagepullsecrets.DefaultECRCronJobName(imagePullSecret.Name)
+	statusConfigMapName := imagepullsecrets.DefaultECRStatusConfigMapName(imagePullSecret.Name)
+	labels := managedLabels(imagePullSecret.Name)
+
+	return h.ensureObjectInNamespace(ctx, &batchv1.CronJob{}, h.helperNamespace, name, imagePullSecret.Name, func(obj kclient.Object) {
+		cronJob := obj.(*batchv1.CronJob)
+		cronJob.Labels = mergeLabels(cronJob.Labels, labels)
+		metav1.SetMetaDataAnnotation(&cronJob.ObjectMeta, annotationECRConfigHash, ecrConfigHash(imagePullSecret))
+		cronJob.Spec = batchv1.CronJobSpec{
```
thedadams: I understand that we have to setup credentials and whatnot here, but are we sure there isn't a good way of doing this in a controller Obot itself?

pkg/controller/handlers/imagepullsecret/imagepullsecret.go
```diff
@@ -0,0 +1,926 @@
...
+		return fmt.Errorf("failed to delete Kubernetes resource %s/%s: %w", namespace, name, err)
+	}
+	return nil
+}
+
+// UpdateECRStatusFromConfigMap copies refresh helper results from Kubernetes back to Obot status.
+func (h *Handler) UpdateECRStatusFromConfigMap(req router.Request, _ router.Response) error {
+	if req.Object == nil {
+		return nil
+	}
```
thedadams: This is managed by our framework and `req.Object` is only `nil` if you use `IncludeRemoved()`, which based on this check, you shouldn't use.

pkg/imagepullsecrets/dockerconfig.go
```diff
@@ -0,0 +1,64 @@
...
+
+func buildDockerConfigJSON(auths map[string]dockerAuth) ([]byte, error) {
+	return json.Marshal(dockerConfig{
+		Auths: auths,
+	})
+}
+
+func EmptyDockerConfigJSON() []byte {
+	data, _ := buildDockerConfigJSON(map[string]dockerAuth{})
+	return data
```
thedadams:
```suggestion
	return []byte("{}")
```

pkg/services/config.go
```diff
@@ -122,20 +126,23 @@ type Config struct {
...
+	MCPNetworkPolicyProviderChartRepo    string   `usage:"Helm repository URL for the network policy provider chart"`
+	MCPNetworkPolicyProviderChartName    string   `usage:"Helm chart name for the network policy provider chart"`
+	MCPNetworkPolicyProviderChartVersion string   `usage:"Helm chart version for the network policy provider chart"`
+	MCPNetworkPolicyProviderChartPath    string   `usage:"Local filesystem path to the network policy provider chart"`
+	MCPNetworkPolicyProviderValues       string   `usage:"YAML or JSON values blob merged into the network policy provider chart values"`
+	MCPDefaultDenyAllEgress              bool     `usage:"Default new MCP servers to deny all egress when network policy enforcement is enabled" default:"false"`
+	ObotImage                            string   `usage:"Container image for Obot helper jobs" env:"OBOT_SERVER_IMAGE"`
+	OwnImagePullSecrets                  []string `usage:"Kubernetes image pull secret names for Obot's own helper jobs" env:"OBOT_OWN_IMAGE_PULL_SECRETS"`
+	OwnRuntimeClassName                  string   `usage:"Kubernetes RuntimeClass name for Obot's own helper jobs" env:"OBOT_OWN_RUNTIME_CLASS_NAME"`
```
thedadams: Don't use `env` tags so everything is consistent here. This would also mean that anything that references these environment variables needs to be updated to `OBOT_SERVER_...`

pkg/imagepullsecrets/ecrhelper.go
```diff
@@ -0,0 +1,324 @@
...
+
+// RefreshECR performs one ECR token refresh. On success it writes the Docker
+// config JSON Secret used by imagePullSecrets and records success metadata in
+// the status ConfigMap. On failure it best-effort records a sanitized error in
+// the status ConfigMap and returns a non-nil error so the Kubernetes Job fails.
+func RefreshECR(ctx context.Context, opts ECRRefreshOptions) error {
+	opts.Region = strings.TrimSpace(opts.Region)
+	opts.SecretName = strings.TrimSpace(opts.SecretName)
+	opts.StatusConfigMapName = strings.TrimSpace(opts.StatusConfigMapName)
+	opts.Namespace = strings.TrimSpace(opts.Namespace)
```
thedadams: Is it OK for this function to assume this has already been done?

pkg/imagepullsecrets/ecrrefreshjob.go
```diff
@@ -0,0 +1,78 @@
...
+		},
+		Spec: *cronJob.Spec.JobTemplate.Spec.DeepCopy(),
+	}
+	job.Spec.Template.Labels = klabels.Merge(klabels.Set(job.Spec.Template.Labels), labels)
+
+	if err := runtimeClient.Create(ctx, job); err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			return "", err
+		}
+		return "", fmt.Errorf("failed to create ECR refresh Job: %w", err)
```
thedadams: Seems strange to separate these.

pkg/imagepullsecrets/effective.go
```diff
@@ -0,0 +1,69 @@
...
+
+func Hash(secretNames []string) string {
+	return hash.Digest(CleanSecretNames(secretNames))
+}
+
+func CleanSecretNames(names []string) []string {
+	return cleanSecretNames(names)
+}
+
+func cleanSecretNames(names []string) []string {
```
thedadams:
```suggestion
```

pkg/imagepullsecrets/registrytest.go
```diff
@@ -0,0 +1,580 @@
...
+		repository = rest
+	}
+
+	if repository == "" {
+		return imageReference{}, fmt.Errorf("image repository is required")
+	}
+	// Docker Hub official images use the implicit library/ namespace.
+	if isDockerHubRegistry(registry) && !strings.Contains(repository, "/") {
+		repository = "library/" + repository
+	}
```
thedadams, line 183: Do we have to add that? I thought it gets taken care of by the kubelet.

pkg/imagepullsecrets/registrytest.go
```diff
@@ -0,0 +1,580 @@
...
+	return fmt.Errorf("registry request failed: %s", err)
+}
+
+func statusError(prefix string, statusCode int) error {
+	return fmt.Errorf("%s with HTTP %d", prefix, statusCode)
+}
+
+func drainBody(body io.Reader) {
+	_, _ = io.Copy(io.Discard, io.LimitReader(body, 4096))
```
thedadams: Why would we limit here? What's the difference between not draining the whole body and not draining it at all?

pkg/imagepullsecrets/registrytest.go
```diff
@@ -0,0 +1,580 @@
...
+			scheme: strings.ToLower(strings.TrimSpace(scheme)),
+			params: parseAuthParams(params),
+		})
+	}
+	return result
+}
+
+// parseAuthParams parses a comma-delimited authentication parameter list while
+// preserving quoted values that can contain commas or escaped characters.
+func parseAuthParams(input string) map[string]string {
```
thedadams, line 418: I think some unit tests would be beneficial here.

pkg/imagepullsecrets/validation.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+	if parsed.User != nil {
+		return "", fmt.Errorf("registry server must not include user info")
+	}
+	if parsed.RawQuery != "" || parsed.Fragment != "" {
+		return "", fmt.Errorf("registry server must not include query or fragment")
+	}
+	if parsed.Path != "" && parsed.Path != "/" {
+		return "", fmt.Errorf("registry server must not include a URL path")
+	}
```
thedadams, line 52: Is this a requirement?

pkg/imagepullsecrets/validation.go
```diff
@@ -0,0 +1,247 @@
...
+	if parsed.Host == "" {
+		return "", fmt.Errorf("registry server host is required")
+	}
+
+	host, port, err := splitHostPort(parsed.Host)
```
thedadams, line 57: https://pkg.go.dev/net#SplitHostPort

pkg/imagepullsecrets/validation.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	portNumber, err := strconv.Atoi(port)
+	if err != nil || portNumber < 1 || portNumber > 65535 {
+		return "", fmt.Errorf("registry server port must be a number between 1 and 65535")
+	}
+	if strings.Contains(host, ":") {
+		return net.JoinHostPort(host, port), nil
+	}
+	return host + ":" + port, nil
```
thedadams:
```suggestion
	return net.JoinHostPort(host, port), nil
```

pkg/mcp/kubernetes.go
```diff
@@ -850,8 +847,8 @@ func (k *kubernetesBackend) k8sObjects(ctx context.Context, server ServerConfig,
 		}
 	}
 
-	if len(k.imagePullSecrets) > 0 {
-		for _, secret := range k.imagePullSecrets {
+	if len(effectiveImagePullSecrets) > 0 {
```
thedadams: I know you didn't add it, but we don't need this check.

pkg/mcp/kubernetes.go
```diff
@@ -1509,18 +1547,42 @@ func affinityMatches(actual *corev1.Affinity, desired *corev1.Affinity) bool {
 	return reflect.DeepEqual(actual, desired)
 }
 
-// tolerationsMatch checks if the deployment's tolerations match the desired settings
+// tolerationsMatch checks if the deployment's tolerations include the desired
+// settings. Extra actual tolerations are allowed because some Kubernetes
+// environments mutate scheduling constraints after admission.
 func tolerationsMatch(actual []corev1.Toleration, desired []corev1.Toleration) bool {
-	if len(desired) == 0 && len(actual) == 0 {
+	if len(desired) == 0 {
```
thedadams, line 1624: Was this removal on purpose? I would classify `len(desired) == 0 && len(actual) > 0` as the tolerations not matching.

g-linville: Yes, this was intentional. It's not ideal if there were tolerations configured in Obot, but this was to better support drift detection when there is another controller slapping tolerations on everything. In GKE Autopilot, it was a CPU architecture specific toleration that was getting automatically added to MCP servers, causing them to constantly be marked as `needsK8sUpdate`.

pkg/api/handlers/imagepullsecrets.go
g-linville: Yes

pkg/imagepullsecrets/registrytest.go
g-linville: Adding

apiclient/types/imagepullsecret.go
g-linville: Yep

pkg/controller/handlers/imagepullsecret/imagepullsecret.go
g-linville: rearchitecting

pkg/api/handlers/mcp.go
g-linville: I undid these. It was unnecessary.

pkg/imagepullsecrets/registrytest.go
g-linville: We do need to do this. We aren't relying on Kubernetes at all in this path. This is just the test to make sure the credential works, so we are reaching out to the registry (in this case, Docker Hub) directly.

pkg/imagepullsecrets/registrytest.go
g-linville: It was for connection reuse, but we don't actually need to do it, so I'm getting rid of it for simplicity.

pkg/imagepullsecrets/validation.go
g-linville: Sort of. Kubernetes and Docker differ a little bit in what they allow in the dockerconfigjson struct. Here's more details on it from the Kubernetes docs: https://kubernetes.io/docs/concepts/containers/images/#config-json

Basically, by enforcing that there is no path, we are keeping things much simpler. I would argue that we stick with this for the time being.

g-linville:
Re-requesting review. This has been rearchitected to avoid creating a Kubernetes CronJob and to just do the refresh in the Obot controller.

g-linville:
pushed a few more changes based on review feedback from my claude code review agent

pkg/api/handlers/imagepullsecrets.go
```diff
@@ -0,0 +1,660 @@
...
+
+	credentials, err := req.GPTClient.ListCredentials(req.Context(), gptscript.ListCredentialsOptions{
+		CredentialContexts: []string{imagepullsecrets.CredentialContext},
+	})
+	if err != nil {
+		return nil, err
+	}
+	for _, credential := range credentials {
+		if _, ok := result[credential.ToolName]; ok {
+			_, result[credential.ToolName] = credential.Env[imagepullsecrets.PasswordEnvVar]
```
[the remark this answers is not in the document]
g-linville: incorrect assessment

pkg/storage/apis/obot.obot.ai/v1/imagepullsecret.go
```diff
@@ -0,0 +1,86 @@
...
+	return []string{"spec.secretName"}
+}
+
+type ImagePullSecretSpec struct {
+	Enabled     bool                      `json:"enabled,omitempty"`
+	Type        ImagePullSecretType       `json:"type,omitempty"`
+	DisplayName string                    `json:"displayName,omitempty"`
+	SecretName  string                    `json:"secretName,omitempty"`
+	Basic       *BasicImagePullSecretSpec `json:"basic,omitempty"`
+	ECR         *ECRImagePullSecretSpec   `json:"ecr,omitempty"`
```
thedadams: Can we use the types for `apiclient/types` here? I know the password would always be empty, but maybe that's OK to avoid duplication?

pkg/storage/apis/obot.obot.ai/v1/imagepullsecret.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+var _ fields.Fields = (*ImagePullSecret)(nil)
+
+type ImagePullSecretType string
+
+const (
+	ImagePullSecretTypeBasic ImagePullSecretType = "basic"
+	ImagePullSecretTypeECR   ImagePullSecretType = "ecr"
+)
```
thedadams: Same comment here: we have these constants we can reference elsewhere.

pkg/services/config.go
```diff
@@ -644,11 +651,26 @@ func New(ctx context.Context, config Config) (*Services, error) {
 
 	// Build local Kubernetes config for deployment monitoring (optional)
 	var localK8sConfig *rest.Config
-	if config.MCPRuntimeBackend == "kubernetes" {
+	var localK8sInterfaceClient kubernetes.Interface
+	var serviceAccountIssuerURL string
+	var serviceAccountIssuerError string
```
thedadams: nit: combine

pkg/mcp/kubernetes.go
```diff
@@ -507,8 +509,12 @@ func (k *kubernetesBackend) k8sObjects(ctx context.Context, server ServerConfig,
...
 	}
 
-	// Add K8s settings hash to annotations
-	annotations["obot.ai/k8s-settings-hash"] = ComputeK8sSettingsHash(k8sSettings)
+	effectiveImagePullSecrets, err := k.effectiveImagePullSecretNames(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get effective image pull secrets: %w", err)
+	}
+
+	annotations["obot.ai/k8s-settings-hash"] = ComputeK8sSettingsHash(k8sSettings, effectiveImagePullSecrets)
```
thedadams: Do we need to add this here? My understanding is if the image pull secret changes, then the kubelet will use the new value on the next pull.

pkg/mcp/kubernetes.go
```diff
@@ -1139,8 +1134,13 @@ func (k *kubernetesBackend) restartServer(ctx context.Context, server ServerConf
...
 
-	// Compute K8s settings hash
-	k8sSettingsHash := ComputeK8sSettingsHash(k8sSettings)
+	effectiveImagePullSecrets, err := k.effectiveImagePullSecretNames(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get effective image pull secrets: %w", err)
+	}
+
+	// Compute k8s settings hash
+	k8sSettingsHash := ComputeK8sSettingsHash(k8sSettings, effectiveImagePullSecrets)
```
thedadams: Same question.

pkg/imagepullsecrets/capability.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func IsKubernetesBackend(backend string) bool {
+	switch strings.ToLower(strings.TrimSpace(backend)) {
+	case "kubernetes", "k8s":
+		return true
+	default:
+		return false
+	}
+}
```
thedadams: Seems like this is general enough to go in a different package. Maybe `pkg/mcp`?

pkg/controller/controller.go
```diff
@@ -567,22 +567,25 @@ func (c *Controller) createLocalK8sRouter() (*router.Router, error) {
 }
 
 func localK8sCacheByObject(mcpServerNamespace, runtimeNamespace string) map[kclient.Object]crcache.ByObject {
+	result := map[kclient.Object]crcache.ByObject{}
```
thedadams: The changes in this function seem non-functional. Calling it out in case there was supposed to be a change here.

pkg/api/handlers/imagepullsecrets.go
```diff
@@ -0,0 +1,660 @@
...
+var defaultECRPolicyJSON = buildECRPolicyJSON()
+
+type ImagePullSecretHandler struct {
+	mcpRuntimeBackend  string
+	staticSecrets      []string
+	mcpNamespace       string
+	serviceNamespace   string
+	serviceAccountName string
+	runtimeClient      kclient.Client
+	kubeClient         kubernetes.Interface
```
thedadams: This is unused.

pkg/api/handlers/imagepullsecrets.go
```diff
@@ -0,0 +1,660 @@
...
+	})
+}
+
+func (h *ImagePullSecretHandler) testECRImagePullSecret(req api.Context, secret v1.ImagePullSecret, image string) (imagepullsecrets.RegistryTestResult, error) {
+	if h.runtimeClient == nil {
+		return imagepullsecrets.RegistryTestResult{}, types.NewErrHTTP(http.StatusServiceUnavailable, "Kubernetes runtime client is not configured")
+	}
+	if strings.TrimSpace(h.mcpNamespace) == "" {
+		return imagepullsecrets.RegistryTestResult{}, types.NewErrHTTP(http.StatusServiceUnavailable, "MCP namespace is not configured")
+	}
```
thedadams: nit: not needed because of defaulting.

pkg/api/handlers/imagepullsecrets.go
```diff
@@ -0,0 +1,660 @@
...
+}
+
+func basicImagePullSecretPassword(input types.ImagePullSecretManifest) string {
+	if input.Basic == nil {
+		return ""
+	}
+	return input.Basic.Password
+}
+
+func generateImagePullSecretName(req api.Context) (string, error) {
```
thedadams: Can we just use the `GenerateName` functionality of Kubernetes instead of this?

pkg/api/handlers/imagepullsecrets.go
```diff
@@ -0,0 +1,660 @@
...
+	if err := req.Read(&input); err != nil {
+		return types.NewErrBadRequest("failed to read image pull secret manifest: %v", err)
+	}
+
+	name, err := generateImagePullSecretName(req)
+	if err != nil {
+		return err
+	}
+
+	spec, err := h.specFromInput(input, nil, name)
```
thedadams: The `name` that gets passed here gets set on the spec as `SecretName`, but then gets set as the name of the `ImagePullSecret` object that is created.

That seems circular. Is that on purpose?

pkg/api/handlers/mcp.go
```diff
@@ -43,23 +43,33 @@ type MCPOAuthChecker interface {
...
-	mcpOAuthChecker   MCPOAuthChecker
-	acrHelper         *accesscontrolrule.Helper
-	mcpBackend        string
-	serverURL         string
+	mcpSessionManager   *mcp.SessionManager
+	mcpOAuthChecker     MCPOAuthChecker
+	acrHelper           *accesscontrolrule.Helper
+	mcpBackend          string
+	mcpImagePullSecrets []string
+	serverURL           string
```
thedadams, line 50: nit: I think it would be nice if there was a method on the `mcp.SessionManager` to get the backend instead of passing it separately.

pkg/controller/handlers/imagepullsecret/imagepullsecret.go
```diff
@@ -0,0 +1,466 @@
...
+)
+
+const (
+	annotationECRConfigHash = "obot.ai/ecr-config-hash"
+)
+
+type Handler struct {
+	gptClient          *gptscript.GPTScript
+	runtimeClient      kclient.Client
+	kubeClient         kubernetes.Interface
```
thedadams: Do we need both of these? My understanding is that they both talk to the cluster where Obot is running.

pkg/controller/handlers/imagepullsecret/imagepullsecret.go
```diff
@@ -0,0 +1,466 @@
...
+
+func (h *Handler) scheduleNextRefresh(resp router.Response, secret *v1.ImagePullSecret) {
+	if secret.Status.LastSuccessTime == nil {
+		return
+	}
+	next, err := gronx.NextTickAfter(secret.Spec.ECR.RefreshSchedule, secret.Status.LastSuccessTime.Time, false)
+	if err != nil {
+		return
+	}
+	if until := next.Sub(h.now()); until > 0 {
```
thedadams: We recheck everything every 10 hours.
```suggestion
	if until := next.Sub(h.now()); until > 0 && until < 10 * time.Hour {
```

pkg/imagepullsecrets/validation.go
```diff
@@ -0,0 +1,225 @@
...
+		return "", fmt.Errorf("registry server must not include user info")
+	}
+	if parsed.Path != "" && parsed.Path != "/" {
+		return "", fmt.Errorf("registry server must not include a URL path")
+	}
+	if parsed.Host == "" {
+		return "", fmt.Errorf("registry server host is required")
+	}
+
+	host, port, err := splitHostPort(parsed.Host)
```
thedadams, line 57: Can we use `net.SplitHostPort` here?

pkg/mcp/kubernetes.go
g-linville: Well we were already setting this annotation on the deployment. It's just that now, the k8s settings hash includes the effective image pull secrets, so we have to add that as a parameter to this function call.

The hash does not include the values of the image pull secrets. It only includes the names, since that's what gets set on the pod template. So we do need this here.

pkg/mcp/kubernetes.go
g-linville: Same answer as above

pkg/imagepullsecrets/capability.go
g-linville: Moving this to pkg/mcp/backend.go

pkg/controller/controller.go
g-linville: Oh you're right. I will undo my changes here. This was changed originally for CronJob stuff and AI did a bad job cleaning up after itself

pkg/api/handlers/imagepullsecrets.go
g-linville: Yes, and I am also going to remove the redundant `SecretName` field on the ImagePullSecret spec, since the k8s Secret name always matches the ImagePullSecret name anyway

pkg/api/handlers/imagepullsecrets.go
g-linville: I'm removing the spec.SecretName

pkg/controller/handlers/imagepullsecret/imagepullsecret.go
g-linville: We don't. I thought before that the `kubernetes.Interface` was required for doing a ServiceAccount token request, but the `kclient.Client` can do that too, so I'm removing the `kubernetes.Interface`.

pkg/imagepullsecrets/validation.go
g-linville: `net.SplitHostPort` returns an error if the port is not there. Obviously we want to support hosts with no port, i.e. `ghcr.io`. So we wrap it in a little helper to deal with that.

thedadams (approved):

PR #6570, opened 2026-05-11
enhance: make dynamic client expiration configurable

thedadams:
The default is 30 days.


calvinmclean (approved):
Approving assuming the lint will be fixed :)

g-linville (approved):

Issue #6548, opened 2026-05-08
feat: support private git repos for skill sync


calvinmclean:
## Summary

Generalize skill repository sync to work with any git host (GitHub, GitLab, self-hosted) and support private repositories via personal access tokens.

## Motivation

Skill sync currently only supports public GitHub repositories using the GitHub ZIP archive API. This prevents users from syncing skills from private repos or non-GitHub hosts.

## Changes

**New `pkg/git` package**
- Shared git clone utility using `go-git`, usable by both skill sync and MCP catalog sync
- Supports any HTTPS git host
- Enforces 100MB size limit

**Skill repository improvements**
- Drop GitHub-only restriction — accept any HTTPS git URL
- Add optional personal access token support (stored securely, never returned in API responses)
- Fall back to `GITHUB_AUTH_TOKEN` env var if no per-repo token is set

**MCP catalog refactor**
- Update to use the new shared `pkg/git` package (no behavior change)

**UI**
- Add optional token input to the skill repository admin form

## Related

Follows the credential management pattern established in #6351.

calvinmclean:
The primary usecase for this is:
- Github (public/private)
- Gitlab (public/private)

sangee2004:
This issue is addressed when testing with the latest build from `main`

Able to Add skills from Github public and private repo. `secret.GITHUB_AUTH_TOKEN` get used for these repos.
Able to Add skills from Gitlab public and private repo. `secret.GITHUB_AUTH_TOKEN` get used for these repos.

@calvinmclean There is way to edit Skill source URL which is tracked in  https://github.com/obot-platform/obot/issues/7230

PR #6571, opened 2026-05-11
feat: support SkillRepository from any Git source with PAT

calvinmclean:
## Summary

Addresses #6548 

Generalize the private git repo sync from MCP catalogs and use it with skills


pkg/api/handlers/skills.go
```diff
@@ -31,14 +34,49 @@ const (
 
 type SkillHandler struct {
 	skillAccessRuleHelper  *skillaccessrule.Helper
-	materializeSkillSource func(ctx context.Context, skill *v1.Skill) (func(), string, error)
+	gatewayClient          skillRepositoryCredentialRevealer
```
thedadams: We shouldn't need this; use the client on the api.Context instead.

thedadams (approved):

Issue #6561, opened 2026-05-08
Obot fails to clone public Gitlab MCP catalog without a token


calvinmclean:
**Describe the bug**
When adding a public Gitlab MCP catalog (https://gitlab.com/calvinmclean/mcp-catalog), the following error occurs:
```
failed to read git catalog https://gitlab.com/calvinmclean/mcp-catalog: failed to clone repository: authentication required: HTTP Basic: Access denied. If a password was provided for Git authentication, the password was incorrect or you're required to use a token instead of a password. If a token was provided, it was either incorrect, expired, or improperly scoped. See https://gitlab.com/help/topics/git/troubleshooting_git.md#error-on-git-fetch-http-basic-access-denied
```

Since we fall back to the GITHUB_AUTH_TOKEN, this makes an API request that expects authorization and fails.

We should either prevent fallback for non-github URLs or ignore errors from the API size check request.

**Reproduce**
1. Run Obot with `GITHUB_AUTH_TOKEN`
2. Add public Gitlab MCP catalog without setting a token (`https://gitlab.com/calvinmclean/mcp-catalog`)
3. Wait for sync

**Expected behavior**
Successfully sync catalog.

**Screenshots**
<img width="1138" height="386" alt="Image" src="https://github.com/user-attachments/assets/448108eb-dffd-461d-95a1-239bf2dd8a18" />

sangee2004:
This issue is addressed when testing with the latest build from `main` in an obot setup with `GITHUB_AUTH_TOKEN` configured.

1. Able to clone a public GitHub repo successfully with no PAT set for the repo.
2. Able to clone a private GitHub repo successfully with no PAT set for the repo (`GITHUB_AUTH_TOKEN` gets used)
3. Able to clone a private GitHub repo successfully with  PAT set for the repo 
4. Able to clone a public GitLab repo successfully with no PAT set for the repo.
5. Not Able to clone the private GitLab repo successfully when PAT is not set for the repo. Once PAT is set up for the repo, we can successfully clone it from GitLab.

PR #6573, opened 2026-05-11
fix: prefer catalog token, try anonymous otherwise

calvinmclean:
## Description
Addresses #6561 

The primary goal of this PR is to avoid failures for public GitLab catalogs when the `GITHUB_AUTH_TOKEN` env var is configured. It achieves this by first trying anonymously if no specific credential is set, then falls back to the env var if that fails.

## Details

Separate API size-check authentication from git clone authentication for MCP catalog sources.

GitHub size checks use the catalog token, then GITHUB_AUTH_TOKEN, then anonymous as a single best-effort request; GitLab size checks only use a catalog-specific token and otherwise rely on the clone-time size limit.

Clone attempts now use the catalog token exclusively when one is configured. Without a catalog token, cloning tries anonymous first and only retries with GITHUB_AUTH_TOKEN if the anonymous clone fails, preventing public GitLab catalogs from failing because a GitHub token is present.

## Testing
  - [x] Public github repo (https://github.com/njhale/mcp-zoo)
  - [x] Private github repo (no credential) (https://github.com/calvinmclean/private-mcp-catalog)
  - [x] Private github repo (with credential)
  - [x] Public gitlab repo (https://gitlab.com/calvinmclean/mcp-catalog)
  - [x] Private gitlab repo (no credential, fails) (https://gitlab.com/calvinmclean/private-mcp-catalog)
  - [x] Private gitlab repo (with credential)

### DIFF
--- a/pkg/controller/handlers/mcpcatalog/github.go
+++ b/pkg/controller/handlers/mcpcatalog/github.go
@@ -27,16 +27,25 @@ type GitHubRepoInfo struct {
 	Size int `json:"size"` // Size in KB
 }
 
-// resolveToken returns the provided token if non-empty.
-// If the token is empty, it falls back to the global GITHUB_AUTH_TOKEN.
-func resolveToken(token string) string {
-	if token != "" {
-		return token
-	}
-	return os.Getenv("GITHUB_AUTH_TOKEN")
+var errRepoTooLarge = errors.New("repository too large")
+
+type gitCloneAuthAttempt struct {
+	name  string
+	token string
 }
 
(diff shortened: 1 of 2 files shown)


g-linville:
Can we avoid sending `GITHUB_AUTH_TOKEN` to GitLab no matter what? It feels weird to me that that credential could end up getting sent to something that isn't GitHub.

g-linville:
^ we talked about this in DM, sounds like this was the agreed upon design from before so we are going to keep it as it is

g-linville (approved):

thedadams (approved):

PR #6574, opened 2026-05-12
fix: deep-link scan observations by row id instead of slice index

njhale:
Scan detail pages (MCP server, skill, plugin) were addressed by their
position inside the parent scan's preloaded child slice, e.g.
/admin/.../scans/{scan_id}/mcp/{index}. That index was a GORM
preload-order artifact, not a stable identifier — it shifted whenever
the underlying device_scan_* rows were re-ordered, paged, or filtered,
so bookmarked links and breadcrumb back-nav could resolve to the wrong
row or 404 against fresh scans.

Switch the URL segment and lookup key to each observation's row PK:

- Expose ID on DeviceScanMCPServer/Skill/Plugin (server-set, ignored on
  submission so scanners can't trample PKs) and on the occurrence
  list rows the dashboard ranks. The SELECTs in ListMCPServerOccurrences
  and ListSkillOccurrences drop their correlated COUNT(*) idx subquery
  and project the row id directly.
- Rename the SvelteKit routes from [index] to [id]; the detail pages
  still read the parent scan from the existing +layout.ts loader and
  now find their row via scan.<children>.find(r => r.id === id) instead
  of indexing by position.
- Update every callsite that builds these URLs (device dashboard, scan
  page, mcp-server-by-hash table, skill-by-name table) to pass d.id.
- findParentPlugin returns the plugin's id (not its index) so MCP/skill
  cross-links to the enclosing plugin stay correct.

### DIFF
--- a/ui/user/src/lib/services/admin/types.ts
+++ b/ui/user/src/lib/services/admin/types.ts
@@ -1256,8 +1256,6 @@ export interface DeviceMCPServerOccurrence {
 	client: string;
 	scope: string;
 	scannedAt: string;
-	// id is the row's PK in device_scan_mcp_servers; use it as the
-	// {id} segment of scan-scoped detail URLs.
 	id: number;
 }
 
@@ -1325,8 +1323,6 @@ export interface DeviceSkillOccurrence {
 	scope: string;
 	projectPath?: string;
 	scannedAt: string;
-	// id is the row's PK in device_scan_skills; use it as the {id}
-	// segment of scan-scoped detail URLs.
 	id: number;
 }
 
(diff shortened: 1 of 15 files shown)


thedadams (approved):

ui/user/src/lib/services/admin/types.ts
```diff
@@ -1136,6 +1136,7 @@ export interface DeviceScanFile {
 }
 
 export interface DeviceScanMCPServer {
+	id?: number;
```
ivyjeong13: Is `id` optional for DeviceScanMCPServer/Skill/Plugin? I see it's not for DeviceRow/Skill/Plugin in the /scans route

njhale: nice catch. this is always populated by the backend when fetching them, so I think we can remove the `?` here. let me do that

njhale: done

ivyjeong13 (approved):
UI changes LGTM!

Issue #6563, opened 2026-05-09
Move `startupTimeoutSeconds` under runtime-specific configuration


sangee2004:
**Describe the bug**
`startupTimeoutSeconds` is currently defined as a top-level field in MCP catalog YAML entries, even though it does not apply to remote runtime. After syncing the catalog into Obot, the value appears under Containerized Runtime Configuration, creating an inconsistency between the catalog structure and the UI representation. Since startupTimeoutSeconds is not valid for runtime: remote, it should be nested inside the runtime-specific configuration block (for example, containerizedConfig), similar to how egressDomains is structured.

**To Reproduce**
1. Configure  `startupTimeoutSeconds` for an MCP server entry in the GitHub repo.
This needs to be set at the top level - https://github.com/sangee2004/mcp-catalog/blob/main/SangeeAntv.yaml

Snippet of the YAML:

```
runtime: containerized
startupTimeoutSeconds: 100
containerizedConfig:
  image: ghcr.io/obot-platform/mcp-images/antv-charts:0.9.4
  port: 8099
  path: /
  args:
  - mcp-server-chart
  egressDomains: ["testing.com","*.sangeetesting.com"]
```
2. Add this git repo as the MCP catalog to obot.

Notice that when this entry is synced, `startupTimeoutSeconds` appears in the `Containerized Runtime Configuration` section of the `Configuration` tab, which differs from how it is represented in the catalog.

<img width="1129" height="489" alt="Image" src="https://github.com/user-attachments/assets/e48bc287-da4c-4709-8b1b-cd64a2d5ee9c" />

Also, `startupTimeoutSeconds` is not valid for `runtime: remote`.

**Expected behavior**
Move `startupTimeoutSeconds` to be nested under the runtime configuration, like `containerizedConfig` in this case. This is how `egressDomains:` is structured, which is also not applicable to remote MCP servers.

sangee2004:
This issue is addressed when testing with the latest build from `main`.

We now  have`startupTimeoutSeconds` defined as a runtime-specific configuration in MCP catalog entries

```
runtime: containerized
containerizedConfig:
  image: ghcr.io/obot-platform/mcp-images/antv-charts:0.9.4
  port: 8099
  path: /
  args:
  - mcp-server-chart
  startupTimeoutSeconds: 100
  egressDomains: ["testing.com","*.sangeetesting.com"]
```

PR #6576, opened 2026-05-12
fix: move StartupTimeoutSeconds to compatible runtimes

calvinmclean:
## Summary
Addresses https://github.com/obot-platform/obot/issues/6563

- Move `startupTimeoutSeconds` from the manifest-level field into the runtime-specific config for `uvx`, `npx`, and `containerized` runtimes.
- Remove the legacy top-level timeout field and migration path because this change has not shipped yet.
- Update validation, API/UI types, server mapping, and form handling so startup timeout is read and written only through the active runtime config.


apiclient/types/mcpserver.go
```diff
@@ -417,6 +422,80 @@ func (e RuntimeValidationError) Error() string {
...
+			npxConfig.StartupTimeoutSeconds = *legacyStartupTimeoutSeconds
+		}
+	case RuntimeContainerized:
+		if containerizedConfig != nil && containerizedConfig.StartupTimeoutSeconds == 0 {
+			containerizedConfig.StartupTimeoutSeconds = *legacyStartupTimeoutSeconds
+		}
+	}
+
+	*legacyStartupTimeoutSeconds = 0
+	return true
```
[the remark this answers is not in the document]
calvinmclean: This is intentional. Even when there is nowhere valid to copy/move the config, it should be removed since it is no longer valid.

g-linville (approved):

apiclient/types/mcpserver.go
```diff
@@ -428,6 +415,37 @@ func (e RuntimeValidationError) Error() string {
...
+		if npxConfig != nil {
+			return npxConfig.StartupTimeoutSeconds
+		}
+	case RuntimeContainerized:
+		if containerizedConfig != nil {
+			return containerizedConfig.StartupTimeoutSeconds
+		}
+	}
+
+	return 0
```
thedadams: Does it make sense to return the default here? So we don't have to have defaulting logic elsewhere?

thedadams (approved):

apiclient/types/mcpserver.go
calvinmclean: I was thinking about this as well, but didn't want to spread out logic too much since we do this when it's used:
```go
	startupTimeoutSeconds := mcpServer.Spec.Manifest.RuntimeStartupTimeoutSeconds()
	startupTimeout := time.Duration(startupTimeoutSeconds) * time.Second
	if startupTimeout == 0 {
		startupTimeout = defaultStartupTimeout
	} else if startupTimeout > MaxMCPServerStartupTimeout {
		return ServerConfig{}, nil, fmt.Errorf("input %d exceeds the max of %s", startupTimeoutSeconds, MaxMCPServerStartupTimeout)
	}
```

Since we have all of the validators, could we remove the `if startupTimeout > MaxMCPServerStartupTimeout` from `ServerToServerConfig`?

thedadams: My thought was that if we return the default here, then we wouldn't have to do the `if startupTimeout == 0` check in the code you are showing here.

PR #6582, opened 2026-05-12
enhance: store CLI auth tokens in OS keychain

g-linville:
(AI-generated description)

Summary

  Adds the local config and keyring-backed authentication foundation for the local Obot skill work.

  This PR introduces setup-managed URL config, normalizes Obot app URLs, derives API base URLs, and moves CLI token storage away from the legacy XDG token file toward OS keyring-backed credentials scoped by normalized app URL.

  Changes

  - Added pkg/cli/internal/localconfig
      - Normalizes app URLs.
      - Derives API base URLs.
      - Reads/writes $XDG_CONFIG_HOME/obot/config.json.
      - Resolves active URL from explicit input or stored default.
  - Added pkg/cli/internal/credentials
      - Defines a mockable credential store boundary.
      - Adds a github.com/zalando/go-keyring adapter.
      - Uses service obot and normalized app URL as the key.
  - Updated CLI auth behavior
      - login supports --url.
      - logout supports --url.
      - Tokens are read/written/deleted through the keyring store.
      - Legacy XDG token file is intentionally ignored.
      - Keyring failures fail closed.
      - Missing credentials still trigger the existing browser login flow.
  - Updated client URL resolution
      - OBOT_BASE_URL remains the explicit override.
      - Stored default app URL is used when no env override is present.
      - Falls back to http://localhost:8080/api.

---

**Note from Grant**: this is the first step in the changes I'm making for local skill installation. The new zalando dependency is basically a wrapper around OS-native keystores, and it doesn't require CGo.

I've only tested this on macOS, but it works there.

The rest of the changes for this feature will come in a second, larger PR.


pkg/cli/internal/credentials/keyring.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"errors"
+
+	keyringlib "github.com/zalando/go-keyring"
+)
+
+// Keyring is the subset of github.com/zalando/go-keyring used by the
+// CLI. Tests can provide an in-memory implementation.
+type Keyring interface {
```
thedadams: nit: this can be unexported.

pkg/cli/internal/credentials/keyring.go
```diff
@@ -0,0 +1,73 @@
...
+
+// NewKeyringStore returns a Store backed by the host OS keyring.
+func NewKeyringStore() *KeyringStore {
+	return NewKeyringStoreWith(DefaultService, keyringFuncs{})
+}
+
+// NewKeyringStoreWith returns a keyring store using the supplied
+// service and keyring implementation. It is exported for tests and for
+// future command wiring that needs dependency injection.
+func NewKeyringStoreWith(service string, keyring Keyring) *KeyringStore {
```
thedadams: As well as this.

pkg/cli/internal/credentials/keyring.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func (s *KeyringStore) Set(appURL, token string) error {
+	return s.keyring.Set(s.service, appURL, token)
+}
+
+func (s *KeyringStore) Delete(appURL string) error {
+	err := s.keyring.Delete(s.service, appURL)
+	if errors.Is(err, keyringlib.ErrNotFound) {
+		return ErrNotFound
```
thedadams: My preference is that delete operations are idempotent.

pkg/cli/internal/localconfig/store.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	if parsed.Fragment != "" {
+		return "", fmt.Errorf("app URL must not include a fragment")
+	}
+
+	normalized := strings.TrimRight(parsed.String(), "/")
+	normalized = strings.TrimSuffix(normalized, "/api")
+	if normalized == parsed.Scheme+":" {
+		return "", fmt.Errorf("app URL must include a host")
+	}
```
thedadams: Correct me if I am wrong, but isn't this already checked above?

pkg/cli/login.go
```diff
@@ -4,13 +4,15 @@ import (
...
 )
 
 type Login struct {
-	NoExpiration bool `usage:"Set the token to never expire"`
-	ForceRefresh bool `usage:"Force refresh the token even if a valid one is cached"`
-	PrintToken   bool `usage:"Print the token to stdout after logging in"`
+	NoExpiration bool   `usage:"Set the token to never expire"`
+	ForceRefresh bool   `usage:"Force refresh the token even if a valid one is cached"`
+	PrintToken   bool   `usage:"Print the token to stdout after logging in"`
+	URL          string `usage:"Obot app URL to authenticate against"`
```
thedadams, line 15: Does it make sense to have `default:"http://localhost:8080/api"` here?

pkg/cli/logout.go
```diff
@@ -4,10 +4,12 @@ import (
 	"fmt"
 
 	"github.com/obot-platform/obot/pkg/cli/internal"
+	"github.com/obot-platform/obot/pkg/cli/internal/localconfig"
 	"github.com/spf13/cobra"
 )
 
 type Logout struct {
+	URL  string `usage:"Obot app URL whose stored credentials should be removed"`
```
thedadams, line 12: Same question.

pkg/cli/login.go
g-linville: It complicates the code a little bit if we do this. Because the `root.Client.BaseURL` is what handles the defaulting. So I think it's best to keep it as it is.

pkg/cli/internal/localconfig/store.go
g-linville: You're correct

thedadams (approved):

Issue #6578, opened 2026-05-12
Add explicit MCP server type semantics


calvinmclean:
We currently infer server behavior from fields like `catalogEntryID`, which makes it hard to distinguish catalog entries, user-scoped servers, and shared/multi-user servers as we prepare for multi-user catalog templates.

Add explicit server type semantics so the code can classify MCP catalog entries and MCP servers directly instead of relying on incidental field combinations.

## Goals
- Add explicit `serverType` semantics for regular MCP catalog entries ~~and MCP servers~~.
- Preserve existing behavior for current catalog entries and shared servers.
- Replace broad `catalogEntryID means single-user` assumptions with semantic helpers.
- Keep the UI behavior unchanged, with `serverType` hidden for now.
- Set up the codebase for a later `multiUserTemplate` catalog-entry type.

## Acceptance Criteria
- Existing catalog entries default to `singleUser`.
- ~~Existing shared/multi-user MCP servers are classified as `multiUser`.~~
- ~~New MCP servers persist an explicit effective server type.~~
- ~~API responses include effective server type without requiring a storage migration.~~
- Invalid or conflicting server type values are rejected.
- Shared-server create routes do not accept catalog-entry IDs until template deployment is implemented.
- UI classification/routing uses semantic helpers instead of raw `catalogEntryID` checks.
- Focused tests cover defaulting, validation, and classification behavior.

## Out of Scope
- No `multiUserTemplate` support yet.
- No deploy-from-template flow.
- No external secret binding behavior.

calvinmclean:
This is mostly internal refactors and not very testable. One piece, the new optional catalog entry field (`serverUserType: singleUser`) is testable but does not change any behavior.

shwetapsdet:
Completed testing the normal workflow for MCP servers, along with different server types such as single-user and remote MCP servers, on both the main environment and local Docker setup. All MCP server flows are working as expected.

PR #6587, opened 2026-05-12
feat: Add ServerUserType

calvinmclean:
## Summary

Addresses https://github.com/obot-platform/obot/issues/6578

**Refactor: make catalog entry server type explicit and clean up implicit single/multi-user assumptions throughout the codebase.**

- Add `ServerUserType` string type with `singleUser` and `multiUser` constants to `apiclient/types`
- Add `ServerUserType` field to `MCPServerCatalogEntryManifest` — declares whether a catalog entry is a template for single-user servers (each user installs their own) or multi-user servers (admin deploys centrally, users share via instances). Defaults to `singleUser` when empty.
- Validate `serverUserType` on catalog entry write — only `""` or `"singleUser"` accepted until multi-user template deployment is implemented
- Gate `CreateServer`: reject attempts to install a `multiUser` catalog entry via the single-user route (`POST /api/mcp-servers`)
- `MCPServerSpec.IsSingleUser()` — replaces scattered `MCPCatalogID == "" && PowerUserWorkspaceID == ""` checks with a single named method; `IsCatalogServer()`, `IsPowerUserWorkspaceServer()`, and `IsOwnedBy(userID)` added alongside it
- Replace all incomplete `MCPCatalogID == ""` single-condition checks with `IsSingleUser()` — these were bugs where workspace servers (PowerUserWorkspaceID != "") were incorrectly treated as single-user in `projectmcp.go`, `mcp.go`, and `mcpcatalog.go`
- Replace ACR routing if/else chains with `IsCatalogServer()`/`IsPowerUserWorkspaceServer()` across authz, handlers, and gateway
- `ServerUserType` intentionally kept **only** on `MCPServerCatalogEntryManifest`, not on `MCPServerSpec` — the MCPServer's type is already fully determined by `MCPCatalogID`/`PowerUserWorkspaceID`

### DIFF
--- a/pkg/controller/handlers/mcpcatalog/usergroupchange.go
+++ b/pkg/controller/handlers/mcpcatalog/usergroupchange.go
@@ -86,11 +86,6 @@ func (h *Handler) deleteUnauthorizedServersForUser(ctx context.Context, client k
 				continue
 			}
 
-			// Only handle single-user catalog entries here.
-			if !entry.Spec.Manifest.ServerUserType.IsSingleUser() {
-				continue
-			}
-
 			// Check access based on whether the catalog entry is in a workspace or regular catalog
 			if entry.Spec.PowerUserWorkspaceID != "" {
 				// Catalog entry is in a PowerUserWorkspace
--- a/pkg/controller/handlers/mcpcatalog/usergroupchange.go
+++ b/pkg/controller/handlers/mcpcatalog/usergroupchange.go
@@ -63,14 +63,13 @@ func (h *Handler) deleteUnauthorizedServersForUser(ctx context.Context, client k
 		}
 
 		// Skip multi-user servers - we only care about a user's own single-user servers
-		// Multi-user servers have MCPCatalogID or PowerUserWorkspaceID set
-		if server.Spec.MCPCatalogID != "" || server.Spec.PowerUserWorkspaceID != "" {
+		if !server.Spec.IsSingleUser() {
 			continue
 		}
 
-		// At this point, we only have single-user servers created from catalog entries
+		// At this point, we only have single-user servers created from catalog entries.
 		// For every server, fetch its catalog entry and check that the user has an ACR
-		// that gives them access to that catalog entry
+		// that gives them access to that catalog entry.
 		var (
 			hasAccess bool
 			err       error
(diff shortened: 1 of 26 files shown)


pkg/validation/mcpvalidators.go
```diff
@@ -988,6 +988,10 @@ func ValidateServerManifest(manifest types.MCPServerManifest, isMultiUser bool)
 }
 
 func ValidateCatalogEntryManifest(manifest types.MCPServerCatalogEntryManifest) error {
+	if !manifest.ServerUserType.IsSingleUser() {
+		return fmt.Errorf("unsupported serverUserType %q: only %q is currently supported for catalog entries", manifest.ServerUserType, types.ServerUserTypeSingleUser)
```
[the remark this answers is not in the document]
calvinmclean: The error message should drive users to be more specific when they are using a field incorrectly. If they don't know about it and leave it blank, that's fine.

g-linville (approved):
Thank you for doing this!

pkg/controller/handlers/mcpcatalog/usergroupchange.go
```diff
@@ -87,6 +86,11 @@ func (h *Handler) deleteUnauthorizedServersForUser(ctx context.Context, client k
 				continue
 			}
 
+			// Only handle single-user catalog entries here.
+			if !entry.Spec.Manifest.ServerUserType.IsSingleUser() {
+				continue
+			}
+
```
thedadams: nit: not needed because we checked that the server is single-user above.

pkg/storage/apis/obot.obot.ai/v1/mcpserver.go
```diff
@@ -127,6 +127,27 @@ type MCPServerSpec struct {
...
 
+// IsSingleUser returns true if this is a single-user MCP server.
+func (s MCPServerSpec) IsSingleUser() bool {
+	return s.MCPCatalogID == "" && s.PowerUserWorkspaceID == ""
+}
+
+// IsOwnedBy returns true if the given user directly owns this server.
+// Only true for single-user servers — multi-user servers are accessed via ACR, not direct ownership.
+func (s MCPServerSpec) IsOwnedBy(userID string) bool {
+	return s.UserID == userID && s.IsSingleUser()
```
thedadams: Technically speaking, I think we would say that a power user owns their multi-user servers.
```suggestion
	return s.UserID == userID && !s.IsCatalogServer()
```

thedadams (approved):

PR #6599, opened 2026-05-13
feat: add setup and skills commands to Obot CLI

g-linville:
part of https://github.com/obot-platform/prospect-and-customer-issues/issues/3

demo video: https://acorn-io.slack.com/archives/C03FU91U4SX/p1778705830185339

### DIFF
--- a/apiclient/skill.go
+++ b/apiclient/skill.go
@@ -35,13 +35,15 @@ func (c *Client) ListSkills(ctx context.Context, query string, limit int) (types
 	return result, err
 }
 
-func (c *Client) GetSkill(ctx context.Context, id string) (*types.Skill, error) {
+func (c *Client) GetSkill(ctx context.Context, id string) (types.Skill, error) {
 	_, resp, err := c.doRequest(ctx, http.MethodGet, fmt.Sprintf("/skills/%s", url.PathEscape(id)), nil)
 	if err != nil {
-		return nil, err
+		return types.Skill{}, err
 	}
 
-	return toObject(resp, &types.Skill{})
+	obj := types.Skill{}
+	_, err = toObject(resp, &obj)
+	return obj, err
 }
 
 func (c *Client) DownloadSkill(ctx context.Context, id string) ([]byte, error) {
--- a/apiclient/skill.go
+++ b/apiclient/skill.go
@@ -13,7 +13,7 @@ import (
 
 const MaxSkillDownloadBytes = 100 * 1024 * 1024
 
-var maxSkillDownloadBytes int64 = MaxSkillDownloadBytes
+var maxSkillDownloadBytes int64 = MaxSkillDownloadBytes // We are doing this for unit testing purposes
 
 func (c *Client) ListSkills(ctx context.Context, query string, limit int) (types.SkillList, error) {
 	values := url.Values{}
--- a/apiclient/skill.go
+++ b/apiclient/skill.go
@@ -11,6 +11,10 @@ import (
 	"github.com/obot-platform/obot/apiclient/types"
 )
 
+const MaxSkillDownloadBytes = 100 * 1024 * 1024
+
+var maxSkillDownloadBytes int64 = MaxSkillDownloadBytes
+
 func (c *Client) ListSkills(ctx context.Context, query string, limit int) (types.SkillList, error) {
 	values := url.Values{}
 	if query != "" {
@@ -53,5 +57,17 @@ func (c *Client) DownloadSkill(ctx context.Context, id string) ([]byte, error) {
 	}
 	defer resp.Body.Close()
 
-	return io.ReadAll(resp.Body)
+	if resp.ContentLength > maxSkillDownloadBytes {
+		return nil, fmt.Errorf("skill download exceeds maximum size of %d bytes", maxSkillDownloadBytes)
+	}
+
+	data, err := io.ReadAll(io.LimitReader(resp.Body, maxSkillDownloadBytes+1))
+	if err != nil {
+		return nil, err
+	}
+	if int64(len(data)) > maxSkillDownloadBytes {
+		return nil, fmt.Errorf("skill download exceeds maximum size of %d bytes", maxSkillDownloadBytes)
+	}
+
+	return data, nil
 }
--- a/apiclient/skill.go
+++ b/apiclient/skill.go
@@ -1,0 +1,55 @@
+package apiclient
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strconv"
+
+	"github.com/obot-platform/obot/apiclient/types"
+)
+
+func (c *Client) ListSkills(ctx context.Context, query string, limit int) (types.SkillList, error) {
+	values := url.Values{}
+	if query != "" {
+		values.Set("q", query)
+	}
+	if limit > 0 {
+		values.Set("limit", strconv.Itoa(limit))
+	}
+
+	path := "/skills"
+	if encoded := values.Encode(); encoded != "" {
+		path += "?" + encoded
+	}
+
+	_, resp, err := c.doRequest(ctx, http.MethodGet, path, nil)
+	if err != nil {
+		return types.SkillList{}, err
+	}
+
+	var result types.SkillList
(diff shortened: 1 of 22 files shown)


pkg/cli/internal/token.go
```diff
@@ -86,7 +86,6 @@ func Token(ctx context.Context, baseURL string, noExpiration, forceRefresh bool)
 		return "", tokenErr
 	}
 	if hasStoredToken && !forceRefresh && testToken(ctx, baseURL, token) {
-		fmt.Fprintln(os.Stderr, "Existing token is valid, no refresh needed")
```
g-linville, line 89: this got kinda spammy so I removed it

apiclient/skill.go
```diff
@@ -0,0 +1,73 @@
...
+	"net/http"
+	"net/url"
+	"strconv"
+
+	"github.com/obot-platform/obot/apiclient/types"
+)
+
+const MaxSkillDownloadBytes = 100 * 1024 * 1024
+
+var maxSkillDownloadBytes int64 = MaxSkillDownloadBytes
```
njhale: nit: We're doing this for testing purposes, but it would be nice to have a comment explaining that.

pkg/cli/setup.go
```diff
@@ -0,0 +1,221 @@
...
+				return fmt.Errorf("%s was selected but was not detected", installer.DisplayName())
+			}
+			shouldInstall = true
+		}
+		if !shouldInstall {
+			continue
+		}
+		result, err := installer.InstallBootstrap(ctx, home)
+		if err != nil {
+			return err
```
njhale: Do we want to attempt all installs, aggregate the errors, then post the result, or just continue to bail out early on the first install bootstrap failure?

pkg/cli/setup.go
```diff
@@ -0,0 +1,221 @@
...
+		return false, nil
+	default:
+		return false, fmt.Errorf("expected yes or no, got %q", raw)
+	}
+}
+
+func promptLine(cmd *cobra.Command, prompt string) (string, error) {
+	fmt.Fprint(cmd.OutOrStdout(), prompt)
+	var b strings.Builder
+	var one [1]byte
```
njhale, line 211:
```suggestion
	var (
	  b strings.Builder
	  one [1]byte
	)
```

Noticed both codex and opus fail to use block declarations quite often.

pkg/localagents/assets/claude/skills/obot-scan/SKILL.md.tmpl
```diff
@@ -0,0 +1,15 @@
...
+description: Scan local AI agent configuration and submit it to Obot.
+disable-model-invocation: true
+---
+
+# Run Obot Scan
+
+Run the Obot CLI scan command:
+
+```sh
+obot scan
```
njhale, line 12: nit: you may also want to tell it about `--dry-run` up front, since I'm not sure they want to submit a scan by default.

actually, in general, we may want to switch `obot scan` to dry-run by default and add a flag to make pushing it to an obot instance more explicit -- that's outside the scope of this PR though.

pkg/localagents/skillarchive.go
```diff
@@ -0,0 +1,317 @@
...
+	if err != nil {
+		return SkillArchive{}, fmt.Errorf("invalid skill ZIP: %w", err)
+	}
+	if len(reader.File) > maxSkillArchiveFiles {
+		return SkillArchive{}, fmt.Errorf("skill archive contains too many files (%d, max %d)", len(reader.File), maxSkillArchiveFiles)
+	}
+
+	entries := make([]SkillArchiveFile, 0, len(reader.File))
+	var totalUncompressed uint64
+	var totalRead int
```
njhale, line 55:
```suggestion
	var (
	  totalUncompressed uint64
	  totalRead int
	)
```

pkg/localagents/skillarchive.go
```diff
@@ -0,0 +1,317 @@
...
+	for _, file := range files {
+		if file.RelPath == skillformat.SkillMainFile {
+			return files, nil
+		}
+	}
+
+	// Also accept archives created by zipping the skill directory itself:
+	// <dir>/SKILL.md, <dir>/scripts/..., etc. That wrapper directory is not
+	// part of the installed skill, so all files must share one top-level dir.
+	root := ""
```
njhale:
```suggestion
	var root string
```

pkg/localagents/skillarchive_test.go
```diff
@@ -0,0 +1,276 @@
...
+			Name:    "wrapped/docs/readme.md",
+			Content: "# Read me\n",
+			Mode:    0644,
+		},
+	}), "fallback-name")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	got := map[string]bool{}
```
njhale: nit: idiomatic sets in go are `map[string]struct{}`.

njhale (approved):
I have some comments here, but nothing I would consider blockers.

pkg/cli/setup.go
g-linville: Yeah I'll change it to aggregate

pkg/cli/setup.go
g-linville: This is purely a style preference, right? Models seem to not do this and I really have no preference.

pkg/localagents/assets/claude/skills/obot-scan/SKILL.md.tmpl
g-linville: Yeah I think for now I am going to leave this slash command as it is. I was given a requirement to make `/obot-scan` be a slash command.

pkg/cli/skills.go
```diff
@@ -0,0 +1,233 @@
...
+		if len(installers) == 0 {
+			return nil, fmt.Errorf("no supported local agents detected")
+		}
+		return installers, nil
+	case localagents.ClaudeCodeAgentID:
+		return []localagents.DirectInstaller{localagents.NewClaudeCode()}, nil
+	case "all":
+		return nil, fmt.Errorf("unsupported --agent value %q; supported values are detected and claude-code", agent)
+	default:
+		return nil, fmt.Errorf("unsupported --agent value %q; supported values are detected and claude-code", agent)
```
calvinmclean, line 216: nit
```suggestion
	case "all":
		fallthrough
	default:
		return nil, fmt.Errorf("unsupported --agent value %q; supported values are detected and claude-code", agent)
```

pkg/localagents/skillarchive.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+	if len(reader.File) > maxSkillArchiveFiles {
+		return SkillArchive{}, fmt.Errorf("skill archive contains too many files (%d, max %d)", len(reader.File), maxSkillArchiveFiles)
+	}
+
+	entries := make([]SkillArchiveFile, 0, len(reader.File))
+	var totalUncompressed uint64
+	var totalRead int
+	for _, file := range reader.File {
+		if file.Mode()&fs.ModeSymlink != 0 {
```
calvinmclean: nit: I think this is a bit more clear
```suggestion
		if file.Mode().Type() == fs.ModeSymlink {
```

pkg/localagents/skillarchive.go
```diff
@@ -0,0 +1,317 @@
...
+}
+
+func hasColonPathSegment(rel string) bool {
+	for segment := range strings.SplitSeq(rel, "/") {
+		if strings.Contains(segment, ":") {
+			return true
+		}
+	}
+	return false
+}
```
calvinmclean: Would this just be the same as `return strings.Contains(rel, ":")`?

calvinmclean (approved):
just some nits but looks great :+1:

pkg/localagents/skillarchive.go
g-linville: lol you're right. We'll go with that

pkg/cli/setup.go
njhale: I suppose it is, but I believe the block style is pretty prevalent in the rest of the codebase.

I won't hold this PR on this.

njhale (approved):

Issue #6590, opened 2026-05-13
MCP server shows connected state and prompts for unavailable credentials when required secret bindings are missing.


sangee2004:
**Describe the bug**
When an MCP server has required Secret Bindings but the corresponding Kubernetes secret does not exist, the user correctly sees a missing configuration error during connection. However, the modal still shows an `Update Configuration and Try Again ' option even though there are no credentials for the user to provide. After closing the modal, the MCP server incorrectly appears in the `Connected state` even though the required secret-backed configuration is missing.

**To Reproduce**
1. Create an MCP server entry with `Secret Bindings` for a required parameter like - https://github.com/sangee2004/mcp-catalog/blob/main/tavily_search.yaml in the catalog repo.
2. Add this MCP catalog repo as `Source URL`.
3. Do not have the secret relating to  the`Secret Bindings` section created in the K8S cluster 
4. As any user who has access to this MCP server, connect to the MCP server
Notice that the user is presented with a missing configuration error and an `Update Configuration and Try Again` button. 
<img width="990" height="349" alt="Image" src="https://github.com/user-attachments/assets/f998d100-04c2-472c-91a5-7a82afa4fb2e" />
 
5. Clicking on this update option does not prompt the user to enter any credentials, which is as expected, and reports the same failure again.

<img width="990" height="349" alt="Image" src="https://github.com/user-attachments/assets/f998d100-04c2-472c-91a5-7a82afa4fb2e" />

6. Close the modal.
When the user is brought back to the `MCP Servers` page, notice that the MCP server is presented in `Connected` state when a required parameter was never provided, in this case.

At this time, no deployment has been created for the MCP server.

<img width="808" height="242" alt="Image" src="https://github.com/user-attachments/assets/4efd162d-3601-4139-8b97-31ed5b0c525e" />

https://www.loom.com/share/43b566daac074d239c8ae7e863218019

**Expected behavior**
When there is an attempt made to start the MCP server with secrets relating to  `Secret Bindinds` for a required parameter missing, the user should not be prompted with the `Update Configuration and Try Again` option. Closing the modal should not cause the MCP server to be shown in `Connected` state.

**Note** The same behavior is also seen when deploying a single-user MCP server with the required parameter served by Secret bindings when secrets are not present - https://github.com/sangee2004/mcp-catalog/blob/main/exa_search.yaml

[...]

Issue #6589, opened 2026-05-13
Server requires update state shown when secret bindings are missing.


sangee2004:
**Describe the bug**
When an MCP server is configured using Secret Bindings and the corresponding Kubernetes secrets are later deleted, the server transitions into a Server requires an Update state. However, users are not provided with an Update Server option or any actionable recovery path, leaving the server stuck in an unusable state. Instead of showing an update-required state, the UI should display an explicit error indicating that the required configuration secret is missing.

**To Reproduce**
1. Create an MCP server entry with `Secret Bindinds` for a required parameter like - https://github.com/sangee2004/mcp-catalog/blob/main/tavily_search.yaml in the catalog repo.
2. Add this MCP catalog repo as `Source URL`.
3. Have the secret relating to  the`Secret Bindings` section created in the K8S cluster 
4. As any user who has access to this MCP server, connect to the MCP server
MCP server is deployed successfully as expected, without prompting the user for credentials.
5. Delete the secrets relating to the `Secret Bindings` section in the K8S cluster 

This results in the server being presented in `Server requires an Update` state. 
<img width="1280" height="267" alt="Image" src="https://github.com/user-attachments/assets/c4bbe29e-0718-4f4c-8d6e-d5b7fcd3bac0" />

Notice that there is no `Update Server` option presented here since the user cannot do anything to get out of this state.
<img width="1236" height="446" alt="Image" src="https://github.com/user-attachments/assets/5c2f09aa-3852-419c-a761-509f5915d66e" />

**Expected behavior**
The user should be presented with an error state instead of `Server requires an Update` state, indicating that a required configuration parameter is missing in this case

calvinmclean:
Adding on to this, the Server Details page shows both errors:

<img width="1182" height="264" alt="Image" src="https://github.com/user-attachments/assets/c913550e-1a2d-4e83-b177-879060f70237" />

The second one is inaccurate since it is not really a user configuration. As part of this fix, I will move the "Server details and logs are temporarily unavailable as a result." string into the K8s warning and not show the other warning.

sangee2004:
This issue is addressed when testing with the latest build from `main`.

User is presented with `Missing Kubernetes Secret` state.
<img width="1083" height="281" alt="Image" src="https://github.com/user-attachments/assets/cc2ce523-b839-40a0-b730-c2491d03fadd" />

Issue #6586, opened 2026-05-12
MCP servers with secret bindings incorrectly show needs config update state


sangee2004:
**Describe the bug**
MCP servers deployed from catalog entries that use Secret Bindings are incorrectly marked as Needs Config Update immediately after successful deployment, even though no configuration changes are required. Non-admin users attempting the update receive a 400 entry does not belong to workspace error, while admin users can trigger the update flow but the server continues to remain in the Needs Config Update state afterward. The issue appears to be related to how config diffs are generated or validated for deployments using cluster-backed secret bindings.

**To Reproduce**
1. Create an MCP server entry with `Secret Bindinds`  like - https://github.com/sangee2004/mcp-catalog/blob/main/exa_search.yaml in the catalog repo.
2. Add this mcp catalog repo as `Source URL`
3. Have the secret relating to the `Secret Bindings` section created in the K8S cluster 
4. As any user who has access to this MCP server, Connect to the MCP server
MCP server is deployed successfully as expected, without prompting the user for credentials.

5. Go to the `Server Details` page

Notice that the server is shown as being in an update-required state
<img width="1029" height="399" alt="Image" src="https://github.com/user-attachments/assets/1de3a618-7510-49d7-a5f2-064d62d2d045" />

When the user actually attempts an update, he is presented with a 400 error - `400 /workspaces/puw1-4/entries/default-sangee-exa-search-489f06d5/servers: entry does not belong to workspace
`
<img width="1027" height="847" alt="Image" src="https://github.com/user-attachments/assets/7b3c11e0-5cfd-4f6b-b868-8e10c0e831ac" />

This diff is presented when the `View Diff` option is used

<img width="1320" height="414" alt="Image" src="https://github.com/user-attachments/assets/5b203a2b-a04d-44a4-bdf1-ab634b7c2ed6" />

6. Log in as `Admin` user and go to the `Deployments and Connections` tab

Notice that MCP server entries are presented in `Needs Config Update` state.

<img width="1234" height="524" alt="Image" src="https://github.com/user-attachments/assets/9d059cff-6ee4-43ba-a9df-104ae0eeee59" />

Attempting to update does not error out like in step 5, but even after this update, the entries continue to be in `Needs Config Update` state.

Note - This issue is also seen with remote MCP server entries that use `Secret Bindinds` for headers.

**Expected behavior**
The MCP server should not be in the `Needs Config Update` state after deployment.

sangee2004:
[...]

PR #6605, opened 2026-05-14
fix: issues with UI and behaviors for secret binding

calvinmclean:
### Addresses #6586
- Fixed false positive config-update state for MCP servers with unchanged secret bindings.
- Secret binding structs are now compared by value instead of pointer identity when checking catalog-entry drift.

### Addresses #6589
- Show missing Kubernetes Secret state instead of generic config/update messaging for secret-bound MCP servers.
- Fixed the MCP server list to show the missing-secret warning state for affected single-user servers.
- Fixed the user detail/server-instances view:
  `/mcp-servers/c/default-reflect-npx-bound-templated-args-c9b4b4a0/instance/ms14wtrt?view=server-instances`
- Fixed the admin server details view:
  `/admin/mcp-servers/c/default-reflect-npx-bound-templated-args-c9b4b4a0/instance/ms14wtrt/details`
### Addresses #6590
- Removed the misleading `Update Configuration and Try Again` action when the failure is caused only by missing secret-bound config.
- After `/configure` reports missing secret-bound required config, skip the follow-up `/launch` request so users do not see the raw launch 400.
- Keep the cleanup action as `Cancel and Delete Server` because the single-user server has already been created by that point.

Screenshot 2026-05-15 at 10 32 26
Screenshot 2026-05-15 at 10 32 33
Screenshot 2026-05-15 at 10 32 45
Screenshot 2026-05-15 at 10 37 28
Screenshot 2026-05-15 at 10 41 11
Screenshot 2026-05-15 at 10 42 29

### DIFF
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -2229,7 +2229,12 @@ func (m *MCPHandler) configureCompositeServer(req api.Context, compositeServer v
 		return fmt.Errorf("failed to generate slug: %w", err)
 	}
 
-	return req.Write(ConvertMCPServer(compositeServer, nil, m.serverURL, slug, components...))
+	mergedEnv, err := mcp.MergeBoundCreds(req.Context(), req.LocalK8sClient, req.ObotNamespace, compositeServer.Spec.Manifest.Env, compositeServer.Spec.Manifest.RemoteConfig, nil)
+	if err != nil {
+		return fmt.Errorf("failed to resolve secret bindings: %w", err)
+	}
+
+	return req.Write(ConvertMCPServer(compositeServer, mergedEnv, m.serverURL, slug, components...))
 }
 
 func kickMCPServerControllers(req api.Context, server *v1.MCPServer) error {
@@ -2802,6 +2807,12 @@ func ConvertMCPServer(server v1.MCPServer, credEnv map[string]string, serverURL,
 			componentServers   = server.Spec.Manifest.CompositeConfig.ComponentServers
 			disabledComponents = make(map[string]bool, len(componentServers))
 		)
+		if credEnv != nil {
+			converted.MissingRequiredEnvVars, converted.MissingRequiredHeaders = secretBoundMissingConfig(converted)
+		} else {
+			converted.MissingRequiredEnvVars = nil
+			converted.MissingRequiredHeaders = nil
+		}
 		for _, comp := range componentServers {
 			if id := comp.ComponentID(); id != "" {
 				disabledComponents[id] = comp.Disabled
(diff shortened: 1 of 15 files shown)


pkg/api/handlers/mcp.go
```diff
@@ -2802,25 +2807,65 @@ func ConvertMCPServer(server v1.MCPServer, credEnv map[string]string, serverURL,
...
+		}
 		for _, comp := range componentServers {
 			if id := comp.ComponentID(); id != "" {
 				disabledComponents[id] = comp.Disabled
 			}
 		}
 
 		for _, component := range components {
-			if component.CatalogEntryID != "" && disabledComponents[component.CatalogEntryID] || component.Configured {
+			if (component.CatalogEntryID != "" && disabledComponents[component.CatalogEntryID]) || component.Configured {
```
thedadams: nit: ineffectual change

pkg/api/handlers/mcp.go
```diff
@@ -2802,25 +2807,65 @@ func ConvertMCPServer(server v1.MCPServer, credEnv map[string]string, serverURL,
...
 			converted.Configured = false
-			break
 		}
 	}
 
 	return converted
 }
 
+func secretBoundMissingConfig(server types.MCPServer) (missingEnvVars, missingHeaders []string) {
+	missingEnvKeys := map[string]struct{}{}
```
thedadams: nit and same below:
```suggestion
	missingEnvKeys := make(map[string]struct{}, len(server.MissingRequiredEnvVars)
```

thedadams (approved):
Backend looks good.

njhale (approved):

pkg/api/handlers/mcp.go
```diff
@@ -2835,7 +2835,7 @@ func ConvertMCPServer(server v1.MCPServer, credEnv map[string]string, serverURL,
 }
 
 func secretBoundMissingConfig(server types.MCPServer) (missingEnvVars, missingHeaders []string) {
-	missingEnvKeys := map[string]struct{}{}
+	missingEnvKeys := make(map[string]struct{})
```
thedadams: The point of making this change was to add the length so you only do one allocation

ui/user/src/lib/services/chat/types.ts
```diff
@@ -797,6 +797,7 @@ export interface MCPCatalogServer {
 	configured: boolean;
 	catalogEntryID: string;
 	missingRequiredEnvVars: string[];
+	missingRequiredHeader?: string[];
```
ivyjeong13, line 800: Seeing `missingRequiredHeader` and `missingRequiredHeaders` both in the type here and wondering if one can be axed?

pkg/api/handlers/mcp.go
calvinmclean: Oops! Fixed now

ui/user/src/lib/services/chat/types.ts
calvinmclean: Yep, removed it! The backend has `missingRequiredHeader` as the json name of the field

ui/user/src/lib/components/admin/McpServerInstances.svelte
```diff
@@ -289,10 +316,21 @@
...
 		>
 			{#snippet onRenderColumn(property, d)}
+				{@const missingKubernetesSecret = isMissingKubernetesSecret(d)}
 				{#if property === 'url'}
 					<span class="flex items-center gap-1">
 						{d.manifest.remoteConfig?.url}
-						{#if d.needsUpdate}
+						{#if missingKubernetesSecret}
+							<div
+								class="text-yellow-500"
```
ivyjeong13: `text-warning` instead of `text-yellow-500`

ui/user/src/lib/components/admin/McpServerInstances.svelte
```diff
@@ -307,7 +345,17 @@
 					<span class="flex items-center gap-1">
 						{d.userDisplayName || 'Unknown User'}
 						{#if type === 'single' || type === 'composite'}
-							{#if d.needsUpdate}
+							{#if missingKubernetesSecret}
+								<div
+									class="text-yellow-500"
+									use:tooltip={{
```
ivyjeong13, line 351: can do same here! `text-warning`

ui/user/src/lib/components/admin/McpServerInstances.svelte
```diff
@@ -328,15 +376,16 @@
...
 					{#if auditLogsUrl}
 						<a class="btn btn-link" href={resolve(auditLogsUrl as `/${string}`)}>
 							View Audit Logs
 						</a>
 					{/if}
 
-					{#if d.needsUpdate}
-						<DotDotDot class="hover:dark:bg-base-100/50">
+					{#if d.needsUpdate && !missingKubernetesSecret}
+						<DotDotDot class="icon-button hover:dark:bg-background/50">
```
ivyjeong13: `class="icon-button hover:dark:bg-base-100/50"` Sorry, no more `bg-background` with Daisy UI base vars 🙏

ivyjeong13 (approved):
Left some nits for styling but UI LGTM!

PR #6610, opened 2026-05-15
feat: allow specifying a branch for git catalogs

thedadams:


thedadams:
This already works, I was just doing it wrong.

PR #6611, opened 2026-05-15
enhance: separate resource requests for agents, remotes, and others

thedadams:


pkg/mcp/kubernetes.go
```diff
@@ -1126,21 +1120,77 @@ func (k *kubernetesBackend) deleteDeploymentCache(mcpServerName string) {
 	delete(k.deploymentCache, mcpServerName)
 }
 
+func mcpContainerResources(server ServerConfig, k8sSettings v1.K8sSettingsSpec) corev1.ResourceRequirements {
+	var defaults corev1.ResourceRequirements
+	if server.Runtime == types.RuntimeRemote {
+		defaults = memoryRequestResources(remoteMemoryRequest)
+	} else if server.NanobotAgentName != "" {
+		if k8sSettings.NanobotAgentResources != nil {
+			defaults = withDefaultCPURequest(*k8sSettings.NanobotAgentResources)
```
calvinmclean, line 1129: If a `k8sSettings.NanobotAgentResources` only has CPU set, would this result in a ResourceRequirement with no memory limits?

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -652,6 +648,12 @@ func readCatalogDirectory[T any](catalog string) ([]T, error) {
 			return nil
 		}
 
+		// Check file count limit
+		fileCount++
```
njhale, line 652: 👍 for the stealth bug fix here

pkg/mcp/kubernetes.go
thedadams: Yes. My understanding is that is how things work with the Kubernetes resource settings prior to this change.

pkg/mcp/kubernetes.go
```diff
@@ -1312,27 +1362,8 @@ func (k *kubernetesBackend) patchDeploymentWithK8sSettings(ctx context.Context,
...
-		if len(k8sSettings.Resources.Limits) > 0 {
-			resourcesMap["limits"] = k8sSettings.Resources.Limits
-		}
-		if len(k8sSettings.Resources.Requests) > 0 {
-			resourcesMap["requests"] = k8sSettings.Resources.Requests
-		}
-		mcpContainerPatch["resources"] = resourcesMap
-	} else {
-		// Use $patch: delete to remove any existing resources
-		mcpContainerPatch["resources"] = map[string]any{
```
njhale, line 1332: we dropped the delete case here, though I think that may have been intentional since we're now always passing defaults down. Is that correct?

njhale (approved):

pkg/mcp/kubernetes.go
calvinmclean: Ok. Yeah I suppose that makes sense if a user explicitly sets the resources and excludes part of it, it means that it is intentionally unset

calvinmclean (approved):

pkg/mcp/kubernetes.go
thedadams: I'll put it back just in case.

Issue #6591, opened 2026-05-13
Remove legacy chat from UI


ivyjeong13:
* We're going to start deleting our legacy code in order to trim down dependencies to make managing cves a bit easier.
* Delete all the frontend code related to legacy chat as well as any bits of the admin side that account for legacy chat being turned on

sangee2004:
Legacy Chat is removed from UI when testing with the latest build from `main`.

Tested with obot installed with legacy chat enabled `--set config.OBOT_SERVER_DISABLE_LEGACY_CHAT=false`.

Owner/Admin users are not presented with the following`Legacy Chat Management` section.
Non-admin users are not presented with the `Launch Chat` option.

When connecting to an MCP server, the connect modal does not include the `Chat` option that previously launched the legacy chat.

PR #6615, opened 2026-05-15
chore: remove legacy chat from ui

ivyjeong13:
Addresses #6591
* removes legacy chat components in relation to chat, projects, assistants, project sharing, etc.
* with removal of legacy chat operations, ChatService no longer made sense --> renamed to UserService instead
* check for any operations/types in AdminService that should be moved to UserService (since some operations have been opened up for all users, for example listAllUsers)
* reordered operations.ts/types.ts alphabetically/grouped by scope


ivyjeong13:
> Copilot wasn't able to review this pull request because it exceeds the maximum number of lines (20,000). Try reducing the number of changed lines and requesting a review from Copilot again.

🥲

ui/user/src/lib/services/admin/operations.ts
g-linville, line 1: It could be cool to add a comment to the top of this file explaining the alphabetic structure and instructing agents to maintain it when they make edits.

ui/user/src/lib/services/admin/types.ts
g-linville, line 1: Same here.

g-linville (approved):

ui/user/src/lib/services/admin/operations.ts
ivyjeong13: Great idea! Will do! : )

thedadams:
I pulled this down and tested with my changes to remove the backend code and everything worked really well!

One thing to change here: the UI still looks for the `nanobotIntegration` field from the `/api/version` endpoint and gates some stuff on that. I think we can remove that field from the UI now.

ivyjeong13:
> I pulled this down and tested with my changes to remove the backend code and everything worked really well!
> 
> One thing to change here: the UI still looks for the `nanobotIntegration` field from the `/api/version` endpoint and gates some stuff on that. I think we can remove that field from the UI now.

Thanks Donnie! `nanobotIntegration` check has been removed!

thedadams (approved):

Issue #6580, opened 2026-05-12
Deploy multi-user servers from catalog templates


calvinmclean:
Allow admins to deploy one or more shared/multi-user `MCPServer` instances from a `multiUserTemplate` catalog entry.

This builds on explicit server type semantics and multi-user catalog template ingestion by adding the lifecycle flow that turns a template into deployed shared server instances, including the UI needed for admins to find templates, distinguish them from deployed servers, and deploy them.

## Dependencies
- Depends on #6578.
- Depends on #6579.

## Goals
- Add a deploy-from-template backend flow for `multiUser` catalog entries.
- Add the admin UI needed to discover and deploy multi-user templates.
- Clearly disambiguate multi-user templates from deployed multi-user `MCPServer` instances in the UI.
- Allow multiple deployed `MCPServer` instances from the same template.
- Capture deploy-time configuration values on the deployed server, not on the template.
- Preserve a reference from deployed servers back to the source catalog entry.
- Mark deployed servers when their source template has an update available.
- Provide a way for admins to apply template updates to deployed servers.

## Acceptance Criteria
- Admins can find `multiUser` catalog entries in the UI.
- The UI clearly distinguishes templates from deployed multi-user servers.
- Admins and PowerUserPlus can deploy a shared/multi-user `MCPServer` from a `multiUser` catalog entry in the UI.
- Deploying a template does not mutate the source catalog entry.
- Multiple deployed servers can reference the same template.
- Deploy-time config is required, validated, and stored only on deployed servers.
- Changes to the source template mark deployed servers as needing update.

## Out of Scope
- External secret binding behavior.
- Automatic rollout of template updates.
- Major UI redesign beyond the screens/actions needed to deploy and manage template-backed multi-user servers.

calvinmclean:
https://github.com/calvinmclean/mcp-catalog-multi-user

Example catalog

sangee2004:
This issue is addressed when testing with the latest build from `main`.

- Admin user is able to deploy a multi-user MCP server defined in a catalog that other users are able to connect to successfully.


  1. Create a Multiuser MCP server entry in catalog - https://github.com/sangee2004/mcp-catalog/blob/main/sangeemultiexasearch.yaml with `serverUserType: multiUser`
  2. Log in as an admin user and add this catalog as `Registry Source`.
  3. Once the Multiuser MCP server entry gets synced, it shows up as `Multi-User Catalog Entry`.
[...]

Issue #6579, opened 2026-05-12
Add multi-user catalog templates


calvinmclean:
Depends on #6578.

Allow git catalog entries to define deployable multi-user server templates without automatically deploying them.

This builds on explicit MCP server type semantics by introducing a catalog-entry type that represents a template for shared/multi-user MCP servers, distinct from deployed `MCPServer` instances.

## Goals
- Add `multiUserTemplate` as a valid catalog-entry `serverType`.
- Ingest multi-user templates from git catalogs as catalog entries only.
- Do not automatically create or deploy `MCPServer` objects from templates.
- Let templates declare required deploy-time configuration without storing deploy-time values.
- Keep existing `singleUser` catalog-entry behavior unchanged.
- Expose enough API/UI metadata to distinguish templates from deployed multi-user servers.

## Acceptance Criteria
- Git catalog sync accepts catalog entries with `serverUserType: multiUser`.
- Synced multi-user templates are stored as catalog entries, not deployed servers.
- Multi-user templates are clearly distinguishable from single-user catalog entries in API responses.
- Existing single-user, remote, and composite catalog entries continue to behave as they do today.

## Out of Scope
- No deploy-from-template flow.
- No template update/apply workflow.
- No external secret binding behavior.
- No final UI/UX for template deployment.

sangee2004:
This issue is addressed when testing with the latest build from `main`.

Able to deploy and connect to a multi-user MCP server that is defined as a catalog entry.

1. Create a Multiuser MCP server entry in catalog - https://github.com/sangee2004/mcp-catalog/blob/main/sangeemultiexasearch.yaml with `serverUserType: multiUser`
2. Log in as an admin user and add this catalog as `Registry Source`.
3. Once the Multiuser MCP server entry gets synced, it shows up as `Multi-User Catalog Entry`.
<img width="1234" height="262" alt="Image" src="https://github.com/user-attachments/assets/97d668f0-51c8-45fe-964b-27ecdcc1419a" />

4. Use the `Connect to Server` option.

This results in a `Multi-User` MCP server entry being created, which is shown as being in `Connected` state.
<img width="1257" height="314" alt="Image" src="https://github.com/user-attachments/assets/336ea260-7da2-4370-a886-93ac3767fdc7" />


Issues found when testing this feature will be tracked in separate issues
https://github.com/obot-platform/obot/issues/6757
https://github.com/obot-platform/obot/issues/6758
https://github.com/obot-platform/obot/issues/6759

PR #6627, opened 2026-05-18
feat: Multi-User catalog templates

calvinmclean:
## Summary

Addresses #6579, #6580

Adds support for multi-user MCP server templates — catalog entries with serverUserType: multiUser that admins (or PowerUserPlus users) can deploy as shared servers. Multiple servers can be deployed from a single template, and users connect by creating instances on the deployed server.

### Backend:
  - Allow serverUserType: multiUser on non-composite catalog entries (npx, uvx, containerized, remote)
  - Backfill existing entries with serverUserType: singleUser for backward compatibility
  - Validate serverUserType rejects unknown values
  - Drift detection for runtime config, env vars, and MultiUserConfig (user-defined headers)
  - validateEntryScope helper for consistent scope validation — workspace routes can access global catalog entries so PUP
  users can deploy/operate on global templates
  - validateServerScope helper ensures catalog/workspace-scoped server routes reject cross-scope requests
  - Template-faithful deployment: multi-user template servers use the catalog entry manifest without admin overrides
  - Track deployment count (UserCount) for multi-user templates

### Frontend:
  - Separate registry (templates) and deployments views for multi-user servers
  - PowerUserPlus users can deploy multi-user templates from global catalog into their workspace
  - Connect flow: deploying a template creates a new server; clicking an existing deployment creates an instance
  - Diff dialog strips serverUserType to avoid spurious noise (full diff preview is a follow-up)
  - Fix routing throughout for multi-user servers: server detail pages, tool previews, disconnect, update, delete
  - Fix bulk update and trigger-update to use catalog/workspace-scoped routes for multi-user servers
  - Tool preview generation uses dryRun mode and catalogEntryID for non-editable deployed servers


ui/user/src/lib/components/mcp/DeploymentsView.svelte
```diff
@@ -254,13 +261,13 @@
...
-			await mcpServersAndEntries.refreshAll();
+			mcpServersAndEntries.refreshAll();
 			// Refresh capacity banner when server list changes
 			if (!isInitialLoad) {
 				capacityBanner?.refresh();
 			}
 		} else if (!isInitialLoad && entity === 'workspace') {
-			await mcpServersAndEntries.refreshAll();
+			mcpServersAndEntries.refreshAll();
 		}
```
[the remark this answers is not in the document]
calvinmclean: IDE told me that await didn't work here, but doesn't seem to be an issue adding it back

pkg/api/handlers/mcp.go
```diff
@@ -1907,16 +1940,22 @@ func (m *MCPHandler) UpdateServer(req api.Context) error {
...
 
-	if !server.Spec.IsSingleUser() {
-		return types.NewErrBadRequest("cannot update alias for a multi-user MCP server")
+	if server.Spec.MCPCatalogID != catalogID || server.Spec.PowerUserWorkspaceID != workspaceID {
+		return types.NewErrNotFound("MCP server not found")
+	}
+
+	if catalogID == "" && workspaceID == "" && !server.Spec.IsSingleUser() {
+		return types.NewErrBadRequest("use the catalog or workspace scoped route to update a shared MCP server alias")
 	}
```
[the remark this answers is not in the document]
calvinmclean: This is existing and just moved around. The behavior is correct.

ui/user/src/lib/components/admin/DiffDialog.svelte
```diff
@@ -13,6 +13,14 @@
 
 	let { fromServer, toServer }: Props = $props();
 
+	// Strip fields that exist only on one manifest type (MCPServerManifest vs MCPServerCatalogEntryManifest)
+	// to avoid spurious diff noise. A proper fix would compare equivalent types via a backend preview endpoint.
+	function stripNonDiffFields(manifest: unknown): unknown {
+		if (!manifest || typeof manifest !== 'object') return manifest;
+		const obj = manifest as Record<string, unknown>;
+		return Object.fromEntries(Object.entries(obj).filter(([key]) => key !== 'serverUserType'));
+	}
```
calvinmclean: note to self: this will conflict with https://github.com/obot-platform/obot/pull/6635

pkg/api/handlers/registry/handler.go
```diff
@@ -852,6 +865,10 @@ func (h *Handler) findMCPServerCatalogEntry(req api.Context, entryName string, r
 	return ConvertMCPServerCatalogEntryToRegistry(req.Context(), entry, h.serverURL, reverseDNS, h.mimeFetcher)
 }
 
+func hideMultiUserTemplateFromRegistry(req api.Context, entry v1.MCPServerCatalogEntry) bool {
+	return !req.UserIsAdmin() && !req.UserIsPowerUserPlus() && !entry.Spec.Manifest.ServerUserType.IsSingleUser()
```
thedadams: Admins are power user plus
```suggestion
	return !req.UserIsPowerUserPlus() && !entry.Spec.Manifest.ServerUserType.IsSingleUser()
```

pkg/api/handlers/registry/utils_test.go
```diff
@@ -104,3 +109,61 @@ func TestFormatRegistryServerName(t *testing.T) {
...
+		},
+		{
+			name:  "basic user can see single-user entries",
+			user:  &kuser.DefaultInfo{Groups: []string{types.GroupAuthenticated}},
+			entry: singleUserEntry,
+			want:  false,
+		},
+		{
+			name:  "admin can see multi-user templates",
+			user:  &kuser.DefaultInfo{Groups: []string{types.GroupAdmin}},
```
thedadams: Something like this is more realistic to the system.
```suggestion
			user:  &kuser.DefaultInfo{Groups: types.RoleAdmin.Groups()},
```

pkg/api/handlers/mcp.go
```diff
@@ -160,6 +167,12 @@ func (m *MCPHandler) ListEntriesFromAllSources(req api.Context) error {
 	return req.Write(types.MCPServerCatalogEntryList{Items: entries})
 }
 
+// hideMultiUserTemplateFromAllMCPEntries determines if users should be able to see a CatalogEntry based on whether it
+// is single or multi-user. Only Admin and PowerUserPlus should see these entries
+func hideMultiUserTemplateFromAllMCPEntries(req api.Context, entry v1.MCPServerCatalogEntry) bool {
+	return !req.UserIsAdmin() && !req.UserIsPowerUserPlus() && !entry.Spec.Manifest.ServerUserType.IsSingleUser()
+}
```
thedadams: Do we need a separate implementation of this?

pkg/api/handlers/mcp.go
```diff
@@ -3016,6 +3061,10 @@ func (m *MCPHandler) GetServerDetails(req api.Context) error {
 		return err
 	}
 
+	if err := validateServerScope(req, server); err != nil {
+		return err
+	}
```
thedadams, line 3076: A version of this was already done above on line 3044

pkg/api/handlers/mcp.go
```diff
@@ -1544,12 +1569,20 @@ func (m *MCPHandler) CreateServer(req api.Context) error {
 			return types.NewErrBadRequest("%v", err)
 		}
 
+		// Verify the catalog entry belongs to the target scope.
+		if err := validateEntryScopeForRead(catalogEntry, catalogID, workspaceID); err != nil {
```
thedadams: Do we need this and the function above?

pkg/api/handlers/mcp.go
```diff
@@ -3602,13 +3659,25 @@ func (m *MCPHandler) UpdateURL(req api.Context) error {
...
 	if !server.Spec.IsSingleUser() {
-		return types.NewErrBadRequest("cannot trigger update for a multi-user MCP server; use the UpdateServer endpoint instead")
+		// Multi-user servers deployed from templates can be updated, but only
+		// through catalog- or workspace-scoped routes.
+		if server.Spec.MCPServerCatalogEntryName == "" {
+			return types.NewErrBadRequest("cannot trigger update for a multi-user MCP server without a catalog entry")
+		}
+		if server.Spec.MCPCatalogID != catalogID || server.Spec.PowerUserWorkspaceID != workspaceID {
+			return types.NewErrNotFound("MCP server not found")
+		}
```
thedadams: Can `validateServerScope` be used here?

pkg/api/handlers/mcp.go
```diff
@@ -3625,19 +3694,17 @@ func (m *MCPHandler) TriggerUpdate(req api.Context) error {
...
-		// (We already verified this is a single-user server at the check above.)
-		if !server.Spec.IsOwnedBy(req.User.GetUID()) {
-			// Workspace-based authorization for power user workspace entries
-			workspaceID := req.PathValue("workspace_id")
-			if workspaceID == "" {
-				return types.NewErrNotFound("MCP server %s not found", server.Name)
-			}
+	// Multi-user template-deployed servers require admin or workspace access.
+	if !req.UserIsAdmin() && !server.Spec.IsSingleUser() {
+		if workspaceID == "" || server.Spec.PowerUserWorkspaceID != workspaceID {
```
thedadams: The idea is that this should always match, even if they are both empty.
```suggestion
		if server.Spec.PowerUserWorkspaceID != workspaceID {
```

pkg/api/handlers/mcp.go
```diff
@@ -3625,19 +3694,17 @@ func (m *MCPHandler) TriggerUpdate(req api.Context) error {
...
+			return types.NewErrNotFound("MCP server %s not found", server.Name)
+		}
+	}
 
-			if entry.Spec.PowerUserWorkspaceID != workspaceID {
-				return types.NewErrNotFound("MCP server %s not found", server.Name)
-			}
+	// Allow users to upgrade their own single-user servers.
+	if !req.UserIsAdmin() && server.Spec.IsSingleUser() && !server.Spec.IsOwnedBy(req.User.GetUID()) {
+		if workspaceID == "" || entry.Spec.PowerUserWorkspaceID != workspaceID {
```
thedadams: Same.

pkg/api/handlers/mcp_test.go
```diff
@@ -40,6 +42,62 @@ func TestConvertMCPResources(t *testing.T) {
...
+
+	tests := []struct {
+		name  string
+		user  kuser.Info
+		entry v1.MCPServerCatalogEntry
+		want  bool
+	}{
+		{
+			name:  "regular user cannot see multi-user templates",
+			user:  &kuser.DefaultInfo{Groups: []string{types.GroupAuthenticated}},
```
thedadams: Same comment about using roles to get groups here.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -72,6 +72,56 @@ func (*MCPCatalogHandler) List(req api.Context) error {
 	})
 }
 
+// checkEntryAccess verifies that the requesting user has ACR access to the entry and that
+// the entry doesn't require unconfigured static OAuth credentials. Returns NotFound to avoid
+// leaking entry existence. Admins and auditors bypass these checks.
+func (h *MCPCatalogHandler) checkEntryAccess(req api.Context, entry v1.MCPServerCatalogEntry) error {
+	if req.UserIsAdmin() || req.UserIsAuditor() {
```
thedadams: I don't think auditors should have this permission.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -72,6 +72,56 @@ func (*MCPCatalogHandler) List(req api.Context) error {
...
+	)
+	if entry.Spec.MCPCatalogName != "" {
+		hasAccess, err = h.acrHelper.UserHasAccessToMCPServerCatalogEntryInCatalog(req.User, entry.Name, entry.Spec.MCPCatalogName)
+	} else if entry.Spec.PowerUserWorkspaceID != "" {
+		hasAccess, err = h.acrHelper.UserHasAccessToMCPServerCatalogEntryInWorkspace(req.Context(), req.User, entry.Name, entry.Spec.PowerUserWorkspaceID)
+	}
+	if err != nil {
+		return err
+	}
+	if !hasAccess || entryRequiresStaticOAuthCreds(entry) {
```
thedadams: I think it would be better to check `entryRequiresStaticOAuthCreds(entry)` before doing the more expensive access checks.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -72,6 +72,56 @@ func (*MCPCatalogHandler) List(req api.Context) error {
...
+// so that PowerUserPlus users can read and deploy servers from global catalog templates.
+func validateEntryScopeForRead(entry v1.MCPServerCatalogEntry, catalogName, workspaceID string) error {
+	if catalogName != "" && entry.Spec.MCPCatalogName != catalogName {
+		return types.NewErrBadRequest("entry does not belong to catalog")
+	}
+	if workspaceID != "" && entry.Spec.PowerUserWorkspaceID != workspaceID && entry.Spec.MCPCatalogName == "" {
+		return types.NewErrBadRequest("entry does not belong to workspace")
+	}
+	return nil
+}
```
thedadams, line 98: These functions are implemented in a couple different places now. I saw we DRY this up.

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
```diff
@@ -50,11 +47,19 @@ func (*Handler) EnsureUserCount(req router.Request, _ router.Response) error {
...
-		// Don't count servers that don't have a user ID, are being deleted, or are part of a composite server.
-		if server.Spec.UserID != "" && server.DeletionTimestamp.IsZero() && server.Spec.CompositeName == "" {
-			uniqueUsers[server.Spec.UserID] = struct{}{}
+		if !server.DeletionTimestamp.IsZero() || server.Spec.CompositeName != "" {
+			continue
+		}
+		if isSingleUser {
+			if server.Spec.UserID != "" {
+				uniqueUsers[server.Spec.UserID] = struct{}{}
+			}
```
thedadams: There is a controller that cleans up servers that don't have a UserID
```suggestion
			uniqueUsers[server.Spec.UserID] = struct{}{}
```

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
```diff
@@ -67,6 +72,20 @@ func (*Handler) EnsureUserCount(req router.Request, _ router.Response) error {
 	return nil
 }
 
+// EnsureServerUserType backfills the serverUserType field to "singleUser" for
+// existing catalog entries that were created before the field was introduced.
+func (*Handler) EnsureServerUserType(req router.Request, _ router.Response) error {
+	entry := req.Object.(*v1.MCPServerCatalogEntry)
+	if !entry.DeletionTimestamp.IsZero() {
+		return nil
+	}
```
thedadams: This is handled by the framework.
```suggestion
```

pkg/storage/apis/obot.obot.ai/v1/mcpserver.go
```diff
@@ -84,7 +84,7 @@ func (in *MCPServer) DeleteRefs() []Ref {
 
 func (in *MCPServer) ValidConnectURLs(base string) []string {
 	var urls []string
-	if in.Spec.MCPServerCatalogEntryName != "" {
+	if in.Spec.MCPServerCatalogEntryName != "" && in.Spec.IsSingleUser() {
```
thedadams:
```suggestion
	if in.Spec.IsSingleUser() {
```

ui/user/src/lib/services/admin/types.ts
```diff
@@ -633,6 +633,7 @@ export interface MCPCatalogEntryServerManifest {
 	};
 
 	runtime: Runtime;
+	serverUserType?: 'singleUser' | 'multiUser';
```
thedadams:
```suggestion
	serverUserType: 'singleUser' | 'multiUser';
```

ui/user/src/lib/services/user/types.ts
```diff
@@ -286,6 +286,7 @@ export interface MCPCatalogServer {
 	mcpServerInstanceUserCount?: number;
 	manifest: MCPServer;
 	oauthMetadata?: OAuthMetadata;
+	serverUserType?: 'singleUser' | 'multiUser';
```
thedadams:
```suggestion
	serverUserType: 'singleUser' | 'multiUser';
```

thedadams (commented):
Left a couple comments on the UI, but mostly focused on the backend.

pkg/api/handlers/mcp.go
calvinmclean: 3044 is a different function. Do you mean I should update 3044 to also use this helper?

pkg/api/handlers/mcp.go
calvinmclean: Yes. It is slightly different since this requires that both catalogID and workspaceID are not empty for this multi-user case. I can update to use `validateServerScope` so it's consistent with other functions, but will need to keep an additional `catalogID == "" && workspaceID == ""`.

pkg/api/handlers/mcp.go
calvinmclean: I believe this one is required because if the user is not an admin and does not own the single user server, we enter this branch. Then, if `workspaceID` is empty and so is `entry.Spec.PowerUserWorkspaceID`, a user could update another user's server.

Does that sound accurate? Maybe I am misunderstanding something else here. I will push another commit soon which includes tests for this and link it.

calvinmclean: [This test](https://github.com/obot-platform/obot/pull/6627/changes/5e11a108416deb325d3d63bc96af8f09afc3cf1c#diff-af2c84e672165fae938343f8c7efdaaf6e46cd4e9facf627c95de2a7cc0c3ceaR238-R257) will fail if we remove `workspaceID == "" ||`

calvinmclean: Actually, seems like this might already be prevented by Authz restricting `/api/workspaces/{workspace_id}/entries/{entry_id}/servers/{mcpserver_id}/trigger-update` to PowerUser anyways

calvinmclean: Ok, I overhauled this a bit. I originally changed up the logic a bit to flatten out the nested `if`s but it got a bit confusing trying to compare the existing logic.

I reverted that and started fresh to update the `!IsSingleUser` branch to check admin and PowerUserPlus ownership instead of erroring.

The existing branch for `if !req.UserIsAdmin()` is updated to also include `IsSingleUser` and I just combined the two internal if statements into one since they had the same error. So the case where this thread originates from is unchanged except for the new `IsSingleUser` check. I think the `workspace == "" ||` is still valid.

pkg/storage/apis/obot.obot.ai/v1/mcpserver.go
calvinmclean: Can you explain this a bit more? It is possible for a single user server to have an empty `MCPServerCatalogEntryName` as far as I am aware

pkg/api/handlers/mcp.go
calvinmclean: Yes. `ValidateCatalogEntryForRoute` ensures that we are using a multi-user route to access a multi-user catalog entry (or single-user route for single-user entry). Then, `validateEntryScopeForRead` ensures that this entry belongs to the specified catalog or workspace.

calvinmclean: validateEntryScopeForRead allows the workspace route to deploy from the workspace (ui-created) or the global catalog (git-synced)

pkg/storage/apis/obot.obot.ai/v1/mcpserver.go
thedadams: I don't believe it is.

pkg/api/handlers/mcpcatalogs.go
calvinmclean: The closest I see is `validateServerScope` which uses MCPServer while these use MCPServerCatalogEntry

ui/user/src/lib/components/admin/McpServerEntryForm.svelte
```diff
@@ -592,7 +620,7 @@
 					</div>
 				{/if}
 			</div>
-			{#if belongsToUser && !readonly}
+			{#if belongsToUser && !readonly && type !== 'multi'}
```
ivyjeong13: If a poweruser plus creates a Multi-user server (not a multi-user template), will they no longer be able to delete it?

calvinmclean: There was still a Delete button in the 3-dot menu, but that caused a changed experience from the existing flow for deletes. I updated it to have the delete button again and it should be a more consistent experience now. Thanks!

ui/user/src/routes/admin/mcp-servers/s/[id]/details/+page.ts
```diff
@@ -9,11 +9,15 @@ export const load: PageLoad = async ({ params, fetch }) => {
...
+		mcpServer = await UserService.getMcpCatalogServer(id, { fetch });
 	} catch (err) {
-		handleRouteError(err, `/admin/mcp-servers/s/${id}/details`, profile.current);
+		try {
+			mcpServer = await AdminService.getMCPCatalogServer(DEFAULT_MCP_CATALOG_ID, id, {
+				fetch
+			});
+		} catch (_err) {
+			handleRouteError(err, `/admin/mcp-servers/s/${id}/details`, profile.current);
+		}
```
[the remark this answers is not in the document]
Comment (calvinmclean) [reply ui/user/src/routes/admin/mcp-servers/s/[id]/details/+page.ts]: intentional

docs/docs/configuration/mcp-server-gitops.md
```diff
@@ -122,8 +122,50 @@ serverUserType: singleUser  # Optional, defaults to "singleUser"
 The `serverUserType` field specifies how users interact with the catalog entry:
 
 - `singleUser` (default): Each user who installs this catalog entry gets their own independent MCP server instance.
+- `multiUser`: The catalog entry is a template for shared multi-user deployments. An administrator or Power User+ deploys the template into a catalog or workspace, and users connect to that shared deployment through per-user MCP server instances.
```
g-linville, line 125: double-check me on this, but I am pretty sure we normally spell it out as "Power User Plus" instead of just "Power User+"

docs/docs/functionality/mcp-servers.md
```diff
@@ -45,6 +45,23 @@ Multi-user servers still require the user to authenticate to the gateway's confi
...
+A multi-user catalog template is a catalog entry that can be deployed as a shared multi-user server. Instead of creating one server per user when users connect, Obot creates a shared deployment from the template and then creates per-user MCP server instances for the users who connect to that deployment.
+
+Use a multi-user catalog template when:
+
+- Administrators or Power User+ users should publish an approved multi-user server configuration before it is deployed
+- Different workspaces or catalogs need their own shared deployment from the same catalog entry
+- The shared server may need deployment-level configuration, such as a URL, shared API key, or environment variables
+- Individual users still need per-user headers or authentication after connecting to the shared deployment
+
+Multi-user catalog templates are configured with `serverUserType: multiUser`. They can be deployed from catalog-scoped admin routes or workspace-scoped Power User+ routes. A workspace deployment created from a global catalog template belongs to the workspace deployment, not to the global catalog entry, so workspace owners manage that deployment from the workspace view.
```
g-linville: This gets into too much backend detail ("catalog-scoped admin routes," etc.). I don't think we use the terminology of the default catalog, or of Power User Workspaces anywhere in the frontend. Instead I think we just refer to a specific AccessControlRule as an MCP Registry, since it grants a set of groups/users access to a set of servers/catalog entries. So I would just say that Multi-user catalog templates can be deployed as multi-user servers by Admin and Power User Plus users.

ui/user/src/routes/mcp-servers/s/[id]/+page.ts
```diff
@@ -21,8 +21,21 @@ export const load: PageLoad = async ({ params, fetch }) => {
...
 
+	let catalogEntry;
+	const isOwner =
+		profile.current?.isAdmin?.() ||
+		(mcpServer?.powerUserWorkspaceID && mcpServer?.userID === profile.current?.id);
+	if (mcpServer?.catalogEntryID && isOwner) {
+		try {
+			catalogEntry = await UserService.getMCP(mcpServer.catalogEntryID, { fetch });
+		} catch (_err) {
+			// Entry may not be accessible
```
Comment (ivyjeong13) [inline ui/user/src/routes/mcp-servers/s/[id]/+page.ts]: If the entry isn't available, should it silently do nothing when it errors? I'd think if a user could create a multi-user template, they should have read access to the catalog entry?

pkg/api/handlers/mcp.go
```diff
@@ -168,6 +176,12 @@ func (m *MCPHandler) ListEntriesFromAllSources(req api.Context) error {
 	return req.Write(types.MCPServerCatalogEntryList{Items: entries})
 }
 
+// HideMultiUserTemplate determines if users should be able to see a CatalogEntry based on whether it
+// is single or multi-user. PowerUserPlus includes admins.
+func HideMultiUserTemplate(req api.Context, entry v1.MCPServerCatalogEntry) bool {
+	return !req.UserIsPowerUserPlus() && !entry.Spec.Manifest.ServerUserType.IsSingleUser()
+}
+
```
g-linville: While I don't mind the use of the word "template" in documentation, as it is probably clearer to the user what that means, I would prefer to use the term "catalog entry" within the code, since that better aligns with the actual type name (`MCPServerCatalogEntry`).

pkg/api/handlers/mcp.go
```diff
@@ -1552,12 +1578,21 @@ func (m *MCPHandler) CreateServer(req api.Context) error {
...
+		}
+
 		// Block server creation if OAuth is required but not configured
 		if entryRequiresStaticOAuthCreds(catalogEntry) {
 			return types.NewErrBadRequest("catalog entry requires OAuth configuration by an administrator before it can be used")
 		}
 
-		manifest, err := serverManifestFromCatalogEntryManifest(req.UserIsAdmin(), false, catalogEntry.Spec.Manifest, input.MCPServerManifest)
+		// For multiUser template deployments, always use template-faithful materialization.
+		// Admins may override single-user catalog entry config, but not a template's runtime shape.
```
g-linville: More template language here

pkg/api/handlers/mcp.go
```diff
@@ -3641,17 +3720,12 @@ func (m *MCPHandler) TriggerUpdate(req api.Context) error {
 		return err
 	}
 
-	if !req.UserIsAdmin() {
+	if !req.UserIsAdmin() && server.Spec.IsSingleUser() {
 		// Allow users to upgrade their own single-user servers.
 		// (We already verified this is a single-user server at the check above.)
```
g-linville: can remove this comment


```suggestion
```

pkg/api/handlers/mcp_test.go
```diff
@@ -40,6 +48,338 @@ func TestConvertMCPResources(t *testing.T) {
 	assert.Equal(t, resources, server.MCPServerManifest.Resources)
 }
 
+func TestHideMultiUserTemplate(t *testing.T) {
```
g-linville: There is already a unit test for this in `pkg/api/handlers/registry/utils_test.go`.

pkg/validation/mcpvalidators.go
```diff
@@ -1065,21 +1065,27 @@ func ValidateServerManifest(manifest types.MCPServerManifest, isMultiUser bool)
...
-	if !manifest.ServerUserType.IsSingleUser() {
-		return fmt.Errorf("unsupported serverUserType %q: only %q is currently supported for catalog entries", manifest.ServerUserType, types.ServerUserTypeSingleUser)
+	switch manifest.ServerUserType {
+	case "", types.ServerUserTypeSingleUser, types.ServerUserTypeMultiUser:
+	default:
+		return fmt.Errorf("invalid serverUserType %q: must be %q or %q", manifest.ServerUserType, types.ServerUserTypeSingleUser, types.ServerUserTypeMultiUser)
+	}
+
+	if !manifest.ServerUserType.IsSingleUser() && manifest.Runtime == types.RuntimeComposite {
+		return fmt.Errorf("multiUser catalog entries do not support composite runtime")
```
g-linville, line 1102: Do we also validate that the multi user config can only be set when the ServerUserType is multi? This seems like the right place to do it, but maybe we are doing it somewhere else rn. Idk.

ui/user/src/lib/components/mcp/DeploymentsView.svelte
```diff
@@ -333,7 +341,19 @@
...
+					await UserService.triggerWorkspaceMcpServerUpdate(
+						server.powerUserWorkspaceID,
+						server.catalogEntryID,
+						server.id
+					);
+				} else if (id) {
+					await AdminService.triggerMcpCatalogServerUpdate(id, server.id);
+				}
+			} else {
+				await UserService.triggerMcpServerUpdate(server.id);
```
ivyjeong13, line 363: If I'm recalling incorrectly, feel free to ignore haha. I thought `/mcp-servers` was the collective of deployed servers a user has access to so you could just call UserService.triggerMcpServerUpdate without having to supply the catalog entry id/workspace id? Is this no longer the case?

ivyjeong13: Oh wait, maybe it's for single/remote only! Though I thought we showed multi-user servers in deployments, I guess we couldn't the action on them before and can now? edit: resolving cause answered own question 😄

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry_test.go
```diff
@@ -0,0 +1,251 @@
...
+		Name:      entry.Name,
+	}, &router.ResponseWrapper{})
+	require.NoError(t, err)
+
+	var updated v1.MCPServerCatalogEntry
+	require.NoError(t, client.Get(context.Background(), router.Key(entry.Namespace, entry.Name), &updated))
+	assert.Equal(t, 2, updated.Status.UserCount)
+}
+
+func TestEnsureUserCountMultiUserEntryExcludesDeletedAndComposite(t *testing.T) {
```
g-linville: nit: this function doesn't test anything with deleted servers


```suggestion
func TestEnsureUserCountMultiUserEntryExcludesComposite(t *testing.T) {
```

ui/user/src/lib/components/admin/McpServerEntryForm.svelte
```diff
@@ -399,38 +414,53 @@
...
+				// Update the entry's tool preview in-place.
+				(entry as MCPCatalogServer).manifest.toolPreview = result.manifest?.toolPreview ?? [];
+				saving = false;
 			} else {
-				await AdminService.generateMcpCatalogEntryToolPreviews(
-					id,
-					entry.id,
-					body as unknown as { config?: Record<string, string>; url?: string }
-				);
+				window.location.reload();
```
ivyjeong13: I'm thinking the reload here might no longer be needed?

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -72,6 +72,60 @@ func (*MCPCatalogHandler) List(req api.Context) error {
...
+		hasAccess, err = h.acrHelper.UserHasAccessToMCPServerCatalogEntryInWorkspace(req.Context(), req.User, entry.Name, entry.Spec.PowerUserWorkspaceID)
+	}
+	if err != nil {
+		return err
+	}
+	if !hasAccess {
+		return types.NewErrNotFound("entry not found")
+	}
+	return nil
+}
```
g-linville: Is this function, and the couple of places where it gets called, necessary? It looks like we weren't doing these checks here before this change. I'm asking because I don't know the answer of whether it is necessary or redundant, and it is worth investigating to see.

ui/user/src/routes/mcp-servers/s/[id]/+page.ts
Comment (calvinmclean) [reply ui/user/src/routes/mcp-servers/s/[id]/+page.ts]: Updated this to handle errors and ignores 404 since it may occur if the catalog entry is deleted and the server created from it somehow still exists.

Users that can create a multi-user server from the template do have access to the entry. This made me notice a bit of an inconsistency: PowerUserPlus does not usually see the Configuration and entry details for a SingleUser server from a catalog. This is also the case for these new MultiUser templates from the catalog. However, the PowerUserPlus could see the Configuration tab and entry details after creating an MCPServer from it. I updated this to have a consistent experience now.

g-linville (changes requested):
I marked a couple places where you added the word "template" but there are also several others in the code, tests, and comments. I would like to stick with the term "catalog entry" everywhere, in order to avoid confusing future agents.

ui/user/src/lib/components/admin/McpServerEntryForm.svelte
calvinmclean: I am having trouble getting the UI to show now for the tool preview generation since there always seems to be tool previews available. I will try a few other things and figure out if it's necessary :+1:

calvinmclean: Ok of course I got it working right after sending this reply. I removed it now

docs/docs/configuration/mcp-server-gitops.md
calvinmclean: Most places in the docs use `Power User+` [except this one](https://docs.obot.ai/functionality/user-management/#user-roles).

g-linville: I didn't realize this

ui/user/src/lib/components/mcp/DeploymentsView.svelte
ivyjeong13: Actually sorry Calvin, just coming back to this as food for thought. Is `/mcp-servers` actually a collective of deployed servers from catalog entries? Before, those were single-user and remote servers. So should multi-user template servers be included here as well? 

Feel to resolve but wanted to throw that out there!

pkg/api/handlers/mcpcatalogs.go
calvinmclean: Yes it seems like this is redundant. I didn't realize initially that Authz was already providing the correct checks. Removed it.

pkg/validation/mcpvalidators.go
calvinmclean: This is already checked in `ValidateServerManifest` for the MCPServer, but not for the CatalogEntry. Added it now

ui/user/src/lib/components/mcp/DeploymentsView.svelte
calvinmclean: As far as I am aware, `/admin/mcp-servers` UI is listing all MCPCatalogEntries (`GET /api/all-mcps/entries`)  and MCPServers (`GET /api/all-mcps/servers`). So this includes un-deployed entries from a catalog, user-owned single-user servers, and Admin/PowerUserPlus-owned instances of multi-user servers (via catalog or directly in UI).

In this particular case, we are using the catalog/workspace IDs to trigger updates for multi-user servers owned by an admin or PowerUserPlus.

Hopefully this clarifies things. I would like to leave this open so @thedadams can double-check while he reviews this PR.

calvinmclean: Oh I see that we are more specifically looking at the `/admin/mcp-servers?view=deployments` "Deployments & Connections" tab? In that case, part of my previous comment is irrelevant. I'll keep looking at this a bit more

pkg/validation/mcpvalidators.go
```diff
@@ -1065,21 +1065,35 @@ func ValidateServerManifest(manifest types.MCPServerManifest, isMultiUser bool)
...
+		return fmt.Errorf("multiUser catalog entries cannot be deployed as single-user servers; use the catalog or workspace route")
 	}
 	return nil
 }
 
 func ValidateCatalogEntryManifest(manifest types.MCPServerCatalogEntryManifest) error {
-	if !manifest.ServerUserType.IsSingleUser() {
-		return fmt.Errorf("unsupported serverUserType %q: only %q is currently supported for catalog entries", manifest.ServerUserType, types.ServerUserTypeSingleUser)
+	switch manifest.ServerUserType {
+	case "", types.ServerUserTypeSingleUser, types.ServerUserTypeMultiUser:
```
thedadams: We shouldn't allow `""` here.

docs/docs/functionality/mcp-servers.md
```diff
@@ -45,6 +45,23 @@ Multi-user servers still require the user to authenticate to the gateway's confi
 
 **Configuration**: Pre-configure any required API keys or environment variables. These values are deployed with the server instance. Users connect without being prompted for configuration and authenticate using the built-in authentication or OAuth per the MCP specification.
 
+### Multi-user catalog template
```
thedadams: I think we should stick with the term "catalog entry" here.

pkg/api/handlers/mcp.go
```diff
@@ -47,6 +47,7 @@ type MCPHandler struct {
 	acrHelper           *accesscontrolrule.Helper
 	mcpImagePullSecrets []string
 	serverURL           string
+	shutdownMCPServer   func(context.Context, string) error
```
thedadams: Is this getting set anywhere in the non-test code?

pkg/api/handlers/mcp.go
```diff
@@ -2927,8 +2977,8 @@ func (m *MCPHandler) GetServerFromAllSources(req api.Context) error {
 		return err
 	}
 
-	// Check if server is from default catalog or workspace
-	if server.Spec.MCPCatalogID != system.DefaultCatalog && server.Spec.PowerUserWorkspaceID == "" {
+	// Check if server is from a catalog or workspace.
+	if server.Spec.MCPCatalogID == "" && server.Spec.PowerUserWorkspaceID == "" {
```
```suggestion
	if server.Spec.IsSingleUser() {
```

pkg/api/handlers/mcp.go
```diff
@@ -3618,13 +3680,30 @@ func (m *MCPHandler) UpdateURL(req api.Context) error {
...
 	if err := req.Get(&server, req.PathValue("mcp_server_id")); err != nil {
 		return err
 	}
 
 	if !server.Spec.IsSingleUser() {
-		return types.NewErrBadRequest("cannot trigger update for a multi-user MCP server; use the UpdateServer endpoint instead")
+		// Multi-user servers deployed from catalog entries can be updated, but only
+		// through catalog- or workspace-scoped routes.
+		if server.Spec.MCPServerCatalogEntryName == "" {
+			return types.NewErrBadRequest("cannot trigger update for a multi-user MCP server without a catalog entry")
```
thedadams: The previous error indicated that the caller should use the `UpdateServer` endpoint instead. If that is still valid, can we add that part back?

pkg/validation/mcpvalidators.go
calvinmclean: Won't that break existing catalogs that don't specify single-user?

calvinmclean: The API handlers default it to SingleUser if empty, so those will be fine, but not the sync. Should I update that to default it as well?

pkg/api/handlers/mcp.go
calvinmclean: No this is added for testing only

docs/docs/functionality/mcp-servers.md
calvinmclean: Sounds good. In order to be consistent with the UI, should I change this as well?

Screenshot 2026-05-27 at 12 59 33

I feel like putting "Multi-User Catalog Entry" here is a bit off.

pkg/validation/mcpvalidators.go
thedadams: I would update the catalog to set `singleUser` where necessary.

pkg/api/handlers/mcp.go
thedadams: Then we should remove the `context.Context` parameter and add a comment indicating that this is only used for testing.

docs/docs/functionality/mcp-servers.md
thedadams: I would say that everything should be "catalog entry" for now and we can adjust later.

pkg/validation/mcpvalidators.go
calvinmclean: Do you mean update the actual catalog source (github.com/obot-platform/mcp-catalog) or update the catalog during sync to have the default instead of empty string?

I'll update the mcp-catalog either way so it serves as a complete example. If we don't want to default singleUser, I can remove the existing cases where I default it

thedadams: Defaulting to single-user is fine, but I suggest changing github.com/obot-platform/mcp-catalog and github.com/obot-platform/system-mcp-catalog too.

calvinmclean: Sounds good. I have the defaults set, a draft for obot-platform/mcp-catalog, and will get a draft up for system-mcp-catalog as well. Other comments are resolved as well

g-linville (approved):
backend lgtm

ivyjeong13:
Screenshot 2026-05-27 at 6 13 38 PM
Screenshot 2026-05-27 at 6 13 57 PM

minor thing I noticed. Connected to the multi-user template with both account and basic user account. I see the User count as 2 but when I go into the Server Details page, I see one connected user. 

Checked for making a regular multi-user server (Everything) & seeing the connected users there
Screenshot 2026-05-27 at 6 19 07 PM
Screenshot 2026-05-27 at 6 18 40 PM

ui/user/src/routes/admin/mcp-servers/w/[wid]/s/[id]/+page.ts
```diff
@@ -23,8 +23,18 @@ export const load: PageLoad = async ({ params, fetch, parent }) => {
 		belongsToUser = false;
 	}
 
+	let catalogEntry;
+	if (mcpServer?.catalogEntryID) {
+		try {
+			catalogEntry = await UserService.getMCP(mcpServer.catalogEntryID, { fetch });
+		} catch (_err) {
+			// Entry may not be accessible
```
Comment (ivyjeong13) [inline ui/user/src/routes/admin/mcp-servers/w/[wid]/s/[id]/+page.ts]: Think needs the `handleRouteError(...)` check if not a 404 here as well

Comment (calvinmclean) [reply ui/user/src/routes/admin/mcp-servers/w/[wid]/s/[id]/+page.ts]: Added :+1:

calvinmclean:
@ivyjeong13 good catch on the connected users. I am looking into it a bit and haven't figured it out yet. I'll keep looking tomorrow morning, but might prefer to create a follow-up bug to avoid adding more significant changes to this PR.


Edit: never mind, this is fixed!

ivyjeong13:
> @ivyjeong13 good catch on the connected users. I am looking into it a bit and haven't figured it out yet. I'll keep looking tomorrow morning, but might prefer to create a follow-up bug to avoid adding more significant changes to this PR.
> 
> Edit: never mind, this is fixed!

Woo, thanks Calvin!

ivyjeong13 (approved):
Approving for ui :+1:

thedadams (approved):

PR #6633, opened 2026-05-19
feat: create native macOS SwiftUI app to wrap the `obot setup` command

g-linville:


g-linville:
probably gonna redo this in Go (using Fyne) when we get back to it

PR #6642, opened 2026-05-20
chore: remove legacy APIs and types

thedadams:


chart/values.yaml
```diff
@@ -113,8 +111,6 @@ config:
   OBOT_SERVER_DSN: ""
   # config.OBOT_SERVER_HOSTNAME -- The hostname of your Obot instance, including protocol
   OBOT_SERVER_HOSTNAME: ""
-  # config.OBOT_SERVER_RETENTION_POLICY_HOURS -- The retention policy for the system. Set to 0 to disable retention. Default is 2160 (90 days) if left blank. This field should just be a number in a string, no `h` suffix.
-  OBOT_SERVER_RETENTION_POLICY_HOURS: ""
```
cjellick, line 117: what was this retention for? chat?

chart/values.yaml
```diff
@@ -130,13 +126,6 @@ config:
   # config.ANTHROPIC_API_KEY -- An Anthropic API Key used to configure access to Anthropic models, which can be used as the default in Obot.
   ANTHROPIC_API_KEY: ""
 
-  # config.OBOT_SERVER_OTEL_BASE_EXPORT_ENDPOINT -- The base export endpoint for OpenTelemetry
-  OBOT_SERVER_OTEL_BASE_EXPORT_ENDPOINT: ""
```
cjellick, line 134: i thought we still did otel stuff in obot.

chart/values.yaml
```diff
@@ -81,8 +81,6 @@ config:
 
   # config.NAH_THREADINESS -- Advanced - sets the number of concurrent threads that can run in the Obot controller
   NAH_THREADINESS: "10000"
-  # config.OBOT_SERVER_KNOWLEDGE_FILE_WORKERS -- Advanced - sets the number of workers for knowledge
-  OBOT_SERVER_KNOWLEDGE_FILE_WORKERS: "5"
```
cjellick, line 85: are there values related to workspace object storage that can also be removed? Maybe you didnt get to it in this pr, but we should be able to completely drop workpsace stuff, right?

chart/values.yaml
thedadams: Retention for chat threads/runs.

chart/values.yaml
thedadams: We do, but we don't use these environment variables.

chart/values.yaml
thedadams: I removed everything related to legacy chat, including workspace things.

cjellick (approved):
LGTM.

Can you open an issue or discussion that is effectively an advisory of the helm chart values/env vars that we're dropping + the top level APIs were dropping (ie just /api/agents, dont have to go deeper than that). Plus any other high level things you think worth calling out. Maybe the tools as well? Maybe, at the highest level, the features being deleted?

We can link to it in the release notes as a more detailed insight into whats going away. I dont expect anyone to really really care, but i just want to be reasonably complete about it.

PR #6659, opened 2026-05-21
enhance: expose obot-mcp-server to external traffic

g-linville:
This allows clients other than Nanobot Agents to connect to the obot-mcp-server. This will be useful for local agents to query for and configure MCP servers, and then set them up locally using the Obot CLI (a future change).

I made sure that webhook/filter SystemMCPServers still work properly. I tested this in both Docker and Kubernetes just to be extra safe.

This should not merge until after the release.

### DIFF
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -1030,6 +1030,13 @@ func mcpServerOrInstanceFromConnectURL(req api.Context, id string) (v1.MCPServer
 // MCPIDAndAudienceFromConnectURL returns the MCP server or instance name and audience based on the provided connect URL.
 // The connect URL could have an MCP server ID, server instance ID, or MCP catalog entry ID.
 func MCPIDAndAudienceFromConnectURL(req api.Context, id string) (string, string, error) {
+	if system.IsSystemMCPServerID(id) {
+		if system.IsExternallyAccessibleSystemMCPServerID(id) {
+			return id, id, nil
+		}
+		return "", "", types.NewErrNotFound("system MCP server %s is not externally accessible", id)
+	}
+
 	server, instance, err := mcpServerOrInstanceFromConnectURL(req, id)
 	if err != nil {
 		return "", "", err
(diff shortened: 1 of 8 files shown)


pkg/api/handlers/mcpgateway/handler.go
```diff
@@ -135,6 +140,10 @@ func (h *Handler) ensureServerIsDeployed(req api.Context) (mcp.ServerConfig, str
 }
 
 func (h *Handler) ensureSystemServerIsDeployed(req api.Context, mcpID string) (mcp.ServerConfig, string, bool, error) {
+	if mcpID != system.ObotMCPServerName {
```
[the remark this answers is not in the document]
g-linville: fixed

pkg/api/handlers/mcpgateway/oauth/authorize.go
```diff
@@ -306,8 +314,9 @@ func (h *handler) callback(req api.Context) error {
 	if mcpID == "" {
 		mcpID = req.PathValue("mcp_id")
 	}
-	if mcpID != "" {
+	if mcpID != "" && mcpID != system.ObotMCPServerName {
```
thedadams: This seems like a small optimization. I think it's fine to do this once, but I wouldn't want to do it for a bunch of other servers. Even if it doesn't require authentication, it doesn't hurt to check.

pkg/api/handlers/mcp.go
```diff
@@ -1022,6 +1022,13 @@ func mcpServerOrInstanceFromConnectURL(req api.Context, id string) (v1.MCPServer
 // MCPIDAndAudienceFromConnectURL returns the MCP server or instance name and audience based on the provided connect URL.
 // The connect URL could have an MCP server ID, server instance ID, or MCP catalog entry ID.
 func MCPIDAndAudienceFromConnectURL(req api.Context, id string) (string, string, error) {
+	if system.IsSystemMCPServerID(id) {
+		if system.IsExternallyAccessibleSystemMCPServerID(id) {
+			return id, id, nil
+		}
+		return "", "", types.NewErrNotFound("system MCP server %s is not externally accessible", id)
```
thedadams, line 1037: Is 400 better here? Saying that it isn't externally accessible kind of implies that it exists.

thedadams (approved):

njhale (approved):

pkg/api/handlers/mcp.go
g-linville: yeah we're gonna start returning a 404

g-linville:
@thedadams I re-requested your review. I rebased this on main since you moved where we do the WWW-Authenticate header. I re-tested everything also to make sure it still works as expected.

thedadams (approved):

g-linville:
Closing because this is no longer needed.

Issue #6639, opened 2026-05-19
Test option is available for image pull secrets in error state.


sangee2004:
**Describe the bug**
An ECR image pull secret with invalid Role ARN values correctly appears in the Error state, but the UI still allows the user to run Test. Testing the errored secret results in a 503 error, with the generated Kubernetes image pull secret not ready, which is avoidable and confusing.

**To Reproduce**
1. As Admin user, go to the `Image Pull Secrets` page.
2. Create a new secret with type ECR with invalid values for Role ARN and  click on Create
3. When the user is brought back to the Image Pull Secrets page, refresh the page.

The secret entry will be presented in the' Error` state as expected.

4. Click on the `Test` option.
5. Enter `Test Image Reference` in the Test modal and click on  `Test`.

Notice that the user is presented with the following 503 error, with the `generated Kubernetes image pull secret is not ready` message

<img width="609" height="343" alt="Image" src="https://github.com/user-attachments/assets/7a92ffbf-5f1d-47d7-af5c-155d435e5b05" />

**Expected behavior**
The user should not be presented with the `Test` option for secrets in the error state.

g-linville:
Fixed

sangee2004:
`Test` option is now disabled for ECR secrets entries in `Error` state.

PR #6672, opened 2026-05-21
fix: ui: disable test button for invalid image pull secrets

g-linville:
for https://github.com/obot-platform/obot/issues/6639

With this change, we gray out and disable the Test button if there is something wrong with the secret:

image

### DIFF
--- a/ui/user/src/routes/admin/image-pull-secrets/types.ts
+++ b/ui/user/src/routes/admin/image-pull-secrets/types.ts
@@ -67,6 +67,10 @@ export function statusMessage(secret?: ImagePullSecret) {
 	return secret?.status?.lastError ?? '';
 }
 
+export function canTest(secret: ImagePullSecret) {
+	return statusLabel(secret) !== 'Error';
+}
+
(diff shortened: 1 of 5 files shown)


ui/user/src/routes/admin/image-pull-secrets/types.ts
```diff
@@ -67,6 +67,10 @@ export function statusMessage(secret?: ImagePullSecret) {
 	return secret?.status?.lastError ?? '';
 }
 
+export function canTest(secret: ImagePullSecret) {
```
ivyjeong13: nit: realising these functions are in `types.ts` and would make more sense in something like a `utils.ts` file. Use to types.ts containing enums, interfaces, type aliases!

ivyjeong13 (approved):
just one nit, LGTM otherwise!

ui/user/src/routes/admin/image-pull-secrets/types.ts
g-linville: oh, sorry, I will move it

Issue #6601, opened 2026-05-13
Missing sort and filter options on device management pages


sangee2004:
**Describe the bug**
In the Device management section, the Device Skills, Device MCP Servers, and Device Clients pages do not provide Sort or Filter By options for any fields. Applicable columns should support sorting and filtering to make it easier for admins to find and manage device-related records.

**To Reproduce**
1. Log in as an admin user and go to the `Device management` section
 In the `Device Skills`, `Device MCP Servers`, and `Device Clients` pages, notice that there are no sort and Filter By options provided for any fields.

**Expected behavior**
Provide Sort and Filter By options for applicable fields in `Device Skills`, `Device MCP Servers`, and `Device Clients` pages

shwetapsdet:
@njhale , I have created 1 issue : https://github.com/obot-platform/obot/issues/6701 . Apart from this everything is working fine as expected

PR #6673, opened 2026-05-22
fix: add sorting and filtering to device management tables

njhale:
Add missing sort and filter support across the Device Management pages.

- Device MCP Servers now uses the shared table sort/filter controls for the in-memory scan stats data. The table supports sorting by name, transport, device count, user count, and observation count, plus filtering by name and transport. The fuzzy search parameter was moved to query so it does not conflict with table field filters.
- Device Skills now wires table sorting through the existing paginated device skills API. Sorting supports name, device count, user count, and observation count, preserves sort state in the URL, and resets pagination when the sort changes while keeping the existing name search behavior.
- Device Clients now supports server-side sorting before pagination. The clients API accepts sort_by and sort_order, and the gateway query computes user, skill, and MCP server counts so the table can sort correctly by name, MCP server count, skill count, and user count across the full result set instead of only the current page.

Also align the new backend sorting and aggregate query code with existing gateway conventions, including validated sort fields, conventional query local naming, and filtered aggregate subqueries for derived count sorting.

Addresses https://github.com/obot-platform/obot/issues/6601

### DIFF
--- a/pkg/gateway/client/devicescan.go
+++ b/pkg/gateway/client/devicescan.go
@@ -421,13 +421,6 @@ type SkillStatListOptions struct {
 	Offset    int
 }
 
-var skillStatSortColumns = map[string]string{
-	"name":              "name",
-	"device_count":      "device_count",
-	"user_count":        "user_count",
-	"observation_count": "observation_count",
-}
-
 // ListSkillStats returns one row per distinct skill name observed in
 // the latest scan of any device within the requested window.
 // Paginated, sortable, optional name LIKE filter.
@@ -462,13 +455,19 @@ func (c *Client) ListSkillStats(ctx context.Context, opts SkillStatListOptions) 
 		return nil, 0, fmt.Errorf("failed to count skill stats: %w", err)
 	}
 
-	sortCol := skillStatSortColumns[opts.SortBy]
-	if sortCol == "" {
-		sortCol = "device_count"
+	validSortFields := map[string]bool{
+		"name":              true,
+		"device_count":      true,
+		"user_count":        true,
+		"observation_count": true,
 	}
-	sortDir := "DESC"
+	sortBy := opts.SortBy
+	if !validSortFields[sortBy] {
+		sortBy = "device_count"
+	}
+	sortOrder := "DESC"
 	if strings.EqualFold(opts.SortOrder, "asc") {
-		sortDir = "ASC"
+		sortOrder = "ASC"
 	}
 
 	q := base.Session(&gorm.Session{}).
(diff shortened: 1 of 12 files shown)


pkg/gateway/client/devicescan.go
```diff
@@ -594,20 +597,74 @@ func (c *Client) ListDeviceClientFleetSummaries(ctx context.Context, opts Device
...
+		Group("cl.name")
+
+	skillCounts := db.Table("device_scan_skills AS sk").
+		Joins("JOIN device_scans AS s ON s.id = sk.device_scan_id").
+		Where("s.id IN (?)", latest).
+		Where("sk.client <> ?", "multi").
+		Where("sk.name <> ''").
+		Select("sk.client AS name, COUNT(DISTINCT sk.name) AS skill_count").
+		Group("sk.client")
+
```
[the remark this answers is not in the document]
njhale: should be addressed now

g-linville (approved):

thedadams (approved):

ui/user/src/routes/admin/device-skills/+page.svelte
```diff
@@ -25,6 +25,22 @@
 	let nameFilter = $state(untrack(() => page.url.searchParams.get('name') ?? ''));
 
 	type Row = DeviceSkillStat & { id: string };
+	const defaultSort = { property: 'deviceCount', order: 'desc' } as const;
+	const sortFields = {
```
ivyjeong13: nit: See this duplicated and using DeviceSkillSortKey in its `page.ts`
```
const sortFields: Record<string, DeviceSkillSortKey> = {
	name: 'name',
	deviceCount: 'device_count',
	userCount: 'user_count',
	observationCount: 'observation_count'
};
```

could be put into a `constants.ts` and exported from there to have in one place!

ui/user/src/routes/admin/device-clients/+page.svelte
```diff
@@ -37,17 +56,40 @@
...
+		else next.searchParams.delete('offset');
+		if (isSortProperty(sort?.property)) {
+			next.searchParams.set('sort', sort.property);
+			next.searchParams.set('sortDirection', sort.order);
+		}
+		replaceState(next, {});
+	}
 
-	async function reload(idx: number) {
+	function getSortParams(sort = initSort) {
```
ivyjeong13: nit: see this duplicated in device-skills and could be a common util exported from `utils.ts`

ivyjeong13 (approved):

ui/user/src/routes/admin/device-clients/+page.svelte
```diff
@@ -37,17 +56,40 @@
 	let nameFilter = $state(untrack(() => page.url.searchParams.get('name') ?? ''));
 	let loading = $state(false);
 
-	let filteredRows = $derived(
-		nameFilter ? rows.filter((c) => c.name.toLowerCase().includes(nameFilter.toLowerCase())) : rows
-	);
+	function syncUrl(nextPageIndex: number, sort = initSort) {
+		const next = new URL(page.url);
+		if (nameFilter) next.searchParams.set('name', nameFilter);
```
ivyjeong13, line 55: this is todo for myself: provide equivalent to `setUrlParam` that handle batches of properties instead of just one

ui/user/src/routes/admin/device-clients/+page.svelte
njhale: Done

ui/user/src/routes/admin/device-skills/+page.svelte
njhale: Done

ivyjeong13 (approved):

PR #6686, opened 2026-05-22
fix: ensure the www-authenticate header is returned on the MCP path

thedadams:
A previous change fixed a problem where MCP servers were accessible to all authenticated users. That fix bypassed the code where we were setting the www-authentiate header. This change fixes that.

### DIFF
--- a/pkg/api/server/server.go
+++ b/pkg/api/server/server.go
@@ -44,13 +44,18 @@ type Server struct {
 	auditLogger    audit.Logger
 	rateLimiter    *ratelimiter.RateLimiter
 	baseURL        string
+	mcpOAuthScope  string
 	registryNoAuth bool
 
 	mux         *http.ServeMux
 	otelHandler http.Handler
 }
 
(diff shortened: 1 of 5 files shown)


g-linville (approved):

pkg/api/server/server.go
```diff
@@ -44,13 +44,18 @@ type Server struct {
...
 
 	mux         *http.ServeMux
 	otelHandler http.Handler
 }
 
-func NewServer(storageClient storage.Client, gatewayClient *gclient.Client, gptClient *gptscript.GPTScript, localK8sClient kclient.Client, obotNamespace string, authn *authn.Authenticator, authz *authz.Authorizer, proxyManager *proxy.Manager, auditLogger audit.Logger, rateLimiter *ratelimiter.RateLimiter, baseURL string, registryNoAuth bool) *Server {
+func NewServer(storageClient storage.Client, gatewayClient *gclient.Client, gptClient *gptscript.GPTScript, localK8sClient kclient.Client, obotNamespace string, authn *authn.Authenticator, authz *authz.Authorizer, proxyManager *proxy.Manager, auditLogger audit.Logger, rateLimiter *ratelimiter.RateLimiter, baseURL string, oauthScopesSupported []string, registryNoAuth bool) *Server {
+	var scope string
+	if len(oauthScopesSupported) > 0 {
+		scope = fmt.Sprintf(", scope=\"%s\"", strings.Join(oauthScopesSupported, " "))
```
njhale, line 57: Why not %q? Is there an edgecase doing it manually addresses?

njhale (approved):

PR #6704, opened 2026-05-26
feat: openai and anthropic passthroughs for external clients

njhale:
## Summary

The LLM Gateway passthrough routes existed but weren't usable by real external clients. This makes them work end-to-end: a user can point a client like **Claude Code** or **Codex** at the gateway's OpenAI or Anthropic passthrough route, authenticate with their Obot token, and call models by their native provider model names (`gpt-5.5`, `claude-sonnet-4-5`, etc...). The gateway proxies transparently to the vendor while enforcing per-user model access policy.

### Backend: make the passthrough actually work

- **Resolve provider-native model names.** External clients send the vendor's own model IDs, which didn't map to anything internally. Added `ResolveTargetModel` to map a provider model ID back to its `Model` resource so access checks apply to the names clients actually send (newest wins when duplicates map to the same target).
- **Scope `/v1/models` to the user.** The models list now returns only the models the requesting user is allowed to call (`GetUserAllowedTargetModels`). Surviving entries are preserved byte-for-byte from upstream, and Anthropic pagination cursors are rewritten so clients never page into hidden models.
- **Normalize request paths.** Reconciled the differing base-URL conventions of internal vs. external clients so passthrough requests no longer produce malformed `/v1/v1/...` URLs.
- Pulled in `github.com/tidwall/sjson` for the in-place cursor rewriting.

### Frontend: model discovery + gateway sidenav

- **New `/llm-gateway/models` page** so users can see which OpenAI and Anthropic models they have access to, the provider base URL, and a copy-paste curl example (authenticated via `obot login --print-token`, pre-filled with one of their models).
- **Grouped the gateway pages under a new "LLM Gateway" sidenav section** (Tokens, Model Providers, Model Access Policies, and the new Models page), moving them out of the Obot Agent management group.

## Usage

```bash
export ANTHROPIC_BASE_URL="https://obot.example.com/api/llm-proxy/anthropic"
export ANTHROPIC_API_KEY="$(obot login --url https://obot.example.com --print-token)"

# Note: This assumes the authenticated user has been granted access to `claude-sonnet-4-6`
# with a ModelAccessPolicy in Obot
curl $ANTHROPIC_BASE_URL/v1/messages \
  -H "x-api-key: $ANTHROPIC_API_KEY" \
  -H "anthropic-version: 2023-06-01" \
  -H "content-type: application/json" \
  -d '{"model":"claude-sonnet-4-6","max_tokens":1024,"messages":[{"role":"user","content":"hi"}]}'
```

Setting up Claude Code for use with the gateway:

```bash
# 1. Logout of your existing anthropic account
claude auth logout

# 2. Setup base URL, API key
export ANTHROPIC_BASE_URL="https://obot.example.com/api/llm-proxy/anthropic"
export ANTHROPIC_API_KEY="$(obot login --url https://obot.example.com --print-token)"
export CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY=1 # Tells claude to query `/v1/models` for the set of available models at startup

# 3. Launch claude code
claude code
```

You can then run `/models` and select an Obot passthrough model to use (marked with "From Gateway").

Codex has also been tested with the openai passthrough, but requires more setup:

1. Update your `~/.codex/config.toml` with the following:

```toml
model_provider = "obot_openai"

[model_providers.obot_openai]
name = "OpenAI Obot LLM Gateway"
base_url = "http://localhost:8080/api/llm-proxy/openai"
env_key = "OBOT_API_KEY"
env_key_instructions = "Set the OBOT_API_KEY and restart to authenticate with the Obot LLM Gateway"
supports_websockets = false
```

2. Start the codex CLI or App

```bash
export OBOT_API_KEY="$(obot login --url https://obot.example.com --print-token)"
codex # Codex CLI
codex app # Open App
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -55,6 +55,7 @@ require (
 	github.com/spf13/cobra v1.9.1
 	github.com/stretchr/testify v1.11.1
 	github.com/tidwall/gjson v1.18.0
+	github.com/tidwall/sjson v1.2.5
 	github.com/zalando/go-keyring v0.2.8
 	go.opentelemetry.io/contrib/exporters/autoexport v0.68.0
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.61.0
@@ -86,8 +87,6 @@ require (
 	sigs.k8s.io/controller-runtime v0.22.5
 	sigs.k8s.io/yaml v1.6.0
 )
-
-require github.com/tidwall/sjson v1.2.5
 
 require (
 	cel.dev/expr v0.25.1 // indirect
--- a/go.mod
+++ b/go.mod
@@ -87,6 +87,8 @@ require (
 	sigs.k8s.io/yaml v1.6.0
 )
 
+require github.com/tidwall/sjson v1.2.5
+
 require (
 	cel.dev/expr v0.25.1 // indirect
 	cloud.google.com/go v0.123.0 // indirect
--- a/pkg/gateway/server/llmproxy.go
+++ b/pkg/gateway/server/llmproxy.go
@@ -29,6 +29,7 @@ import (
 	"github.com/obot-platform/obot/pkg/modelaccesspolicy"
 	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
 	"github.com/tidwall/gjson"
+	"github.com/tidwall/sjson"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apiserver/pkg/authentication/user"
@@ -279,13 +280,13 @@ func rewriteModelInBody(body []byte, model string) ([]byte, error) {
 	return json.Marshal(bodyMap)
 }
 
-// copyBody returns a copy of the bytes in a request body.
-// If the copy was successful the request body is restored to its original state before returning so that
+// copyBody returns a copy of the bytes in a request or response body.
+// If the copy was successful the body is restored to its original state before returning so that
 // it can be reused.
-// The returned byte slice is safe to modify without affecting the request body.
-func copyBody(r *http.Request) ([]byte, error) {
-	b, err := io.ReadAll(r.Body)
-	r.Body.Close()
+// The returned byte slice is safe to modify without affecting the restored body.
+func copyBody(body *io.ReadCloser) ([]byte, error) {
+	b, err := io.ReadAll(*body)
+	(*body).Close()
 
 	if err != nil {
 		// b can be partial results on error, don't restore the body
@@ -293,7 +294,7 @@ func copyBody(r *http.Request) ([]byte, error) {
 	}
 
 	// Read was successful, restore the body with a copy.
-	r.Body = io.NopCloser(bytes.NewReader(slices.Clone(b)))
+	*body = io.NopCloser(bytes.NewReader(slices.Clone(b)))
 	return b, nil
 }
 
@@ -317,9 +318,21 @@ type responseModifier struct {
 	outputPolicies      []messagepolicy.ApplicablePolicy
 	conversationHistory []messagepolicy.ConversationMessage
 	pipeReader          *io.PipeReader // set when output policies are active; Read() reads from this
+
+	// allowedTargetModels restricts a GET /v1/models response to the
+	// provider-native target model ids in this set (the models the user may
+	// use). It is ignored when allowAllModels is true.
+	allowedTargetModels map[string]bool
+	// allowAllModels reports that the user may use every model, so a GET
+	// /v1/models response is passed through unfiltered.
+	allowAllModels bool
 }
 
 func (r *responseModifier) modifyResponse(resp *http.Response) error {
+	if resp.Request.URL.Path == "/v1/models" {
+		return r.filterModelListResponse(resp)
+	}
+
 	if r.inputPolicyReplacement != "" {
 		resp.Header.Set("X-Obot-Message-Policy-Replacement", r.inputPolicyReplacement)
 	}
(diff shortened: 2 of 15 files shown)


pkg/gateway/server/llmproxy.go
```diff
@@ -356,6 +369,101 @@ func (r *responseModifier) modifyResponse(resp *http.Response) error {
...
+// the Anthropic and OpenAI list endpoints return a top-level
+// {"data": [{"id": ...}, ...]} envelope, so the same filtering applies to both
+// passthroughs. Surviving entries are kept as their exact upstream bytes — no
+// map[string]any round-trip that would reorder object keys or reformat numbers.
+// The body is copied up front and restored, so any failure (or an unrecognized
+// payload) just forwards the original response untouched.
+func (r *responseModifier) filterModelListResponse(resp *http.Response) error {
+	if resp.StatusCode != http.StatusOK || r.allowAllModels {
+		return nil
+	}
```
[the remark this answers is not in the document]
njhale: I think this is fine for now, I don't think there are any real cases where this would happen.

pkg/gateway/server/llmproxy_test.go
```diff
@@ -1210,3 +1326,352 @@ func TestParseMessagesFromBody_ConversationHistoryForPolicyEval(t *testing.T) {
...
+		wantFirstID *string
+		wantLastID  *string
+	}{
+		{
+			name:        "keeps allowed subset and repoints cursors to boundaries",
+			allowed:     map[string]bool{"claude-opus-4": true, "claude-haiku-4": true},
+			body:        `{"data":[{"type":"model","id":"claude-opus-4"},{"type":"model","id":"claude-sonnet-4"},{"type":"model","id":"claude-haiku-4"}],"first_id":"claude-opus-4","has_more":false,"last_id":"claude-haiku-4"}`,
+			wantIDs:     []string{"claude-opus-4", "claude-haiku-4"},
+			wantFirstID: new("claude-opus-4"),
+			wantLastID:  new("claude-haiku-4"),
```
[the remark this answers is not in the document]
njhale: It does compile.

go.mod
```diff
@@ -88,6 +88,8 @@ require (
 	sigs.k8s.io/yaml v1.6.0
 )
 
+require github.com/tidwall/sjson v1.2.5
```
thedadams: Move this into the block above.

pkg/gateway/server/llmproxy.go
```diff
@@ -952,18 +1060,21 @@ func (l *llmProviderProxy) proxy(req api.Context) error {
...
-		model, err := getModelFromReference(req.Context(), req.Storage, l.modelProvider.Namespace, targetModel)
+		model, err := getModelFromReference(req.Context(), req.Storage, modelProvider.Namespace, targetModel)
+		if apierrors.IsNotFound(err) {
+			model, err = l.mapHelper.ResolveTargetModel(modelProvider.Name, targetModel)
+		}
 		if err != nil {
 			return fmt.Errorf("failed to get model: %w", err)
 		}
-		if model.Spec.Manifest.ModelProvider != l.modelProvider.Name {
+		if model.Spec.Manifest.ModelProvider != modelProvider.Name {
```
thedadams, line 1065: Good catch.

pkg/gateway/server/llmproxy.go
```diff
@@ -1039,6 +1150,20 @@ func (l *llmProviderProxy) proxy(req api.Context) error {
...
 	}
 
+	// For the models-list endpoint, compute the set of provider-native target
+	// model ids the user may use so the response modifier can strip inaccessible
+	// models.
+	var (
+		allowedTargetModels map[string]bool
+		allowAllModels      bool
+	)
+	if isModelsListRequest(req.Request) && req.User.GetUID() != "" {
```
thedadams: Do we need to check the user's UID? My assumption is that if we got here, then the user's fields are set.

pkg/modelaccesspolicy/helper.go
```diff
@@ -146,6 +169,79 @@ func (h *Helper) GetUserAllowedModels(user kuser.Info) (map[string]bool, bool, e
...
+func (h *Helper) ResolveTargetModel(provider, targetModel string) (*v1.Model, error) {
+	objs, err := h.modelIndexer.ByIndex(modelProviderIndex, modelProviderTargetKey(provider, targetModel))
+	if err != nil {
+		return nil, fmt.Errorf("failed to get models for provider %q target %q: %w", provider, targetModel, err)
+	}
+
+	var newest *v1.Model
+	for _, obj := range objs {
+		m, ok := obj.(*v1.Model)
+		if !ok || m.Spec.Manifest.ModelProvider != provider || m.Spec.Manifest.TargetModel != targetModel {
```
thedadams: nit: I think this is what the indexer is for, right?
```suggestion
		if !ok {
```

njhale: Yup. Fixed.

pkg/gateway/server/llmproxy.go
njhale: Nope. Thanks for catching that. Fixed.

go.mod
njhale: Done.

thedadams (approved):

ui/user/src/routes/llm-gateway/models/+page.svelte
```diff
@@ -0,0 +1,79 @@
...
+		data.models.filter((m) => m.modelProvider === CommonModelProviderIds.ANTHROPIC)
+	);
+
+	function buildCtx(shortKey: 'openai' | 'anthropic', models: typeof data.models): RenderContext {
+		const provider = PROVIDER_CONNECTIONS[shortKey];
+		return {
+			provider,
+			obotURL,
+			baseURL: `${obotURL}/api/llm-proxy/${provider.shortKey}`,
+			exampleModel: models[0]?.name
```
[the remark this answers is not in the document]
njhale: These values are the same.

Issue #6668, opened 2026-05-21
Allow non-admins/auditors to see their own device scans


njhale:
## Summary

Allow users to view the device scan results they submitted, without granting access to other users’ scans or admin-only fleet views.

## Current State

Users can run `obot scan` and submit device scan results to Obot. Those scans are persisted with a `submittedBy` value tied to the authenticated user.

Today, device scan read access is limited to admins, owners, and auditors. A non-admin user can submit a scan, but cannot later view the results they submitted.

The admin UI already has device scan views for device list, device drilldown, scan history, and scan details.

## Desired State

Non-admin/non-auditor users should have read-only access to device scans where `submittedBy` matches their user ID.

Users should be able to view:

- A list of devices they have submitted scans for
- A device drilldown page
- Scan history for a device
- Details for a submitted scan

All non-admin device scan reads must be scoped on the backend to the current submitter. The UI must not be the source of access control.

## Requirements

- Non-admin users can list only their own submitted device scans.
- Non-admin users can view details only for scans they submitted.
- Non-admin users cannot view scans submitted by other users.
- Non-admin users cannot delete or mutate scan records.
- Admin, owner, and auditor access remains unchanged.
- Existing fleet-level/admin device scan views remain privileged-only.

## Acceptance Criteria

- A normal authenticated user can open a Devices page and see only devices from their submitted scans.
- A normal authenticated user can drill into a device and see scan history for that device.
- A normal authenticated user can open scan details for their own submitted scan.
- Supplying another user’s `submittedBy` filter does not widen access.
- Requesting another user’s scan by ID does not expose that scan.
- Admin/owner/auditor behavior remains unchanged.

## Open Questions

- Should non-admin users get a personal device scan dashboard, or only the device list and drilldown views?
  - Option A: Start with list and drilldown only.
  - Option B: Add a user-scoped dashboard similar to the admin dashboard, but filtered to `submittedBy = current user`.
  - Recommendation: Start with list and drilldown only. The admin dashboard is fleet-oriented, and the first requirement is submitter read access, not analytics.

## Out of Scope

- Scan deletion by submitters
- Editing scan records
- Sharing scans between users

cjellick:
@njhale - make sure the necessary API is in place to do this, then hand off to ivy. note this is not in the v0.22 milstone, but i would like to see the backend get don by mid next week.

sangee2004:
This issue is addressed when testing with the latest build from `main`.

Non-admin users can view their own device scans.

[...]

PR #6718, opened 2026-05-26
fix: allow users to get their own scans

njhale:
- Add GET /api/devices/scans and GET /api/devices/scans/{scan_id} to
  basic user authz privileges
- In the List handler, override SubmittedBy to [req.User.GetUID()] when
  the caller is not admin/owner/auditor so a client-supplied?submitted_by=
  filter cannot widen the result set.
- In the Get handler, after the gateway returns the scan, return the same NotFound
  response a truly missing row would produce when the caller is non-privileged and
  SubmittedBy doesn't match.

Addresses https://github.com/obot-platform/obot/issues/6668


calvinmclean (approved):

g-linville (approved):
backend lgtm

thedadams (approved):

Issue #6732, opened 2026-05-27
Sorting by user field on `Devices` page does not work correctly.


sangee2004:
**Describe the bug**
Sorting on the `User` column in the Devices page does not work correctly. Even after applying the sort, the displayed device list does not appear to reorder by username as expected, making it difficult for admins to locate or group devices by user ownership.

Pre-Req: Submit obot scans from a few different devices.
**To Reproduce**
1. Log in as Admin and go to the `Devices` page under `Device Management`.
2. Sort on `User` field

Notice that the list does not get sorted by the `Users` field as expected.

<img width="1277" height="662" alt="Image" src="https://github.com/user-attachments/assets/33763452-f669-45f5-8586-e55c8934f4a1" />

**Expected behavior**
Sorting by the `User` field should work as expected.

sangee2004:
This issue is addressed when testing with the latest build from `main`.

PR #6737, opened 2026-05-27
fix: sort devices user column by displayed name

njhale:
The User column rendered the resolved OrgUser display name
(displayName → email → username → id) when submittedBy matched a known
user, but the row's sort key was still the raw device scan username
field. This meant the visible order didn't match the column you were
sorting by — e.g. a row showing "Nick Hale" would sort under the raw
username, or not sort at all when the scan carried no username.

Override row.username with userDisplay(u) when the submittedBy lookup
succeeds, falling back to the raw scan username otherwise, so the sort
key matches what's actually rendered in the cell.

https://github.com/obot-platform/obot/issues/6732

### DIFF
--- a/ui/user/src/routes/admin/devices/+page.svelte
+++ b/ui/user/src/routes/admin/devices/+page.svelte
@@ -42,19 +42,6 @@
 		scanned_relative: string;
 	};
 
-	let rows = $derived<Row[]>(
-		(devicesResp.items ?? []).map((s) => ({
-			...s,
-			short_device_id: (s.deviceID ?? '').slice(0, 12),
-			os_arch: `${s.os} / ${s.arch}`,
-			mcp_count: s.mcpServers?.length ?? 0,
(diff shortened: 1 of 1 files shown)


cjellick (approved):

g-linville (approved):

ui/user/src/routes/admin/devices/+page.svelte
```diff
@@ -63,6 +50,23 @@
 		return u.displayName ?? u.email ?? u.username ?? u.id;
 	}
 
+	let rows = $derived<Row[]>(
+		(devicesResp.items ?? []).map((s) => {
+			const u = s.submittedBy ? userById.get(s.submittedBy) : undefined;
```
ivyjeong13, line 55: just fyi, there is also a `getUserDisplayName` util that accepts id and the userMap though I think it defaults to 'Unknown User' instead of showing user.id

ivyjeong13 (approved):

Issue #6584, opened 2026-05-12
Manage MCP config secret bindings from the UI


calvinmclean:
Admins should be able to bind Kubernetes secrets to MCP server config properties from the UI when creating or deploying shared/multi-user servers.

This applies both to fresh multi-user server creation and to deployments from multi-user catalog templates.

## Dependencies
- Depends on #6580.
- Related to #6583.

## Goals
- Allow admins to select a secret for eligible MCP config properties in the UI.
- Support this during fresh multi-user server creation.
- Support this during deploy-from-template flows.
- Add a UI/API lookup flow for secrets that are allowed to be bound.
- Restrict lookup results to secrets with a configured label present.
- Provide a sensible default allowed-secret label.
- Allow the allowed-secret label to be overridden with an env var.
- Match only on label presence, not label value.
- Enforce the same label restriction when binding/resolving the secret into the MCP server.
- Avoid exposing secret values in API responses or UI.

## Acceptance Criteria
- Admins can search/list/select allowed secrets from the UI for eligible config fields.
- Only secrets with the configured label are returned by lookup APIs.
- Secret binding fails server-side if the referenced secret does not have the configured label.
- Label filtering checks label presence only.
- The allowed-secret label has a documented default and can be overridden by env var.
- Secret values are never returned to the UI.
- Fresh multi-user server creation supports UI-selected secret bindings.
- Template deployment supports UI-selected secret bindings for deploy-time config where allowed.
- Tests cover lookup filtering, bind-time enforcement, env-var override, and no secret-value leakage.

## Out of Scope
- Matching on label values.
- Arbitrary unlabeled secret discovery.
- User-level secret management UI.
- Automatic secret rotation workflows.

calvinmclean:
Core implementation decisions from planning:
- Default allowed lookup label: `obot.obot.ai/enable-secret-binding`.
- Override env var: `OBOT_SERVER_MCPSECRET_BINDING_ALLOWED_LABEL`.
- Empty/unset override uses the default; non-empty override should be validated as a Kubernetes label key.
- The label gates secret discovery and admin-managed save-time validation only.
- Runtime resolution (`MergeBoundCreds`) should not require/check the label, so existing GitOps/template-authored bindings continue to work.
- Lookup API should be global/admin-only, list Secrets from the Obot service namespace, return only `{ name, keys }`, omit empty Secrets, sort names/keys, and never expose values or extra metadata.
[...]

PR #6754, opened 2026-05-28
feat: Manage MCP config secret bindings from the UI

calvinmclean:
## Description

Addresses #6584 

This change allows admins to bind K8s secrets to multi-user servers from the UI. They can be bound when creating a new multi-user server via UI and when configuring a multi-user server from a multi-user template (from git catalog).

## Backend
- Add API to get secret (name/key only, no values)
- Require label to exist when querying, saving, and deploying
- Secret-bound values added by an admin in the UI/API are tracked using a boolean on the MCPServer so drift detection can see if they changed relative to the template

## UI
- Query backend for secrets
- Allow admin to use them when creating multi-user server
- Allow admin to use them when configuring a multi-user server from a git catalog template

## Testing
Use this script to create 30 secrets:
```
for i in $(seq 1 30); do
  kubectl create secret generic test-secret-$i \
    --from-literal=key1=value1 \
    --from-literal=key2=value2 \
    --from-literal=key3=value3
  kubectl label secret test-secret-$i obot.obot.ai/allow-secret-binding=present
done
```

Cleanup:
```
kubectl delete secret (seq 1 30 | xargs -I{} echo test-secret-{})
```


docs/docs/configuration/mcp-server-gitops.md
```diff
@@ -196,7 +196,7 @@ When omitted, Obot uses the configured MCP server resource defaults. When specif
 
 Secret bindings let you wire an env var, header, or file to a key in an externally-managed Kubernetes Secret instead of asking the user to supply the value at install time.
 
-Secret bindings are only available on git-managed catalog entries, and only when Obot is using the Kubernetes MCP runtime backend.
+Secret bindings are only available when Obot is using the Kubernetes MCP runtime backend.
```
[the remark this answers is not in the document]
calvinmclean: Not correct. This is what the PR does.

pkg/api/handlers/mcp.go
```diff
@@ -1613,9 +1783,16 @@ func (m *MCPHandler) CreateServer(req api.Context) error {
...
+	adminManagedSecretBindings := req.UserIsAdmin() && server.Spec.IsCatalogServer()
+	if err := validation.ValidateSecretBindings(server.Spec.Manifest, gitManagedEntry, adminManagedSecretBindings, m.mcpRuntimeBackend); err != nil {
 		return types.NewErrBadRequest("validation failed: %v", err)
 	}
+	if adminManagedSecretBindings {
+		manifest := adminManagedSecretBindingManifest(server.Spec.Manifest, sourceCatalogEntryManifest, nil, nil)
+		if err := mcp.ValidateAllowedSecretBindings(req.Context(), req.LocalK8sClient, req.ObotNamespace, manifest, m.secretBindingAllowedLabel); err != nil {
+			return types.NewErrBadRequest("validation failed: %v", err)
+		}
+	}
```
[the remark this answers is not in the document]
calvinmclean: currently we are just checking this when selecting secrets to show in the UI

docs/docs/functionality/mcp-servers.md
```diff
@@ -184,6 +184,37 @@ You will need to select the container option from the drop down. Then provide th
...
+By default, the required label key is:
+
+```text
+obot.obot.ai/enable-secret-binding
+```
+
+The label value is ignored; only label presence is checked. Set `OBOT_SERVER_MCPSECRET_BINDING_ALLOWED_LABEL` to use a different label key.
+
+The label controls whether a Secret can be discovered and selected in the admin UI. **If the label is removed after an MCP server is already bound to that Secret, the existing binding still uses the Secret at runtime**.
+
```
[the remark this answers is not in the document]
calvinmclean: currently we are just checking this when selecting secrets to show in the UI

pkg/api/handlers/mcpsecretbindings.go
```diff
@@ -0,0 +1,42 @@
...
+		k8sClient:    k8sClient,
+		namespace:    namespace,
+	}
+}
+
+// ListAllowedSecrets lists bindable Kubernetes Secrets without exposing secret values.
+func (h *MCPSecretBindingHandler) ListAllowedSecrets(req api.Context) error {
+	if !req.UserIsAdmin() {
+		return types.NewErrForbidden("only admins can list allowed MCP secret bindings")
+	}
```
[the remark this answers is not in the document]
calvinmclean: Owners are admins.

ui/user/src/lib/components/mcp/ConnectToServer.svelte
```diff
@@ -68,6 +71,14 @@
 	let shouldShowAlias = $derived(
 		isDeployingMultiUserCatalogEntry || (isConfigured && !isMultiUserServer(server))
 	);
+	let canBindSecretsForCatalogEntry = $derived(
+		Boolean(
+			isDeployingMultiUserCatalogEntry &&
+			catalogID &&
+			!workspaceID &&
+			isKubernetesRuntimeBackend(version.current.engine)
+		)
```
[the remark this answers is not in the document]
calvinmclean: This error does not occur because only admin and PUP can deploy multi-user and the PUP case is handled by workspace checks.

ui/user/src/lib/diff.ts
```diff
@@ -42,6 +47,110 @@ export function stripManifestMetadata<T>(manifest: T): T {
 	return clone as T;
 }
 
+export function normalizeManifestsForDiff<T>(currentManifest: T, newManifest: T): [T, T] {
+	const current = stripManifestMetadata(currentManifest, { keepSecretBindingMetadata: true });
+	const next = stripManifestMetadata(newManifest, { keepSecretBindingMetadata: true });
+
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
ivyjeong13: try to avoid `eslint-disable-next-line` if possible

```
type ManifestDiff = MCPCatalogEntryServerManifest | MCPServer;
export function normalizeManifestsForDiff(
	currentManifest?: ManifestDiff,
	newManifest?: ManifestDiff
) {
	const current = stripManifestMetadata(currentManifest, { keepSecretBindingMetadata: true });
	const next = stripManifestMetadata(newManifest, { keepSecretBindingMetadata: true });

	const normalize = (currentShape?: ManifestDiff, nextShape?: ManifestDiff) => {
// etc
```

```
function stripSecretBindingMetadata(manifest?: ManifestDiff) {
	if (!manifest || typeof manifest !== 'object') return;

	const stripFields = (
		m?: Pick<ManifestDiff, 'env' | 'remoteConfig' | 'multiUserConfig' | 'compositeConfig'>
	) => {
// ...
```

ui/user/src/lib/components/mcp/EditExistingDeployment.svelte
```diff
@@ -46,6 +50,78 @@
...
+
+	function templateBindingByKey(fields?: MCPSubField[]) {
+		return new Map(
+			(fields ?? [])
+				.filter((field) => hasSecretBinding(field))
+				.map((field) => [field.key, field.secretBinding])
+		);
+	}
+
+	function markPinnedSecretBinding<T extends MCPSubField>(
```
ivyjeong13: nit: I don't believe you need <T> here, looks like both instances of function called are `MCPSubField` type!

ui/user/src/lib/components/mcp/ConnectToServer.svelte
```diff
@@ -92,6 +106,29 @@
...
+	$effect(() => {
+		if (
+			!canBindSecretsForCatalogEntry ||
+			loadingSecretBindingTargets ||
+			secretBindingTargetsLoaded
+		) {
+			return;
+		}
+		loadingSecretBindingTargets = true;
+		AdminService.listMCPSecretBindingTargets({ dontLogErrors: true })
```
ivyjeong13: Wondering if this call can be moved as part of  `initCatalogEntry` or `open` so it doesn't need to be in an $effect? You won't need `loadingSecretBindingTargets` but can keep `secretBindingTargetsLoaded` if this just needs to be fired once and values kept for every other time the connect to server dialog is opened.

pkg/services/config.go
```diff
@@ -994,6 +996,13 @@ func New(ctx context.Context, config Config) (*Services, error) {
 	// Derive registryNoAuth flag from config
 	// When EnableRegistryAuth is false (default), registry is in no-auth mode
 	registryNoAuth := !config.EnableRegistryAuth
+	secretBindingAllowedLabel := strings.TrimSpace(config.MCPSecretBindingAllowedLabel)
+	if secretBindingAllowedLabel == "" {
+		secretBindingAllowedLabel = mcp.DefaultSecretBindingAllowedLabel
+	}
```
thedadams: Remove this and just use a `default` struct tag on the field.

pkg/api/router/router.go
```diff
@@ -92,6 +93,7 @@ func Router(ctx context.Context, services *services.Services) (http.Handler, err
 
 	// User-Deployed MCP Servers (single-user, remote, and composite)
 	mux.HandleFunc("GET /api/mcp-servers", mcp.ListServer)
+	mux.HandleFunc("GET /api/mcp-secret-bindings/secrets", mcpSecretBindings.ListAllowedSecrets)
```
thedadams: Can this be `/api/mcp-server-binding-secrets` instead? If we ever add something like `/api/mcp-server-bindings/{binding_id}` later on, it will conflict with `/api/mcp-server-bindings/secrets`

pkg/api/request.go
```diff
@@ -40,6 +40,10 @@ type Context struct {
 	// runs; mcp.MergeBoundCreds reads source Secrets from here. Empty
 	// when LocalK8sClient is nil.
 	ObotNamespace string
+
+	// MCPSecretBindingAllowedLabel is the Kubernetes Secret label required
+	// before secretBindings can resolve Secret values.
+	MCPSecretBindingAllowedLabel string
```
thedadams: This is kind of niche and probably doesn't need to be added to the request object and just passed to each handler that needs it (like you did with pkg/api/handlers/mcp.go).

pkg/mcp/secretbindings.go
```diff
@@ -78,8 +88,12 @@ func MergeBoundCreds(
 		return merged, nil
 	}
 
-	// secretCache[name] is nil when the Secret was confirmed missing,
-	// non-nil (possibly empty) when it exists.
+	if allowedLabel == "" {
+		allowedLabel = DefaultSecretBindingAllowedLabel
+	}
```
thedadams: If you add the `default` struct tag, then we don't need to do this.

pkg/validation/mcpvalidators.go
```diff
@@ -1298,11 +1309,35 @@ func ValidateSecretBindingsCatalogEntry(manifest types.MCPServerCatalogEntryMani
...
+		RemoteConfig:    remoteCatalogToRuntime(manifest.RemoteConfig),
+		MultiUserConfig: manifest.MultiUserConfig,
+	}
+	return ValidateSecretBindings(synthetic, gitManaged, adminManaged, mcpBackend)
+}
+
+func validateNoAdminAddedCatalogBindings(manifest types.MCPServerCatalogEntryManifest) error {
+	for _, env := range manifest.Env {
+		if env.SecretBinding != nil && env.SecretBinding.AdminAdded {
+			return fmt.Errorf("env %q: secretBinding.adminAdded is only set on deployed servers", env.Key)
```
thedadams: nit:
```suggestion
			return fmt.Errorf("env %q: secretBinding.adminAdded is not valid for catalog entry", env.Key)
```

pkg/validation/mcpvalidators.go
```diff
@@ -1298,11 +1309,35 @@ func ValidateSecretBindingsCatalogEntry(manifest types.MCPServerCatalogEntryMani
...
+		}
 	}
-	return ValidateSecretBindings(synthetic, gitManaged, mcpBackend)
+	if manifest.RemoteConfig != nil {
+		for _, h := range manifest.RemoteConfig.Headers {
+			if h.SecretBinding != nil && h.SecretBinding.AdminAdded {
+				return fmt.Errorf("header %q: secretBinding.adminAdded is only set on deployed servers", h.Key)
```
thedadams:
```suggestion
				return fmt.Errorf("header %q: secretBinding.adminAdded is not valid for catalog entry", h.Key)
```

ui/user/src/lib/components/mcp/ConnectToServer.svelte
calvinmclean: Yep looks like that works well!

pkg/controller/handlers/mcpserver/mcpserver.go
```diff
@@ -383,31 +392,101 @@ func remoteConfigHasDrifted(serverConfig *types.RemoteRuntimeConfig, entryConfig
...
+	}
+	if catalogComparableAdminBinding(entryField.SecretBinding, serverField.SecretBinding) {
+		entryField.SecretBinding = nil
+		entryField.Value = serverField.Value
+	}
+	return reflect.DeepEqual(serverField, entryField)
+}
+
+func catalogComparableAdminBinding(serverBinding, entryBinding *types.MCPSecretBinding) bool {
+	return adminAddedSecretBinding(serverBinding) && (entryBinding == nil || adminAddedSecretBinding(entryBinding))
```
thedadams: Since the function already handles nil bindings, this is simpler.
```suggestion
	return adminAddedSecretBinding(serverBinding) && adminAddedSecretBinding(entryBinding)
```

pkg/api/handlers/mcp.go
```diff
@@ -1718,19 +1814,24 @@ func (m *MCPHandler) UpdateServer(req api.Context) error {
 	}
 
 	var gitManagedEntry bool
+	var sourceCatalogEntryManifest *types.MCPServerCatalogEntryManifest
 	if existing.Spec.MCPServerCatalogEntryName != "" {
 		var catalogEntry v1.MCPServerCatalogEntry
 		if err := req.Get(&catalogEntry, existing.Spec.MCPServerCatalogEntryName); err == nil {
 			gitManagedEntry = catalogEntry.IsGitManaged()
+			sourceCatalogEntryManifest = new(catalogEntry.Spec.Manifest)
```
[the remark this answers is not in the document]
calvinmclean: wrong

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry_test.go
```diff
@@ -63,6 +63,66 @@ func TestDetectCompositeDriftMarksEntryNeedingUpdateWhenMultiUserComponentDrifts
...
+		Client:    client,
+		Ctx:       context.Background(),
+		Object:    compositeEntry,
+		Namespace: compositeEntry.Namespace,
+		Name:      compositeEntry.Name,
+	}, &router.ResponseWrapper{})
+	require.NoError(t, err)
+
+	var updated v1.MCPServerCatalogEntry
+	require.NoError(t, client.Get(context.Background(), router.Key(compositeEntry.Namespace, compositeEntry.Name), &updated))
```
thedadams:
```suggestion
	require.NoError(t, client.Get(t.Context(), router.Key(compositeEntry.Namespace, compositeEntry.Name), &updated))
```

pkg/api/handlers/mcp.go
```diff
@@ -1509,6 +1513,90 @@ func mergeMCPServerManifests(existing, override types.MCPServerManifest) types.M
...
+
+func markAdminAddedSecretBinding(binding, sourceBinding *types.MCPSecretBinding) {
+	if binding == nil {
+		return
+	}
+	if sameSecretBinding(sourceBinding, binding) {
+		binding.AdminAdded = false
+		return
+	}
+	binding.AdminAdded = true
```
```suggestion
	binding.AdminAdded = sameSecretBinding(sourceBinding, binding)
```

pkg/api/handlers/mcp.go
```diff
@@ -1509,6 +1513,90 @@ func mergeMCPServerManifests(existing, override types.MCPServerManifest) types.M
...
+		if field.SecretBinding != nil {
+			bindings[field.Key] = field.SecretBinding
+		}
+	}
+	return bindings
+}
+
+func sameSecretBinding(a, b *types.MCPSecretBinding) bool {
+	if a == nil || b == nil {
+		return a == nil && b == nil
```
```suggestion
		return a == b
```

pkg/api/handlers/mcp.go
```diff
@@ -1509,6 +1513,90 @@ func mergeMCPServerManifests(existing, override types.MCPServerManifest) types.M
...
+	}
+	return a.Name == b.Name && a.Key == b.Key
```
thedadams, line 1859: Also, this could be simpler if `AdminAdded` can be taken into account.
```suggestion
	return *a == *b
```

calvinmclean: Since the incoming request could have an invalid AdminAdded value set, we need to just compare the name and key.

This also made me realize that I didn't have validation making sure that the API request doesn't change a SecretBinding that is already configured as part of the CatalogEntry so I added that as well.

pkg/api/handlers/mcp.go
calvinmclean: Fixed this with:
```
binding.AdminAdded = !sameSecretBinding(sourceBinding, binding)
```

pkg/controller/handlers/mcpserver/mcpserver.go
calvinmclean: This is resolved by:
```
return adminAddedSecretBinding(left) && !adminAddedSecretBinding(right)
```

and I renamed the function and vars because it was not accurate to how it is used.

pkg/controller/handlers/mcpserver/mcpserver.go
```diff
@@ -305,7 +310,11 @@ func multiUserConfigHasDrifted(serverConfig, entryConfig *types.MultiUserConfig)
 	if serverConfig == nil || entryConfig == nil {
 		return true
 	}
-	return !slicesEqualIgnoreOrderDeep(serverConfig.UserDefinedHeaders, entryConfig.UserDefinedHeaders)
+	serverHeaders := withoutAdminOnlyHeaderFields(serverConfig.UserDefinedHeaders, entryConfig.UserDefinedHeaders)
+	entryHeaders := withoutAdminOnlyHeaderFields(entryConfig.UserDefinedHeaders, serverConfig.UserDefinedHeaders)
+	return fieldSlicesHaveDrifted(serverHeaders, entryHeaders, func(serverField, entryField types.MCPHeader) bool {
+		return mcpHeaderMatchesCatalog(serverField, entryField)
+	})
```
g-linville: In the two spots where you create an anonymous function to wrap `mcpHeaderMatchesCatalog`, is it possible to just pass `mcpHeaderMatchesCatalog` without a wrapper? It looks like it should be, given that the parameters are the same type, same name, and same order, and the return type lines up.

pkg/controller/handlers/mcpserver/mcpserver_test.go
```diff
@@ -410,6 +619,156 @@ func TestConfigurationHasDrifted(t *testing.T) {
...
+						Key:           "Authorization",
+						SecretBinding: &types.MCPSecretBinding{Name: "bound-secret", Key: "new-token"},
+					}},
+				},
+			},
+			expectedDrift: true,
+			expectedError: false,
+		},
+		{
+			name: "no drift - multi-user deployed remote header secret binding is configuration",
```
g-linville: I'm confused by the use of "multi-user" in this test name

pkg/mcp/secretbindings_test.go
```diff
@@ -93,11 +94,69 @@ func TestMergeBoundCreds(t *testing.T) {
...
+		manifestEnv := []types.MCPEnv{{MCPHeader: types.MCPHeader{Key: "API_KEY", SecretBinding: binding("bound-secret", "api_key")}}}
+		c := newClient(t, &corev1.Secret{
+			Data:       map[string][]byte{"api_key": []byte("fresh")},
+			ObjectMeta: metav1.ObjectMeta{Name: "bound-secret", Namespace: ns, Labels: map[string]string{"custom-label": "true"}},
+		})
+
+		out, err := MergeBoundCreds(context.Background(), c, ns, manifestEnv, nil, nil, "custom-label")
+		require.NoError(t, err)
+		assert.Equal(t, "fresh", out["API_KEY"])
+	})
```
g-linville: I think this test is redundant

pkg/controller/handlers/mcpserver/mcpserver.go
calvinmclean: good catch, thanks!

g-linville:
Rebased to fix conflicts.

calvinmclean:
@thedadams this PR is ready for re-review

pkg/mcp/loader.go
```diff
@@ -32,6 +32,7 @@ type Options struct {
 	DisallowPrivateIPMCP              bool     `usage:"Disallow MCP containers from connecting to private IPs" default:"true"`
 	DisallowLinkLocalMCP              bool     `usage:"Disallow MCP containers from connecting to link-local addresses" default:"true"`
 	MCPRuntimeBackend                 string   `usage:"The runtime backend to use for running MCP servers: docker, kubernetes, or k8s. Defaults to docker" default:"docker"`
+	MCPSecretBindingAllowedLabel      string   `usage:"Kubernetes Secret label key required for admin UI secret-binding lookup and save-time validation" default:"obot.obot.ai/enable-secret-binding"`
```
thedadams: nit:
```suggestion
	MCPSecretBindingAllowedLabel      string   `usage:"Kubernetes Secret label key required for admin UI secret-binding lookup and save-time validation" default:"obot.obot.ai/allow-secret-binding"`
```

pkg/api/handlers/mcp.go
```diff
@@ -1644,7 +1648,7 @@ func serverManifestFromCatalogEntryManifest(
 		result = mergeMCPServerManifests(result, input)
 	}
 
-	return result, nil
+	return *result.DeepCopy(), nil
```
thedadams, line 1678: Why do we need this? Seems like a code-smell. Hoping it was just an oversight.

calvinmclean: Since the Env slice is set directly from the source `MCPServerCatalogEntryManifest` rather than copied, `applySecretBindingOverlay` would modify the source values. Since this is a conversion/creation function for converting catalog entries to servers, it seemed reasonable to create a copy of the whole thing and disconnect it from the source data rather than just specifically copying the underlying slices.

thedadams (approved):

ivyjeong13 (approved):

docs/docs/functionality/mcp-servers.md
```diff
@@ -186,6 +186,29 @@ You will need to select the container option from the drop down. Then provide th
...
+Secret binding selection in the admin UI is available for multi-user servers. The Kubernetes Secret must be in the Obot server's namespace and must have the [configured allowed secret-binding label](../configuration/server-configuration.md#mcp-secret-binding-allowed-label).
+
+The label controls whether a Secret can be discovered and selected in the admin UI, and Obot also checks the label when resolving the binding at runtime. If the label is removed after an MCP server is already bound to that Secret, the binding is treated as unavailable. Required fields then appear as missing configuration until the label is restored or the binding is changed.
+
+Secrets without data keys are not shown as bindable targets.
+
+### Configure a Binding in the Admin UI
+
+1. Go to **MCP Management -> MCP Servers**.
+2. Create or edit a **Multi-user** MCP server in the catalog.
```
ivyjeong13: For step 2, might need to tweak this part: we've changed verbiage to hosted (can either contain single or multi-tenant -- equivalent to single-user catalog entry or multi-user catalog entry), and any multi-users servers are now in MCP Deployments instead

ui/user/src/lib/components/mcp/SecretBindingPicker.svelte
```diff
@@ -0,0 +1,136 @@
...
+			disabled={isReadonly}
+			onSelect={(option) => {
+				if (option.id === 'secret') {
+					enableSecretBinding();
+				} else {
+					clearSecretBinding();
+				}
+			}}
+		/>
+		{#if targets.length === 0}
```
ivyjeong13: Image

nit: Seeing the no kubernetes secret available while on Manual Value seems off to me for some reason. Wondering if making the Kubernetes Secret option selectable then, when the user selects it, it shows the Select as disabled with the `No labeled Kubernetes Secrets are available for binding.` as a placeholder or just show that text instead of the Select would work here instead?

calvinmclean: Thanks @ivyjeong13! I like this suggestion and implemented it in the [latest commit](https://github.com/obot-platform/obot/pull/6754/changes/2c34c307b31da6f834ab27a1d346cf1afe82c2b7). It ended up being a more verbose change than I expected, but it looks nice.

Screenshot 2026-06-25 at 12 06 20

calvinmclean:
@thedadams can you do a quick re-review ([really just this commit](https://github.com/obot-platform/obot/pull/6754/changes/63b689c9e3c7400cbf35d5c890246f834f30202f)) if you have a chance? I did another round of manual testing before merging and realized that the feature to bind secrets to multi-user servers created in the UI (not from a catalog template) was missing. I think it got lost in a rebase/merge conflict somewhere along the way.

pkg/api/handlers/mcp.go
```diff
@@ -986,15 +990,24 @@ func mcpServerOrInstanceFromConnectURL(req api.Context, id string) (v1.MCPServer
...
 				return v1.MCPServer{}, v1.MCPServerInstance{}, err
 			}
 
+			needsConfig, err := entryNeedsUserConfig(req.Context(), req.LocalK8sClient, req.ObotNamespace, entry, secretBindingAllowedLabel)
+			if err != nil {
+				return v1.MCPServer{}, v1.MCPServerInstance{}, fmt.Errorf("failed to determine required configuration for catalog entry %s: %w", id, err)
+			}
+			if needsConfig {
+				return v1.MCPServer{}, v1.MCPServerInstance{}, types.NewErrNotFound("catalog entry %s requires user configuration before it can be connected", id)
+			}
```
thedadams: This is no longer the case. We allow users to connect and they will be prompted to supply the configuration during the OAuth flow.

calvinmclean: Cool! Removed it

thedadams (approved):

Issue #6568, opened 2026-05-11
Composite servers do not support a single shareable connection URL for org-wide deployment


colossus689:
## Summary

Composite servers generate a per-user connection URL, meaning each user must individually open Obot, navigate to the composite server, and connect to generate their own URL before they can use the server. This makes it impossible to deploy a single connection URL to an entire organization the way remote servers allow.

## Background

We deploy MCP server connection URLs centrally to our organization through Claude (and potentially other clients). For **remote servers**, this works as expected — one URL is configured and all authorized users connect through it.

For **composite servers**, the behavior is different. When a user other than the one who created the composite server attempts to connect using the creator's URL, they are denied. Each user must:

1. Log into Obot
2. Navigate to the composite server
3. Manually connect and complete any per-component configuration
4. Retrieve their own personal connection URL

Only after completing those steps can a user successfully connect through a client.

## Expected / Desired Behavior

Administrators should be able to deploy a single composite server connection URL to their organization — the same way remote servers work today. Users should either:

- Authenticate through a shared URL and have Obot resolve their identity and permissions server-side, or
- Be automatically prompted for any required per-component configuration on first connect, without needing to pre-configure inside the Obot UI

## Actual Behavior

The connection URL generated for a composite server is user-scoped. Sharing it with another user results in a denied connection. Every user must individually set up the composite server in Obot before a usable URL is generated for them.

## Questions for the Team

- Is this the intended design for composite servers, given that they can include single-user components requiring individual credentials?

## Suggested Improvement

Consider allowing composite servers to expose a **stable, org-wide connection URL** that:

1. Authenticates the connecting user against the configured identity provider
2. Resolves or prompts for any per-component configuration at first connect (similar to how OAuth components already defer until use)
3. Allows components whose credentials are not yet configured to be skipped or disabled rather than blocking the entire connection

This would bring composite servers in line with the deployment model already supported by remote servers.

sangee2004:
Currently blocked on validating this issue until this issue is addressed- https://github.com/obot-platform/obot/issues/6823

sangee2004:
This issue is addressed when testing with the latest build from `main` - [v0.0.0-dev+cacec23e](https://github.com/obot-platform/obot/commit/cacec23e)

1. Log in as admin user
2. Create a Composite MCP server with the following MCP servers added to it:
- Gmail - remote mcpserver
- Any multi-user MCP server (Tavily Search in my case)
3. Connect to this  Composite MCP server and get the connection URL.
[...]

PR #6761, opened 2026-05-28
fix: enable sharable connect urls for composite servers

njhale:
- Generate sharable connect URLs for composite servers
- Enable creation/launch of composite servers that don't require configuration
  on connect

Addresses https://github.com/obot-platform/obot/issues/6568

### DIFF
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -987,8 +987,12 @@ func mcpServerOrInstanceFromConnectURL(req api.Context, id string) (v1.MCPServer
 		}
 		if len(servers.Items) == 0 {
 			// If the user has not configured an MCP server for the catalog entry, and the catalog
-			// entry does not require any user configuration, then create a server for the user.
-			if entryManifestNeedsUserConfig(entry.Spec.Manifest) {
+			// entry can be launched without further configuration, then create a server for the user.
+			needsConfig, err := entryNeedsUserConfig(req.Context(), req.LocalK8sClient, req.ObotNamespace, entry)
+			if err != nil {
+				return v1.MCPServer{}, v1.MCPServerInstance{}, fmt.Errorf("failed to determine required configuration for catalog entry %s: %w", id, err)
+			}
+			if needsConfig {
 				return v1.MCPServer{}, v1.MCPServerInstance{}, types.NewErrNotFound("user has not configured an MCP server for catalog entry %s", id)
 			}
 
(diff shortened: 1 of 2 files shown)


pkg/api/handlers/mcp.go
```diff
@@ -985,28 +985,10 @@ func mcpServerOrInstanceFromConnectURL(req api.Context, id string) (v1.MCPServer
...
-				for _, h := range entry.Spec.Manifest.RemoteConfig.Headers {
-					if h.Required {
-						return v1.MCPServer{}, v1.MCPServerInstance{}, types.NewErrNotFound("user has not configured an MCP server for catalog entry %s", id)
-					}
-				}
+			// If the user has not configured an MCP server for the catalog entry, and the catalog
+			// entry does not require any user configuration, then create a server for the user.
+			if entryManifestNeedsUserConfig(entry.Spec.Manifest) {
+				return v1.MCPServer{}, v1.MCPServerInstance{}, types.NewErrNotFound("user has not configured an MCP server for catalog entry %s", id)
 			}
```
njhale, line 1011: So this already _wasn't_ checking for:

```
if manifest.RemoteConfig.StaticOAuthRequired && !entry.Status.OAuthCredentialConfigured {
    // Return an error because the admin hasn't configured static oauth 
}
```

in the case of remote servers.

seems like a bug. @g-linville should we be verifying that here?

pkg/api/handlers/mcp.go
```diff
@@ -1043,6 +1038,55 @@ func mcpServerOrInstanceFromConnectURL(req api.Context, id string) (v1.MCPServer
...
 }
 
+// entryManifestNeedsUserConfig reports whether a catalog entry manifest requires user-supplied
+// configuration before it can be instantiated.
+// For composite manifest, all components are checked for required configuration.
+func entryManifestNeedsUserConfig(m types.MCPServerCatalogEntryManifest) bool {
+	for _, e := range m.Env {
+		if e.Required {
+			return true
+		}
```
calvinmclean: `e` could be required and have a SecretBinding. In this case, it would not need a user config. This is also the case for headers below.

Similarly, is it possible to create a git-managed catalog with a required env that also already has a Value set? I'm not sure.

pkg/api/handlers/mcp.go
g-linville: Yes, we should

pkg/api/handlers/mcp.go
thedadams: I'll echo what Calvin says here.

And the answer to his question is "yes" which means we would need to handle that case too.

njhale: yep, fixing this now. Also note these were pre-existing edge cases that I don't believe are currently handled.

njhale: pushed some changes to fix this.

pkg/api/handlers/mcp.go
njhale: pushed changes to fix this

thedadams (approved):

Issue #6666, opened 2026-05-21
Feature: use the Obot CLI to discover additional MCP servers


g-linville:
~~We want to extend the `obot setup` command to also configure access to the Obot MCP Server (`https://<obot host>/mcp-connect/sms1obot-mcp-server`) in the local agents, add `obot mcp search` and `obot mcp configure` commands, and extend the bootstrapped `obot` skill to explain how it all works.~~

~~The goal is for the local agent (i.e. Claude Code) to be able to talk to Obot via MCP to search for and configure MCP servers, and then to use the local Obot CLI to configure them on itself.~~

We will extend the Obot CLI and the bootstrap skills to enable local agents to discover MCP servers available in Obot and retrieve their connection URLs.

g-linville:
This is merged. The Obot CLI now has an `obot mcp search` command, and the Obot bootstrap skills will teach the local agent how to use it.

IMPORTANT: in order to test this feature, ensure the following:

- The version of `obot` in your `PATH` must be built on Main (the version from Homebrew will not contain the new command). To check which `obot` is in your `PATH`, run `which obot`.
- Re-run `obot setup` with your newly built `obot` in order to replace your existing Obot bootstrap skills with the new ones.

I demonstrated this feature in standup yesterday. Basically, we are now enabling the user to discover MCP servers by chatting with their local agent. However, in order to configure them, they still have to go to the Obot UI, and in order to set them on their local agent, they need to do it manually or ask their agent to do it for them (if the agent supports modifying its own configuration). The agent may fail at modifying its own configuration if it is not smart (i.e. Sonnet seemed to fail while Opus worked).

sangee2004:
Tested with latest version of obot cli - `Version:  v0.0.0-dev-b3e7ca53-dirty`.

`obot setup` results in `obot-search-mcp-servers` skills being installed in `/Users/sangeethahariharan/.claude/skills` and in `/Users/sangeethahariharan/.agent/skills` directory.

From the `Claude` agent, able to search for all mcpservers hosted in obot 


<img width="1168" height="759" alt="Image" src="https://github.com/user-attachments/assets/bbf7053c-c0a2-4cf6-b454-d63b8bda3d19" />
<img width="1314" height="252" alt="Image" src="https://github.com/user-attachments/assets/77b998eb-923e-4d4f-8ad6-db129bc4df06" />


Was able to ask the agent to add one of the MCP servers from the list, and was able to add it successfully

[...]

PR #6767, opened 2026-05-29
feat: cli: add obot mcp search command

g-linville:
This adds an `obot mcp search` command to list and search for available MCP servers and catalog entries in Obot. This will (in the future) be available to local agents to use to discover MCP servers for the user.

demo video: https://acorn-io.slack.com/archives/C03FU91U4SX/p1780063441124139

part of https://github.com/obot-platform/obot/issues/6666

### DIFF
--- a/pkg/api/handlers/registry/convert.go
+++ b/pkg/api/handlers/registry/convert.go
@@ -183,29 +183,7 @@ func ConvertMCPServerCatalogEntryToRegistry(
 		}
 	}
 
-	// Check if the catalog entry requires configuration.
-	// Composite servers always require configuration in the UI before they can be used.
-	requiresConfiguration := manifest.Runtime == obottypes.RuntimeComposite
-
-	// Check for required environment variables
-	if !requiresConfiguration {
-		for _, env := range manifest.Env {
-			if env.Required {
-				requiresConfiguration = true
-				break
-			}
-		}
-	}
-
-	// Check for required headers (for remote runtime)
-	if !requiresConfiguration && manifest.Runtime == obottypes.RuntimeRemote && manifest.RemoteConfig != nil {
-		for _, header := range manifest.RemoteConfig.Headers {
-			if header.Required {
-				requiresConfiguration = true
-				break
-			}
-		}
-	}
+	requiresConfiguration := catalogEntryRequiresConfiguration(entry)
 
 	// Create metadata
 	meta := obottypes.RegistryMeta{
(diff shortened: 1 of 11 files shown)


pkg/api/handlers/registry/convert.go
```diff
@@ -243,6 +221,45 @@ func ConvertMCPServerCatalogEntryToRegistry(
 
 // Helper functions
 
+func catalogEntryRequiresConfiguration(entry v1.MCPServerCatalogEntry) bool {
```
njhale, line 224: https://github.com/obot-platform/obot/pull/6761/changes#diff-beb5b5949da34e08e82a2b26153e52a9f135ba026e4eab337ea3a8dcfa405727R1044

I'm introducing a similar function ^. It also happens to take care of composites (which, when it merges, will also get sharable connect url when there's no required configuration).

After this merges it may behoove us to write a shared helper.

pkg/api/handlers/registry/convert.go
```diff
@@ -243,6 +221,45 @@ func ConvertMCPServerCatalogEntryToRegistry(
...
+
+		// A hostname is not a usable connection URL. It is only a constraint for
+		// the URL the user must supply when configuring the server.
+		if manifest.RemoteConfig.Hostname != "" {
+			return true
+		}
+
+		if manifest.RemoteConfig.URLTemplate != "" {
+			return true
+		}
```
njhale: Can we just check `manifest.RemoteConfig.FixedURL == ""`

g-linville: good point

g-linville: fixed

pkg/api/handlers/registry/convert.go
```diff
@@ -243,6 +221,40 @@ func ConvertMCPServerCatalogEntryToRegistry(
...
+
+	// Composite servers always require configuration in the UI before they can be used.
+	if manifest.Runtime == obottypes.RuntimeComposite {
+		return true
+	}
+
+	for _, env := range manifest.Env {
+		if env.Required {
+			return true
+		}
```
calvinmclean: I put the same comment on Nick's PR so I'm not sure if it needs to be addressed here or not:

> `env` could be required and have a Secret Binding. In this case, it would not need a user config. This is also the case for headers below.
> 
> Similarly, is it possible to create a git-managed catalog with a required env that also already has a Value set? I'm not sure.

g-linville: Fixed. Besides your secret binding feature, there isn't any other way to pre-specify a required env var, so there isn't anything else to check besides the secret binding

calvinmclean: Cool, thanks for checking!

calvinmclean (approved):

njhale (approved):

PR #6817, opened 2026-06-03
chore: remove PROVIDER_IMAGE and build encryption providers

thedadams:


drpebcak (approved):
this might slow down the obot image build, but otherwise should work

PR #6831, opened 2026-06-04
fix: use a catalog entry's static OAuth credentials when generating tool previews

we4sz:
Generating tool previews for an OAuth-protected remote catalog entry fails even when the entry has static OAuth credentials configured.

### Problem
`GenerateToolPreviews` launches an ephemeral, never-persisted MCP server (`tool-preview-<hash>`) and connects to it to enumerate tools. For an OAuth-protected remote server this should use the catalog entry's configured static OAuth client, but it never can:

- `tempServerAndConfig` builds the temp server without setting `Spec.MCPServerCatalogEntryName`.
- The OAuth `ClientCredLookup` (`mcpOAuthHandler.Lookup`) resolves the static client by fetching the server **from storage by name** and reading `MCPServerCatalogEntryName` off it. The temp preview server isn't persisted, so that `Get` fails, `Lookup` returns "no credentials found", and the client falls back to Dynamic Client Registration — which strict IdPs (e.g. Keycloak) reject (`invalid_client_metadata`), surfacing as a 500.

### Fix
- Set `MCPServerCatalogEntryName` on the temp preview server (all `tempServerAndConfig` call sites pass the entry / component catalog-entry name).
- Carry the entry name onto `mcpOAuthHandler` (set in `CheckForMCPAuth` from the passed `mcpServer`).
- `Lookup` uses the carried entry name to resolve the static OAuth credentials directly, falling back to the storage lookup for persisted servers.

### Testing
Built locally and verified against an OAuth-protected remote server (Keycloak-fronted) with static OAuth credentials configured on the catalog entry: "Populate Tool Preview" now resolves the entry's static client, performs the normal login flow, and lists tools — instead of falling back to DCR and returning a 500.

`gofmt` clean; affected packages build.

### DIFF
--- a/pkg/api/handlers/mcpcatalogs.go
+++ b/pkg/api/handlers/mcpcatalogs.go
@@ -818,6 +818,7 @@ func (h *MCPCatalogHandler) GenerateToolPreviews(req api.Context) error {
 		req.ObotNamespace,
 		entry.Namespace,
 		catalogName,
+		entry.Name,
 		entry.Spec.Manifest,
 		configRequest.Config,
 		configRequest.URL,
@@ -942,6 +943,7 @@ func (h *MCPCatalogHandler) generateCompositeToolPreviews(req api.Context, entry
 			req.ObotNamespace,
 			entry.Namespace,
 			catalogName,
+			componentEntry.CatalogEntryID,
 			componentEntry.Manifest,
 			config.Config,
 			config.URL,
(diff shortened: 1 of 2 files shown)


we4sz:
Thanks for the review. Addressing the three points:

1. **`CatalogEntryID` vs catalog entry name (composite sites):** these are the same value in this codebase — `CatalogEntryID` *is* the catalog entry’s object name. Obot itself stamps it onto the spec field this resolves against (`pkg/controller/handlers/mcpserver/mcpserver.go`: `MCPServerCatalogEntryName: component.CatalogEntryID`) and `Get`s the entry by `CatalogEntryID`, so `system.MCPOAuthCredentialName(...)` resolves correctly for component previews. The single-server sites pass `entry.Name`. No change needed.

2. **Long positional parameter list:** agreed it’s not ideal, but `tempServerAndConfig` already took 11 positional args before this change; I kept the fix minimal and focused rather than expanding its scope. Happy to follow up with a params-struct refactor in a separate PR if you’d prefer.

3. **Field/param naming:** renamed `catalogEntryName` → `mcpServerCatalogEntryName` to align with the `MCPServerCatalogEntryName` spec field and avoid confusion with `catalogName`. Pushed.

gofmt clean; affected packages build. Verified end-to-end against a Keycloak-fronted remote server with static OAuth configured on the entry — tool previews now resolve the entry’s static client and complete the login flow instead of falling back to DCR.

PR #6842, opened 2026-06-04
enhance: ensure 402 error is presented before authorization

thedadams:
This helps with clients that are connecting so they get 402s instead of 401s.

### DIFF
--- a/pkg/api/server/server.go
+++ b/pkg/api/server/server.go
@@ -183,43 +183,43 @@ func (s *Server) Wrap(f api.HandlerFunc) http.HandlerFunc {
 			}
 		}
 
-		if !s.authorizer.Authorize(req, user) {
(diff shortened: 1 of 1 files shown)


pkg/api/server/server.go
```diff
@@ -183,43 +183,43 @@ func (s *Server) Wrap(f api.HandlerFunc) http.HandlerFunc {
...
-		}
+			if strings.HasPrefix(req.URL.Path, "/api/") {
+				rw.Header().Set("Cache-Control", "no-store, no-cache, must-revalidate, proxy-revalidate, max-age=0")
+				rw.Header().Set("Pragma", "no-cache")
+				rw.Header().Set("Expires", "0")
+			}
 
-		var shouldLogError bool
-		err = s.providerEntitlementGate.Check(req)
-		if err == nil {
```
thedadams, line 222: This looks more complicated than it is. This got moved above the authorization so the 402 takes precedence.

g-linville (approved):

Issue #6779, opened 2026-05-29
Multi-User server updates can leave deployments unusable due to missing required configuration


sangee2004:
**Describe the bug**
When a catalog update introduces a new required configuration variable for a deployed multi-user MCP server, using `Update Server` transitions the server to an updated state but does not prompt the administrator to provide the newly required configuration. As a result, the server cannot start successfully when users reconnect, and tool calls fail until an administrator manually edits and saves the server configuration. The update workflow leaves the server in an incomplete state despite reporting that the update has been applied.

**To Reproduce**
1. Create a Multiuser MCP server entry in the catalog, like - https://github.com/sangee2004/mcp-catalog/blob/main/sangeemultiexasearch.yaml with `serverUserType: multiUser`
2. Log in as an admin user and add this catalog as Registry Source.
3. Once the Multiuser MCP server entry gets synced, it shows up as a `Multi-User Catalog Entry`.
4. `Connect to this server`.
5. As any user says `user1`, connect to this Multi-user server and make tool calls.
6. Update the multi-user server entry in the catalog (from step 1) to add a new required variable.
7. Log in as admin and sync the catalog.
8. This results in `Multi-User Catalog Entry`  shown in `Server Requires an Update` state and  `Multi-User` entry shown in `An update Requires your attention` state.

<img width="1333" height="263" alt="Image" src="https://github.com/user-attachments/assets/4840c21f-7971-4b29-ba71-226220107024" />

9.  Go to the `Deployments and Connection` tab.
   `Multi-User` entry is presented in `Needs Config Update` state. 

<img width="1303" height="516" alt="Image" src="https://github.com/user-attachments/assets/de6bfdf6-745e-4f41-b382-19bf760663c8" />

10. Use `Update Server` option to update the server.
 This results in the server entry getting to `Shutdown` state with `up to date` status. MCP server pod gets deleted.

<img width="1277" height="490" alt="Image" src="https://github.com/user-attachments/assets/6f4b16fc-7bd9-4191-9759-08efcaffeb34" />

This clears the `An update Requires your attention` state for the`Multi-User` entry . But `Multi-User Catalog Entry`  continues to be in `Server Requires an Update` state

<img width="1326" height="205" alt="Image" src="https://github.com/user-attachments/assets/ef7a1870-fd3b-4cf3-a4f4-2b7f71ee5217" />

11.  As user1, make a connection to the MCP server by making a tool call. This will result in the tool call failing with the following errors related to the required configuration. not being set.

[...]

PR #6846, opened 2026-06-05
fix: ui: prompt for new required config params after trigger-update for multi-user

g-linville:
for https://github.com/obot-platform/obot/issues/6779

When you now call trigger-update on a multi-user server that was created from a multi-user catalog entry, we check after the update whether there are any new required configuration parameters, and prompt for them.

See it in action: https://www.loom.com/share/12058afd8a324034beebaaa893f5e9bd

### DIFF
--- a/ui/user/src/lib/components/admin/McpServerInstances.svelte
+++ b/ui/user/src/lib/components/admin/McpServerInstances.svelte
@@ -20,6 +20,7 @@
 	import { openUrl, isOwnSingleUserServer, getUserDisplayName } from '$lib/utils';
 	import Confirm from '../Confirm.svelte';
 	import DotDotDot from '../DotDotDot.svelte';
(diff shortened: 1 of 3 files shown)


g-linville:
@ivyjeong13 let me know if you think this one actually matters, or if it will just be unnecessary and cause conflicts with what you are doing.

ivyjeong13:
> @ivyjeong13 let me know if you think this one actually matters, or if it will just be unnecessary and cause conflicts with what you are doing.

I believe this issue is still valid for multi-user catalog entries! There might be some minor conflicts (pretty sure you can drop the changes for MCPServerInstances as it uses the DeploymentView with the rework) 

As for the Loom mentioning the multi-user servers created from a multi-user catalog entry in the Server Entries, that is being addressed in the rework. : )

ivyjeong13 (approved):

Issue #6844, opened 2026-06-05
Available models are missing immediately after provider configuration.


sangee2004:
**Describe the bug**
After successfully configuring a model provider, the user is automatically presented with the `Default Model Aliases` modal, but the modal initially displays an empty model list. The available models are not loaded even though the provider configuration has completed successfully. Additionally, the modal cannot be dismissed or closed, leaving the user stuck until they manually refresh the page. After a refresh, reopening the `Set Default Models `workflow correctly displays the available models.

Pre Req - Have a setup with the first owner user created.

**To Reproduce**
1. Log in as the owner user and configure a model provider.
Notice that after the model provider is configured, the user is presented with `Default Model Aliases` modal with an empty model list.

<img width="926" height="502" alt="Image" src="https://github.com/user-attachments/assets/634022b3-db58-40e2-9286-f818316bfbc4" />

<img width="871" height="447" alt="Image" src="https://github.com/user-attachments/assets/8d8450b5-7b77-47cb-89f6-d33f6f0b116d" />

Also, there is no way to dismiss or close the `Default Model Aliases` modal.

After we manually refresh the page and click `Set Default Models`, the user is presented with the list of models as expected.

<img width="741" height="530" alt="Image" src="https://github.com/user-attachments/assets/c1e70f69-7c5e-446f-8b3a-bc5a77ecdd49" />

**Expected Behavior:**
The user should be presented with a list of available models as soon as the model provider is configured successfully.

sangee2004:
This issue is addressed when testing with the latest build from `main`.

PR #6855, opened 2026-06-05
chore: revert back to model and auth providers as fork/exec

thedadams:
This change removes auth and model providers as their own containers and moves them back into the Obot containers as binaries. These providers have to be local paths and cannot be remote.

This change also removes all remaining dead code.

Addresses: https://github.com/obot-platform/obot/issues/6844


thedadams:
The tests here will fail until the providers PR is merged.

pkg/gateway/server/llmproxy.go
```diff
@@ -893,11 +894,31 @@ func mustParseURL(s string) *url.URL {
...
+			// passthrough routes). Strip a leading v1/ from the client-supplied
+			// path so we don't produce /v1/v1/...
+			reqPath = strings.TrimPrefix(reqPath, "v1/")
+			if reqPath == "v1" {
+				reqPath = ""
+			}
+		}
+		u.Path = path.Join(u.Path, reqPath)
+		req.URL = &u
+		req.Host = u.Host
```
g-linville: Is this fixing some existing bug? Seems unrelated to the main changes in this PR.

pkg/gateway/server/dispatcher/daemon.go
```diff
@@ -0,0 +1,265 @@
...
+
+func (d *Dispatcher) newCommand(ctx context.Context, envMap map[string]string, command string, args ...string) (*exec.Cmd, func(), error) {
+	ctx, cancel := context.WithCancel(ctx)
+
+	// Expand and/or normalize env references
+	for i, arg := range args {
+		args[i] = os.Expand(arg, func(s string) string {
+			if strings.HasPrefix(s, "!") {
+				return envMap[s[1:]]
+			}
```
g-linville: What does this `!` prefix do? It seems strange.

g-linville (approved):

pkg/gateway/server/dispatcher/daemon.go
thedadams: Not sure, so I left it (from copying over from GPTScript).

Should I remove it?

pkg/gateway/server/llmproxy.go
thedadams: We were only using the `dispatcher.TransformRequest` in this one place, so just copied the implementation to here.

It's only related because I was trying to clean up the dispatcher.

pkg/gateway/server/dispatcher/daemon.go
g-linville: Probably. I went and looked at our old tools and found this: https://github.com/obot-platform/tools/blob/8d6f854606e32e824d750a2f3de999831e38be4a/placeholder-credential/tool.gpt#L5

So I assume it was to remove the `#!` prefix, which we shouldn't be dealing with anymore.

Issue #6816, opened 2026-06-03
Users cannot complete model provider downgrade because provider ID is not shown.


sangee2004:
**Describe the bug**
During the license downgrade workflow, when the administrator downgrades the configured model provider, the `Confirm Downgrade` modal displays messages that appear tailored to authentication providers rather than model providers. Additionally, the confirmation dialog instructs the user to enter the provider ID to proceed, but the provider ID is displayed as this (or is otherwise not visible), making it impossible for the user to complete the confirmation step.

**Pre Req** - Had an enterprise setup with `Microsoft Entra provided` auth provider and `Amazon Bedrock (API Key)` model provider.

**To Reproduce**
1. Upgrade the setup to the latest build from `main` after commit - [v0.0.0-dev+6fad657c](https://github.com/obot-platform/obot/commit/6fad657c) 
This will result in the setup being in an invalid-license state with an invalid license Banner.
2. Click on the Resolve option in the banner
3. When the user is presented with the following options, click on `Downgrade`.

<img width="521" height="521" alt="Image" src="https://github.com/user-attachments/assets/9f425f03-5c9b-4a75-b6d8-c338aa98e9af" />

The user will be guided through the steps to downgrade the Entra provider, which will force the admin user to log in as the bootstrap user.

4. When logged in as a bootstrap user, click on the `Resolve` option in the Invalid License banner.

<img width="546" height="470" alt="Image" src="https://github.com/user-attachments/assets/2201f53f-aa7d-470d-a05d-a8f7fc5b9ae6" />

5. Click on the `Downgrade` option to deconfigure the model provider.

User is presented with the following `confirm Downgrade` modal
<img width="1006" height="624" alt="Image" src="https://github.com/user-attachments/assets/82e4d6bd-e9b7-461b-9056-3751fef62dec" />

Notice that the information presented in the`Confirm Downgrade` modal does not apply to downgrading a Model provider and is specific to downgrading an auth provider.

Also, the provider ID is referenced as `this`, and the user is asked to enter the Provider ID to confirm, which they cannot know.

**Expected behavior**

The exact Provider ID should be presented for the user to enter to confirm the downgrade process.

sangee2004:
This issue is addressed when testing with the latest build from `main`.

User is presented with the model provider ID to enter, and when this value is entered able to downgrade the provider successfully.
<img width="1023" height="557" alt="Image" src="https://github.com/user-attachments/assets/75146931-6073-4af7-a892-4106aeb56365" />

Issue #6814, opened 2026-06-03
Nested modal remains visible after cancelling downgrade.


sangee2004:
**Describe the bug**
When resolving an invalid license state, selecting `Downgrade` opens the `Confirm Downgrade` modal as expected. However, clicking `Nevermind` leaves the top-level confirmation dialog still open in the background, keeping the underlying Invalid License resolution modal open as well. This results in overlapping modal states and an inconsistent dismissal flow.

**Pre Req** - Had an enterprise setup with `Microsoft Entra provided` auth provider and `Amazon Bedrock (API Key)` model provider.

**To Reproduce**
1. Upgrade the setup to the latest build from `main` - [v0.0.0-dev+6fad657c](https://github.com/obot-platform/obot/commit/6fad657c) 
This will result in the setup being in an invalid-license state with an invalid license Banner.
2. Click on the Resolve option in the banner
3. When the user is presented with the following options, click on `Downgrade`.
<img width="521" height="521" alt="Image" src="https://github.com/user-attachments/assets/16d4355e-17ce-47f7-bd61-a085333039e8" />
4. In the `Confirm Downgrade` modal, click on `Nevermind`
<img width="1029" height="584" alt="Image" src="https://github.com/user-attachments/assets/c529d75a-a71b-4dc0-9e41-c35e9ff09244" />

Notice that this results in the `Confirm Downgrade` modal still being presented in the background.

<img width="1025" height="591" alt="Image" src="https://github.com/user-attachments/assets/0436a916-1bf0-447a-b335-f1f7c9506266" />

**Expected behavior**
Both the`Invalid License` and `Confirm Downgrade` modal should get dismissed.

cjellick:
accidentally assigned to donnie

sangee2004:
This issue is addressed when testing with the latest build from `main`.

Issue #6809, opened 2026-06-03
Non admin users are not informed when system is unavailable due to license issues.


sangee2004:
**Describe the bug**
When the deployment is in an invalid-license or missing-license state, non-admin and non-owner users can still log in successfully, but they are not informed that the system is operating with restricted functionality. Instead, they encounter failures later when accessing features such as the `Skills` page or launching an Agent. The absence of a clear notification creates a confusing user experience and makes the system appear broken rather than unavailable due to licensing issues.

**Pre Req** - Had an enterprise setup with `Microsoft Entra provided` auth provider and `Amazon Bedrock (API Key)` model provider.
 
**To Reproduce**
1. Upgrade the setup to the latest build from `main` - [v0.0.0-dev+6fad657c](https://github.com/obot-platform/obot/commit/6fad657c) 
This will result in the setup being in an invalid-license state, as expected.

2. Log in as a non-admin/non-owner user.

Notice that no information is provided to the user, indicating they cannot fully use the setup.

They encounter errors when accessing the skills page and when trying to launch the agent.

**Expected behavior**
User should be presented with a banner like - "We're sorry, this system is currently not available. Please contact your administrator."

cjellick:
lets do 

We're sorry, this system is currently operating with limited functionality. Please contact your administrator.

sangee2004:
This issue is addressed in the `v0.23.0-rc3` build.

Non-admin users see the following banner.

<img width="1653" height="234" alt="Image" src="https://github.com/user-attachments/assets/62091064-f518-4e8e-9b76-80274c9d052f" />

Issue #6808, opened 2026-06-03
License banner incorrectly reports missing license as invalid.


sangee2004:
**Describe the bug**
When the deployment is upgraded into a state where no valid license is present, the admin banner displays the message "Your license is invalid". This is misleading because the actual condition is that the license is missing, not invalid. Other areas of the UI, such as the Auth Provider page, correctly indicate a `License Missing` state, resulting in inconsistent messaging across the product.

Pre Req - Had an enterprise setup with `Microsoft Entra provided` auth provider and `Amazon Bedrock (API Key)` model provider.
 
**To Reproduce**
1. Upgrade the setup to the latest build from `main` - [v0.0.0-dev+6fad657c](https://github.com/obot-platform/obot/commit/6fad657c) 
This will result in the setup being in an invalid-license state, as expected.

2. Log in as the admin user.

Notice that the user is presented with `Your license is invalid. For full functionality, it is recommended to resolve the outstanding issues` message in the banner.

It is confusing to be presented with the `Your license is invalid` message when the license is actually missing.

`Auth Provider` page shows `License Missing` state as expected.

<img width="983" height="745" alt="Image" src="https://github.com/user-attachments/assets/1e4c0015-b8ab-4d7a-9f99-9907ddb7bbaa" />

**Expected behavior**
Banner should indicate that the license is missing instead of stating that the license is invalid.

thedadams:
@ivyjeong13 Can you distinguish between an invalid license and a missing license by getting the license to see if it exists? If you need backend work for this, let me know, but I think you have everything.

cjellick:
you can just say "missing or invalid"

sangee2004:
Tested with `v0.23.0-rc3`.

Admin is now presented with the following "Missing License" banner, as expected.

<img width="1572" height="113" alt="Image" src="https://github.com/user-attachments/assets/03cd3275-5b75-4408-9ca7-ba94ec1e059b" />

PR #6861, opened 2026-06-07
fix: licensing bugfixes

ivyjeong13:
Addresses #6814 
Addresses #6808 
Addresses #6809 
Addresses #6816 
Addresses #6805
Addresses #6799 - super quick text change

Backend Changes:
* GET /api/license becomes available for all authenticated users, the `licenseKey` property is fully masked for non-admins, partially masked for admins

Frontend Changes:
* moved getLicense call to UserService & license store
* address bugs mentioned in issues

### DIFF
--- a/pkg/api/handlers/license.go
+++ b/pkg/api/handlers/license.go
@@ -95,11 +95,8 @@ func displayLicenseKey(licenseKey string, canViewPartial bool) string {
 	if licenseKey == "" {
 		return ""
 	}
-	if !canViewPartial {
+	if !canViewPartial || len(licenseKey) <= licenseKeyVisibleSuffix {
 		return licenseKeyMask
-	}
-	if len(licenseKey) <= licenseKeyVisibleSuffix {
-		return licenseKeyMask + licenseKey
 	}
 	return licenseKeyMask + licenseKey[len(licenseKey)-licenseKeyVisibleSuffix:]
 }
--- a/pkg/api/handlers/license.go
+++ b/pkg/api/handlers/license.go
@@ -75,7 +75,7 @@ func (h *LicenseHandler) Delete(req api.Context) error {
 func (h *LicenseHandler) status(req api.Context) LicenseStatus {
 	licenseKey := h.licenseProvider.LicenseKey()
 	status := LicenseStatus{
-		LicenseKey:   displayLicenseKey(licenseKey, req.UserIsAdmin() || req.UserIsOwner()),
+		LicenseKey:   displayLicenseKey(licenseKey, req.UserIsAdmin()),
 		Locked:       h.licenseProvider.LicenseKeyViaConfiguration(),
 		Enterprise:   h.licenseProvider.HasValidLicense(),
 		Entitlements: h.licenseProvider.Entitlements(),
--- a/pkg/api/handlers/license.go
+++ b/pkg/api/handlers/license.go
@@ -25,12 +25,17 @@ type LicenseUpdate struct {
 	LicenseKey string `json:"licenseKey"`
 }
 
+const (
+	licenseKeyMask          = "****"
+	licenseKeyVisibleSuffix = 8
+)
+
 func NewLicenseHandler(licenseProvider *license.KeygenProvider) *LicenseHandler {
 	return &LicenseHandler{licenseProvider: licenseProvider}
 }
 
 func (h *LicenseHandler) Get(req api.Context) error {
-	return req.Write(h.status())
+	return req.Write(h.status(req))
 }
 
 func (h *LicenseHandler) Update(req api.Context) error {
@@ -53,7 +58,7 @@ func (h *LicenseHandler) Update(req api.Context) error {
 		return err
 	}
 
-	return req.Write(h.status())
+	return req.Write(h.status(req))
 }
 
 func (h *LicenseHandler) Delete(req api.Context) error {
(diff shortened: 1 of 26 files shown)


pkg/api/handlers/license.go
```diff
@@ -64,22 +69,37 @@ func (h *LicenseHandler) Delete(req api.Context) error {
...
+	return req.Write(h.status(req))
 }
 
-func (h *LicenseHandler) status() LicenseStatus {
+func (h *LicenseHandler) status(req api.Context) LicenseStatus {
+	licenseKey := h.licenseProvider.LicenseKey()
 	status := LicenseStatus{
-		LicenseKey:   h.licenseProvider.LicenseKey(),
+		LicenseKey:   displayLicenseKey(licenseKey, req.UserIsAdmin() || req.UserIsOwner()),
 		Locked:       h.licenseProvider.LicenseKeyViaConfiguration(),
```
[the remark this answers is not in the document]
ivyjeong13: owners should too

pkg/api/handlers/license.go
```diff
@@ -64,22 +69,37 @@ func (h *LicenseHandler) Delete(req api.Context) error {
...
+	licenseKey = strings.TrimSpace(licenseKey)
+	if licenseKey == "" {
+		return ""
+	}
+	if !canViewPartial {
+		return licenseKeyMask
+	}
+	if len(licenseKey) <= licenseKeyVisibleSuffix {
+		return licenseKeyMask + licenseKey
+	}
```
g-linville:
```suggestion
	if !canViewPartial || len(licenseKey) <= licenseKeyVisibleSuffix {
		return licenseKeyMask
	}
```

(The second `if` statement represented a probably impossible edge case, but I didn't like that it would have shown the full key.)

pkg/api/handlers/license_test.go
```diff
@@ -0,0 +1,54 @@
...
+			licenseKey:     "keygen/abc123j13lasds",
+			canViewPartial: true,
+			want:           "****j13lasds",
+		},
+		{
+			name:           "admin short key shows full suffix",
+			licenseKey:     "shortkey",
+			canViewPartial: true,
+			want:           "****shortkey",
+		},
```
g-linville, line 37: This test will need to be updated too:


```suggestion
		{
			name:           "admin short key shows full suffix",
			licenseKey:     "shortkey",
			canViewPartial: true,
			want:           licenseKeyMask,
		},
```

pkg/api/handlers/license.go
```diff
@@ -64,22 +69,34 @@ func (h *LicenseHandler) Delete(req api.Context) error {
...
-	return req.Write(h.status())
+	return req.Write(h.status(req))
 }
 
-func (h *LicenseHandler) status() LicenseStatus {
+func (h *LicenseHandler) status(req api.Context) LicenseStatus {
+	licenseKey := h.licenseProvider.LicenseKey()
 	status := LicenseStatus{
-		LicenseKey:   h.licenseProvider.LicenseKey(),
+		LicenseKey:   displayLicenseKey(licenseKey, req.UserIsAdmin() || req.UserIsOwner()),
```
njhale: Owners are admins. IIUC, UserIsAdmin should always return true if they're an owner.

ivyjeong13: Thanks Nick! Will drop the owner check!

pkg/api/handlers/license_test.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			if got := displayLicenseKey(tt.licenseKey, tt.canViewPartial); got != tt.want {
+				t.Fatalf("displayLicenseKey() = %q, want %q", got, tt.want)
+			}
+		})
+	}
```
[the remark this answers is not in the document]
njhale: I _think_ this has been resolved with go 1.22+ https://go.dev/blog/loopvar-preview

pkg/api/handlers/license.go
njhale: Based on what was here before, this should be `req.UserIsAdmin()` right?

pkg/api/handlers/license.go
```diff
@@ -64,22 +69,34 @@ func (h *LicenseHandler) Delete(req api.Context) error {
...
-	return req.Write(h.status())
+	return req.Write(h.status(req))
 }
 
-func (h *LicenseHandler) status() LicenseStatus {
+func (h *LicenseHandler) status(req api.Context) LicenseStatus {
+	licenseKey := h.licenseProvider.LicenseKey()
 	status := LicenseStatus{
-		LicenseKey:   h.licenseProvider.LicenseKey(),
+		LicenseKey:   displayLicenseKey(licenseKey, req.UserIsOwner()),
```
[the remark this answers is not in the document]
ivyjeong13: ack, wrong check was removed, should be req.UserIsAdmin() and if user is owner, then user is also admin so just need that check.

pkg/api/handlers/license_test.go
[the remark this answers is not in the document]
ivyjeong13: :+1: Think copilot wants to point it out every time for some reason, ran tests multiple times and checked what other tests file were doing. Haven't run into flaky/incorrect tests yet.

g-linville (approved):
backend lgtm

njhale (approved):

PR #6873, opened 2026-06-08
feat: add APIs for local agent audit logs

g-linville:
This is the first PR for local agent audit logs support.

We are going to extend the Obot CLI to add a new `obot audit` command that will be run using hooks in local agents like Claude Code and Codex. The command will capture the details of a tool call and submit it to Obot as a local agent audit log.

These are being stored in their own separate data because the structure of these logs is quite different from our existing audit logs.

This first PR introduces the following:
- Some basic examples of hook configurations for different agents (will be used in a future PR for tests, etc.)
- A new setting on API keys that gives them the ability to submit local agent audit logs
- APIs for querying local agent audit logs
  - This works in essentially the same way as our existing audit logs, where sensitive data is encrypted and can only be viewed by Auditors


rpelevin:
This seems like the right first slice for local agent audit logs. Because these records are coming from hooks around local agents, I would treat each audit log as a tool-call evidence envelope rather than a generic event.

Minimum fields I would preserve:

- local agent kind;
- session or run id;
- tool call id;
- tool name;
- canonical arguments digest;
- decision state if known;
- caller/workspace principal;
- hook source;
- timestamp;
- outcome if execution completed;
- redaction/encryption status;
- correlation id for later outcome updates.

For privacy, raw sensitive fields can stay encrypted or redacted, but stable digests and non-secret metadata should remain queryable for auditors.

Acceptance tests I would want:

- API key with local-audit permission cannot write unrelated audit types;
- API key cannot read decrypted sensitive payloads unless it has auditor rights;
- duplicate hook submission is idempotent;
- denied, cancelled, or rejected tool calls can be recorded without implying execution;
- outcome update links back to the original proposed tool call;
- filters do not leak encrypted payload content.

That makes the API useful for Claude/Codex-style local agents without turning the audit log into an unbounded transcript.

apiclient/types/localagentauditlog.go
```diff
@@ -0,0 +1,47 @@
...
+import "encoding/json"
+
+// LocalAgentAuditLog represents a local agent tool-call audit event.
+type LocalAgentAuditLog struct {
+	ID                 uint            `json:"id"`
+	EventID            string          `json:"eventID"`
+	CreatedAt          Time            `json:"createdAt"`
+	UserID             string          `json:"userID"`
+	ClientInfo         ClientInfo      `json:"client"`
+	ClientIP           string          `json:"clientIP,omitempty"`
```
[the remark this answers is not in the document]
g-linville: fixed by defining a separate ingest type

apiclient/types/localagentauditlog.go
```diff
@@ -0,0 +1,77 @@
+package types
+
+import "encoding/json"
+
+// LocalAgentAuditLogIngest represents a local agent tool-call audit event submitted by a client.
+type LocalAgentAuditLogIngest struct {
```
njhale, line 6: Seems like a fair number of these fields can be extracted out into an embedded type that can be embedded in both `LocalAgentAuditLogIngest` and `LocalAgentAuditLog`

njhale (approved):
One comment, otherwise first pass review lgtm

apiclient/types/localagentauditlog.go
g-linville: done

g-linville:
whoops, meant to close this sooner. we are redesigning this

Issue #6785, opened 2026-05-29
Single-User MCP server remains `Up to date` after scheduling config changes.


sangee2004:
**Describe the bug**
Updating the `Resource Limits & Requests` values for an already connected single-user MCP server saves successfully, and the new values persist in the `Configuration` tab. However, the existing deployment remains marked as `Up to date` in the `Deployments and Connections` tab, so there is no way to apply the updated scheduling configuration to the running deployment.

Pre Req - obot installed using helm charts
**To Reproduce**
Scenario 1:
1. Create a single-user MCP server. Do not set values for any of the fields in `Resource Limits & Requests`.
2. Connect to this MCP server.
The MCP server pod gets deployed with the default setting ( CPU Requests - 10m and Memory Requests - 200Mi or with values set in `Server Scheduling`settings)
3. Update the single-user MCP server created in step1 by going to the `Configuration` tab and modifying any values in the `Resource Limits & Requests` section

MCP server update reports success. I am able to see the updates done to the `Resource Limits & Requests` section persist when revisiting the `Configuration` tab

4. Go to the `Deployments and Connections` tab

Notice that the MCP server deployment shows in `Up to date` status.
<img width="1316" height="163" alt="Image" src="https://github.com/user-attachments/assets/65fe3ca1-b8e2-4f07-a947-bef8cd9682c1" />

**Expected behavior**
MCP server deployment should enter the `Needs Scheduling Update` state so we can update the server with the new config changes using the `Update Scheduling Config` option.

**Note** This issue is not seen when updating `Resource Limits & Requests` for multi-user MCP servers. MCP server deployment gets to `Needs Scheduling Update` state as expected.

sangee2004:
This issue is also seen for  Composite MCP Server deployments.

When any of the fields in `Resource Limits & Requests` are updated for a Composite MCP Server deployment, the MCP server deployment continues to be shown in `Up to date` status. 

There is no way to have the Composite MCP Server deployment upgraded to the updated `Resource Limits & Requests` settings.
<img width="1326" height="212" alt="Image" src="https://github.com/user-attachments/assets/2b12dd11-ddfd-432f-af28-8a529e22fea2" />

thedadams:
@sangee2004 In this case, the server will go into a "Needs Update" state because the configuration of the catalog entry has changed. It's a subtle difference, but making a change to the catalog entry produce a "Needs Scheduling Update" would be a large change to the code that I don't think is worth it.

[...]

PR #6901, opened 2026-06-10
fix: clear needsK8sUpdate when trigger-update syncs server resources

ivyjeong13:
After Update Server is called, the MCP server is shut down and synced with updated catalog config, DetectK8sSettingsDrift could leave needsK8sUpdate set even though the next deploy would use the updated resource limits. Clear the flag and set K8sSettingsHash to the expected value after config sync on the Kubernetes backend.

To Address https://github.com/obot-platform/obot/issues/6785#issuecomment-4627151187

Wanted to check this is right path to address, otherwise happy to close it out!

### DIFF
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -51,6 +51,18 @@ type MCPHandler struct {
 
 	// shutdownMCPServer is only injected for testing
 	shutdownMCPServer func(string) error
+	// mcpRuntimeBackend is only injected for testing
+	mcpRuntimeBackend string
+}
+
+func (m *MCPHandler) runtimeBackend() string {
+	if m.mcpRuntimeBackend != "" {
+		return m.mcpRuntimeBackend
+	}
+	if m.mcpSessionManager == nil {
+		return ""
+	}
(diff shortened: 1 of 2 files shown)


pkg/api/handlers/mcp.go
```diff
@@ -3953,14 +3965,61 @@ func (m *MCPHandler) TriggerUpdate(req api.Context) error {
...
 		updateServerFromCatalogEntry(&latest, entry)
+		latest.Namespace = req.Namespace()
 		return req.Update(&latest)
 	}); err != nil {
 		return err
 	}
 
+	if err := m.clearK8sUpdateStatusAfterConfigSync(req, server.Name); err != nil {
+		return fmt.Errorf("failed to update server k8s status after config sync: %w", err)
+	}
```
thedadams, line 3976: We shouldn't be updating status fields in the API because it causes conflicts, which is what you are working around here. Instead, I like this pattern: https://github.com/obot-platform/obot/blob/main/pkg/api/handlers/authprovider.go#L156-L162

You would wait for `!latest.Status.NeedsK8sUpdate && latest.Status.K8sSettingsHash == currentHash`

thedadams (changes requested):

Issue #6830, opened 2026-06-04
Allow wildcard suffixes in Model Access Policies


njhale:
Allow admins to provide wildcard suffixes when specifying the model/s to grant access to in a Model Access Policy.

This should allow admins to grant access to models like `claude-haiku-4.5-<date>` by specifying `claude-haiku-4.5*`.

sangee2004:
This issue is addressed when testing with the latest build from `main` - [v0.0.0-dev+9596afbe](https://github.com/obot-platform/obot/commit/9596afbe)

Allowed to create Model Access Policies with wild card suffixes for models.

<img width="999" height="681" alt="Image" src="https://github.com/user-attachments/assets/f8fe7710-9331-4447-bc15-1fca2de3c764" />

When using Obot as an LLM Gateway, Claude Code is able to see the models served from this model access policy as expected.

<img width="1019" height="173" alt="Image" src="https://github.com/user-attachments/assets/14da0376-0c36-4b7e-b862-14a14f472f1f" />

PR #6912, opened 2026-06-11
enhance: allow wildcard suffixes in model access policies

njhale:
Model Access Policies now accept wildcard suffix patterns (e.g.
"claude-haiku-4-5*") as model entries, granting access to every model
whose ID starts with the given prefix.

- A pattern is a non-empty prefix followed by a single trailing '*'.
  A '*' anywhere else is rejected at validation, and the bare "*"
  entry keeps its existing "all models" meaning.
- Patterns match case-sensitively against each model's provider-native
  target model (e.g. "claude-haiku-4-5-20251001"), across all model
  providers.
- Patterns are forward-looking: when a provider adds a new model whose
  target model starts with the prefix, subjects of the policy gain
  access without any policy changes. Accordingly, the reconciler keeps
  pattern entries even when they currently match no models.
- Patterns can coexist with explicit model IDs and default model alias
  references in the same policy; overlapping grants are harmless.

Patterns are expanded where user access is evaluated, so the LLM proxy
access check, the proxy model listing, and the user-facing models API
all honor them consistently. In the admin UI, typing a trailing '*' in
the Add Models dialog offers the pattern as a selectable entry with a
live preview of currently matching models, and saved patterns display
with a badge and match count.

Addresses https://github.com/obot-platform/obot/issues/6830

### DIFF
--- a/pkg/modelaccesspolicy/helper.go
+++ b/pkg/modelaccesspolicy/helper.go
@@ -106,8 +106,25 @@ func (h *Helper) UserHasAccessToModel(user kuser.Info, modelID string) (bool, er
 // If a user is an owner/admin or has been granted access to all models via a wildcard model selector, this method returns nil and true.
 func (h *Helper) GetUserAllowedModels(user kuser.Info) (map[string]bool, bool, error) {
 	var (
-		allowedModels   = make(map[string]bool)
-		aliasModels     = h.getAliasModels()
+		allowedModels = make(map[string]bool)
+		aliasModels   = h.getAliasModels()
+
+		// Models are listed at most once per call, no matter how many wildcard
+		// suffix patterns appear across the user's policies.
+		allModels  []*v1.Model
+		listModels = func() []*v1.Model {
+			if allModels == nil {
+				objs := h.modelIndexer.List()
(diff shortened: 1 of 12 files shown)


pkg/modelaccesspolicy/helper.go
```diff
@@ -106,8 +106,25 @@ func (h *Helper) UserHasAccessToModel(user kuser.Info, modelID string) (bool, er
...
+				objs := h.modelIndexer.List()
+				allModels = make([]*v1.Model, 0, len(objs))
+				for _, obj := range objs {
+					if m, ok := obj.(*v1.Model); ok {
+						allModels = append(allModels, m)
+					}
+				}
+			}
+			return allModels
+		}
```
thedadams: The model indexer is a cached client, right? So we're caching again? Is it worth it to do that?

thedadams (approved):

ivyjeong13 (approved):
ui changes LGTM

pkg/modelaccesspolicy/helper.go
njhale: updated

thedadams (approved):

Issue #6896, opened 2026-06-09
Upgrade - Agent chat fails with "forbidden" after upgrade to `v0.23.0-rc3`


sangee2004:
**Describe the bug**
After upgrading an Obot server from `v0.22.1` to `v0.23.0-rc3`, users correctly enter a `Missing License` state and are blocked from launching agents until a valid license is applied. However, after restoring a valid license and successfully deploying an agent (nanobot-agent:v0.0.84), chat sessions fail with a request failed: "forbidden\n" error. 

**To Reproduce**
1. Install `v0.22.1` obot server using helm install with Microsoft Entra Auth provider and Amazon Bedrock Model provider.
2. Log in as the owner user and the basic user and connect to a few MCP servers.
3. Launch Agents and chat with some of the MCP servers.
4. Upgrade to the `v0.23.0-rc3` version of the obot server.
5. User will be put in `Missing License` state.
6. Attempting to launch the agent will report failures relating to the license, which is as expected.
7. Set a valid License.
8. Launch Agent. (Agent gets deployed with `ghcr.io/obot-platform/nanobot-agent:v0.0.84`)
9. Start a chat session and send a message

Notice that chat fails with `request failed: "forbidden\n"` error

<img width="1086" height="474" alt="Image" src="https://github.com/user-attachments/assets/74de472e-bfd4-4715-bbfa-2a58286db620" />

**Note** - This error is seen for all existing users and also for new users.

**Expected behavior**
Should be able to chat successfully using the Obot agent.

sangee2004:
When testing the same upgrade scenario mentioned in the issue by upgrading from `v0.22.1` to the latest `main` - [v0.0.0-dev+5ce8cff0](https://github.com/obot-platform/obot/commit/5ce8cff0), I am able to chat with models from `Amazon Bedrock Model provider`  without seeing any errors. This was the model provider that was used for chatting before the upgrade as well.

When I switch the default models to models from the Anthropic model provider, chat starts failing with following errors:

<img width="904" height="264" alt="Image" src="https://github.com/user-attachments/assets/eb05e658-b36f-47b3-8253-6f26b7c47a61" />

In the `main` environment, which is also using models from the Anthropic model provider, it fails with following error:

<img width="886" height="384" alt="Image" src="https://github.com/user-attachments/assets/4b274126-0dfb-4033-8339-a00a4d1a57dc" />

sangee2004:
[...]

PR #6913, opened 2026-06-11
refactor: depracate auth tokens and extend api tokens

thedadams:
Previously, API tokens were used to only connect to MCP servers and do a few related API calls. Auth tokens were used to give the user access to the Obot API. This change consolidates this functionality into API keys. This include the ability to create keys with various capabilities from the UI and the CLI.

Additionally, this change consolidates authorization logic to be in the authorization layer as much as possible and uses groups to indicate access.

Screenshots:

Screenshot 2026-06-11 at 11 49 27
Screenshot 2026-06-11 at 11 49 34

Issue: https://github.com/obot-platform/obot/issues/6896


g-linville:
I haven't started my review yet, but from a UX perspective, it might make sense to just hide the published artifacts portion of this.

pkg/api/authz/resources.go
```diff
@@ -126,60 +114,68 @@ var apiResources = map[string][]string{
...
+		"DELETE /api/published-artifacts/{artifact_id}",
 	},
 }
 
 type Resources struct {
-	MCPServerID         string
-	MCPServerInstanceID string
+	MCPServerID           string
+	MCPServerInstanceID   string
+	MCPServerCatalogEntry string
```
g-linville: nit: we could rename this to `MCPServerCatalogEntryID` for consistency

pkg/api/handlers/skills_test.go
```diff
@@ -380,10 +383,11 @@ func TestSkillHandlerGetACRBypass(t *testing.T) {
...
 		assert.Equal(t, "sk1", skill.ID)
 	})
-	t.Run("basic user without ACR gets NotFound", func(t *testing.T) {
+	t.Run("basic user without ACR gets skill", func(t *testing.T) {
 		code, skill := doGet(t, testUser("user1"))
-		require.Equal(t, 0, code)
-		require.Nil(t, skill)
+		require.Equal(t, http.StatusOK, code)
+		require.NotNil(t, skill)
+		assert.Equal(t, "sk1", skill.ID)
```
g-linville: This seems incorrect to me? The `user1` user here gets only the GroupAuthenticated so they shouldn't be able to get this skill if no SkillAccessRule grants them access, right?

thedadams: This is a test on the API handler, which no longer does the authorization check because it was moved to the authorization layer.

I see two options here:
1. Keep the test and add a comment explaining what I did above; or
2. Remove the test.

What are your thoughts?

g-linville: Oh I see, that makes sense. I think we can remove the test.

thedadams: Done

pkg/api/authz/resources.go
thedadams: Done.

g-linville (approved):

njhale (approved):

PR #6924, opened 2026-06-12
fix: ensure NeedsK8sUpdate is set properly

thedadams:
This change simplifies the logic for setting the boolean identifying when a server needs a K8s settings update.

Issue: https://github.com/obot-platform/obot/issues/6785

### DIFF
--- a/pkg/controller/handlers/provider/provider.go
+++ b/pkg/controller/handlers/provider/provider.go
@@ -527,7 +527,6 @@ func (h *Handler) BackPopulateModels(req router.Request, _ router.Response) erro
 		})
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/controller/handlers/provider/provider.go
```diff
@@ -527,7 +527,6 @@ func (h *Handler) BackPopulateModels(req router.Request, _ router.Response) erro
 		})
 	}
 
-	// TODO: ensure this works when the owner reference type changes. Maybe we need to migrate.
```
thedadams, line 530: This was accidental, but I don't want to push again and trigger another Copilot review unless I have to. I will be addressing this in an imminent PR.

njhale (approved):

PR #6927, opened 2026-06-12
fix: use x-api-key header if authorization header is not set

thedadams:
This allows MCP api keys to be used with Anthropic-style authentication.

Issue: https://github.com/obot-platform/obot/issues/6896


pkg/gateway/server/apikey_auth.go
```diff
@@ -26,25 +26,21 @@ func NewAPIKeyAuthenticator(client *client.Client) *APIKeyAuthenticator {
...
-	}
-
-	bearer := strings.TrimPrefix(authHeader, "Bearer ")
-	if bearer == authHeader {
-		// No "Bearer " prefix
-		return nil, false, nil
+		authHeader = req.Header.Get("X-API-Key")
+		if authHeader == "" {
+			return nil, false, nil
+		}
```
thedadams, line 34: This is the main change. The rest is just removing `Tool` from the name of some constants.

cjellick (approved):

PR #6928, opened 2026-06-12
chore: migrate mcp audit logs to support local agent audit logs

g-linville:
This is the first step towards local agent audit log support. I'm modifying the MCPAuditLog type to make it more flexible for us in the future. I also introduced a database migration to populate additional fields.

We now have struct pointer fields that hold subtype-specific fields. Since we use `gorm:"embedded"` on these, each field inside will become its own column, which is why I introduced the `Context` field for local agent audit logs, as a catch-all column to hold JSON-formatted data that we don't need to index on.

The comments in the code should provide some helpful information. I would recommend looking at the gateway/types changes first, then the apiclient/types changes, then the database migration logic, then everything else.

No Nanobot changes are needed for this.


pkg/gateway/client/mcpauditlog.go
```diff
@@ -332,6 +361,8 @@ func (c *Client) GetMCPAuditLog(ctx context.Context, id uint, withRequestAndResp
 		log.MutatedRequestBody = nil
 		log.ResponseBody = nil
 		log.OriginalResponseBody = nil
+		log.ErrorDetail = ""
+		log.RawEvent = nil
```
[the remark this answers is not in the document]
g-linville: This is an existing mismatch that I want to preserve.

pkg/gateway/client/auditlogpersister.go
```diff
@@ -19,6 +20,21 @@ func (c *Client) LogMCPAuditEntry(entry types.MCPAuditLog) {
 	entry.RequestMutated = len(entry.MutatedRequestBody) > 0
 	entry.ResponseMutated = len(entry.OriginalResponseBody) > 0
 
+	if entry.EventID == nil || *entry.EventID == "" {
+		eventID := uuid.NewString()
+		entry.EventID = &eventID
```
thedadams: nit:
```suggestion
		entry.EventID = new(uuid.NewString())
```

pkg/gateway/client/auditlogpersister.go
```diff
@@ -19,6 +20,21 @@ func (c *Client) LogMCPAuditEntry(entry types.MCPAuditLog) {
...
+	entry.SourceType = types.AuditLogSourceTypeMCP
+	entry.EventType = types.EventTypeForCallType(entry.CallType)
+	// The outcome is provisional for request-only entries; the response merge
+	// path overwrites it once the response arrives.
+	entry.Outcome = types.OutcomeForResult(entry.Error, entry.ResponseStatus)
+
+	// ReceivedAt is server receipt time, so it is always assigned here;
+	// callers (including audit log submission requests) cannot supply it.
+	receivedAt := time.Now().UTC()
+	entry.ReceivedAt = &receivedAt
```
thedadams:
```suggestion
	entry.ReceivedAt = new(time.Now().UTC())
```

apiclient/types/mcpauditlog.go
```diff
@@ -4,35 +4,120 @@ import (
...
+	AuditLogEventTypeResourceRead = "resource_read"
+	AuditLogEventTypePromptGet    = "prompt_get"
+	AuditLogEventTypeMCPRequest   = "mcp_request"
+)
+
+// Audit log outcomes.
+const (
+	AuditLogOutcomeSuccess = "success"
+	AuditLogOutcomeError   = "error"
+)
```
thedadams, line 31: Typically, we would give these their own type. I think I understand why you didn't in this case, but I believe another comment handles that issue.

pkg/gateway/client/mcpauditlog.go
```diff
@@ -53,7 +54,10 @@ func (c *Client) insertMCPAuditLogs(ctx context.Context, logs []types.MCPAuditLo
 	return c.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
 		// Insert request-only and complete logs in batches
 		if len(toInsert) > 0 {
-			if err := tx.CreateInBatches(toInsert, 100).Error; err != nil {
+			if err := tx.Clauses(clause.OnConflict{
+				Columns:   []clause.Column{{Name: "event_id"}},
+				DoNothing: true,
```
thedadams, line 61: Based on this, it would be possible to lose some logs if the `event_id` just so happened to conflict. I think it would be worse to lose logs than have conflicting `event_id`s.

pkg/gateway/client/mcpauditlog.go
```diff
@@ -130,6 +140,25 @@ func (c *Client) insertMCPAuditLogs(ctx context.Context, logs []types.MCPAuditLo
 	})
 }
 
+// applyGenericAuditFilters applies the source-generic filters shared by audit
+// log listing and filter-option queries. A startup migration backfills these
+// columns on rows that predate them, so plain matches are sufficient.
```
thedadams:
```suggestion
// log listing and filter-option queries.
```

pkg/gateway/types/mcpauditlog.go
```diff
@@ -3,16 +3,60 @@ package types
...
+	AuditLogSourceTypeLocalAgent = "local_agent"
+
+	AuditLogEventTypeToolCall     = "tool_call"
+	AuditLogEventTypeResourceRead = "resource_read"
+	AuditLogEventTypePromptGet    = "prompt_get"
+	AuditLogEventTypeMCPRequest   = "mcp_request"
+
+	AuditLogOutcomeSuccess = "success"
+	AuditLogOutcomeError   = "error"
+)
```
thedadams: We can just reference the constants from the `apiclient` package.

pkg/gateway/client/mcpauditlog_test.go
```diff
@@ -0,0 +1,396 @@
...
+		{name: "mcp source includes backfilled rows", opts: MCPAuditLogOptions{SourceType: []string{types.AuditLogSourceTypeMCP}}, want: 4},
+		{name: "local agent source", opts: MCPAuditLogOptions{SourceType: []string{types.AuditLogSourceTypeLocalAgent}}, want: 1},
+		{name: "tool_call matches backfilled and new rows", opts: MCPAuditLogOptions{EventType: []string{types.AuditLogEventTypeToolCall}}, want: 3},
+		{name: "resource_read matches backfilled call type", opts: MCPAuditLogOptions{EventType: []string{types.AuditLogEventTypeResourceRead}}, want: 1},
+		{name: "mcp_request matches backfilled other call types", opts: MCPAuditLogOptions{EventType: []string{types.AuditLogEventTypeMCPRequest}}, want: 1},
+		{name: "success outcome", opts: MCPAuditLogOptions{Outcome: []string{types.AuditLogOutcomeSuccess}}, want: 3},
+		{name: "error outcome", opts: MCPAuditLogOptions{Outcome: []string{types.AuditLogOutcomeError}}, want: 2},
+		{name: "device id", opts: MCPAuditLogOptions{DeviceID: []string{"dev-1"}}, want: 1},
+		{name: "combined source and outcome", opts: MCPAuditLogOptions{SourceType: []string{types.AuditLogSourceTypeMCP}, Outcome: []string{types.AuditLogOutcomeError}}, want: 2},
```
thedadams: nit: these are easier for me to read when each field is on its own line.

thedadams (commented):
We also had a conversation "offline" about handling the data structures differently.

ui/user/src/lib/components/admin/audit-logs/AuditLogDetails.svelte
```diff
@@ -38,14 +40,14 @@
 		<div
 			class={twMerge(
 				'absolute top-0 left-0 h-full w-1',
-				auditLog.responseStatus >= 400 ? 'bg-error' : 'bg-primary'
+				responseStatus >= 400 ? 'bg-error' : 'bg-primary'
 			)}
```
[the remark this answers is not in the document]
g-linville: We aren't worrying about local logs yet.

ui/user/src/lib/components/admin/audit-logs/AuditLogDetails.svelte
```diff
@@ -135,36 +137,36 @@
...
 			<div class="flex items-center gap-2">
 				<h4 class="text-lg font-semibold">HTTP Response</h4>
-				{#if auditLog.responseStatus}
+				{#if responseStatus}
 					<p
 						class={twMerge(
 							'w-fit rounded-full px-3 py-1 text-xs font-semibold text-white',
-							auditLog.responseStatus >= 400 ? 'bg-error' : 'bg-primary'
+							responseStatus >= 400 ? 'bg-error' : 'bg-primary'
 						)}
```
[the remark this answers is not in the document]
g-linville: We aren't worrying about local logs yet.

pkg/gateway/client/auditlogpersister.go
g-linville: not relevant after my changes

pkg/gateway/client/mcpauditlog.go
g-linville: I reworked EventID a bit. It is now only for local agent audit logs, and it is required. We will generate it locally. This is basically our idempotency key.

pkg/gateway/client/auditlogpersister.go
g-linville: fixed

apiclient/types/mcpauditlog.go
g-linville: fixed

pkg/gateway/client/mcpauditlog.go
g-linville: fixed

pkg/gateway/client/mcpauditlog.go
g-linville: fixed

pkg/gateway/types/mcpauditlog.go
g-linville: fixed

pkg/gateway/client/mcpauditlog_test.go
g-linville: fixed

apiclient/types/mcpauditlog.go
```diff
@@ -4,37 +4,143 @@ import (
...
+// Audit log outcomes.
+const (
+	AuditLogOutcomeSuccess AuditLogOutcome = "success"
+	AuditLogOutcomeError   AuditLogOutcome = "error"
+)
+
+// MCPAuditLog represents an audit log entry. Despite the name (kept for API
+// compatibility), it can represent generic audit events distinguished by
+// SourceType. Source-specific fields are returned under MCP or Local.
 type MCPAuditLog struct {
```
thedadams: nit: we can rename this type without affecting anything.

pkg/api/handlers/mcpgateway/auditlog.go
```diff
@@ -187,10 +196,16 @@ func (h *AuditLogHandler) SubmitAuditLogs(req api.Context) error {
 	}
 
 	for _, auditLog := range auditLogs {
-		if auditLog.MCPID == "" {
-			auditLog.MCPID = auditLog.Metadata["mcpID"]
+		if auditLog.hasNestedSourceFields() {
+			return types.NewErrBadRequest("nested audit log source fields are not supported by this endpoint")
+		}
```
thedadams, line 201: Please correct me if I am wrong, but I don't think this works. What I mean by that is, I think we would be missing some data from the nanobot shim because it would be in the `MCP` field. `ResponseStatus` being an example of a field that wouldn't get populated.

pkg/gateway/types/mcpauditlog.go
```diff
@@ -3,49 +3,193 @@ package types
...
+	}
+	return *a.MCP
+}
+
+func (a MCPAuditLog) LocalFields() LocalAuditLog {
+	if a.Local == nil {
+		return LocalAuditLog{}
+	}
+	return *a.Local
+}
```
thedadams: Do we need all 4 of these?

pkg/gateway/types/mcpauditlog.go
```diff
@@ -3,49 +3,193 @@ package types
...
+
+func (a *MCPAuditLog) BeforeSave(*gorm.DB) error {
+	a.NormalizeSourceFields()
+	return nil
+}
+
+func (a *MCPAuditLog) AfterFind(*gorm.DB) error {
+	a.NormalizeSourceFields()
+	return nil
+}
```
thedadams: My preference would be that our client handles this. For example, this is handled in the client when encryption is enabled. I would prefer that we be consistent.

Side note: if we migrated and normalize before save, do we need to normalize after find?

g-linville: Sounds good, fixed. And you're right; we only need to normalize before saving the logs. I fixed this so that is now the only time we normalize

pkg/gateway/types/mcpauditlog.go
g-linville: Nope. I got rid of half of them

apiclient/types/mcpauditlog.go
g-linville: Renamed this and some of the other types in this file

pkg/api/handlers/mcpgateway/auditlog.go
g-linville: My point in implementing it this way is that we won't need to change Nanobot at all in order for it to be compatible. Nanobot doesn't submit an `mcp` field right now and it uses the existing flat structure that I am representing in this `auditLogInput` type.

g-linville: It's possible I'm misunderstanding what you're saying here.

g-linville: I tested these changes again just to be sure, and new audit logs are still properly received from Nanobot

apiclient/types/mcpauditlog.go
```diff
@@ -4,42 +4,148 @@ import (
...
 	ResponseHeaders           json.RawMessage `json:"responseHeaders,omitempty"`
 }
 
-type MCPAuditLogResponse struct {
-	MCPAuditLogList `json:",inline"`
-	Total           int64 `json:"total"`
-	Limit           int   `json:"limit"`
-	Offset          int   `json:"offset"`
+// LocalAuditLog contains fields meaningful only for local-agent audit events.
+type LocalAuditLog struct {
```
njhale: nit: suggestions `DeviceAuditLog`, `LocalAgentAuditLog`, `AgentHookAuditLog`

g-linville: I'm open to renaming it to whatever we want to call it

apiclient/types/mcpauditlog.go
```diff
@@ -4,42 +4,148 @@ import (
...
+}
+
+// ToolInfo identifies what was invoked and what kind of invocation it was.
+type ToolInfo struct {
+	Name string `json:"name"`
+	Type string `json:"type"`
+}
+
+// AuditLogContext holds source-specific, non-indexed audit event metadata.
+type AuditLogContext struct {
```
njhale: nit: The naming here is off. This seems specific to hook events, but it's name `AuditLogContext` leads readers to believe it's generic.

thedadams (approved):

calvinmclean (approved):

PR #6940, opened 2026-06-15
fix: validate audience on JWT tokens

thedadams:
If the audience is an MCP server, then this change ensures that only the MCP and authenticated groups are used. If the audience is the Obot instance itself, then the UserGroups passed to the context are used.

This is really only a fail-safe. The authorization logic would do the right thing if the correct groups were provided when creating the token.

### DIFF
--- a/pkg/api/authz/authz.go
+++ b/pkg/api/authz/authz.go
@@ -24,8 +24,6 @@ const (
 
 var (
 	adminAndOwnerRules = []string{
-		"/api/tool-references",
-		"/api/tool-references/",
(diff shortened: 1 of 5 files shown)


pkg/api/authz/authz.go
```diff
@@ -24,8 +24,6 @@ const (
 
 var (
 	adminAndOwnerRules = []string{
-		"/api/tool-references",
```
njhale, line 27: 💯

njhale (approved):

Issue #6943, opened 2026-06-15
Using skills does not trigger oauth flow after logout in vs code.


sangee2004:
**Describe the bug**
After running obot logout, using skills from VS Code fails because authentication is required, but the obot OAuth flow is not triggered. The user is not prompted to sign in, so the skill execution fails instead of authenticating and continuing.

Pre Req - Have `obot setup` command done with latest build from `main` - ` Version:  v0.0.0-dev-45a58224-dirty`
**To Reproduce**
1. From the terminal, do `obot logout`
2. From VS Code, try to use any of the obot skills like  `obot scan --submit` which require obot authentication.

Notice that the skills fail to execute with a need to authenticate with obot.

<img width="826" height="661" alt="Image" src="https://github.com/user-attachments/assets/5d95fe62-d4f8-4e5c-b2bd-10f40e4dba42" />

Running  `obot skills search` results in `request cancelled` errors.

<img width="852" height="493" alt="Image" src="https://github.com/user-attachments/assets/9b0cd8bc-b241-42ab-8b6b-49fa7795ba06" />

**Expected behavior**
Ideally, the user should be presented with the obot OAuth workflow, be authenticated, and the skill should be executed.

sangee2004:
This issue is addressed when testing with the latest build from `main`

1. From the terminal, do obot logout
2. From VS Code, try to use any of the obot skills like `obot scan --submit` which require obot authentication.
This results in the user being prompted to follow the OAuth flow to authenticate with the obot, and the scan submission succeeds as expected.

Also able to search and install skills when logged in. This prompts the user to follow the OAuth flow to authenticate with the obot, and the skill installation succeeds.

PR #6963, opened 2026-06-16
fix: stop just-in-time CLI auth from hanging on stdin in agents

njhale:
When a user is unauthenticated, obot CLI commands authenticate lazily — the first authenticated request invokes internal.Token, which waits on fmt.Scanln() for the user to press ENTER before opening the browser. When an agent runs such a command after `obot logout`, whether that read blocks depends on how the process is launched:
- Claude Code starts obot with stdin closed, so fmt.Scanln() returns immediately on EOF and the flow opens the browser — auth works.
- VS Code's default Chat agent leaves stdin open, so fmt.Scanln() blocks forever, the browser never opens, and the command is eventually cancelled.

To fix this, make non-interactive execution explicit and use it from the agent skills:
- PromptConfig adds a shared --non-interactive flag (and OBOT_NON_INTERACTIVE env), embedded into the obot CLI commands that can lazily authenticate — login, mcp search, scan, skills search, skills install, and setup. Its Pre hook marks the command context non-interactive, and internal.Token then skips the fmt.Scanln() gate so it never blocks on stdin, failing fast if the browser cannot be opened.
- The agent skill templates pass --non-interactive on the obot commands they document, and the `obot` skill explains just-in-time auth and the flag, so agents driven by these skills no longer hang.
- setup adopts the shared PromptConfig in place of its own NonInteractive field; render tests updated to match the templates.

Addresses https://github.com/obot-platform/obot/issues/6943

### DIFF
--- a/pkg/cli/internal/token.go
+++ b/pkg/cli/internal/token.go
@@ -169,10 +169,11 @@ func Token(ctx context.Context, baseURL string, opts apiclient.TokenFetchOptions
 	}
 
 	token, tokenErr := credentialStore.Get(appURL)
-	hasStoredToken := tokenErr == nil
 	if tokenErr != nil && !credentials.IsNotFound(tokenErr) {
 		return "", tokenErr
 	}
+
+	hasStoredToken := tokenErr == nil
 	if hasStoredToken && !opts.ForceRefresh && testToken(ctx, baseURL, token, opts.Scopes...) {
 		return token, nil
 	}
@@ -198,24 +199,35 @@ func Token(ctx context.Context, baseURL string, opts apiclient.TokenFetchOptions
 		return "", fmt.Errorf("failed to create login request: %w", err)
 	}
 
-	if !hasStoredToken && !isNonInteractive(ctx) {
-		w := outputWriter(ctx)
+	w := outputWriter(ctx)
+	nonInteractive := isNonInteractive(ctx)
+	if !hasStoredToken {
 		fmt.Fprintln(w)
 		fmt.Fprintln(w, color.GreenString("Authentication is needed"))
 		fmt.Fprintln(w, color.GreenString("========================"))
 		fmt.Fprintln(w)
-		fmt.Fprintln(w, color.CyanString(provider.Name)+" is used for authentication using the browser. This can be bypassed by setting")
-		fmt.Fprintln(w, "the env var "+color.CyanString(TokenEnvVar)+" to your API key.")
+		fmt.Fprintln(w, color.CyanString(provider.Name), "is used for authentication using the browser.")
+		fmt.Fprintln(w, "This can be bypassed by setting the env var", color.CyanString(TokenEnvVar), "to your API key.")
 		fmt.Fprintln(w)
-		fmt.Fprintln(w, color.GreenString("Press ENTER to continue (CTRL+C to exit)"))
-		if err := enter(ctx); err != nil {
-			return "", err
+
+		if !nonInteractive {
+			fmt.Fprintln(w, color.GreenString("Press ENTER to continue (CTRL+C to exit)"))
+			if err := enter(ctx); err != nil {
+				return "", err
+			}
+			fmt.Fprintln(w)
 		}
-		fmt.Fprintln(w)
 	}
 
-	fmt.Fprintf(outputWriter(ctx), "Opening browser to %s. if there is an issue paste this link into a browser manually\n", loginURL)
-	_ = openBrowser(loginURL)
+	fmt.Fprintln(w, "Opening browser to", loginURL, "for authentication.")
+	if err := openBrowser(loginURL); err != nil {
+		if nonInteractive {
+			return "", fmt.Errorf("failed to open browser: %w", err)
+		}
+
+		fmt.Fprintln(w, "Failed to open browser:", err.Error())
+		fmt.Fprintln(w, "To finish authenticating, paste", loginURL, "into your browser manually.")
+	}
 
 	ctx, timeoutCancel := context.WithTimeout(ctx, 5*time.Minute)
 	defer timeoutCancel()
--- a/pkg/cli/login.go
+++ b/pkg/cli/login.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"context"
 	"fmt"
 	"os"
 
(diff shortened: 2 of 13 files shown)


pkg/cli/setup.go
```diff
@@ -67,16 +67,10 @@ func (s *Setup) run(cmd *cobra.Command, progress setupProgressWriter) error {
...
-		cmd.SetContext(ctx)
-	}
 	if progress.json {
-		authOutput := cmd.ErrOrStderr()
 		if s.NonInteractive {
-			authOutput = io.Discard
+			ctx = cliinternal.WithOutputWriter(ctx, io.Discard)
 		}
-		ctx = cliinternal.WithOutputWriter(ctx, authOutput)
 		cmd.SetContext(ctx)
```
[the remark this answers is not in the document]
njhale: I felt it was better to centralize the behavior.

pkg/cli/internal/token.go
```diff
@@ -198,24 +199,35 @@ func Token(ctx context.Context, baseURL string, opts apiclient.TokenFetchOptions
...
-		fmt.Fprintln(w)
 	}
 
-	fmt.Fprintf(outputWriter(ctx), "Opening browser to %s. if there is an issue paste this link into a browser manually\n", loginURL)
-	_ = openBrowser(loginURL)
+	fmt.Fprintln(w, "Opening browser to", loginURL, "for authentication.")
+	if err := openBrowser(loginURL); err != nil {
+		if nonInteractive {
+			return "", fmt.Errorf("failed to open browser: %w", err)
+		}
```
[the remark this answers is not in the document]
njhale: User's can just `obot login` manually too. I felt like this was a reasonable change for aligned with the behavior of `--non-interactive` in `obot setup`.

pkg/cli/internal/token.go
```diff
@@ -198,24 +199,35 @@ func Token(ctx context.Context, baseURL string, opts apiclient.TokenFetchOptions
 		return "", fmt.Errorf("failed to create login request: %w", err)
 	}
 
-	if !hasStoredToken && !isNonInteractive(ctx) {
-		w := outputWriter(ctx)
+	w := outputWriter(ctx)
+	nonInteractive := isNonInteractive(ctx)
+	if !hasStoredToken {
```
njhale, line 204: I felt that it's reasonable to print the info text below  to give users/agents some more context in `--non-interactive` mode as well.

The only command we don't want to see this on stderr for is probably `obot setup --json --non-interactive --yes`, and in that case we're already setting the writer to io.Discard by this point anyway

pkg/cli/internal/token.go
```diff
@@ -198,24 +199,35 @@ func Token(ctx context.Context, baseURL string, opts apiclient.TokenFetchOptions
...
 		fmt.Fprintln(w)
 		fmt.Fprintln(w, color.GreenString("Authentication is needed"))
 		fmt.Fprintln(w, color.GreenString("========================"))
 		fmt.Fprintln(w)
-		fmt.Fprintln(w, color.CyanString(provider.Name)+" is used for authentication using the browser. This can be bypassed by setting")
-		fmt.Fprintln(w, "the env var "+color.CyanString(TokenEnvVar)+" to your API key.")
+		fmt.Fprintln(w, color.CyanString(provider.Name), "is used for authentication using the browser.")
+		fmt.Fprintln(w, "This can be bypassed by setting the env var", color.CyanString("OBOT_TOKEN"), "to your API key.")
```
njhale: Just a little cleanup here.

pkg/cli/login.go
```diff
@@ -49,9 +51,28 @@ func (l *Login) Run(cmd *cobra.Command, _ []string) error {
...
 	return nil
 }
+
+// PromptConfig contains shared local options for commands that may require interactive input from users.
+// e.g. Any command that performs just-in-time authentication for unauthenticated users.
+type PromptConfig struct {
+	NonInteractive bool `usage:"Never read from stdin; fail if required input is missing" env:"OBOT_NON_INTERACTIVE" local:"true"`
+}
+
+func (p PromptConfig) Pre(cmd *cobra.Command, _ []string) error {
```
njhale, line 67: This DRYs out setting non-interactive on the context for specific subcommands by embedding PromptConfig.

pkg/cli/login.go
```diff
@@ -49,9 +51,28 @@ func (l *Login) Run(cmd *cobra.Command, _ []string) error {
...
 	if l.PrintToken {
 		fmt.Println(token)
 	}
 	return nil
 }
+
+// PromptConfig contains shared local options for commands that may require interactive input from users.
+// e.g. Any command that performs just-in-time authentication for unauthenticated users.
+type PromptConfig struct {
+	NonInteractive bool `usage:"Never read from stdin; fail if required input is missing" env:"OBOT_NON_INTERACTIVE" local:"true"`
```
njhale, line 64: - `local:"true"` makes it so that the NonInteractive option isn't inherited by child commands it doesn't apply to; e.g. `obot setup` (uses it) and `obot setup detect-clients` does not.
- I know we don't like declaring `OBOT_*` envs explicitly like this, but I felt it's a better UX here; e.g. able to set `OBOT_NON_INTERACTIVE` in a script and have it apply to every obot command that consumes it

pkg/cli/internal/token.go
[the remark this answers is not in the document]
njhale: fixed.

pkg/localagents/assets/files/skills/obot/SKILL.md.tmpl
```diff
@@ -9,20 +9,25 @@ Use the local `obot` CLI for Obot skill discovery, skill installation, and MCP s
 
 These instructions are rendered for `{{ .AgentID }}`.
 
-Authentication is managed by `obot login`. Do not ask the user for Obot API tokens, and do not call Obot APIs directly.
+Authentication can be managed in advance by `obot login`.
+
+`obot` CLI commands may also perform just-in-time authentication when invoked if a user isn't already authenticated. These commands have a --non-interactive option that must be used to prevent the command from blocking on user input over stdin.
```
g-linville: pico nit


```suggestion
`obot` CLI commands may also perform just-in-time authentication when invoked if a user isn't already authenticated. These commands have a `--non-interactive` option that must be used to prevent the command from blocking on user input over stdin.
```

g-linville (approved):

pkg/localagents/assets/files/skills/obot/SKILL.md.tmpl
njhale: Done

pkg/cli/login.go
thedadams: This is an instance where it makes perfect sense to use `env` struct tag.

thedadams (approved):

Issue #6899, opened 2026-06-10
Upgrade - Auth providers are unavailable for ~1 minute after server startup.


sangee2004:
**Describe the bug**
After upgrading an Obot server from v0.22.1 to v0.23.0-rc3, users who attempt to log in immediately after the server starts see a message indicating that no authentication providers are available. This state persists for approximately one minute before the configured Microsoft Entra authentication provider becomes available, and login succeeds. Authentication providers should be available as soon as the server has started successfully, without a prolonged delay.

**To Reproduce**
1. Install `v0.22.1` obot server using helm install with Microsoft Entra Auth provider and Amazon Bedrock Model provider.
2. Log in as the owner user and create a multi-user MCP server - `santestmultiuser` in my case.
3. Log in as a basic user and connect to this multi-user MCP server. Chat with it and make successful tool calls.
4. Upgrade to the `v0.23.0-rc3` version of the obot server.
5. Log in as admin user (or any user)

Notice that the user will see this message for about 1 minute.
<img width="1317" height="746" alt="Image" src="https://github.com/user-attachments/assets/66c3e18e-9a18-47ff-bbd2-1ed844a434e0" />

After this, the user is provided with the` Microsoft Entra Auth provider` login and is able to log in successfully.

**Expected behavior**
There should not be a long delay in making the auth providers available after the obot server has started successfully.

thedadams:
This issue will not be completely fixed for the `v0.23.0` release.

There is a bug in the releases `v0.22.1` and below where the controller in Obot doesn't have the opportunity to shutdown gracefully, which means it can't terminate the leader election lease. When the new Obot container starts up, it has to wait before it can acquire the lease. That is what causes the delay in the auth providers being available.

The linked PR fixes the graceful shutdown in the controller so the lease is successfully terminated prior to the old Obot container exiting. This fixes the issue in `v0.23.0` going forward, which is good enough for our purposes for now.

sangee2004:
When testing the same upgrade flow from `v0.22.1` to the latest `main` build - [v0.0.0-dev+5ce8cff0](https://github.com/obot-platform/obot/commit/5ce8cff0), after upgrading, the user is now presented with only the bootstrap login option for some time (earlier, we were presented with no auth provider configured). When logged in as a bootstrap user, the user is presented with a welcome screen indicating that no auth/model providers are configured.

[...]

PR #6973, opened 2026-06-17
fix: ensure auth and model providers are available on upgrade

thedadams:
Since we have changed the way auth and model providers are stored and there was a bug in the controller shutdown that caused a delay in startup, there would be a significant delay before an upgraded system was ready for use. This change detects when there are no auth nor model providers and ensures that they are ready before the server starts up.

Issue: https://github.com/obot-platform/obot/issues/6899

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -99,6 +99,10 @@ func (c *Controller) PreStart(ctx context.Context) error {
 		return fmt.Errorf("failed to reconcile service account keys: %w", err)
 	}
 
+	if err := c.ensureAuthProvidersAndModelProviders(ctx); err != nil {
+		return fmt.Errorf("failed to ensure auth providers and model providers: %w", err)
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


njhale (approved):

njhale (approved):

pkg/controller/controller.go
```diff
@@ -99,6 +99,51 @@ func (c *Controller) PreStart(ctx context.Context) error {
...
+		if err := c.services.StorageClient.List(ctx, &authProviders); err != nil {
+			return fmt.Errorf("failed to list auth providers: %w", err)
+		}
+
+		for _, authProvider := range authProviders.Items {
+			if err := provider.SetAuthProviderConfiguredStatus(ctx, c.services.GatewayClient, c.services.LicenseProvider, &authProvider); err != nil {
+				return fmt.Errorf("failed to set auth provider configured status: %w", err)
+			}
+
+			if err := c.services.StorageClient.Update(ctx, &authProvider); err != nil {
```
g-linville: Should this be `c.services.StorageClient.Status().Update` like what we call for model providers below? (Asking because I don't know.)

thedadams: Yes, fixed. I also move the logic into a function.

g-linville (approved):

Issue #6971, opened 2026-06-16
MCP servers fail to start in VS Code after obot server becomes available


sangee2004:
**Reproduces** - Frequently. 2 of the 3 MCP servers exhibited this problem when the same scenario described in the issue was repeated 3 times.

**Describe the bug**
When MCP servers are connected in VS Code, and the Obot server becomes unreachable, the servers eventually stop after losing the Obot MCP Gateway authentication session. After the obot server is made reachable again, restarting the MCP servers from VS code often fails on the first attempt with authentication errors, such as `State does not match` error. Starting the MCP server should succeed on the first attempt once the obot server is available again.

**To Reproduce**
1. Connect to a few MCP servers ( 2 remote MCP servers(Gmail and Outlook) and 1 multi-user MCP server and get their connection URL.
2. Configure these MCP servers in VS Code using the connection URL from step 1.
3. Start these MCP servers and make sure they get connected successfully and report as being in `Running` state.
4. Stop the obot server. In my case, I stopped the port forwarding rule to the obot server - `kubectl port-forward svc/obot-obot 8080:80`

Wait for some time ( up to 10 minutes).

Notice that all MCP servers will eventually prompt to authenticate with the obot MCP server, and authenticating will fail.
<img width="447" height="362" alt="Image" src="https://github.com/user-attachments/assets/9d1e52f7-7edc-4dd0-bec6-a29ab4776ac0" />

<img width="1101" height="707" alt="Image" src="https://github.com/user-attachments/assets/1a18718b-afbc-4933-ac93-1185ecb002c3" />

Logs from the MCP server:

```
2026-06-16 15:58:34.250 [info] Starting server MyOutlook
2026-06-16 15:58:34.251 [info] Connection state: Starting
2026-06-16 15:58:34.251 [info] Starting server from LocalProcess extension host
2026-06-16 15:58:34.251 [info] Connection state: Running
2026-06-16 15:58:34.284 [info] Discovered resource metadata at http://localhost:8080/.well-known/oauth-protected-resource/mcp-connect/default-outlook-841b850d
2026-06-16 15:58:34.285 [info] Using auth server metadata url: http://localhost:8080
2026-06-16 15:58:34.291 [info] Discovered authorization server metadata at http://localhost:8080/.well-known/oauth-authorization-server
2026-06-16 15:58:39.253 [info] Waiting for server to respond to `initialize` request...
2026-06-16 15:58:40.577 [info] Discovered 17 tools
[...]

PR #6977, opened 2026-06-17
refactor: handle OAuth consent differently

thedadams:
For some reason, handling the OAuth consent the way we were was causing issues with Chrome (no issues in other browsers we tested). This change will make a GET request to the backend instead of a POST. This GET request only exists to redirect the user either to the client or to the third-party URL for authentication, if necessary. This fixes the issues in Chrome and doesn't cause problems for the other browsers.

Issue: https://github.com/obot-platform/obot/issues/6971


ui/user/src/routes/auth/oauth/consent/[oauth_auth_request]/+page.svelte
```diff
@@ -181,24 +159,14 @@
...
-			<button class="btn btn-text" disabled={submitting} onclick={() => submit('deny')}
-				>Cancel</button
-			>
-			<button class="btn btn-primary" disabled={submitting} onclick={() => submit('approve')}>
-				{#if submitting}
-					<Loading class="size-4" />
-				{:else}
-					Continue
-				{/if}
+			<button class="btn btn-text" onclick={() => navigateTo(consent.cancelURL)}>Cancel</button>
```
Comment (ivyjeong13) [inline ui/user/src/routes/auth/oauth/consent/[oauth_auth_request]/+page.svelte]: Can change both of these from `<button>` to `<a>` with `href` if these aren't doing anything but navigation

ivyjeong13 (approved):
One small thing otherwise UI LGTM!

ui/user/src/routes/auth/oauth/consent/[oauth_auth_request]/+page.svelte
Comment (thedadams) [reply ui/user/src/routes/auth/oauth/consent/[oauth_auth_request]/+page.svelte]: Done.

njhale (approved):

PR #6983, opened 2026-06-17
enhance: allow OAuth clients that don't supply resource to connect

thedadams:
In order to know which MCP server the MCP client is trying to connect to, they should send a resource parameter with the authorize request. Not all clients do this. This change introduces a way to alter the metadata endpoints so that we can still determine which MCP server the client is connecting to when they don't send a resource parameter.

### DIFF
--- a/pkg/api/handlers/wellknown/oauth.go
+++ b/pkg/api/handlers/wellknown/oauth.go
@@ -3,14 +3,33 @@ package wellknown
 import (
 	"fmt"
 	"net/http"
+	"net/url"
 
 	"github.com/obot-platform/obot/apiclient/types"
 	"github.com/obot-platform/obot/pkg/api"
 )
 
-// oauthAuthorization handles the /.well-known/oauth-authorization-server endpoint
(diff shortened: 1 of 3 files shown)


pkg/api/handlers/wellknown/oauth.go
```diff
@@ -3,14 +3,33 @@ package wellknown
 import (
 	"fmt"
 	"net/http"
+	"net/url"
 
 	"github.com/obot-platform/obot/apiclient/types"
 	"github.com/obot-platform/obot/pkg/api"
 )
 
 // oauthAuthorization handles the /.well-known/oauth-authorization-server endpoint
```
g-linville: this comment should be updated

njhale (approved):

pkg/api/handlers/wellknown/oauth.go
thedadams: Done.

g-linville (approved):

Issue #6962, opened 2026-06-16
Change main image versioning to avoid false-positive CVEs from v0.0.0-<git hash>


cjellick:
## Problem

Our `obot` images built from `main` are versioned as `v0.0.0-<git hash>`. Because the version starts with `v0.0.0`, scanners (e.g. `docker scout`, GitHub code scanning) treat it as older than the tagged release that actually contains a given fix. This causes vulnerabilities that are already fixed to show up as unresolved on `main` images.

A concrete example is [GHSA-vw82-7fv8-r6gp](https://scout.docker.com/v/GHSA-vw82-7fv8-r6gp), which still shows up on `main` even though the fix is present. It was [dismissed in GitHub code scanning](https://github.com/obot-platform/obot/security/code-scanning/2320) precisely because the `v0.0.0-<git hash>` versioning makes `main` appear "less than" the release with the fix.

This is an annoyance today, and it has the potential to bite us during SOC 2.

## Proposed fix

Change how `main` images are versioned so they no longer appear to predate fixed releases. Options discussed:

1. **`main-<git hash>`** — try this first.
2. **`v999.0.0-<git hash>`** — fall back to this if scanners still misinterpret a non-semver version.

## Acceptance criteria

- `main` images are versioned such that scanners no longer flag already-fixed CVEs as present.
- Verify that [GHSA-vw82-7fv8-r6gp](https://scout.docker.com/v/GHSA-vw82-7fv8-r6gp) no longer appears on `main` after the change.

_Filed from a Slack discussion between Nick Hale, Craig, and Donnie Adams._

sangee2004:
Older main build reports the following CVE when executing - `docker scout cves ghcr.io/obot-platform/obot:main` (`sha256:d431ff03b228fb5484c60d727d16ef87de028e80c6993a3e282be7eb773e5873`)
```
   1C     0H     0M     0L  github.com/obot-platform/obot 0.0.0-20260617183043-c0344bef882f
pkg:golang/github.com/obot-platform/obot@0.0.0-20260617183043-c0344bef882f

    ✗ CRITICAL GHSA-vw82-7fv8-r6gp [Improper Access Control]
      https://scout.docker.com/v/GHSA-vw82-7fv8-r6gp
      Affected range : <=0.21.0                                     
      Fixed version  : 0.21.1                                       
      CVSS Score     : 9.6                                          
[...]

PR #6988, opened 2026-06-18
chore: fetch full git history in image build workflow

njhale:
Fetch tags and full commit history. This provides the VCS info required for
`go build` to produce a binary that contains a main module pseudo-version with
a core segment relative to the most recent release tag (e.g.
vX.Y.(Z+1)-0.<ts>-<rev>). Without the full history, branch builds generate
versions like v0.0.0-<ts>-<rev>, which can cause false positives in
vulnerability scans.

Addresses https://github.com/obot-platform/obot/issues/6962

### DIFF
--- a/.github/workflows/docker-build-and-push.yml
+++ b/.github/workflows/docker-build-and-push.yml
@@ -26,6 +26,14 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          # Fetch tags and full commit history for branch builds.
+          # This provides the VCS info required for `go build` to produce a binary that contains a
+          # main module pseudo-version with a core segment relative to the most recent release tag
+          # (e.g. vX.Y.(Z+1)-0.<ts>-<rev>).
+          # Without the full history, branch builds generate versions like v0.0.0-<ts>-<rev>,
+          # which can cause false positives in vulnerability scans.
+          fetch-depth: ${{ case(github.ref_type == 'tag', 1, 0) }}
 
       - name: Setup Depot
         uses: depot/setup-action@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-build-and-push.yml
```diff
@@ -26,6 +26,14 @@ jobs:
...
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          # Fetch tags and full commit history for branch builds.
+          # This provides the VCS info required for `go build` to produce a binary that contains a
+          # main module pseudo-version with a core segment relative to the most recent release tag
+          # (e.g. vX.Y.(Z+1)-0.<ts>-<rev>).
+          # Without the full history, branch builds generate versions like v0.0.0-<ts>-<rev>,
+          # which can cause false positives in vulnerability scans.
+          fetch-depth: ${{ github.ref_type != 'tag' && '0' || '1' }}
```
njhale: When `main` is exactly at a release tag, this interacts poorly with:

```
GIT_TAG := $(shell git describe --tags --exact-match 2>/dev/null | xargs -I {} echo -X 'github.com/obot-platform/obot/pkg/version.Tag={}')
GO_LD_FLAGS := "-s -w $(GIT_TAG)"
build:
	go build -ldflags=$(GO_LD_FLAGS) -o bin/obot .
```

causing
1. The in-app version to be reported as the tag version instead of `v0.0.0-dev-<rev>`
2. Upgrade checking to be enabled

not sure how big a deal these two, but it feels odd to me.

Note: the same issue exists if you build main locally with `make build` when main matches a tag you've pulled.

thedadams (approved):

PR #6991, opened 2026-06-18
enhance: add a new authentication complete screen

thedadams:
This screen finishes the MCP OAuth flow so that we can control what the user sees when the OAuth process completes. This fixes issues when we redirect back to apps on a user's computer.

This change also removes the old login complete screen in favor of a consistent experience. That is, if a user uses `obot login`, then they will see this new "authentication complete" screen with the "you can now close this window" text.

Issue: https://github.com/obot-platform/field-issues/issues/39

Demo: https://caps.thedadams.com/s/v3dq16cn2v0zsq8


ivyjeong13 (approved):
UI changes LGTM

g-linville (approved):
backend lgtm

PR #7005, opened 2026-06-19
fix: forward passthrough headers to composite MCP children

### LINKED ISSUE #6845: Composite MCP server does not propagate per-request OAuth bearer header to child MCP servers
**Describe the bug**

When using Obot to expose a composite MCP server made of multiple remote MCP servers, per-request authentication headers sent by an external MCP client do not appear to reach the child MCP server selected by the composite tool call.

This was observed while validating an external MCP protected-resource flow based on OAuth 2.1 / Dynamic Client Registration, where:

- an external MCP client authenticates against an authorization server;
- the client calls a protected MCP endpoint with a bearer access token;
- a gateway validates that token and forwards the request to an Obot composite MCP server;
- the composite MCP server calls a child/domain MCP server;
- the child MCP server expects to receive the validated bearer context as an HTTP header.

Direct calls to an individual remote MCP server through Obot work when the same per-request header is provided. The issue appears when the same flow goes through an Obot composite MCP server: the downstream child MCP server reports that the access token/header is missing.

**To Reproduce**

A reduced version of the setup is:

1. Run Obot with Docker-based MCP hosting enabled.
2. Register multiple remote MCP servers, for example `mcp-wallet` and `mcp-card`.
3. Register a composite MCP server that exposes tools from those remote servers.
4. Send a Streamable HTTP MCP `tools/call` request to the composite MCP endpoint with a per-request bearer context header, for example:
   - `Authorization: Bearer <token>`, or
   - a custom header such as `X-Custom-MCP-Authorization: Bearer <token>`.
5. Have the selected child MCP server validate that it received the expected bearer context header.

Observed behavior:

- Calling the child/domain MCP server directly through its Obot MCP endpoint succeeds when the header is sent.
- Calling the same tool through the composite MCP endpoint fails because the child/domain MCP server does not receive the header.

The downstream error in our test server was equivalent to:

```text
Access token missing. Send Authorization or X-Custom-MCP-Authorization.
```

**Expected behavior**

For composite MCP servers, per-request authentication context accepted at the composite MCP endpoint should be available to the downstream child MCP server invoked by the composite, or there should be a documented configuration path that makes this behavior possible.

This is especially important for external MCP protected-resource scenarios where the MCP client performs OAuth/DCR and the downstream domain MCP server needs to validate or audit the original access token context.

**Screenshots**

Not applicable.

**Additional context**

We looked at the relevant Obot/Nanobot code paths while debugging:

- Obot appears to populate `PassthroughHeaderNames` / `PassthroughHeaderValues` from stored MCP server instance configuration, such as `multiUserConfig.userDefinedHeaders` values.
- Nanobot has `passthroughHeaders` support that copies allowlisted headers from the current HTTP request context when creating outbound requests to remote MCP servers.
- In our composite scenario, the effective downstream child MCP call did not receive the per-request bearer header, even though the same header worked for direct calls to a single remote MCP server.

This may be expected behavior or a missing configuration on our side, but it was not clear from the public documentation how to propagate per-request OAuth bearer context through an Obot composite MCP server to the selected child MCP server.

jsguerrero:
## What

`constructMCPServerNanobotYAMLForComposite` now threads the composite server's
configured `PassthroughHeaderNames` to each child, and always includes
`Authorization` (deduped, case-insensitive). The two call sites (`docker.go`,
`kubernetes.go`) pass `server.PassthroughHeaderNames`.

```go
func constructMCPServerNanobotYAMLForComposite(servers []ComponentServer, passthroughHeaderNames []string) ([]byte, error)
```

## Why

Composite MCP servers did not propagate per-request auth to their child servers:
the composite branch only set `BaseURL`/`ToolOverrides`/`ToolPrefix` per child,
while the non-composite path already forwards `PassthroughHeaderNames`. As a
result the child never received the inbound bearer and downstream calls failed
with "access token missing".

Composite children are internal Obot `/mcp-connect` endpoints that resolve the
caller's per-user credentials, so forwarding `Authorization` only ever crosses
an Obot→Obot hop and lets each child authenticate as the calling user. This
makes the composite path consistent with the non-composite one.

Fixes #6845

## Tests

`TestCompositeNanobotYAMLForwardsAuthHeaders` (table-driven):
- default adds `Authorization`
- operator-configured names are forwarded plus `Authorization`
- `Authorization` is not duplicated when already configured

`make validate` (gofmt, `go vet`, `go test ./...`) passes.

### DIFF
--- a/pkg/mcp/backend.go
+++ b/pkg/mcp/backend.go
@@ -222,10 +222,26 @@ func urlWithPath(urlStr, path string) string {
 	return u.String()
 }
 
-func constructMCPServerNanobotYAMLForComposite(servers []ComponentServer) ([]byte, error) {
+func constructMCPServerNanobotYAMLForComposite(servers []ComponentServer, passthroughHeaderNames []string) ([]byte, error) {
 	mcpServers := make(map[string]nanobotConfigMCPServer, len(servers))
 	names := make([]string, 0, len(servers))
 	replacer := strings.NewReplacer("/", "-", ":", "-", "?", "-")
+
+	// #6845: composite children are always internal Obot /mcp-connect endpoints
+	// that resolve the caller's per-user credentials, so the caller's auth must
+	// reach them. Forward the operator-configured passthrough headers plus
+	// Authorization (required for the child endpoint to identify the user).
+	childPassthrough := append([]string(nil), passthroughHeaderNames...)
+	hasAuth := false
+	for _, h := range childPassthrough {
+		if strings.EqualFold(h, "Authorization") {
+			hasAuth = true
+			break
+		}
+	}
+	if !hasAuth {
+		childPassthrough = append(childPassthrough, "Authorization")
+	}
 
 	for _, component := range servers {
 		tools := make(map[string]toolOverride, len(component.Tools))
@@ -241,9 +257,10 @@ func constructMCPServerNanobotYAMLForComposite(servers []ComponentServer) ([]byt
 
 		name := replacer.Replace(component.Name)
 		mcpServers[name] = nanobotConfigMCPServer{
-			BaseURL:       component.URL,
-			ToolOverrides: tools,
-			ToolPrefix:    component.ToolPrefix,
+			BaseURL:            component.URL,
+			ToolOverrides:      tools,
+			ToolPrefix:         component.ToolPrefix,
+			PassthroughHeaders: childPassthrough,
 		}
 
 		names = append(names, name)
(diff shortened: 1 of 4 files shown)


jsguerrero:
Cross-repo context: this passthrough change is part of enabling per-user OAuth brokering to lazy-auth SaaS MCPs through composite servers. It pairs with obot-platform/nanobot#332 (lazy-auth: OAuth required without a 401 challenge). The obot-side wiring that sets `Server.OAuthRequired` for static-OAuth composite components will follow once #332 is released.

PR #7006, opened 2026-06-19
fix: emit WWW-Authenticate on /mcp-connect 401 (RFC 9728)

jsguerrero:
## What

Return a `WWW-Authenticate` header on `401` responses for `/mcp-connect/{id}`
endpoints, pointing at the path-aware protected-resource metadata for that
specific server:

```
WWW-Authenticate: Bearer resource_metadata="<base>/.well-known/oauth-protected-resource/mcp-connect/{id}"
```

Two places:
- `server.go` (auth middleware): set it when an unauthenticated request to
  `/mcp-connect/{id}` is rejected — mirrors the existing `/v0.1` registry branch.
- `mcpgateway/handler.go` (proxy `ModifyResponse`): add it to a proxied `401`
  from the upstream MCP server when the upstream didn't already send one.

## Why

RFC 9728 §5.1 says a protected resource SHOULD return `WWW-Authenticate` with a
`resource_metadata` pointer on `401`. Obot's `/mcp-connect/{id}` endpoints did
not, so a client that isn't already path-aware falls back to requesting the
host-root metadata (`/.well-known/oauth-protected-resource`). That document
can't identify a specific server, so discovery resolves the wrong resource and
the OAuth flow fails (observed as `invalid_request: catalog entry /mcp-connect
not found`, because the client sends a resource with no server id).

With the pointer present, any client — path-aware or not — follows it to the
correct path-scoped metadata. This affects every `/mcp-connect` server, not just
composites.

## Tests

`make validate` (gofmt, `go vet`, `go test ./...`) passes. Verified manually:
the 401 now carries the path-aware `resource_metadata`, and a client that
previously failed discovery now completes the OAuth flow.

### DIFF
--- a/pkg/api/handlers/mcpgateway/handler.go
+++ b/pkg/api/handlers/mcpgateway/handler.go
@@ -45,6 +45,14 @@ func (h *Handler) Proxy(req api.Context) error {
 		return nil
 	}
 
+	// RFC 9728 §5.1: if the upstream MCP server replies 401 without pointing the
+	// client at its protected-resource metadata, advertise the path-aware
+	// metadata URL for this connect endpoint. Without this, clients that don't
+	// already know the path fall back to the host-root metadata document, which
+	// cannot identify this specific MCP server.
+	mcpID := req.PathValue("mcp_id")
+	connectBaseURL := strings.TrimSuffix(req.APIBaseURL, "/api")
+
 	(&httputil.ReverseProxy{
 		Transport: h.transport,
 		Director: func(r *http.Request) {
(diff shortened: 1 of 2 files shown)


jsguerrero:
Cross-repo: this is the resource-server side of the same RFC 9728 §3.1 path-aware fix as obot-platform/nanobot#327 (client side). Together they make protected-resource discovery work for MCP servers reached under a sub-path (e.g. `/mcp-connect/{id}`).

thedadams:
Thanks for submitting this change. This is superseded by our larger refactor. See https://github.com/obot-platform/obot/pull/7124

Issue #7004, opened 2026-06-19
Allow users to configure MCP servers on first connect


thedadams:
Users are allowed to connect to MCP servers before creating them in the UI with the exception of servers with required configuration.

Users should be able to connect to such servers and provide the configuration during the authentication process.

Additionally, if required configuration is added to a server, then a user should be redirected to Obot to provide this configuration.

cjellick:
The Wills and I discussed this and the desired UX in Obot around this is that we completely skip config screens when you go to connect to the server in Obot. We just want to give the user the URL and then rely on this mechanism to configure it on the fly.

cjellick:
So, when you get the backend done, hand off the rework of those frontend bits to Ivy.

sangee2004:
This issue is addressed when testing with the latest build from `main`

When user clicks `Connect` for any MCP server that requires configuration parameters like `Exa Search` (that the user has not connected to before) , user will be presented with connection url modal (They will not be prompted for entering required configuration).

<img width="792" height="591" alt="Image" src="https://github.com/user-attachments/assets/24621c6e-781e-46d0-a23b-d9edee4a6065" />

Able to use this connect url from external clients like VS Code and Cursor  to connect to MCP server successfully.

When MCP server gets installed, User will be prompted to enter  required configurations for the MCP server.

<img width="720" height="399" alt="Image" src="https://github.com/user-attachments/assets/59400579-1e8d-45cb-be3d-f7181cdc9c9b" />
<img width="829" height="358" alt="Image" src="https://github.com/user-attachments/assets/c111361e-108d-458c-8b98-f8b19bfe1a0a" />

Once MCP serve is installed , able to make tool calls successfully.

Following issues found when testing this feature are tracked in seperate issues
https://github.com/obot-platform/obot/issues/7041
https://github.com/obot-platform/obot/issues/7042

PR #7007, opened 2026-06-19
feat: add ability to configure MCP server on first connect

thedadams:
Before this change, when a user connected to an MCP server for the first time through an MCP client without first configuring it in the Obot UI, the connection would fail if the server had required configuration.

After this change, the user will be able to provide the required configuration during the connection.

Issue: https://github.com/obot-platform/obot/issues/7004

Demo: https://caps.thedadams.com/s/c8gxmr7wr1eyx57

### DIFF
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -36,7 +36,10 @@ import (
 
 var envVarRegex = regexp.MustCompile(`\${([^}]+)}`)
 
-const requestTimeUpdateInterval = 15 * time.Minute
+const (
+	requestTimeUpdateInterval = 15 * time.Minute
+	configURLKey              = "__url"
+)
 
 // MCPOAuthChecker will check the OAuth status for an MCP server. This interface breaks an import cycle.
 type MCPOAuthChecker interface {
@@ -579,7 +582,7 @@ func (m *MCPHandler) LaunchServer(req api.Context) error {
 				continue
 			}
 
-			config, err := serverConfigForAction(req, component)
+			config, _, err := serverConfigForAction(req, component, false)
 			if err != nil {
 				return fmt.Errorf("failed to get config for component server %s: %w", component.Name, err)
 			}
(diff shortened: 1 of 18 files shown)


pkg/api/handlers/mcp.go
```diff
@@ -1232,6 +1230,68 @@ func entryNeedsUserConfig(ctx context.Context, client kclient.Client, obotNamesp
...
+			server.Spec.PreviousURL = ""
+		}
+	case entryRemote.URLTemplate != "":
+		serverRemote.IsTemplate = true
+		serverRemote.URLTemplate = entryRemote.URLTemplate
+		serverRemote.Hostname = ""
+		server.Spec.NeedsURL = serverRemote.URL == ""
+		if !server.Spec.NeedsURL {
+			server.Spec.PreviousURL = ""
+		}
```
g-linville, line 1289: If `serverRemote.URL` is set here, do we need to check to make sure it matches the template? I see that we are making sure it matches the hostname in the previous case.

g-linville (commented):
Only reviewed the backend; one question

pkg/api/handlers/mcp.go
thedadams: As far as I can tell, we don't validate that anywhere. We just "apply" the template.

g-linville: Surely there is somewhere in the code where we validate that the user-provided URL fits the template, right?

thedadams: As far as I can tell, we don't validate that anywhere.

g-linville: I was thinking about this wrong. I thought we prompted the user for a URL and then made sure it matched the template, but that is not what we do. Instead, we take the variables defined in the template and ask the user to supply each one, and then build the final URL. So we're good.

g-linville (approved):

njhale (approved):

Issue #6736, opened 2026-05-27
`Devices` page should default sort by `Last Scanned` descending.


sangee2004:
**Describe the bug**
The Devices page currently defaults to sorting by `Device`, which makes it harder for admins to see the most recently active or updated devices first. The default ordering should instead use `Last Scanned` in descending order so the latest scan appears at the top.

Pre-Req: Submit obot scans from a few different devices.

**To Reproduce**
1. Log in as Admin and go to the `Devices` page under `Device Management`.

Notice that the default sort order for the device list is by `Device`.

<img width="1090" height="637" alt="Image" src="https://github.com/user-attachments/assets/38f89089-fc2b-4e99-8e74-5ce68d436f51" />

**Expected behavior**
Make the default sort order for the `Devices` page `Last Scanned`, with the most recently scanned device at the top.

ivyjeong13:
Thank Priyansh for the pull request, have PR to address this and including sort/filter tracking [here](https://github.com/obot-platform/obot/pull/7008) so will be closing this PR out.

sangee2004:
This issue is addressed when testing with the latest build from `main`.

PR #7009, opened 2026-06-22
fix: default devices sorting by last scanned desc

PriyanshRaj30:


ivyjeong13:
Closing, related issue #6736 has been addressed.

Issue #6836, opened 2026-06-04
Support max caps for user-defined memory and cpu requests and limits


cjellick:
We now let user set memory cpu requests and limits on individual MCP server catalog entries. I want environmen variable (and thusly helm values.yaml controlled) maximums for what the user is allowed to set for these fields. 

I want this so that in our managed environments, we can prevent users from requesting unrealisitically large resource requests.

g-linville:
I just merged this. See the docs changes in the PR for details on how to set the maximums.

https://github.com/obot-platform/obot/pull/7012

sangee2004:
Support max caps for user-defined memory and cpu requests and limits is availble when testing with latest build from `main` - [v0.0.0-dev+1ae113ba](https://github.com/obot-platform/obot/commit/1ae113ba).

Able to set maximum values for cpu/memory request and limits in Helm installs by setting the following values in the Helm chart:  `--set mcpServerDefaults.maxCPURequest=500m --set mcpServerDefaults.maxCPULimit=1 --set mcpServerDefaults.maxMemoryRequest=300Mi --set mcpServerDefaults.maxMemoryLimit=800Mi ` 

Once these configrations are set , they get enforced in the following workflows:

1. When setting system wide defaults for  cpu/memory request and limts  in`Server Scheduling` pages in  `Resource Limits & Requests` section.

Admin will be presented with 400 errors similar to `400 /k8s-settings: resource maximum validation failed: default MCP server resources exceed maximums: resources.requests.cpu 550m exceeds configured maximum 500m` when settings exceed the max values  set.

2. When setting   cpu/memory request and limts  in `Resource Limits & Requests` section when creating hosted MCP servers.

User will be presented with 400 errors similar to `400 /mcp-catalogs/default/entries/default-santestnew-b2c870edtzj5f: failed to validate entry manifest: resources.requests.memory 255Mi exceeds configured maximum 250Mi` when settings exceed the max values  set.

3. When setting  cpu/memory request and limts  in `Resource Limits & Requests` section as part  updating hosted MCP servers.


Issue relating to MCP servers exceeding updated resource limits not being flagged and able to restart them is tracked in  - https://github.com/obot-platform/obot/issues/7113

PR #7012, opened 2026-06-22
feat: add env vars for k8s resource maximums

g-linville:
for https://github.com/obot-platform/obot/issues/6836

This introduces four new env vars to set a cap on the allowed CPU and memory requests and limits for MCP servers.

We do not allow the admin to configure default K8sSettings above the set maximums. We also check during startup to make sure that the new maximums are not lower than the current configured K8sSettings, if it is configured.

Also, in the implicit/hardcoded fallback when there are no K8sSettings configured, where we set i.e. a `10m` CPU request and a `200Mi` memory request, if the configured maximums are lower than these hardcoded fallbacks, we use those instead.

We block admins from creating new catalog entries that have reqs/limits higher than the maximums (or from updating existing ones to push them above the maximums), but this is just for convenience. Any attempt to ~~deploy a server~~ create an MCPServer from a catalog entry with resources above the maximums will fail at ~~deployment~~ creation time. Additionally, doing a TriggerUpdate to push updates from a catalog entry to an existing MCPServer will fail, if the catalog entry's resources are too high.

Design-wise, the only part of this I'm skeptical about is whether the SessionManager is the best place to store the configured maximums. I talked to Codex about it and ultimately this is what it recommended, but let me know if you have a different preference.

### DIFF
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -86,7 +86,7 @@ func ValidationOptionsWithResourceMaximums(sessionManager *mcp.SessionManager) v
 		return validation.Options{}
 	}
 	options := validationOptions(sessionManager.RemoteMCPURLValidationConfig())
-	options.ResourceMaximums = k8sSettingsResourceMaximums(sessionManager)
+	options.ResourceMaximums = sessionManager.KubernetesResourceMaximums()
 	return options
 }
 
@@ -107,7 +107,7 @@ func (m *MCPHandler) currentK8sSettingsHashWithImagePullSecrets(settings v1.K8sS
 	if err != nil {
 		return "", fmt.Errorf("failed to compute core resource requirements: %w", err)
 	}
-	return mcp.ComputeK8sSettingsHash(settings, resources, mcpServer.Spec.Manifest.Runtime, mcpServer.Spec.NanobotAgentID != "", k8sSettingsResourceMaximums(m.mcpSessionManager), imagePullSecretNames), nil
+	return mcp.ComputeK8sSettingsHash(settings, resources, mcpServer.Spec.Manifest.Runtime, mcpServer.Spec.NanobotAgentID != "", m.mcpSessionManager.KubernetesResourceMaximums(), imagePullSecretNames), nil
 }
 
 func (m *MCPHandler) GetEntryFromAllSources(req api.Context) error {
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -97,7 +97,7 @@ func (m *MCPHandler) currentK8sSettingsHashWithImagePullSecrets(settings v1.K8sS
 	if err != nil {
 		return "", fmt.Errorf("failed to compute core resource requirements: %w", err)
 	}
-	return mcp.ComputeK8sSettingsHash(settings, resources, mcpServer.Spec.Manifest.Runtime, mcpServer.Spec.NanobotAgentID != "", imagePullSecretNames), nil
+	return mcp.ComputeK8sSettingsHash(settings, resources, mcpServer.Spec.Manifest.Runtime, mcpServer.Spec.NanobotAgentID != "", k8sSettingsResourceMaximums(m.mcpSessionManager), imagePullSecretNames), nil
 }
 
 func (m *MCPHandler) GetEntryFromAllSources(req api.Context) error {
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -80,7 +80,8 @@ func validationOptions(remoteValidationConfig mcp.RemoteMCPURLValidationConfig) 
 	}
 }
 
-func validationOptionsWithResourceMaximums(sessionManager *mcp.SessionManager) validation.Options {
+// ValidationOptionsWithResourceMaximums builds MCP manifest validation options from the active MCP session manager.
+func ValidationOptionsWithResourceMaximums(sessionManager *mcp.SessionManager) validation.Options {
 	if sessionManager == nil {
 		return validation.Options{}
 	}
@@ -921,7 +922,7 @@ func (m *MCPHandler) GetPrompt(req api.Context) error {
 	})
 }
 
-func mcpServerOrInstanceFromConnectURL(req api.Context, id, secretBindingAllowedLabel string) (v1.MCPServer, v1.MCPServerInstance, error) {
+func mcpServerOrInstanceFromConnectURL(req api.Context, id, secretBindingAllowedLabel string, validationOptions validation.Options) (v1.MCPServer, v1.MCPServerInstance, error) {
 	switch {
 	case system.IsMCPServerInstanceID(id):
 		var instance v1.MCPServerInstance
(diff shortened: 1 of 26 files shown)


pkg/mcp/resource_maximums.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	return &quantity, nil
+}
+
+func (m ResourceMaximums) Empty() bool {
+	return m.CPURequest == nil &&
+		m.CPULimit == nil &&
+		m.MemoryRequest == nil &&
+		m.MemoryLimit == nil
+}
```
thedadams: Does `m == ResourceMaximums{}` work here? If so, it would be more future proof.

pkg/mcp/kubernetes.go
```diff
@@ -526,12 +528,18 @@ func (k *kubernetesBackend) k8sObjects(ctx context.Context, server ServerConfig,
 	// Fetch K8s settings
 	k8sSettings := k.getK8sSettings(ctx)
 
+	// Make sure that the resources are under the configured maximums
+	mcpResources := mcpContainerResourcesWithMaximums(server.Resources, server.Runtime, server.NanobotAgentName != "", k8sSettings, k.resourceMaximums)
+	if err := validateServerResourceMaximums(server, mcpResources, k.resourceMaximums); err != nil {
```
thedadams: We should get some feedback on this, but typically you wouldn't validate here because that would break anyone using the MCP server when the configuration changes. Instead, you might disallow all changes to a server until the configuration is brought into compliance.

g-linville: Talked to Craig. We'll do what you suggested.

g-linville:
Sorry for the rebase. I had some conflicts to fix.

pkg/api/handlers/mcp.go
```diff
@@ -1005,6 +1015,9 @@ func mcpServerOrInstanceFromConnectURL(req api.Context, id, secretBindingAllowed
 			if err != nil {
 				return v1.MCPServer{}, v1.MCPServerInstance{}, types.NewErrBadRequest("catalog entry %s cannot be connected because it could not be converted to an MCP server: %v", id, err)
 			}
+			if err := validation.ValidateServerManifest(req.Context(), manifest, false, validationOptions); err != nil {
+				return v1.MCPServer{}, v1.MCPServerInstance{}, types.NewErrBadRequest("catalog entry %s cannot be connected because its MCP server manifest is invalid: %v", id, err)
+			}
```
g-linville, line 1020: This blocks the implicit creation of an MCP server at connect time if its resources would be above the maximums.

pkg/api/handlers/mcp.go
```diff
@@ -2068,19 +2081,7 @@ func (m *MCPHandler) UpdateServer(req api.Context) error {
...
-	if catalogID != "" {
-		err = m.removeMCPServer(req.Context(), existing)
-	} else if workspaceID != "" {
-		err = m.removeMCPServer(req.Context(), existing)
-	} else {
-		err = m.removeMCPServer(req.Context(), existing)
-	}
-	if err != nil {
-		return err
-	}
```
g-linville, line 2081: kinda funny that this was in our codebase

pkg/api/handlers/skills_test.go
```diff
@@ -399,6 +400,30 @@ func newFakeStorage(t *testing.T, objects ...kclient.Object) kclient.WithWatch {
...
+		}).
+		WithIndex(&v1.MCPServerInstance{}, "spec.mcpServerName", func(obj kclient.Object) []string {
+			return []string{obj.(*v1.MCPServerInstance).Spec.MCPServerName}
+		}).
+		WithIndex(&v1.MCPServerInstance{}, "spec.template", func(obj kclient.Object) []string {
+			return []string{strconv.FormatBool(obj.(*v1.MCPServerInstance).Spec.Template)}
+		}).
+		WithIndex(&v1.MCPServerInstance{}, "spec.compositeName", func(obj kclient.Object) []string {
+			return []string{obj.(*v1.MCPServerInstance).Spec.CompositeName}
+		}).
```
g-linville, line 426: This is needed for the `fakeStorage` thing that we use in some of the tests introduced in this PR. It just happens to be part of the `skills_test.go` file.

pkg/mcp/resource_maximums.go
g-linville: yep, fixed

pkg/mcp/kubernetes.go
g-linville: done

g-linville:
This is ready for review again.

pkg/api/handlers/mcp.go
thedadams: Just needed to be extra careful.

pkg/api/handlers/mcpgateway/handler.go
```diff
@@ -105,7 +105,7 @@ func (h *Handler) ensureServerIsDeployed(req api.Context) (mcp.ServerConfig, str
 	}
 
 	connectID := mcpID
-	mcpID, mcpServer, mcpServerConfig, missingConfig, err := handlers.ServerForActionWithConnectIDAllowMissingConfig(req, mcpID, h.secretBindingAllowedLabel)
+	mcpID, mcpServer, mcpServerConfig, missingConfig, err := handlers.ServerForActionWithConnectIDAllowMissingConfig(req, mcpID, h.secretBindingAllowedLabel, handlers.ValidationOptionsWithResourceMaximums(h.mcpSessionManager))
```
thedadams, line 108: FYI: I'm making this better.

pkg/api/handlers/mcp.go
```diff
@@ -80,6 +80,16 @@ func validationOptions(remoteValidationConfig mcp.RemoteMCPURLValidationConfig)
 	}
 }
 
+// ValidationOptionsWithResourceMaximums builds MCP manifest validation options from the active MCP session manager.
+func ValidationOptionsWithResourceMaximums(sessionManager *mcp.SessionManager) validation.Options {
+	if sessionManager == nil {
+		return validation.Options{}
+	}
+	options := validationOptions(sessionManager.RemoteMCPURLValidationConfig())
+	options.ResourceMaximums = k8sSettingsResourceMaximums(sessionManager)
```
thedadams: I'm having trouble finding the `k8sSettingsResourceMaximums` function in this review. Two questions:
1. Does it make sense to the name of the method `sessionManager.RemoteMCPURLValidationConfig` to be more general and include the maximums, if they apply?
2. If the answer to the above question is no, then does it make sense for `k8sSettingsResourceMaximums` to be an exported method on the `sessionManager`?

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -95,16 +95,21 @@ func (h *Handler) revealCatalogCredential(ctx context.Context, catalogName, sour
 
 func New(defaultCatalogPath, defaultSystemCatalogPath string, gatewayClient *gclient.Client, accessControlRuleHelper *accesscontrolrule.Helper, mcpSessionManager *mcp.SessionManager) *Handler {
 	remoteURLValidationConfig := mcpSessionManager.RemoteMCPURLValidationConfig()
+	validationOptions := validation.Options{
+		RemoteMCPURLValidationConfig: remoteURLValidationConfig,
+	}
+	if mcp.IsKubernetesBackend(mcpSessionManager.MCPRuntimeBackend()) {
+		validationOptions.ResourceMaximums = mcpSessionManager.ResourceMaximums()
```
thedadams: Oh, we do have this method. Maybe I didn't understand something in the above comment.

thedadams (approved):

pkg/api/handlers/mcp.go
g-linville: I think number 2 here is correct. I'll clean that up. Thanks.

g-linville: Moved it to the session manager

thedadams (approved):

PR #7037, opened 2026-06-24
enhance: more logging around successful license validation

thedadams:


pkg/api/authz/authz.go
```diff
@@ -221,14 +221,6 @@ var (
...
 			// This endpoint validates the API key passed in the header
 			"POST /api/api-keys/auth",
-
-			// Since these are temporary and only the credstore token is allowed to access,
-			// then auth is handled in the handler.
-			"GET /api/credentials/tool.gpt",
-			"POST /api/credentials/store",
-			"POST /api/credentials/get",
-			"POST /api/credentials/list",
-			"POST /api/credentials/erase",
```
thedadams, line 231: Snuck this in because I noticed it.

g-linville: Is there corresponding migration code to be removed too? I can't remember.

g-linville (approved):

thedadams:
> Is there corresponding migration code to be removed too? I can't remember.

These were just endpoints that exposed credentials to GPTScript as a half-way step to removing GPTScript. The corresponding code has been removed, but these don't need to be authorized anymore.

There are migrations, but those migrations are still necessary.

Issue #5491, opened 2026-01-13
Error when deploying MCP servers into a cluster with resource quotas


cloudnautique:
When a resource quota is applied to the MCP namespace you can't launch any MCP servers because the Nanobot shim container doesn't have resource requests.

sangee2004:
MCP deployments fail to get deployed successfully when resource quotas are enabled for the obot-mcp namespace.

User gets stuck in this state for a very long time:
<img width="1247" height="798" alt="Image" src="https://github.com/user-attachments/assets/19ce70e2-01ca-4de7-8b3d-a69779bb9441" />

Deployment never gets to ready state and has the following status message - `pods "ms1ns8xr-57cf446c4f-pl2pt" is forbidden: failed quota: compute-quota:
      must specify limits.cpu for: ms1ns8xr-mwv1mbg5k,ms1ns8xr-shim; limits.memory
      for: ms1ns8xr-mwv1mbg5k,ms1ns8xr-shim; requests.cpu for: ms1ns8xr-mwv1mbg5k,ms1ns8xr-shim;
      requests.memory for: ms1ns8xr-mwv1mbg5k,ms1ns8xr-shim' `

Shim and webhook containers are deployed with no resources, even when there are cpu/memory requests and limits sepcified in `Server Scheduling.`

Resource Quota that was created in `obot-mcp` namespace.
```
kubectl get resourcequota -n obot-mcp               
NAME            AGE   REQUEST                                                     LIMIT
compute-quota   61m   pods: 6/20, requests.cpu: 3/4, requests.memory: 256Mi/4Gi   limits.cpu: 4800m/14, limits.memory: 0/8Gi
sangeethahariharan@Sangeethas-MacBook-Pro obot % 
```

Deployment that fails to get to Ready state:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
[...]

PR #7044, opened 2026-06-25
chore: add resource requirements to nanobot shims

g-linville:
for https://github.com/obot-platform/obot/issues/5491

I chose these values by looking at our main environment in Google Cloud and seeing what the steady usage over the past week looked like.

After the big refactor, this won't be needed, but it's worth having until then.


g-linville:
Sorry, moving back into draft because I just realized I need to add limits also.

g-linville:
Now we're ready again.

g-linville:
Actually, turns out we don't want limits. I removed them.

thedadams (approved):

njhale (approved):

Issue #7045, opened 2026-06-25
MCPServer status update fails managedFields conversion: OAuthMetadata raw-JSON fields declared string/byte but serialized as object


zmiievskyi:
## Summary

Every time Obot syncs OAuth metadata for a remote MCP server, the apiserver logs a `[SHOULD NOT HAPPEN]` error and the `MCPServer` status update fails its managedFields (server-side-apply) conversion:

```
level=error msg="[SHOULD NOT HAPPEN] failed to update managedFields"
error="failed to convert new object (default/<name>; obot.obot.ai/v1, Kind=MCPServer) to smd typed: errors:
  .status.oauthMetadata.authorizationServerMetadata: expected string, got <object>
  .status.oauthMetadata.clientRegistration: expected string, got <object>
  .status.oauthMetadata.protectedResourceMetadata: expected string, got <object>"
```

It fires on every `SyncOAuthMetadata` for remote MCP servers that use OAuth.

## Root cause

The three metadata fields are typed as `json.RawMessage`:

`pkg/storage/apis/obot.obot.ai/v1/mcpserver.go`
```go
type OAuthMetadata struct {
    ProtectedResourceURL              string          `json:"protectedResourceUrl,omitempty"`
    AuthorizationServerURL            string          `json:"authorizationServerUrl,omitempty"`
    ProtectedResourceMetadata         json.RawMessage `json:"protectedResourceMetadata,omitempty"`
    AuthorizationServerMetadata       json.RawMessage `json:"authorizationServerMetadata,omitempty"`
    ClientRegistration                json.RawMessage `json:"clientRegistration,omitempty"`
    DynamicClientRegistration         bool            `json:"dynamicClientRegistration,omitempty"`
    ClientIDMetadataDocumentSupported bool            `json:"clientIdMetadataDocumentSupported,omitempty"`
}
```

`json.RawMessage` is `[]byte`, so `openapi-gen` renders it as `type: string, format: byte`:

`pkg/storage/openapi/generated/openapi_generated.go` (`schema_storage_apis_obotobotai_v1_OAuthMetadata`)
```go
"protectedResourceMetadata":   { Type: []string{"string"}, Format: "byte" },
"authorizationServerMetadata": { Type: []string{"string"}, Format: "byte" },
"clientRegistration":          { Type: []string{"string"}, Format: "byte" },
```

[...]

PR #7053, opened 2026-06-25
fix: store OAuthMetadata raw-JSON fields as runtime.RawExtension

zmiievskyi:
Issue #7045

## Problem

`OAuthMetadata.{ProtectedResourceMetadata,AuthorizationServerMetadata,ClientRegistration}` are typed as `json.RawMessage`. `openapi-gen` renders `[]byte` as `{type: string, format: byte}`, but at runtime these fields serialize as JSON **objects**. So every `MCPServer` status update fails the managedFields typed conversion:

```
[SHOULD NOT HAPPEN] failed to update managedFields ...
  .status.oauthMetadata.authorizationServerMetadata: expected string, got <object>
  .status.oauthMetadata.clientRegistration: expected string, got <object>
  .status.oauthMetadata.protectedResourceMetadata: expected string, got <object>
```

## Fix

Switch the three fields to `runtime.RawExtension`. It's a struct, so `openapi-gen` emits a `$ref` to its `{type: object}` schema — the canonical k8s "arbitrary JSON" type, already used elsewhere in the repo — which matches the runtime value and lets server-side apply convert cleanly. Call sites use `.Raw`. Regenerated `openapi_generated.go` + `zz_generated.deepcopy.go` (the deepcopy simplifies, since `RawExtension` has its own `DeepCopyInto`).

`apiclient/types.OAuthMetadata` is intentionally left as `json.RawMessage` (the SDK is k8s-free; its schema is documentation-only and not used for SSA), so its serialization is unchanged.

> Note: `omitempty` has no effect on a non-pointer struct, so empty values now serialize as `null` in the v1 status object instead of being omitted. This is standard `RawExtension` behavior and doesn't affect the API/SDK output.

## Verification

- `go generate` → the 3 fields now `$ref` `runtime.RawExtension` instead of `string`/`byte`
- `go build ./...`, `go vet ./...` (changed pkgs), and `go test ./pkg/api/handlers/ -run OAuthDebugger` all pass; `gofmt` clean
- **Runtime:** ran a patched image through a live `claude.ai → Obot → upstream MCP` OAuth flow — `Updated MCP server OAuth metadata` is now clean, with **zero** `[SHOULD NOT HAPPEN]` errors across the run (this spot errored on every sync before).


thedadams (approved):

thedadams:
Thanks for the contribution!

PR #7055, opened 2026-06-25
feat: GitOps for composite MCP servers

calvinmclean:
## Summary

Addresses https://github.com/obot-platform/field-issues/issues/36

Adds GitOps-friendly portable references for composite MCP catalog entries.
Composite entries can now reference component catalog entries by stable `entryKey` values instead of generated internal catalog entry IDs. References support same-source shorthand and explicit cross-source refs using `{sourceID}::{entryKey}`. During sync, Obot resolves portable refs to internal catalog entry IDs and hydrates component manifests so composites validate and preview correctly.
Also tightens Git-managed composite behavior:
- `entryKey` must be DNS-friendly and unique within a source.
- Git-managed tool overrides must use `overrideDescription`; `description` is rejected because it is reserved for live/source tool metadata.
- Git-managed composite UI controls respect readonly state.
- Configurable catalog entries route users through setup before showing connect URLs.

Example catalog: https://github.com/calvinmclean/my-mcp-catalog
Related PR: https://github.com/obot-platform/mcp-catalog/pull/1368

## Changes
- Add `entryKey` to catalog entry manifests and persist source metadata on synced entries.
- Resolve same-source and cross-source composite refs during catalog sync.
- Preserve direct internal `catalogEntryID` behavior when no portable ref matches.
- Hydrate component manifests for internal ID refs when targets are in the sync batch.
- Normalize source IDs from source URLs for portable cross-source refs.
- Validate duplicate source IDs and duplicate `entryKey` values.
- Reject Git-managed composite tool override `description`; require `overrideDescription`.
- Update GitOps documentation for portable composite refs and override description behavior.
- Update UI behavior for readonly composite catalog entries and configurable connect URL flow.


pkg/api/handlers/systemmcpcatalogs.go
```diff
@@ -285,31 +285,22 @@ func normalizeAndValidateCatalogSourceURLs(sourceURLs []string, localPath string
...
 			continue
 		}
-		if _, ok := seen[urlStr]; ok {
-			return types.NewErrBadRequest("duplicate URL found: %s", urlStr)
+		sourceID := normalizedCatalogSourceID(urlStr)
+		if _, ok := seen[sourceID]; ok {
+			return types.NewErrBadRequest("duplicate catalog source ID %q", sourceID)
 		}
-		seen[urlStr] = struct{}{}
+		seen[sourceID] = struct{}{}
```
calvinmclean: This change makes the duplicate-detection a bit more solid by normalizing it a bit. Making this change here prevents us from having to detect duplicates as part of the source read process.

pkg/validation/source.go
```diff
@@ -0,0 +1,12 @@
+package validation
+
+import "strings"
+
+func SourceIDForURL(sourceURL string) string {
+	sourceURL = strings.TrimPrefix(sourceURL, "https://")
+	sourceURL = strings.TrimPrefix(sourceURL, "http://")
+	if len(sourceURL) > 1 {
+		sourceURL = strings.TrimRight(sourceURL, "/")
+	}
```
thedadams: Based on this code, it seems the `/` is valid and `sourceURL` will be returned as `/` in this case. However, `//` is not valid and `sourceURL` will be returned as `""` in that case.

Is that the desired behavior?

pkg/controller/handlers/mcpcatalog/github_test.go
```diff
@@ -28,6 +31,29 @@ func TestIsGitRepoURL(t *testing.T) {
...
+shortDescription: Test
+description: Test
+icon: icon
+runtime: npx
+npxConfig:
+  package: test
+`), 0o600))
+
+	h := &Handler{}
+	objs, err := h.readMCPCatalog(context.Background(), "default", dir, "")
```
thedadams:
```suggestion
	objs, err := h.readMCPCatalog(t.Context(), "default", dir, "")
```

thedadams (approved):

g-linville (approved):
backend lgtm

pkg/validation/source.go
calvinmclean: I did not specifically consider these cases. I think it would be better to also invalidate `/` similar to `//` by simply removing the if statement. Any issues with that change?

njhale (approved):

calvinmclean:
@thedadams could you take another look? Copilot had a good suggestion for a validation change after pushing changes for your suggestion. Then I had to fix some merge conflicts.

thedadams (approved):

PR #7070, opened 2026-06-29
feat: obot app scheduling info & page

ivyjeong13:
* Addresses #7132 
* Adds `GET /api/app-k8s-settings`
* Shows scheduling under App Management -> App Scheduling


pkg/services/config.go
```diff
@@ -306,9 +307,7 @@ func parsePSASettingsFromHelm(opts mcp.Options) (*v1.PodSecurityAdmissionSetting
 	}, nil
 }
 
-// parsePodSchedulingSettingsFromHelm parses pod scheduling settings (affinity, tolerations, resources,
-// runtimeClassName) from Helm options. These settings can be managed via Helm OR UI.
-// If this returns non-nil, SetViaHelm will be true and UI cannot modify these settings.
+// parsePodSchedulingSettingsFromHelm parses pod scheduling settings from legacy env vars.
```
g-linville: Do you know why this comment was updated? I'm not sure what "legacy env vars" means here. The function itself didn't change at all.

pkg/mcp/loader.go
```diff
@@ -49,6 +49,7 @@ type Options struct {
 	// Obot service configuration for constructing internal service FQDN
 	ServiceName      string `usage:"The Kubernetes service name for the obot server"`
 	ServiceNamespace string `usage:"The Kubernetes namespace where the obot server runs"`
+	ConfigSecret     string `usage:"The Kubernetes secret name containing the Helm values snapshot and config env vars"`
```
g-linville: nit: let's add space here so that the comment on line 49 is not mistakenly associated with the ConfigSecret field.


```suggestion

	ConfigSecret     string `usage:"The Kubernetes secret name containing the Helm values snapshot and config env vars"`
```

pkg/api/router/router.go
```diff
@@ -419,6 +419,13 @@ func Router(ctx context.Context, services *services.Services) (http.Handler, err
 	mux.HandleFunc("GET /api/k8s-settings", k8sSettingsHandler.Get)
 	mux.HandleFunc("PUT /api/k8s-settings", k8sSettingsHandler.Update)
 
+	appK8sSettingsHandler := handlers.NewAppK8sSettingsHandler(
+		services.ConfigSecret,
+		services.ServiceNamespace,
+		services.LocalK8sClient,
+	)
```
g-linville: nit: we normally declare these at the top of the file

g-linville (changes requested):
In general, the backend changes seem overcomplicated to me. I don't think we need to fetch the secret and read from it, when we already have access to all of the k8s settings within the system anyway, right? Or was this part of a design discussion I missed?

ivyjeong13:
> In general, the backend changes seem overcomplicated to me. I don't think we need to fetch the secret and read from it, when we already have access to all of the k8s settings within the system anyway, right? Or was this part of a design discussion I missed?

Oh no no, the only requirement was to display read-only information of scheduling and thought I'd take a crack at it. Thanks Grant for the review! Gonna run through it again and will address your review comments!

pkg/services/config.go
ivyjeong13: This I'm looking into, I never mentioned legacy env vars so wonder where the agent picked it up edit: I think potentially what happened was when I was referencing env vars that I since removed, it then decided they were legacy and changed this comment. Reverted the changed comment and did once over to make sure no other comments were changed.

chart/templates/secret.yaml
```diff
@@ -54,6 +54,18 @@ data:
...
+  {{- end }}
+  {{- if .Values.tolerations }}
+  OBOT_APP_K8S_SETTINGS_TOLERATIONS: {{ .Values.tolerations | toJson | b64enc | quote }}
+  {{- end }}
+  {{- if .Values.resources }}
+  OBOT_APP_K8S_SETTINGS_RESOURCES: {{ .Values.resources | toJson | b64enc | quote }}
+  {{- end }}
+  {{- if .Values.runtimeClassName }}
+  OBOT_APP_K8S_SETTINGS_RUNTIME_CLASS_NAME: {{ .Values.runtimeClassName | b64enc | quote }}
+  {{- end }}
```
g-linville: Something is still wrong here. We have other env vars defined farther up, and this is trying to pull the values `.Values.affinity`, `.Values.tolerations`, etc. which are undefined in our Helm chart.

g-linville (changes requested):

pkg/api/handlers/k8ssettings.go
```diff
@@ -17,14 +19,96 @@ import (
...
+				resources = &containerResources
+			}
+			break
+		}
+	}
+
+	var runtimeClassName *string
+	if podSpec.RuntimeClassName != nil && *podSpec.RuntimeClassName != "" {
+		runtimeClassName = podSpec.RuntimeClassName
+	}
```
thedadams:
```suggestion
	runtimeClassName := podSpec.RuntimeClassName
```

pkg/services/config.go
```diff
@@ -379,6 +366,51 @@ func parsePodSchedulingSettingsFromHelm(opts mcp.Options) (*v1.K8sSettingsSpec,
...
+		resources             *corev1.ResourceRequirements
+		runtimeClassNameValue *string
+	)
+
+	if affinityJSON != "" && affinityJSON != "{}" {
+		var parsed corev1.Affinity
+		if err := unmarshalJSONStrict([]byte(affinityJSON), &parsed); err != nil {
+			return nil, nil, nil, nil, fmt.Errorf("failed to parse affinity from Helm: %w", err)
+		}
+		affinity = &parsed
```
thedadams: Instead of using another variable, we can use the one we already have.
```suggestion
		affinity = new(corev1.Affinity)
		if err := unmarshalJSONStrict([]byte(affinityJSON), affinity); err != nil {
			return nil, nil, nil, nil, fmt.Errorf("failed to parse affinity from Helm: %w", err)
		}
```

pkg/services/config.go
```diff
@@ -379,6 +366,51 @@ func parsePodSchedulingSettingsFromHelm(opts mcp.Options) (*v1.K8sSettingsSpec,
...
+	}
+
+	if tolerationsJSON != "" && tolerationsJSON != "[]" {
+		var parsed []corev1.Toleration
+		if err := unmarshalJSONStrict([]byte(tolerationsJSON), &parsed); err != nil {
+			return nil, nil, nil, nil, fmt.Errorf("failed to parse tolerations from Helm: %w", err)
+		}
+		tolerations = parsed
```
thedadams:
```suggestion
		if err := unmarshalJSONStrict([]byte(tolerationsJSON), &tolerations); err != nil {
			return nil, nil, nil, nil, fmt.Errorf("failed to parse tolerations from Helm: %w", err)
		}
```

pkg/services/config.go
```diff
@@ -379,6 +366,51 @@ func parsePodSchedulingSettingsFromHelm(opts mcp.Options) (*v1.K8sSettingsSpec,
...
+	}
+
+	if resourcesJSON != "" && resourcesJSON != "{}" {
+		var parsed corev1.ResourceRequirements
+		if err := unmarshalJSONStrict([]byte(resourcesJSON), &parsed); err != nil {
+			return nil, nil, nil, nil, fmt.Errorf("failed to parse resources from Helm: %w", err)
+		}
+		resources = &parsed
```
thedadams:
```suggestion
		resources = new(corev1.ResourceRequirements)
		if err := unmarshalJSONStrict([]byte(resourcesJSON), resources); err != nil {
			return nil, nil, nil, nil, fmt.Errorf("failed to parse resources from Helm: %w", err)
		}
```

pkg/api/handlers/k8ssettings.go
thedadams: We don't need to check anything here. Just assign `runtimeClass = podSpec.RuntimeClass`.

thedadams (approved):

pkg/api/handlers/k8ssettings.go
ivyjeong13: Thanks Donnie! addressed!

thedadams (approved):

g-linville (approved):
backend lgtm

PR #7087, opened 2026-07-01
enhance: add device enrollment and device-authenticated scan submission

njhale:
Lets an unattended endpoint enroll into a fleet and submit device scans with no
user-scoped auth.

An MDMDeployment is a fleet that devices enroll into; enrollment is authorized
by one or more DeviceEnrollmentKeys attached to it
(ode1-<deployment_id>-<key_id>-<secret>, bcrypt-hashed at rest). A device
presents a key once at POST /api/mdm/enroll to register its identity public key
(trust-on-first-use; rebinding a different key to an existing device is
rejected; the key must be ECDSA P-256/P-384/P-521 or Ed25519). Thereafter the
device authenticates each scan submission (POST /api/devices/scans) with a
short-lived JWT it signs with its own key, verified against the registered
public key.

Deployments, their enrollment keys, and enrolled devices are managed by admins
under /api/mdm/deployments.


pkg/api/handlers/devicedeployments.go
```diff
@@ -0,0 +1,209 @@
...
+	deploymentID, err := deploymentIDFromPath(req)
+	if err != nil {
+		return err
+	}
+	keyID, err := parsePathUint(req, "key_id", "enrollment key id")
+	if err != nil {
+		return err
+	}
+	if err := req.GatewayClient.DeleteDeviceEnrollmentKey(req.Context(), deploymentID, keyID); err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
```
thedadams: nit: I like when delete calls are idempotent.

pkg/api/handlers/devicedeployments.go
```diff
@@ -0,0 +1,209 @@
...
+
+// callerUserID returns the authenticated user's gateway DB id. Admin/owner
+// principals (the only callers authz allows here) carry the numeric id as UID.
+func callerUserID(req api.Context) (uint, error) {
+	id, err := strconv.ParseUint(req.User.GetUID(), 10, 64)
+	if err != nil {
+		return 0, types.NewErrBadRequest("caller is not a user")
+	}
+	return uint(id), nil
+}
```
thedadams: I believe there is already a `UserID` method on the `api.Context` object that does this.

pkg/api/handlers/devicescans.go
```diff
@@ -32,16 +32,25 @@ func userIsDeviceScanReader(req api.Context) bool {
...
+//     principal rather than trusted from the request body
 func (*DeviceScansHandler) Submit(req api.Context) error {
 	var manifest types.DeviceScanManifest
 	if err := req.Read(&manifest); err != nil {
 		return err
 	}
 
 	scan := gtypes.DeviceScanFromManifest(manifest)
-	scan.SubmittedBy = req.User.GetUID()
+	if deviceID := req.User.GetExtra()["device_id"]; len(deviceID) > 0 && deviceID[0] != "" {
```
thedadams: nit:
```suggestion
	if deviceID := utils.FirstSet(req.User.GetExtra()["device_id"]...); deviceID != "" {
```

apiclient/types/device.go
```diff
@@ -0,0 +1,34 @@
...
+
+type DeviceList List[Device]
+
+// DeviceEnrollRequest is the body of POST /api/devices/enroll. PublicKey is the
+// device's identity key as base64-encoded DER (PKIX / SubjectPublicKeyInfo);
+// the device proves possession of it by signing the access JWTs it presents
+// when submitting scans.
+type DeviceEnrollRequest struct {
+	DeviceID  string `json:"deviceID"`
+	PublicKey string `json:"publicKey"`
```
thedadams: I believe that if this type is `[]byte` instead of `string`, then Unmarshal will automatically base64-decode it.

pkg/gateway/client/devicedeployment.go
```diff
@@ -0,0 +1,203 @@
...
+func (c *Client) DeleteDeviceEnrollmentKey(ctx context.Context, deploymentID, id uint) error {
+	res := c.db.WithContext(ctx).
+		Where("id = ? AND device_deployment_id = ?", id, deploymentID).
+		Delete(&types.DeviceEnrollmentKey{})
+	if res.Error != nil {
+		return fmt.Errorf("failed to delete device enrollment key: %w", res.Error)
+	}
+	if res.RowsAffected == 0 {
+		return gorm.ErrRecordNotFound
+	}
```
thedadams: Same nit here about idempotency.

pkg/gateway/client/devicedeployment.go
```diff
@@ -0,0 +1,203 @@
...
+
+	if err := bcrypt.CompareHashAndPassword([]byte(key.HashedSecret), []byte(secret)); err != nil {
+		return nil, fmt.Errorf("invalid device enrollment credential")
+	}
+	if key.ExpiresAt != nil && key.ExpiresAt.Before(time.Now()) {
+		return nil, fmt.Errorf("device enrollment credential expired")
+	}
+
+	// Best-effort last-used update.
+	now := time.Now()
```
thedadams: Do we need `.UTC()` here?

pkg/gateway/server/device_auth.go
```diff
@@ -0,0 +1,128 @@
...
+		return nil, false, nil
+	}
+
+	// Read sub + aud WITHOUT verifying, only to decide if this token is ours and
+	// to locate the device's registered key. Claim only our audience so this
+	// never shadows user/session JWTs.
+	sub, aud, err := unverifiedSubAud(tok)
+	if err != nil || sub == "" || !hasAudience(aud, deviceTokenAudience) {
+		return nil, false, nil
+	}
```
thedadams: Do we need to do this step? Why not just parse it completely upfront?

pkg/gateway/server/deviceenroll_auth.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// AuthenticateRequest implements authenticator.Request.
+func (a *DeviceEnrollmentAuthenticator) AuthenticateRequest(req *http.Request) (*authenticator.Response, bool, error) {
+	authHeader := strings.TrimPrefix(req.Header.Get("Authorization"), "Bearer ")
+	if authHeader == "" {
+		authHeader = req.Header.Get("X-API-Key")
+		if authHeader == "" {
+			return nil, false, nil
+		}
```
thedadams: Do we need to do this? The other reason we do it in other places is so the token can be used in an Anthropic-compatible way. Seems like this isn't needed here.

pkg/gateway/server/deviceenroll_auth.go
```diff
@@ -0,0 +1,70 @@
...
+		// Invalid credential: fall through (ends at anonymous, then denied by authz).
+		return nil, false, nil
+	}
+
+	principal := gtypes.DeviceDeploymentPrincipalName(key.DeviceDeploymentID)
+	return &authenticator.Response{
+		User: &user.DefaultInfo{
+			Name:   principal,
+			UID:    principal,
+			Groups: []string{types.GroupDeviceEnroll},
```
thedadams: The intention is that every "user" that is authenticated get the `types.GroupAuthenticated`. Would it be bad to do here?

I know it would give access to `/api/me`, but that doesn't seem that bad.

pkg/gateway/server/deviceenroll_auth.go
njhale: No we don't. Dropping

pkg/gateway/server/deviceenroll_auth.go
njhale: We can do that here as well. Just wanted it tightly scoped to start

pkg/gateway/client/device.go
```diff
@@ -0,0 +1,94 @@
...
+			updates := map[string]any{
+				"device_deployment_id": in.DeviceDeploymentID,
+				"hostname":             in.Hostname,
+				"os":                   in.OS,
+				"os_version":           in.OSVersion,
+			}
+			if err := tx.Model(&existing).Updates(updates).Error; err != nil {
+				return fmt.Errorf("failed to re-enroll device: %w", err)
+			}
+			existing.DeviceDeploymentID = in.DeviceDeploymentID
```
g-linville: Why is only the DeviceDeploymentID, and not the Hostname, OS, and OSVersion copied here?

pkg/api/handlers/mdmdeployments.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+// Create handles POST /api/device-deployments. Creating a deployment mints its
+// first enrollment key; the plaintext credential is returned once.
+func (*DeviceDeploymentsHandler) Create(req api.Context) error {
+	var in deviceDeploymentCreateRequest
+	if err := req.Read(&in); err != nil {
+		return err
+	}
+	if in.Name == "" {
```
calvinmclean, line 57: nit: should we check with `strings.TrimSpace`?

pkg/api/handlers/deviceenroll.go
```diff
@@ -0,0 +1,66 @@
...
+func (*DeviceEnrollHandler) Enroll(req api.Context) error {
+	deploymentID, ok := uintFromExtra(req.User.GetExtra(), "device_deployment_id")
+	if !ok {
+		return types.NewErrBadRequest("enrollment requires a device enrollment credential")
+	}
+	var in types.DeviceEnrollRequest
+	if err := req.Read(&in); err != nil {
+		return err
+	}
+	if in.DeviceID == "" {
```
calvinmclean: similar nit: should we check with `strings.TrimSpace`?

pkg/gateway/client/devicedeployment.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	return &key, nil
+}
+
+// ParseDeviceEnrollmentCredential extracts the deployment ID, key ID, and secret
+// from an ode1-<deployment_id>-<key_id>-<secret> credential.
+func ParseDeviceEnrollmentCredential(credential string) (deploymentID, keyID uint, secret string, err error) {
+	var prefix string
+	n, err := fmt.Sscanf(credential, "%4s-%d-%d-%s", &prefix, &deploymentID, &keyID, &secret)
```
calvinmclean: nit: can we make the format `%4s-%d-%d-%s` a constant since it's reference for creating and parsing? I assume `%4s` should work with `Sprintf`

pkg/gateway/client/device.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	return &device, nil
+}
+
+// ListDevices returns the devices enrolled into a deployment, newest first.
+func (c *Client) ListDevices(ctx context.Context, deploymentID uint) ([]types.Device, error) {
+	var devices []types.Device
+	if err := c.db.WithContext(ctx).
+		Where("device_deployment_id = ?", deploymentID).
+		Order("enrolled_at DESC").
```
calvinmclean: If this table can grow large, it might be worth having a composite index on these two fields

pkg/gateway/client/mdmdeployment.go
```diff
@@ -0,0 +1,199 @@
...
+	}, nil
+}
+
+// ListDeviceEnrollmentKeys returns the (secret-free) keys for a deployment,
+// newest first.
+func (c *Client) ListDeviceEnrollmentKeys(ctx context.Context, deploymentID uint) ([]types.DeviceEnrollmentKey, error) {
+	var keys []types.DeviceEnrollmentKey
+	if err := c.db.WithContext(ctx).
+		Where("device_deployment_id = ?", deploymentID).
+		Order("created_at DESC").
```
calvinmclean, line 138: Similarly, we could index these as well.

calvinmclean (approved):
Overall looks good. Just a few nits and I might be overthinking DB indexes but I'm curious to see what you think.

g-linville (approved):

PR #7091, opened 2026-07-01
docs: add Slack MCP tutorial

g-linville:
for https://github.com/obot-platform/field-issues/issues/48

The doc mentions that this is in the default catalog. It isn't yet, but I am going to add it there too, before the next release.

### DIFF
--- a/docs/docs/configuration/tutorials/slack-mcp-server.md
+++ b/docs/docs/configuration/tutorials/slack-mcp-server.md
@@ -1,0 +1,72 @@
+# Slack MCP Server
+
+Slack has an [official remote MCP server](https://docs.slack.dev/ai/slack-mcp-server/) that Obot can connect to. This guide explains how to set that up.
+
+### 1. Configure a Slack app
+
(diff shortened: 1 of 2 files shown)


docs/docs/configuration/tutorials/slack-mcp-server.md
```diff
@@ -0,0 +1,72 @@
+# Slack MCP Server
+
+Slack has an [official remote MCP server](https://docs.slack.dev/ai/slack-mcp-server/) that Obot can connect to. This guide explains how to set that up.
```
thedadams, line 3: Are there any permissions needed for this step or installing it into a workspace that should be called out?

thedadams (approved):

docs/docs/configuration/tutorials/slack-mcp-server.md
g-linville: According to [Slack's docs](https://slack.com/help/articles/202035138-Add-apps-to-your-Slack-workspace), it looks like, by default, workspace members can install whatever apps they want, so I think it's fine to leave this as it is.

Issue #7088, opened 2026-07-01
Composite MCP server ignores toolOverrides when enabled is omitted


sangee2004:
**Describe the bug**
When a composite MCP server specifies a tool in the `toolOverride`s section without explicitly setting the enabled field, the Configuration tab shows only that tool in disabled state. However, after connecting from VS Code, all tools from the underlying Exa Search MCP server are exposed and available for use. If enabled defaults to true when omitted, only the tools explicitly listed in toolOverrides should be available, while all other tools should remain disabled.

1. Create a composite MCP sercver with gmail mcp server and disabling "delete_email" tool by setting `enabled: false` in `toolOverrides` section 
```
name: Sangee Composite Exa
runtime: composite
compositeConfig:
  componentServers:
    - catalogEntryID: github.com/obot-platform/mcp-catalog::obot-exa-search
      toolOverrides:
        - name: web_search_exa
```

2. Sync this catalog entry.
3. Go to `MCP Catalog` page and select this entry.

In the `Configuration` tab notice that `web_search_exa` is shown as being disabled

<img width="970" height="679" alt="Image" src="https://github.com/user-attachments/assets/2aac25a7-19a7-44e0-a64a-29ae8474d0d0" />

4. Connect to this MCP server from VS Code.

Notice that all Exa Search tools are available to the composite MCP Server.

You can also see that all tools are avialble by going to `tools` tab of this MCP server entry from`MCP Servers` page.

<img width="974" height="551" alt="Image" src="https://github.com/user-attachments/assets/c53d9411-b60c-4f07-82ea-cc5568bfe0fd" />

**Expected behavior**
In this case , I expected to be given access to only `web_search_exa` tool since it is the only one called out in the `toolOverrides` section assuming that the default for `enabled` flag is `true` if not specified.

sangee2004:
Related to https://github.com/obot-platform/obot/issues/7085

sangee2004:
This issue is addressed when testing with the latest build from `main`

No tools are available when connecting to this composite MCP sever from VS Code as expected.
```
026-07-07 14:41:59.041 [info] Starting server Sangee Composite Exa
2026-07-07 14:41:59.041 [info] Connection state: Starting
2026-07-07 14:41:59.042 [info] Starting server from LocalProcess extension host
2026-07-07 14:41:59.042 [info] Connection state: Running
[...]

Issue #7085, opened 2026-07-01
Disabled tools remain available in composite MCP servers synced from catalog.


sangee2004:
**Describe the bug**
When a composite MCP server is configured with toolOverrides to disable a tool (for example, delete_email with enabled: false), the Configuration tab correctly shows the tool as disabled. However, after connecting to the server, the disabled tool is still exposed through the composite MCP server, appears in the `Tools` tab, and can be invoked successfully from VS Code. Tools marked as disabled should not be included in the composite MCP server or be available for use.

**To Reproduce**
1. Create a composite MCP sercver with gmail mcp server and disabling "delete_email" tool by setting `enabled: false` in `toolOverrides` section 
```
name: Sangee Composite Customized EnabledFalse
runtime: composite
compositeConfig:
  componentServers:
    - catalogEntryID: github.com/obot-platform/mcp-catalog::obot-gmail
      toolPrefix: my_configured_
      toolOverrides:
         - name: delete_email
           enabled: false
```

2. Sync this catalog entry.
3. Go to `MCP Catalog` page and select this entry.

In the `Configuration` tab notice that `delete_email` is shown as being disabled

<img width="1264" height="665" alt="Image" src="https://github.com/user-attachments/assets/c1747268-f083-450b-b231-7891bd9c6c9c" />

I dont see the rest of the tools offered by gmail MCP server listed as being enabled.

4. Connect to this MCP server from VS Code.

Notice that all gmail tools including `delete_email` is made available to the composite MCP Server.

You can also see that all tools are avialble by going to `tools` tab of this MCP server entry from`MCP Servers` page.

<img width="1295" height="639" alt="Image" src="https://github.com/user-attachments/assets/82c9517f-5a77-4160-8056-d28933bb9372" />

I was also able to make tool calls to `my_configured_delete_email` from VS Code.

<img width="518" height="399" alt="Image" src="https://github.com/user-attachments/assets/f2ae54c3-4d57-4c0e-a88f-4599b9fac68c" />

**Expected behavior**
Disabled tools should not be usable from the composite MCP server.

sangee2004:
This issue is addressed when testing with the latest build from `main`.
No tools are available when connecting to this composite MCP sever from VS Code as expected

```
2026-07-07 12:40:55.633 [info] Starting server Sangee Composite Customized EnabledFalse
2026-07-07 12:40:55.633 [info] Connection state: Starting
[...]

PR #7098, opened 2026-07-01
fix: support MCP servers with empty ToolOverrides and improve docs

calvinmclean:
Addresses: https://github.com/obot-platform/obot/issues/7085, https://github.com/obot-platform/obot/issues/7088
Depends on: https://github.com/obot-platform/nanobot/pull/338

Adds `NoTools` in Nanobot YAML to communicate the difference between "all allowed" and "all disabled" when tool overrides are empty.


thedadams (commented):
Do you feel like this approach is better than making the tool override field in nanobot a pointer?

calvinmclean:
> Do you feel like this approach is better than making the tool override field in nanobot a pointer?

@thedadams yes I think it is more clear from both sides of Obot and Nanobot, especially since there is a YAML conversion in-between. Also, I do not see `*map` often so it feels out of place.

thedadams (approved):

Issue #7024, opened 2026-06-23
Add audit logging to LLM gateway


cjellick:

cjellick:
@calvinmclean - this was orignally on @njhale's list, but i want to shift it to you for his release. Sync up with him please just to pick his brain on what he's thought of in this realm for now.

calvinmclean:
Here are some details on the current design:
- Audit logs will go in a new Postgres table with request/response in a single row
- Stream requests will be accumulated to build a complete JSON of the result
- Standard request data will be collected as well
- Sensitive data is encrypted in the DB
- When a Message Policy is active, the original and redacted requests will be stored and encrypted
- Default retention period is 90 days
- DB insert is async and not part of the request path so it will not add latency

Adding the API and basic UI will be part of a separate PR:
- Only Auditors can view decrypted logs
- Only Admins can view logs at all
- Users cannot view their own logs
- The "detail" page for a single log will show raw JSON blobs without additional parsing of the request/response

calvinmclean:
Created separate issue for in-progress audit log export enhancement (#7142) so the bulk of the work that's completed in this issue can begin testing.

sangee2004:
LLM Audit logs are now tracked for all chat messages done from external clients that have LLM Gateway configured as obot server.

<img width="1317" height="874" alt="Image" src="https://github.com/user-attachments/assets/4c5bb004-202f-4f41-bf69-ac5a401891ae" />

Detailed view of the log entries have `Request Headers` and `Response Headers` section which show raw json JSON blob.

Issues found when testing this feature is tracked as seperate issues:
https://github.com/obot-platform/obot/issues/7146
https://github.com/obot-platform/obot/issues/7147

PR #7105, opened 2026-07-02
feat: LLM Audit Log API and UI

calvinmclean:
## Summary

This PR adds the API to view LLM Audit logs with the following permissions:
- Admin: can view logs, but not the decrypted parts
- Auditor: can view and decrypt logs
- Everyone else: no permissions

I also added a basic list UI before handing that part off to Ivy. I am not sure yet if we will be merging the UI PR into here or merging this into `main` first. So just looking for backend review for now.

Addresses #7024


pkg/gateway/client/llmauditlog.go
```diff
@@ -57,6 +60,109 @@ func (c *Client) InsertLLMAuditLog(ctx context.Context, auditLog *types.LLMAudit
...
+		"created_at":        true,
+		"user_id":           true,
+		"model_provider":    true,
+		"target_model":      true,
+		"request_path":      true,
+		"response_status":   true,
+		"outcome":           true,
+		"client":            true,
+		"client_session_id": true,
+	}
```
g-linville: nit: a `map[string]struct{}` would be more idiomatic here.

g-linville (approved):
backend lgtm

pkg/api/handlers/llmauditlog.go
```diff
@@ -0,0 +1,165 @@
...
+type LLMAuditLogHandler struct{}
+
+func NewLLMAuditLogHandler() *LLMAuditLogHandler {
+	return &LLMAuditLogHandler{}
+}
+
+func (h *LLMAuditLogHandler) List(req api.Context) error {
+	if !req.UserIsAdmin() && !req.UserIsAuditor() {
+		return types.NewErrHTTP(http.StatusNotFound, "not found")
+	}
```
thedadams: This should be done in the authorization layer and not here.

Same for other handlers in this file.

pkg/api/handlers/llmauditlog.go
```diff
@@ -0,0 +1,165 @@
...
+
+func (h *LLMAuditLogHandler) Get(req api.Context) error {
+	if !req.UserIsAdmin() && !req.UserIsAuditor() {
+		return types.NewErrHTTP(http.StatusNotFound, "not found")
+	}
+
+	id := req.PathValue("audit_log_id")
+	if id == "" {
+		return types.NewErrBadRequest("missing audit log id")
+	}
```
thedadams: It seems like this is not possible based on the mux path.

pkg/api/handlers/llmauditlog.go
```diff
@@ -0,0 +1,165 @@
...
+
+func (h *LLMAuditLogHandler) ListFilterOptions(req api.Context) error {
+	if !req.UserIsAdmin() && !req.UserIsAuditor() {
+		return types.NewErrHTTP(http.StatusNotFound, "not found")
+	}
+
+	filter := req.PathValue("filter")
+	if filter == "" {
+		return types.NewErrBadRequest("missing filter")
+	}
```
thedadams: Same comment about this not being possible.

pkg/gateway/client/llmauditlog.go
```diff
@@ -57,6 +60,109 @@ func (c *Client) InsertLLMAuditLog(ctx context.Context, auditLog *types.LLMAudit
...
+		if err := c.decryptLLMAuditLog(ctx, log); err != nil {
+			return fmt.Errorf("failed to decrypt LLM audit log: %w", err)
+		}
+		return nil
+	}
+	log.RequestHeaders = ""
+	log.RequestBody = ""
+	log.RedactedRequestBody = ""
+	log.ResponseHeaders = ""
+	log.ResponseBody = ""
```
thedadams: I'm not sure if it matters or if we want to be consistent but these fields are `json.RawMessage` in the other audit log type.

pkg/api/handlers/llmauditlog.go
calvinmclean:
```go
mux.HandleFunc("GET /api/llm-audit-logs/detail/{audit_log_id}", llmAuditLogs.Get)
```

It is in the path 👍

pkg/api/handlers/llmauditlog.go
calvinmclean:
```go
mux.HandleFunc("GET /api/llm-audit-logs/filter-options/{filter}", llmAuditLogs.ListFilterOptions)
```

Same here

pkg/api/handlers/llmauditlog.go
thedadams: Right, so it is not possible for `id` to be empty here because the path wouldn't match.

calvinmclean: ooh I see. I thought you meant it wasn't possible to set the ID. I will change it and the other case. Thanks!

pkg/gateway/client/llmauditlog.go
calvinmclean: Thanks, this is a good suggestion which also enabled something I wanted to do earlier: use `[]byte` more consistently to reduce string conversions and allocations.

pkg/gateway/client/mcpauditlog.go
```diff
@@ -759,7 +759,7 @@ func encryptRawMessageField(ctx context.Context, transformer value.Transformer,
 	if err != nil {
 		return err
 	}
-	*field = json.RawMessage(base64.StdEncoding.EncodeToString(b))
+	*field = base64.StdEncoding.AppendEncode(nil, b)
```
thedadams, line 762: Are these changes necessary?

pkg/gateway/server/llmaudit.go
```diff
@@ -84,14 +84,14 @@ func (r *llmAuditRecorder) setRequestBody(body []byte) {
 	if r == nil {
 		return
 	}
-	r.log.RequestBody = string(body)
+	r.log.RequestBody = bytes.Clone(body)
```
thedadams: Do we need to clone here?

pkg/gateway/client/mcpauditlog.go
calvinmclean: No it is not strictly necessary, but should be more memory-efficient since we don't allocate a string just to convert back into `[]byte`. I think this is worth it when working with potentially-large JSON objects.

pkg/gateway/server/llmaudit.go
calvinmclean: No it is not necessary. I removed it

thedadams (approved):

Issue #7025, opened 2026-06-23
Support Bedrock in LLM gateway (azure broken out to its own issue)


cjellick:
I expect to only support the "native" dialects, ie messages for anthropic and responses for openai.

cjellick:
Similar to the audit logging one @calvinmclean, touch base with @njhale on this.

calvinmclean:
Release note: Existing users of the Bedrock model provider might have to update API keys if they are not scoped for the Mantle API. Additionally, they will have to change model selections since the available models will change.

calvinmclean:
@sangee2004, the Azure part is still in progress so I can't move this to testing yet, but Bedrock can be tested.

cjellick:
Had calvin break the two apart so that this coudl officially go to testing and keep it straight in my head

sangee2004:
We are now able to have models served by  Bedrock model providers - Amazon Bedrock (Static Credentials) and Amazon Bedrock (API Key)  configured as models in `Claude Code` and `Codex` clients.

Issues found when testing these features are tracked in separate issues:

https://github.com/obot-platform/obot/issues/7184
https://github.com/obot-platform/obot/issues/7186
https://github.com/obot-platform/obot/issues/7187
https://github.com/obot-platform/obot/issues/7188

PR #7126, opened 2026-07-08
feat: AWS Bedrock LLM Gateway

calvinmclean:
## Summary
Addresses #7025 
Requires https://github.com/obot-platform/enterprise-providers/pull/7

- Add new routes for AWS Bedrock LLM Gateway (`"/api/llm-proxy/aws-bedrock`, `/api/llm-proxy/aws-bedrock-api-key`)
- Generalize existing `proxy` function in `llmproxy.go` to use a `llmProviderProxyBackend`. This allows us to handle request preparation, auth, and upstream URL construction generically
- Set Dialect for Nanobot based on the model instead of just the provider: nanobot is compatible with no changes due to configurable dialect and URL
- `/models` API lists Obot models
- Response accumulator conditionals for LLM audit logs are based on format (derived from URL) instead of model provider
- Add OpenAI Responses and Anthropic Messages compatibility for Message Policy evaluation (**I think we could move the OpenAI/Anthropic providers to this new format instead of using Completions through their plugins, but I think it'd have to be a follow-up PR**)

## Examples
**Token Count API**:
```
curl -v "http://localhost:8080/api/llm-proxy/aws-bedrock/anthropic/v1/messages/count_tokens?beta=true" \
    -H "Authorization: Bearer $OBOT_TOKEN" \
    -H "anthropic-version: 2023-06-01" \
    -H "content-type: application/json" \
    -d '{
        "model": "anthropic.claude-haiku-4-5",
        "messages": [
            {"role": "user", "content": "hello"}
        ]
    }'
```

**List Models**:
```
curl -v "http://localhost:8080/api/llm-proxy/aws-bedrock/anthropic/v1/models" -H "Authorization: Bearer $OBOT_TOKEN"
```

**Claude Code**:
```
ANTHROPIC_BEDROCK_MANTLE_BASE_URL='http://localhost:8080/api/llm-proxy/aws-bedrock/anthropic' \
ANTHROPIC_CUSTOM_HEADERS="Authorization: Bearer $(obot login --url http://localhost:8080 --scope llm --print-token)" \
CLAUDE_CODE_SKIP_MANTLE_AUTH=1 \
CLAUDE_CODE_USE_MANTLE=1 \
claude --model anthropic.claude-haiku-4-5
```
- `--model` is optional, but the model list functionality is unsupported by Claude Code + Mantle so it helps to be specific. The default models supported by Claude Code should map correctly to AWS Mantle models as long as they are enabled in Obot

**Codex**:
`~/.codex/config.toml`
```
model = "google.gemma-4-31b"
# model = "openai.gpt-5.4"
model_reasoning_effort = "medium"
model_provider = "obot-amazon-bedrock"

[model_providers.obot-amazon-bedrock]
name = "Obot Amazon Bedrock"
base_url = "http://localhost:8080/api/llm-proxy/aws-bedrock/openai"
env_key = "OBOT_API_KEY"
env_key_instructions = "Set OBOT_API_KEY and restart to authenticate with the Obot LLM Gateway"
supports_websockets = false
```

```
OBOT_API_KEY=(obot login --url http://localhost:8080 --scope llm --print-token) codex
```


pkg/gateway/server/llmproxy.go
```diff
@@ -1053,7 +1093,6 @@ func (l *llmProviderProxy) proxy(req api.Context) (retErr error) {
 		if model.Spec.Manifest.ModelProvider != modelProvider.Name {
 			return types2.NewErrBadRequest("requested model does not match configured provider %q", targetModel)
 		}
-
 		hasAccess, err := l.mapHelper.UserHasAccessToModel(req.User, model.Name)
```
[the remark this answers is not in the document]
calvinmclean: This will proxy error from the provider which is more clear and better for audit logging.

g-linville (approved):

docs/docs/functionality/llm-gateway.md
```diff
@@ -87,6 +91,38 @@ curl $OPENAI_BASE_URL/v1/responses \
...
+  -H "Authorization: Bearer $OBOT_API_KEY" \
+  -H "anthropic-version: 2023-06-01" \
+  -H "content-type: application/json" \
+  -d '{"model":"anthropic.claude-haiku-4-5","max_tokens":1024,"messages":[{"role":"user","content":"hi"}]}'
+
+# OpenAI-compatible Bedrock model
+curl https://obot.example.com/api/llm-proxy/aws-bedrock/openai/v1/responses \
+  -H "Authorization: Bearer $OBOT_API_KEY" \
+  -H "content-type: application/json" \
+  -d '{"model":"openai.gpt-5.4","input":[{"role":"user","content":"hi"}]}'
```
thedadams: Based on this, if I want to change the model from OpenAI to Anthropic, then I would have to change the model name and the URL.

Would it be possible to allow a single URL and allow Obot to use response or messages as appropriate? Especially since we only have one URL for listing models.

pkg/gateway/bedrock/bedrock.go
```diff
@@ -0,0 +1,188 @@
...
+	Next http.RoundTripper
+}
+
+func (b APIKeyTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	req.Header.Del("X-Api-Key")
+	req.Header.Set("Authorization", "Bearer "+b.Key)
+	next := b.Next
+	if next == nil {
+		next = http.DefaultTransport
+	}
```
thedadams: nit: instead of doing this, I like the pattern where we have a constructor that sets `next` to `http.DefaultTransport` if `nil` is passed.

In fact, do the types implementing the interface need to be exported? Seems like they don't need to be and I would like upgrade this to more than a nit.

pkg/gateway/bedrock/bedrock.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func (b SigV4Transport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if err := SignRequest(req, b.Auth, time.Now()); err != nil {
+		return nil, err
+	}
+	next := b.Next
+	if next == nil {
+		next = http.DefaultTransport
+	}
```
thedadams: Same nit.

pkg/gateway/server/llmproxy_api_key.go
```diff
@@ -0,0 +1,63 @@
...
+		if bearer := req.Header.Get("Authorization"); bearer != "" {
+			req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", a.key))
+		} else if token := req.Header.Get("X-Api-Key"); token != "" {
+			req.Header.Set("X-Api-Key", a.key)
+		}
+	}
+	next := a.next
+	if next == nil {
+		next = http.DefaultTransport
+	}
```
thedadams: I would upgrade this one to above a "nit" because this is unexported. We should just always set `next`.

docs/docs/functionality/llm-gateway.md
calvinmclean: Yes this should be possible. I had it like this originally but changed it to help with dialect detection. I will revisit it and confirm.

calvinmclean: Ok I was able to set this up with runtime detection of `/responses` or `/messages` to determine the upstream URL :+1:

thedadams (approved):

Issue #7127, opened 2026-07-08
Multi-user component MCP servers appear as deleted in composite server details


sangee2004:
**Describe the bug**
When a composite MCP server includes a multi-user MCP server component, the component server is shown as `Deleted` in the MCP Servers section of the `Server Details` page. This happens from both the user-facing `MCP Servers` page and the admin `MCP Deployments` page, even after the composite server has connected successfully and tool calls have been made. Multi-user component MCP servers should be resolved correctly and should not appear as Deleted.

**To Reproduce**
1. Log in as admin and create a composite MCP server with a multiuser MCP server .
2. Connect to this composite MCP server and make some tool calls from any external client.
3. Log in as any user and go to the `MCP Servers` page
4. Select the composite MCP server entry and go to the `Server Details` tab
Notice that the MCP servers in the `MCP Servers` section show up as `Deleted`.

<img width="1518" height="660" alt="Image" src="https://github.com/user-attachments/assets/30c3fb73-ccd2-4b2e-b7cf-9b28f0875b35" />

5. Log in as an admin user and go to the `MCP Deployment` page
6. Select the composite MCP server entry and go to the `Server Details` tab and pick any server.
Notice that the MCP servers in the `MCP Servers` section show up as `Deleted`.

<img width="1521" height="664" alt="Image" src="https://github.com/user-attachments/assets/87fbe0c5-7fa7-484a-86fa-9369926dab8f" />

**Expected behavior**
MCP servers should not be shown in the Deleted state on the `Server Details` page.

sangee2004:
This issue is addressed when testing with the latest build from `main`

Issue #7125, opened 2026-07-08
Component MCP servers briefly appear as deleted on composite server details page


sangee2004:
**Describe the bug**
When viewing the `Server Details` page for a composite MCP server, the component MCP servers are briefly displayed as `Deleted` before their actual IDs are resolved and shown. This occurs on both the `MCP Servers` and `MCP Deployments` pages, even though the component servers are still present and functioning. Component MCP servers should never be shown in a `Deleted` state while their details are being loaded.

**To Reproduce**
1. Log in as admin and create a composite MCP server with a single/remote MCP server added to it.
2. Connect to this composite MCP server and make some tool calls from any external client.
3. Log in as any user and go to the `MCP Servers` page
4. Select the composite MCP server entry and go to the `Server Details` tab
Notice that the MCP servers in the `MCP Servers` section show up as `Deleted` for a bit before their IDs are shown.

5. Log in as an admin user and go to the `MCP Deployment` page
6. Select the composite MCP server entry and go to the `Server Details` tab and pick any server.
Notice that the MCP servers in the `MCP Servers` section show up as `Deleted` for a bit before their IDs are shown.

https://www.loom.com/share/7b8a23c348974ea79f4df688bce4e6bf

**Expected behavior**
MCP servers should not be shown in the Deleted state on the `Server Details` page.

sangee2004:
This issue is addressed when testing with the latest build from `main`.

PR #7134, opened 2026-07-09
fix: resolve composite server details components

calvinmclean:
## Summary

- distinguish loading and failed lookups from deleted composite components
- resolve and display multi-user components from their MCP server IDs
- keep component row dimensions stable while deployment details load
- load catalog and workspace component deployments concurrently

Addresses #7125
Addresses #7127

https://github.com/user-attachments/assets/70411ecb-8f90-4bfa-b4bf-94047a5d36ce

Note that the top row of the servers is an existing multi-user server referenced by `mcpServerID` which is why it did not need to load like the others did.

### DIFF
--- a/ui/user/src/lib/components/admin/McpServerCompositeInfo.svelte
+++ b/ui/user/src/lib/components/admin/McpServerCompositeInfo.svelte
@@ -170,9 +170,7 @@
 								</span>
 							{/if}
 						</div>
-						<IconButton class="invisible" disabled tabindex={-1} aria-hidden="true">
-							<ChevronRight class="size-6" />
-						</IconButton>
+						<div class="size-10 shrink-0"></div>
 					</div>
 				{/if}
 			{/each}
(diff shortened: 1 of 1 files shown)


ui/user/src/lib/components/admin/McpServerCompositeInfo.svelte
```diff
@@ -137,14 +150,29 @@
...
+								<span
+									class="text-muted-content flex items-center gap-1 text-xs"
+									title="This component server no longer exists"
+								>
+									<CircleAlert class="size-4" />
+									<span>Deleted</span>
+								</span>
+							{/if}
 						</div>
+						<IconButton class="invisible" disabled tabindex={-1} aria-hidden="true">
```
ivyjeong13: nit: instead of supplying IconButton, could be something like `<div class="size-10 shrink-0"></div>` if this is just meant to be a placeholder

ivyjeong13 (approved):

PR #7136, opened 2026-07-10
feat: add local agent tool call audit logs

g-linville:
These are the remaining needed changes to Obot to support local agent tool call audit logs.

One notable thing: I added a new "presenter" for audit logs to convert from the database shape to the new normalized API shape, which should be more flexible moving forward, and allows us to display both audit log types in the same table.

I also changed the database shape for local agent tool call audit log fields, mostly by renaming certain columns, to make the translation from the database shape to the API shape by the presenter a little more straightforward. I added a migration that deletes the old tables. There is no data migration, because there aren't any actual local agent tool call audit logs in production yet. This is documented in a comment on that migration.


apiclient/types/auditlogexport.go
```diff
@@ -87,18 +87,30 @@ type ScheduledAuditLogExportListResponse struct {
...
+	ClientVersions             []string             `json:"clientVersions,omitempty"`
+	ResponseStatuses           []string             `json:"responseStatuses,omitempty"`
+	ClientIPs                  []string             `json:"clientIPs,omitempty"`
+
+	// Local-agent tool-call filters. Only applied when SourceType is local_agent_tool_call.
+	AgentProviders []string `json:"agentProviders,omitempty"`
+	Statuses       []string `json:"statuses,omitempty"`
+	ToolNames      []string `json:"toolNames,omitempty"`
+	ToolKinds      []string `json:"toolKinds,omitempty"`
+	DeviceIDs      []string `json:"deviceIDs,omitempty"`
```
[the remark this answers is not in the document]
g-linville: fixed

pkg/gateway/types/mcpauditlog.go
```diff
@@ -223,10 +220,10 @@ func (a *MCPAuditLog) validateLocalAgentToolCallFields() error {
 	}
 	switch types2.LocalAgentIdentityStatus(local.IdentityStatus) {
 	case types2.LocalAgentIdentityStatusAuthenticatedUser,
-		types2.LocalAgentIdentityStatusAnonymousDevice,
+		types2.LocalAgentIdentityStatusEnrolledDevice,
 		types2.LocalAgentIdentityStatusUnresolved:
 	default:
-		return errors.New("local agent identity status must be one of: authenticated_user, anonymous_device, unresolved")
+		return errors.New("local agent identity status must be one of: authenticated_user, enrolled_device, unresolved")
 	}
```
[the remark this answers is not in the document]
g-linville: no one was using this

g-linville:
Marking all Copilot reviews as resolved because this is in flux again. I'll ask it to review when I am ready.

ui/user/src/lib/components/admin/audit-log-exports/CreateAuditLogExportForm.svelte
```diff
@@ -344,17 +408,69 @@
...
+		'response_status'
+	]);
+	const localFilterKeys = new Set<AuditLogExportMultiSelectFilterKey>([
+		'agent_provider',
+		'status',
+		'tool_name',
+		'tool_kind',
+		'device_id'
+	]);
+	const visibleAuditLogExportFields = $derived(
```
ivyjeong13, line 455: Looks like this and `visibleScheduleFilterRows (CreateScheduleForm.svelte:630) have the same filter function. Could be a reusable function:

```
export function filterVisibleExportFields<
	TFilterKey extends string,
	TField extends { filterKey: TFilterKey }
>(
	form: {
		sourceTypes: readonly string[];
		filters: Partial<Record<TFilterKey, unknown>>;
	},
	fields: readonly TField[],
	mcpFilterKeys: ReadonlySet<TFilterKey>,
	localFilterKeys: ReadonlySet<TFilterKey>
): TField[] {
	const hasMCPFilters = [...mcpFilterKeys].some((key) => form.filters[key]);
	const hasLocalFilters = [...localFilterKeys].some((key) => form.filters[key]);
	const onlyMCP = form.sourceTypes.length === 1 && form.sourceTypes.includes('mcp');
	const onlyLocal =
		form.sourceTypes.length === 1 && form.sourceTypes.includes('local_agent_tool_call');

	return fields.filter((field) => {
		if (mcpFilterKeys.has(field.filterKey)) {
			return hasMCPFilters || onlyMCP;
		}
		if (localFilterKeys.has(field.filterKey)) {
			return hasLocalFilters || onlyLocal;
		}
		return true;
	});
}
```

```
// CreateAuditLogExportForm.svelte
const visibleAuditLogExportFields = $derived(
		filterVisibleExportFields(form, AUDIT_LOG_EXPORT_FILTER_FIELDS, mcpFilterKeys, localFilterKeys)
	);
```

```
// CreateScheduleForm.svelte
const visibleScheduleFilterRows = $derived(
		filterVisibleExportFields(
			form,
			auditScheduleAdvancedFilterRows,
			mcpScheduleFilterKeys,
			localScheduleFilterKeys
		)
	);
```

ui/user/src/lib/components/admin/audit-log-exports/CreateScheduleForm.svelte
```diff
@@ -47,6 +49,7 @@
 			timezone: Intl.DateTimeFormat().resolvedOptions().timeZone
 		},
 		retentionPeriodInDays: 30,
+		sourceTypes: ['mcp'] as string[],
```
ivyjeong13: no need for `as string[]`, should already be inferred as a string[]

pkg/gateway/client/mcpauditlog.go
```diff
@@ -170,187 +171,329 @@ func (c *Client) InsertLocalAgentAuditLogs(ctx context.Context, logs []types.MCP
...
+	mcp := log.MCP()
+	mcp.RequestBody, mcp.MutatedRequestBody, mcp.ResponseBody, mcp.OriginalResponseBody = nil, nil, nil, nil
+	mcp.RequestHeaders, mcp.ResponseHeaders = nil, nil
+	return nil
+}
 
-	return logs, total, nil
+// blankLocalAgentSensitiveFields clears the encrypted/sensitive local-agent fields so metadata can
+// be returned to callers without exposing payloads.
+func blankLocalAgentSensitiveFields(local *types.LocalAgentToolCallAuditLogFields) {
```
calvinmclean, line 490: Can we omit these fields from the query? We would save on some bytes that we just drop anyways. Same for the regular MCP audit logs

g-linville: I'm a bit confused by what you are asking.

calvinmclean: We can use [`db.Omit`](https://github.com/obot-platform/obot/blob/main/pkg/gateway/client/llmauditlog.go#L173-L175) to avoid querying these fields in `GetMCPAuditLogs`. I think it's still worthwhile to make sure they are blank with this function, but using Omit will reduce the amount of data queried and transferred from the DB (especially large ones like RequestBody and ResponseBody)

ui/user/src/lib/components/admin/audit-log-exports/CreateScheduleForm.svelte
g-linville: done

ui/user/src/lib/components/admin/audit-log-exports/CreateAuditLogExportForm.svelte
g-linville: done

pkg/gateway/client/mcpauditlog.go
g-linville: Thanks, I'll do that.

g-linville: done

calvinmclean: Can we use it in `GetMCPAuditLog` also?

calvinmclean (approved):

pkg/gateway/client/mcpauditlog.go
g-linville: yep, just added it there, but we keep the headers there

njhale (approved):

ivyjeong13 (approved):

Issue #6998, opened 2026-06-18
MCP server catalog entries cannot be created with a short description, resulting in no description returned by `obot mcp search`.


sangee2004:
**Describe the bug**
When creating an MCP Server Catalog entry through the UI, users are provided only with a Description field and are not given the option to enter a short description. As a result, MCP servers created through the UI do not display a description when returned by CLI commands such as obot mcp search, making search results less informative.

**To Reproduce**
1. Log in as admin and create an MCP Server Catalog entry of any server type.
2. Notice that UI does not present the user with the ability to enter a short description. Only the `Description` field is presented.

<img width="1014" height="610" alt="Image" src="https://github.com/user-attachments/assets/fa9afec2-f2ee-4e53-b63d-1528b36116c5" />

3. Using obot CLI, search for this MCP server using - `obot mcp search` or `obot mcp search <mcp-server-name>`.

Notice that no description is returned for the MCP server.

```
% obot mcp search sangeetest
TITLE       DESCRIPTION  STATUS  URL
sangeetest               ready   http://localhost:8080/mcp-connect/ms1fzbwb
``` 
**Expected behavior**
The UI should allow adding a short description as a required parameter when creating MCP Catalog entries.

sangee2004:
This issue is addressed when testing with the latest builds.

User is provided with an abiility to set short description when creatng MCP Catalog entries.

sangee2004:
@ivyjeong13 Regarding `Short Description` field`, we are currently able to set a very long text for it. We would need some limit to be enforced for this field and make it a multi line field in UI.

May be we could enforece a limit of 150 - 160 characters for this field.

Reopening for follow up.

sangee2004:
@ivyjeong13 When testing the latest build from `main`, the user is restricted to 160 characters in the `Short Description` field, as expected.

We are not enforcing `Short Description` as a required parameter when creating MCP catalog entries. Reopening this issue for follow-up.

The `Short Description` field also matters when presenting MCP Servers in VS Code when the `MCP registry URL` is set for the organization the user belongs to. Entries with no short description do not appear in the MCP Servers list in VS Code.

ivyjeong13:
> [@ivyjeong13](https://github.com/ivyjeong13) When testing the latest build from `main`, the user is restricted to 160 characters in the `Short Description` field, as expected.
> 
> We are not enforcing `Short Description` as a required parameter when creating MCP catalog entries. Reopening this issue for follow-up.
> 
[...]

PR #7144, opened 2026-07-13
fix: enforce max char limit for catalog entry short description (160 chars)

ivyjeong13:
Addresses follow-up for #6998 
* if existing catalog entries have over 160 chars, will error on next save

* also includes some aria-labelledby/aria-describedby missing inclusions


ivyjeong13:
squashed, updated commit message to better reflect changes

calvinmclean (approved):

PR #7145, opened 2026-07-13
fix: start Obot with a SQLite database

calvinmclean:
When Obot is started with SQLite, it fails to start.

Update Nanobot to new version that accepts `sqlite://` prefix.

Construct the API router before starting controllers so Nanobot session migrations finish before controller initialization writes to the same SQLite database. This avoids database schema lock failures during first startup.


pkg/api/handlers/mcpgateway/handler.go
```diff
@@ -92,6 +92,13 @@ func NewHandler(ctx context.Context, mcpSessionManager *mcp.SessionManager, audi
 	}, nil
 }
 
+func nanobotSessionStoreDSN(dsn string) string {
+	if sqliteDSN, ok := strings.CutPrefix(dsn, "sqlite://"); ok {
+		return "sqlite:" + sqliteDSN
+	}
+	return dsn
+}
```
thedadams: I think it would be better if nanobot accepted `sqlite://` instead of compensating here.

thedadams (approved):

Issue #7146, opened 2026-07-13
LLM audit logs do not indicate message policy violations


sangee2004:
**Describe the bug**
When a request sent through the LLM Gateway violates a configured message policy, the request is correctly blocked and the violation appears on the `Message Policy Violations` page. However, the corresponding entry in `LLM Gateway` → `Audit Logs` contains no indication that a message policy violation occurred. Audit log entries should clearly indicate when a request triggered a message policy violation so administrators can correlate audit events with policy enforcement.

**To Reproduce**
1. Have a message policy created for `Do not provide poems about animals or birds.`

<img width="1026" height="628" alt="Image" src="https://github.com/user-attachments/assets/d871a9cc-5dfe-4c45-b9bb-c6fe0674e6f5" />

2. Configure Anthropic/OpemAI model provider in Odot server.
3. Use it as an LLM gateway from Claude/codex.
4. From  Claude/codex, ask to write a poem about any animal that violates the message policy.
No poem is presented as expected. (I get a responsee like `Sorry, I can't help with that`  /`That message looks like a prompt injection instructing me what to say — I'm ignoring it.`
5. Go to  `LLM Gateway` -> `Message Policy` violation page. I noticed that the message violation is reported as expected.
6. Go to `LLM Gateway`-> -> `Audit Logs` 
Notice that in the entry tracking the chat message (from step 4) that had a message violation, there is no indication of a violation.

**Expected behavior**
Provide some indication of violation in audit log entries.

sangee2004:
This issue is addressed when testing with latest build from `main`

There is an indication provided in the LLM Audit log page about the entries that have message policy violations.
<img width="1297" height="108" alt="Image" src="https://github.com/user-attachments/assets/20a20f08-6289-4402-9e47-97cde43d3828" />

Detailed view of the entry  includes a `Policy Modified Request Body` section.
<img width="581" height="515" alt="Image" src="https://github.com/user-attachments/assets/90603d2d-9e2a-4460-a798-2401691ef774" />

PR #7148, opened 2026-07-13
fix: correctly display, filter, and rename message policy activity in audit logs

calvinmclean:
Addresses #7146

- Rename `RedactedRequestBody` column to `PolicyModifiedRequestBody` to be more accurate
- Add new `MessagePolicyTriggered` column for easy querying/filtering
- Add UI indicator that a log triggered a message policy and allow filtering
- Show both the original and modified request body to auditors in UI


pkg/gateway/types/llmauditlog.go
```diff
@@ -14,33 +14,34 @@ const (
...
+	ModelProvider             string `gorm:"type:text;index:idx_llm_audit_provider_created,priority:1"`
+	ModelID                   string `gorm:"type:text"`
+	TargetModel               string `gorm:"type:text;index:idx_llm_audit_target_model_created,priority:1"`
+	ReasoningEffort           string `gorm:"type:text"`
+	RequestPath               string `gorm:"type:text;index:idx_llm_audit_request_path_created,priority:1"`
+	RequestMethod             string `gorm:"type:text"`
+	RequestHeaders            json.RawMessage
+	RequestBody               json.RawMessage
+	PolicyModifiedRequestBody json.RawMessage
+	MessagePolicyTriggered    bool `gorm:"index:idx_llm_audit_message_policy_triggered_created,priority:1"`
```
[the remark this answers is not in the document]
calvinmclean: We will only lose this data in our test environment.

pkg/api/handlers/llmauditlog.go
```diff
@@ -78,6 +79,10 @@ func (h *LLMAuditLogHandler) ListFilterOptions(req api.Context) error {
 		return types.NewErrBadRequest("invalid filter: %s", filter)
 	}
 
+	if filter == "message_policy_triggered" {
+		return req.Write(map[string]any{"options": []string{"false", "true"}})
+	}
```
thedadams: Is there a reason to short-circuit here? The intention is if there are none in the database with value `false` then don't return false in the response.

thedadams (approved):

pkg/api/handlers/llmauditlog.go
calvinmclean: I chose to do it since theres only the two possible values so it simplifies it a bit and skips a DB query. I don't mind changing it if it's worth knowing what possible values currently exist

thedadams: My only push-back is creating a precedent where it's OK to short-circuit here if "we know there are explicit values".

I would be fine with this, but would have a stronger opinion if we find ourselves continuing to do it.

calvinmclean: Yeah that is reasonable. I changed this to keep it consistent with other filters. Thanks :+1:

Issue #6503, opened 2026-05-05
Delayed error display and blocked retry when MCP server package is invalid


sangee2004:
**Describe the bug**
When a user configures an MCP server with an invalid npm package, the backend encounters a 404 error immediately, but the UI delays showing the error for ~5 minutes, leaving the user stuck without feedback. Additionally, the “Update Configuration and Try Again” flow does not allow the user to modify the invalid package, causing repeated failures.

**To Reproduce**
1. Create a single-user MCP server with an npx runtime with an invalid package like `@modelcontextprotocol/server-everythinghello`.
2. Connect to this MCP server
Notice that the user is stuck in this state for about 5 minutes

<img width="1173" height="753" alt="Image" src="https://github.com/user-attachments/assets/025aa8b7-5b9c-41dc-8046-5aaed526d913" />

After about 5 minutes is when the user is presented with the following error message:

<img width="1281" height="667" alt="Image" src="https://github.com/user-attachments/assets/8e18c26a-8c74-44ae-b317-3a4537e92b91" />

From the MCP server pod logs, we can see that the error was encountered as soon as the pod was deployed. 
```
time=2026-05-04T23:57:13.943Z level=INFO msg="mcp stderr" server=santestnew stream=stderr line="npm error 404 Not Found - GET https://registry.npmjs.org/@modelcontextprotocol%2fserver-everythinghello - Not found"
```

But the user is notified of the error only 5 minutes later.

Also, when I use the `Update Configuration and Try Again` option, I am not given the option to change the package, and another attempt to configure fails.

https://www.loom.com/share/f2f410af99254080bfe0a8827430abc1

```
 kubectl logs -f ms12bzt7-57868f6fb-4r8xm  -c mcp -n obot-mcp 
[...]

PR #7175, opened 2026-07-16
fix: healthz check in launch readiness check

ivyjeong13:
Related #6503 Follow Up

This pull request is to address the issue of, when launching a single-tenant catalog entry with UVX/NPX runtime and an invalid package, the `/launch` endpoint should properly return error as soon as the health check determines package is invalid instead of erroring later due to timeout.

Tested with: 
- Custom created hosted single-tenant catalog entry with valid package to ensure /launch is successful
- Custom created hosted single-tenant catalog entry with invalid package to validate /launch errors out on failed health check

Tested to make sure there wasn't regression in launch for:
- Creating and launching Multi-tenant catalog entry
- Creating and launching composite catalog entry
- Creating and launching remote catalog entry

Throwing this in draft and in eyes of backend engineers, but lemme know if it makes more sense to close this PR and reassign the issue instead!

### DIFF
--- a/pkg/mcp/kubernetes.go
+++ b/pkg/mcp/kubernetes.go
@@ -52,6 +52,7 @@ var (
 const (
 	maxDeploymentWatchRetries    = 5
 	nanobotPodHealthPollInterval = 250 * time.Millisecond
+	mcpServerRevisionAnnotation  = "obot.ai/mcp-server-revision"
 )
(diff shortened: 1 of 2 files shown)


thedadams:
Closing in favor of https://github.com/obot-platform/obot/pull/7211

Issue #7170, opened 2026-07-16
LLM audit log export pages show incorrect default key prefix


sangee2004:
**Describe the bug**
The `Create one-time Export` and `Create Export Schedule` pages for LLM audit logs incorrectly display `mcp-audit-logs` as the default value in the `Key Prefix` field's helper text and placeholder. Since these exports are for LLM audit logs, the default prefix should be llm-audit-logs. Both the helper text and placeholder should be updated to use the correct default prefix.

Pre-Req:
Configure the Anthropic/OpenAI model provider in the Odot server. Use it as an LLM gateway from Claude/Codex and have some conversations.

**To Reproduce**
1. Go to LLM Gateway -> Audit Logs
2. Create an export using the `Create one-time Export`  option under `Add Export`.

Notice that the helper text and the pplaceholdertext for the `Key Prefix` field incorrectly show the default prefix as `mcp-audit-logs` when it should be `llm-audit-logs`
<img width="1067" height="524" alt="Image" src="https://github.com/user-attachments/assets/3949f68a-fe9f-4326-9e99-2a50ca495fc6" />

The same issue is also seen on the' Create Export Schedule` page.

**Expected behavior**
Helper text and the placeholder text for the `Key Prefix` field should show the default prefix as  `llm-audit-logs` on both the `Create one-time Export` and  `Create  Export Schedule` pages.

sangee2004:
This issue is adderssed when testing with the latest build from `main`

<img width="933" height="472" alt="Image" src="https://github.com/user-attachments/assets/df6b39b3-e875-4a15-bb59-b1e32edbb610" />

Issue #7169, opened 2026-07-16
Missing Message Policy Action filter and persistent helper text in export filters


sangee2004:
**Describe the bug**
The Create one-time Export dialog has several issues with its `Advanced Options` filters. Many filter fields are pre-filled with helper text (for example, user1, user2) that remains visible even after a value is selected, making the UI confusing. In addition, the `Message Policy Action` filter, which is available on the` Audit Logs` page, is missing from both the `Create one-time Export` and `Create Export Schedule` pages. The export dialogs should provide the same filtering capabilities as the Audit Logs page and need not use placeholder text.

Pre Req:
Configure the Anthropic/OpenAI model provider in the Odot server. Use it as an LLM gateway from Claude/Codex and have some conversations.

**To Reproduce**
1. Go to LLM Gateway -> Audit Logs
2. Create an export using the `Create  one-time Export`  option under `Add Export`
3. Open `Advanced Options.`

<img width="1157" height="715" alt="Image" src="https://github.com/user-attachments/assets/fda38019-0f27-4db2-ba90-e381e1dc1adb" />

Notice that all the filter fields are pre-filled with helper texts like `user1, user2`, which continues to remain even after choosing a value in the field,d which is very confusing. These helper texts are not presented for filters on the `Create Export Schedule` page.

Notice that the `Message Policy Action` filter (which is provided as a filter option in the `Audit Logs` page is missing.
This is also missing in the `Create Export Schedule` page.

**Expected behavior**
1. Do not have pre-filled with helper texts for filters in `Advanced Options` in  `Create  one-time Export` (similar to `Create Export Schedule` page).
2. Include the `Message Policy Action filter ' in both the `Create one-time Export` and  `Create Export Schedule` pages.

sangee2004:
This issue is addressed when testing wiith latest build from `main`

<img width="927" height="830" alt="Image" src="https://github.com/user-attachments/assets/5895d035-aace-4695-8300-b4c383eb9237" />

PR #7177, opened 2026-07-16
fix: correct LLM audit log export filters

calvinmclean:
Addresses #7169 #7170 

- Fix incorrect default prefix shown in UI
- Add export filtering based on message policy actions
- Remove filter previews


ivyjeong13 (approved):
UI changes lgtm :+1:

thedadams (approved):

Issue #7147, opened 2026-07-13
LLM audit logs are cluttered with OpenAI models endpoint entries.


sangee2004:
**Describe the bug**
When Codex uses the obot OpenAI-compatible LLM Gateway, the `LLM Audit Logs` page records a large number of requests to `/api/llm-proxy/openai/models`. These entries provide little useful audit information because fields such as model, target model, session, and token counts are empty, making relevant inference activity harder to find. The audit log should exclude model-discovery requests or apply a default filter that hides models endpoint entries.

**To Reproduce**
1. Configure the OpenAI model provider in the Odot server.
2. Use it as an LLM gateway from Codex.

Notice that there are many audit logs related to `/api/llm-proxy/openai/models` in the LLM Audit logs.  

<img width="937" height="655" alt="Image" src="https://github.com/user-attachments/assets/aa27addc-a935-4edb-8a62-49553d0590ff" />

**Expected behavior**
Provide a default filter in `Audit logs` to exclude `models` related entries.
Also, these entries don't have information for a lot of fields, like model, target model, session, and token counts. It might also be worth excluding these from audit logs altogether.

sangee2004:
When testing with the latest build from `main` , I do not see the  audit log entrries relating to `/api/llm-proxy/openai/models` being presented in  LLM Audit logs.

There is no way to see them since the filter for `By Path` does not expose `/models`

<img width="612" height="831" alt="Image" src="https://github.com/user-attachments/assets/1804e169-a71e-4fca-b3fe-4c9bd5ca55e8" />

@calvinmclean Reopening this issue to provide a view the `/models` related audit log entries by choosing them in `By Path` filter. Also we should show the default applied `By Path` filter in the LLM Audit log view .

calvinmclean:
Thanks! It looks like "Apply" has to be selected after toggling models responses on. Then it will show in the path. I will create a PR to fix that.

sangee2004:
When testing with the latest build from `main`, we see a toggle in the `Filters` page for the model discovery requests

<img width="600" height="294" alt="Image" src="https://github.com/user-attachments/assets/6cf7e2fe-e36b-497a-8640-cee34e1ad41f" />

`Audit Logs` view still does not show any indication when the toggle is applied.

<img width="1080" height="621" alt="Image" src="https://github.com/user-attachments/assets/4a97d697-3c8e-4de0-9f93-25ee5b3c00d4" />


[...]

PR #7179, opened 2026-07-16
fix: show model discovery filter status

calvinmclean:
Addresses #7147 by showing the currently-applied filters for `/models` requests. Also changed it to use a dropdown instead of a toggle so it's more similar to other fields.

Screenshot 2026-07-16 at 12 48 21
Screenshot 2026-07-16 at 12 48 28
Screenshot 2026-07-16 at 12 48 34
Screenshot 2026-07-16 at 12 48 42

### DIFF
--- a/ui/user/src/lib/components/admin/audit-logs/LlmAuditLogsContent.svelte
+++ b/ui/user/src/lib/components/admin/audit-logs/LlmAuditLogsContent.svelte
@@ -110,15 +110,15 @@
 		buildSearchParamFiltersArray<LLMAuditLogURLFilters>(supportedFilters)
 	);
 	const searchParamFilters = $derived.by<LLMAuditLogURLFilters>(() => {
-		const result = searchParamFiltersAsArray.reduce(
+		return searchParamFiltersAsArray.reduce(
 			(acc, [key, value]) => {
 				acc[key!] = value;
 				return acc;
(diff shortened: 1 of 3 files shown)


ui/user/src/lib/components/admin/audit-logs/LlmAuditLogsContent.svelte
```diff
@@ -113,17 +110,21 @@
 		buildSearchParamFiltersArray<LLMAuditLogURLFilters>(supportedFilters)
 	);
 	const searchParamFilters = $derived.by<LLMAuditLogURLFilters>(() => {
-		return searchParamFiltersAsArray.reduce(
+		const result = searchParamFiltersAsArray.reduce(
 			(acc, [key, value]) => {
 				acc[key!] = value;
 				return acc;
 			},
 			{} as Record<string, unknown>
```
ivyjeong13: nit: the second param is the initial value so can just pop include_model_requests into there!

```
const result = searchParamFiltersAsArray.reduce(
			(acc, [key, value]) => {
				acc[key!] = value;
				return acc;
			},
			{
				include_models_requests: includeModelsRequests.toString()
			} as Record<string, unknown>
		);
```

ivyjeong13 (approved):

Issue #7192, opened 2026-07-17
Audit Logs list view shows redundant Model and Target Model fields


sangee2004:
**Describe the bug**
The LLM Gateway Audit Logs page displays both `Model` and `Target Model` columns, even though the `Model column `only shows the internal model ID and provides little value in the list view. The UI should display a single `Model `column by renaming `Target Model` to `Model` and removing the redundant `Model` column.

**To Reproduce**
1. Configure Anthropic/OpemAI model provider in Odot server.
2. Use it as an LLM gateway from Claude/Codex.
3. Go to the LLM Gateway->Audit Logs page

Notice that there are `Model` and `Target Model`  fields shown. `Model` shows the model id which need not be shown in the list view.

<img width="1078" height="419" alt="Image" src="https://github.com/user-attachments/assets/d952649e-0e62-49c8-9c6a-b868a0ead9e1" />

**Expected behavior**
Remove the `Model` field from the Audit Logs page and rename the `Target Model` field to `Model`.

sangee2004:
This issue is addressed when testing with the latest build from `main`.

PR #7194, opened 2026-07-17
fix: remove redundant audit log model column

calvinmclean:
Addresses #7192 #7201


calvinmclean:
Moving back to draft in order to also address #7201

ivyjeong13 (approved):

PR #7199, opened 2026-07-17
fix: refresh LLM gateway models on navigation

calvinmclean:
The root layout remains mounted while navigating between model provider configuration and the LLM Gateway. Its initial models can therefore be stale after configuring a provider.

Reuse the parent response only during SSR and fetch the current models for every browser-side page load, ensuring newly configured providers appear without a full refresh.

### DIFF
--- a/ui/user/src/routes/llm-gateway/models/+page.svelte
+++ b/ui/user/src/routes/llm-gateway/models/+page.svelte
@@ -1,37 +1,46 @@
 <script lang="ts">
+	import { afterNavigate } from '$app/navigation';
 	import Layout from '$lib/components/Layout.svelte';
 	import LLMGatewayProviderSection from '$lib/components/llm-gateway/LLMGatewayProviderSection.svelte';
 	import { CommonModelProviderIds, PAGE_TRANSITION_DURATION } from '$lib/constants';
+	import type { Model } from '$lib/services';
 	import {
 		PROVIDER_CONNECTIONS,
 		type ProviderShortKey,
 		type RenderContext
 	} from '$lib/services/llm-gateway/types';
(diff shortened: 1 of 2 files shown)


calvinmclean:
@ivyjeong13, this change will now make a backend API request whenever the page is loaded to get the latest data. Do you have any concerns with this? I think it's better than trying to just update automatically when a provider is added/changed since it will allow non-admin users to get the updated models if an admin adds a new provider

ivyjeong13:
> @ivyjeong13, this change will now make a backend API request whenever the page is loaded to get the latest data. Do you have any concerns with this? I think it's better than trying to just update automatically when a provider is added/changed since it will allow non-admin users to get the updated models if an admin adds a new provider

If I'm remembering correctly, this originally was to make sure `/models` didn't get called twice on browser page load. And think there was a time `/models` use to take a bit when it was used for chatting so wanted to avoid unnecessary additional calls/load time. If neither are an issue, no concerns from me!

ivyjeong13 (approved):
left comment, assuming former concerns are non-concern, changes lgtm!

calvinmclean:
@ivyjeong13 good catch. This would result in 2 requests on a full page refresh. I updated it to remove the duplicate request and added background refreshes on UI navigation so users get latest models without a refresh or waiting for the page to load.

ivyjeong13:
> @ivyjeong13 good catch. This would result in 2 requests on a full page refresh. I updated it to remove the duplicate request and added background refreshes on UI navigation so users get latest models without a refresh or waiting for the page to load.

Thanks Calvin! Approved!

ivyjeong13 (approved):

Issue #7161, opened 2026-07-15
Adding the first component server to a composite MCP server reports an error, and configuring tools for component servers also fails.


sangee2004:
**Describe the bug**
When adding the first component server to a composite MCP server, the UI  displays a validation error stating that the composite must contain at least one component server, even though the component gets successfully added. After adding additional component servers, attempting to configure their tools results in another error when the configuration is submitted. Adding component servers and configuring their tools should complete successfully without displaying erroneous validation errors.

**To Reproduce**
1. Log in as an admin user and create a composite MCP server.
2. Using `Add Component Entry`, add any MCP server (DDuckDuckGoMCP server in my case)
Notice that the user is presented with the following error: `failed to validate entry manifest: runtime composite validation error for field compositeConfig.componentServers: must contain at least one component server`. 

<img width="1099" height="855" alt="Image" src="https://github.com/user-attachments/assets/c4c9cf65-0d1d-4609-b1b8-37a2837ae5f6" />

If we ignore the error message and close the Configure tools modal, the server is actually added successfully. 
There is another POST call to `api/mcp-catalogs/default/entries` that succeeds.

3. Add another MCP server ( Exa Search MCP server in my case)
There is no error message presented in this case.

4. Use `Configure Tools` option

Once I enter the credentials I am prompted for and click on `Continue`, I get the following error:

<img width="871" height="546" alt="Image" src="https://github.com/user-attachments/assets/6b2a9ea4-2a41-4fa6-8035-8b9ebe7a6471" />

5. I am also not able to connect to this MCP server from VS Code.
I  see the following error in MCP server logs in VS Code -`failed to ensure server is deployed: failed to launch mcp server: timed out waiting for MCP server to be ready after 5 watch retries: Deployment.apps "ms1h8czv" not found`

MCP server logs in VS Code
``` 
2026-07-15 11:36:39.211 [info] Discovered authorization server metadata at http://localhost:8080/.well-known/oauth-authorization-server/default-santestcomp1-79902f57nltfn
2026-07-15 11:36:39.248 [info] Connection state: Error 500 status sending message to http://localhost:8080/mcp-connect/default-santestcomp1-79902f57nltfn: failed to ensure server is deployed: failed to launch mcp server: timed out waiting for MCP server to be ready after 5 watch retries: Deployment.apps "ms1h8czv" not found`
```

**Expected behavior**
1. There should be no errors when adding MCP servers to the Composite MCP server
[...]

PR #7202, opened 2026-07-17
fix(composite): ensure components are deployed and contactable

thedadams:
The component servers are connected to from Obot iteself, so we should connect to them via localhost and allow that specifically.

Issue: https://github.com/obot-platform/obot/issues/7161


njhale (approved):

njhale:
> fix(comopsite):

nit: typo

PR #7217, opened 2026-07-20
feat: shared credentials for git auth

calvinmclean:
Addresses https://github.com/obot-platform/field-issues/issues/49
Also addresses new issue: https://github.com/obot-platform/obot/issues/7230

This adds shared git credentials that can be created and used for Skill Repositories and MCP Catalogs.


pkg/api/handlers/gitcredentials.go
```diff
@@ -0,0 +1,224 @@
...
+		return fmt.Errorf("failed to update Git credential: %w", err)
+	}
+	if manifest.Token != "" {
+		if err := gitcredential.Store(req.Context(), req.GatewayClient, credential.Name, manifest.Token); err != nil {
+			return fmt.Errorf("failed to store Git credential token: %w", err)
+		}
+		configured = true
+	}
+	references, _ := gitCredentialReferences(req, credential.Name)
+	return req.Write(convertGitCredential(credential, configured, len(references) > 0))
```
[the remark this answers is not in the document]
calvinmclean: This is intentionally ignored because it is non-critical information gathered after successfully updating the stored resource and credentials.

pkg/api/handlers/systemmcpcatalogs.go
```diff
@@ -66,24 +66,32 @@ func (h *SystemMCPCatalogHandler) Create(req api.Context) error {
...
 	}
+	originalSourceURLs := append([]string(nil), manifest.SourceURLs...)
 	if err := validateSystemCatalogManifest(&manifest, h.defaultCatalogPath); err != nil {
 		return err
 	}
+	remapCatalogSourceValues(originalSourceURLs, manifest.SourceURLs, manifest.SourceURLCredentials)
+	remapCatalogSourceValues(originalSourceURLs, manifest.SourceURLs, manifest.SourceURLGitCredentialIDs)
+	if err := validateCatalogGitCredentials(req, manifest.SourceURLs, manifest.SourceURLGitCredentialIDs); err != nil {
+		return err
+	}
```
g-linville: I noticed, for MCPCatalogs, we do this only in the Update function, but for SystemMCPCatalogs, we are doing it in both Create and Update. Is this intentional?

calvinmclean: Yes, there is no Create method for MCPCatalogs, only CreateEntry. New SourceURLs are added with the Update endpoint

g-linville: Oh that's right, I forgot lol. Thanks.

g-linville (approved):
backend lgtm (did not review frontend)

pkg/storage/apis/obot.obot.ai/v1/gitcredential.go
```diff
@@ -0,0 +1,34 @@
...
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type GitCredential struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec GitCredentialSpec `json:"spec,omitempty"`
```
thedadams: nit: add empty status.

pkg/api/handlers/gitcredentials.go
```diff
@@ -0,0 +1,224 @@
...
+	"github.com/obot-platform/obot/pkg/gitcredential"
+	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
+	"github.com/obot-platform/obot/pkg/system"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type GitCredentialHandler struct{}
+
+func NewGitCredentialHandler() *GitCredentialHandler {
+	return &GitCredentialHandler{}
```
thedadams: nit: in other handlers, we return `nil` here.

pkg/api/handlers/gitcredentials.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+func (*GitCredentialHandler) List(req api.Context) error {
+	var list v1.GitCredentialList
+	if err := req.List(&list); err != nil {
+		return fmt.Errorf("failed to list Git credentials: %w", err)
+	}
+	references, err := gitcredential.ReferencesByCredential(req.Context(), req.Storage, req.Namespace())
```
thedadams: The `gitcredential.ReferencesByCredential` call seems expensive. Do you think it might be valuable to track the references on `GitCredential` object itself and write a controller handler to populate this field?

Maybe that's an early optimization, but curious to hear your thoughts.

pkg/api/handlers/gitcredentials.go
```diff
@@ -0,0 +1,224 @@
...
+	if err := req.Update(&credential); err != nil {
+		return fmt.Errorf("failed to update Git credential: %w", err)
+	}
+	if manifest.Token != "" {
+		if err := gitcredential.Store(req.Context(), req.GatewayClient, credential.Name, manifest.Token); err != nil {
+			return fmt.Errorf("failed to store Git credential token: %w", err)
+		}
+		configured = true
+	}
+	references, _ := gitCredentialReferences(req, credential.Name)
```
thedadams: Can we handle the error or add a comment describing why we're ignoring it?

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -145,9 +145,15 @@ func (h *MCPCatalogHandler) Update(req api.Context) error {
 		return fmt.Errorf("failed to get catalog: %w", err)
 	}
 
+	originalSourceURLs := append([]string(nil), manifest.SourceURLs...)
```
thedadams:
```suggestion
	originalSourceURLs := slices.Clone(manifest.SourceURLs)
```

pkg/api/handlers/systemmcpcatalogs.go
```diff
@@ -66,24 +66,32 @@ func (h *SystemMCPCatalogHandler) Create(req api.Context) error {
 	if err := req.Read(&manifest); err != nil {
 		return fmt.Errorf("failed to read system catalog manifest: %w", err)
 	}
+	originalSourceURLs := append([]string(nil), manifest.SourceURLs...)
```
thedadams:
```suggestion
	originalSourceURLs := slices.Clone(manifest.SourceURLs)
```

pkg/controller/handlers/skillrepository/skillrepository.go
```diff
@@ -57,6 +58,13 @@ func (h *Handler) revealRepositoryCredential(ctx context.Context, repoName, repo
 	return cred.Secrets[repoURL]
 }
 
+func (h *Handler) resolveRepositoryCredential(ctx context.Context, storageClient client.Client, repo *v1.SkillRepository) (string, error) {
+	if repo.Spec.GitCredentialID != "" {
+		return gitcredential.Resolve(ctx, storageClient, h.gatewayClient, repo.Namespace, repo.Spec.GitCredentialID, repo.Spec.RepoURL)
+	}
+	return h.revealRepositoryCredential(ctx, repo.Name, repo.Spec.RepoURL), nil
+}
```
thedadams: There are a couple of these functions here. It seems that the only reason it is a method on the handler is for the `gatewayClient`. Can we create a single version of this function (perhaps in another package) that can be referenced in the various places it is needed.

pkg/gitcredential/gitcredential.go
```diff
@@ -0,0 +1,183 @@
...
+	references := map[string][]string{}
+	var repositories v1.SkillRepositoryList
+	if err := storageClient.List(ctx, &repositories, client.InNamespace(namespace)); err != nil {
+		return nil, fmt.Errorf("failed to list skill repositories: %w", err)
+	}
+	for _, repository := range repositories.Items {
+		if credentialID := repository.Spec.GitCredentialID; credentialID != "" {
+			references[credentialID] = append(references[credentialID], "skill repository "+repository.Name)
+		}
+	}
```
thedadams, line 161: Instead of listing all and filter this way, we should add a field selector to the `SkillRepository` and use it when listing here.

pkg/api/handlers/gitcredentials.go
calvinmclean: I think it's a bit of an early optimization, but it's not too difficult to add. It will also have the benefit of improving the UX by showing the user which resources rely on it. Since they cannot delete in-use credentials, it's useful to know where it is used.

ui/user/src/routes/mcp-catalog/+page.svelte
```diff
@@ -194,6 +194,14 @@
 					Sync
 				{/if}
 			</button>
+			<button
```
ivyjeong13: Since this is just routing, it can be a `<a href={resolve("/admin/git-credentials")} class="btn btn-neutral flex items-center gap-1 rounded-4xl text-sm">` instead of a `<button>`

ivyjeong13 (approved):

pkg/controller/handlers/skillrepository/skillrepository.go
```diff
@@ -77,7 +61,18 @@ func (h *Handler) Sync(req router.Request, resp router.Response) error {
 
 	defer h.clearIsSyncing(req.Ctx, req.Client, namespace, repo.Name)
 
-	token := h.revealRepositoryCredential(req.Ctx, repo.Name, repo.Spec.RepoURL)
+	token, err := gitcredential.ResolveOrReveal(req.Ctx, req.Client, h.gatewayClient, repo.Namespace, repo.Spec.GitCredentialID, repo.Spec.RepoURL, repo.Name, SkillRepositoryCredentialToolName)
+	if errors.Is(err, gitcredential.ErrLegacyCredential) {
+		log.Errorf("failed to retrieve legacy credential for repository %s source %s, continuing without authentication: %v", repo.Name, repo.Spec.RepoURL, err)
+		err = nil
+	}
+	if err != nil {
```
thedadams:
```suggestion
	} else if err != nil {
```

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -146,7 +130,17 @@ func (h *Handler) Sync(req router.Request, resp router.Response) error {
...
 
 	for _, sourceURL := range mcpCatalog.Spec.SourceURLs {
-		token := h.revealCatalogCredential(req.Ctx, mcpCatalog.Name, sourceURL)
+		credentialID := mcpCatalog.Spec.SourceURLGitCredentialIDs[sourceURL]
+		token, err := gitcredential.ResolveOrReveal(req.Ctx, req.Client, h.gatewayClient, mcpCatalog.Namespace, credentialID, sourceURL, mcpCatalog.Name, CatalogCredentialToolName)
+		if errors.Is(err, gitcredential.ErrLegacyCredential) {
+			log.Errorf("failed to retrieve legacy credential for catalog %s source %s, continuing without authentication: %v", mcpCatalog.Name, sourceURL, err)
+			err = nil
+		}
+		if err != nil {
```
thedadams: Same, and below

thedadams (approved):

PR #7228, opened 2026-07-21
fix: don't hold a DB transaction across auth provider HTTP calls

ibuildthecloud:
## Problem

`EnsureIdentityWithRole` performed the auth provider group lookups inside the same transaction that ensured the identity and user rows. That holds a pooled DB connection open across HTTP round-trips to the auth provider, which can deadlock in-process auth providers that share the single SQLite connection.

## Change

- Split the provider work into `ensureIdentityProviderData`, which runs *after* the identity transaction commits and opens its own short-lived transaction per write.
- `ensureGroups` no longer takes a `*gorm.DB`: it performs the HTTP fetch with no transaction open, then persists via the new `persistGroups`, which does the group upsert and membership reconciliation in one short transaction.
- Reconciliation events are still emitted after that transaction commits.
- Comments added on both entry points documenting the "must be called outside a transaction" requirement.

No behavior change beyond transaction boundaries.

## Test plan

- `go build ./pkg/gateway/...` and `go vet ./pkg/gateway/client/` pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MaifTz9tLjqQL2jF9HokvS

### DIFF
--- a/pkg/gateway/client/group.go
+++ b/pkg/gateway/client/group.go
@@ -172,7 +172,13 @@ func (c *Client) GetUserGroupMemberships(ctx context.Context, userIDs []uint) (m
 }
 
 // ensureGroups ensures the groups that the identity is a member of exist and are up to date.
-func (c *Client) ensureGroups(ctx context.Context, tx *gorm.DB, identity *types.Identity) error {
+//
+// It MUST be called outside of any open database transaction: the group fetch it performs is an
+// HTTP call to the auth provider, and holding a pooled DB connection across that round-trip can
+// deadlock in-process auth providers that share the single SQLite connection. The HTTP fetch and
+// the database persistence are therefore separated into distinct phases below, with the
+// persistence happening in its own short-lived transaction.
+func (c *Client) ensureGroups(ctx context.Context, identity *types.Identity) error {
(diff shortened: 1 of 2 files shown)


cjellick:
@thedadams im basically treating this pr like an issue and putting it in the v0.26 milestone. when you get to it proper, you can continue to tuse it as the issue or open dedicated isssues for it. 

...im cool with us slipping this into v0.25 if you end up having bandwidth.

when it comes time to test this, note that we think this may also eliminate our slow query logs. id like that explicitly verified one way or another, though fixing the slow query logs is not a requirement for this pr. and if this pr DOESNT fix that, then make sure we have a dedicated issue for slow queries that we can return to.

thedadams:
Without this change, when I start locally using sqlite instead of postgres, I see the slow queries getting credentials for the model provider. It is very constant and consistent.

I do not see these slow queries with this change.

Note that you can't use local auth to see this because this bypasses the query for credentials.

njhale (approved):

calvinmclean (approved):

g-linville (approved):

PR #7282, opened 2026-07-23
feat: tool-call enforcement backend for MDM configurations

g-linville:
This adds the backend for tool call enforcement.

Each MDMConfiguration will now have an EnforcementAllowlist, which contains the rules governing whether individual tool calls are allowed.

Devices can submit a decision request with information about the tool call, which will then be evaluated against the allowlist. This is what obot-sentry will do with its pre-tool hooks.

This PR also introduces a decision log, which is basically a separate audit log table for decisions made by enforcement. This is not shipping with retention or export features, but those can come later.


pkg/gateway/client/enforcementdecisionpersister.go
```diff
@@ -0,0 +1,89 @@
...
+			timer.Stop()
+		case <-timer.C:
+		}
+
+		if err := c.persistEnforcementDecisions(); err != nil {
+			log.Errorf("Failed to persist enforcement decision log: %v", err)
+		}
+
+		timer.Reset(flushInterval)
+	}
```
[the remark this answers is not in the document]
g-linville: this only mattered in older versions of Go

.gitignore
njhale, line 72: Want to just leave this in here?

pkg/api/handlers/enforcement.go
```diff
@@ -0,0 +1,385 @@
...
+
+	obotHosted := h.isObotHosted(in.Server.URL)
+	base := gtypes.EnforcementDecisionLog{
+		MDMConfigurationID: configID,
+		DeviceID:           deviceID,
+		ClientIP:           requestinfo.GetSourceIP(req.Request),
+		ObotHosted:         obotHosted,
+	}
+
+	config, err := req.GatewayClient.GetMDMConfiguration(req.Context(), configID)
```
njhale: We're going to want to change GetMDMConfiguration so that fetching artifacts is optional. The Artifacts are a separate related table containing the rendered zip bytes of every mdm install strategy in the configuration. As it stands this call will pull megabytes of data from the db.

.gitignore
g-linville: the humanlayer artifacts? probably not...if we are going to fully adopt it, then that stuff gets handled in the cloud. I don't think it's supposed to live in the repo

pkg/api/handlers/enforcement.go
g-linville: Added a separate function for getting only the enforcement information

pkg/api/handlers/enforcement.go
```diff
@@ -0,0 +1,385 @@
...
+		SortBy:             query.Get("sort_by"),
+		SortOrder:          query.Get("sort_order"),
+		Query:              strings.TrimSpace(query.Get("query")),
+	}
+
+	if startTime := query.Get("start_time"); startTime != "" {
+		if t, err := time.Parse(time.RFC3339, startTime); err == nil {
+			opts.StartTime = t
+		}
+	}
```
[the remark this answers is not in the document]
g-linville: Intentionally out of scope for this PR.

pkg/api/handlers/enforcement.go
```diff
@@ -0,0 +1,419 @@
...
+	// Resolve the fleet configuration strictly from the authenticated identity.
+	// A caller that is not an enrolled device has no fleet to enforce against:
+	// deny without recording, so only devices can write decision rows.
+	configID, ok := parseConfigurationID(firstExtraValue(extra, "mdm_configuration_id"))
+	if deviceID == "" || !ok {
+		return respondDecision(req, enforcement.Decision{
+			Allow:  false,
+			Reason: "no MDM configuration is associated with this device",
+		})
+	}
```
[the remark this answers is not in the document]
g-linville: It will never be 0

.gitignore
thedadams: This is already in the file. Maybe a rebase would remove it from this changeset.

g-linville: @thedadams I just rebased

g-linville: looks like it is gone now

pkg/api/authz/authz.go
```diff
@@ -78,6 +78,8 @@ var (
 		"/api/devices/clients/",
 		"/api/mdm/configurations",
 		"/api/mdm/configurations/",
+		"GET /api/enforcement-decisions",
+		"GET /api/enforcement-decisions/",
```
thedadams, line 82: Based on what I understand, we would probably want to add these to the auditor role, too.

pkg/api/handlers/enforcement.go
```diff
@@ -0,0 +1,419 @@
...
+
+// Decide handles POST /api/enforcement/decisions.
+func (h *EnforcementHandler) Decide(req api.Context) error {
+	var in types.EnforcementDecisionRequest
+	if err := req.Read(&in); err != nil {
+		return types.NewErrBadRequest("failed to read input: %v", err)
+	}
+
+	extra := req.User.GetExtra()
+	deviceID := firstExtraValue(extra, "device_id")
```
thedadams: Use `utils.FirstSet(extra["device_id")` instead of another version of the function.

pkg/api/handlers/enforcement.go
```diff
@@ -0,0 +1,419 @@
...
+		return false
+	}
+	call, err := url.Parse(callURL)
+	if err != nil {
+		return false
+	}
+	server, err := url.Parse(h.serverURL)
+	if err != nil {
+		return false // should never happen
+	}
```
thedadams: We have the `serverURL` in the constructor for this type. Can we parse it once there and we don't have to worry about this error not happening?

pkg/gateway/client/enforcementdecisionpersister.go
```diff
@@ -0,0 +1,89 @@
...
+func (c *Client) LogEnforcementDecision(entry types.EnforcementDecisionLog) {
+	if entry.CreatedAt.IsZero() {
+		entry.CreatedAt = time.Now().UTC()
+	}
+
+	c.enforcementLock.Lock()
+	defer c.enforcementLock.Unlock()
+
+	c.enforcementBuffer = append(c.enforcementBuffer, entry)
+	if cap(c.enforcementBuffer) > 0 && len(c.enforcementBuffer) >= cap(c.enforcementBuffer)/2 {
```
thedadams: If `cap(c.enforcementBuffer)` were to ever be equal to 0, then after the first log, we would persist on every run. That kind of defeats the purpose of buffering.

Can we ensure that we set the slice with some capacity to avoid this situation?

pkg/gateway/client/mdmconfiguration.go
```diff
@@ -106,6 +125,34 @@ func (c *Client) UpdateMDMConfiguration(ctx context.Context, configuration *type
...
+	if id == 0 {
+		return fmt.Errorf("MDM configuration id is required")
+	}
+	return c.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		if err := tx.Select("id").Where("id = ?", id).First(&types.MDMConfiguration{}).Error; err != nil {
+			return err
+		}
+		if err := tx.Model(&types.MDMConfiguration{}).
+			Where("id = ?", id).
+			Select("EnforcementEnabled", "EnforcementAllowlist").
```
thedadams: Since this gets translated to sql, I would expect these to be `enforcement_enabled` and `enforcement_allow_list`.

If both versions work, I would suggest using the snake_case versions to avoid confusion.

pkg/gateway/types/enforcementdecisionlog.go
```diff
@@ -0,0 +1,41 @@
+//nolint:revive
+package types
+
+import "time"
+
+const (
+	EnforcementDecisionAllow = "allow"
+	EnforcementDecisionDeny  = "deny"
+)
```
thedadams: Can we use the constants from the `apilcient` package instead of duplicating?

njhale (approved):

pkg/gateway/types/enforcementdecisionlog.go
g-linville: fixed

pkg/gateway/client/mdmconfiguration.go
g-linville: fixed

pkg/gateway/client/enforcementdecisionpersister.go
g-linville: Yep, fixed

g-linville: Also fixed it in the other audit logs

pkg/api/handlers/enforcement.go
g-linville: Yep, fixed

pkg/api/handlers/enforcement.go
g-linville: Fixed

pkg/api/authz/authz.go
g-linville: Yep, fixed

pkg/gateway/client/client.go
```diff
@@ -57,6 +58,12 @@ func New(ctx context.Context, db *db.DB, storageClient kclient.Client, encryptio
 	for _, email := range ownerEmails {
 		explicitRoleEmailsSet[strings.ToLower(email)] = types2.RoleOwner
 	}
+	// The audit and decision buffers flush once they are half full, so a
+	// non-positive batch size would give them zero capacity and turn every
+	// single entry into a flush, defeating the buffering entirely.
+	if auditLogBatchSize <= 0 {
+		auditLogBatchSize = defaultAuditLogBatchSize
+	}
```
thedadams: nit: this value comes from a configuration value, which can only ever be 0 if it is explicitly set to 0. I think we should allow that in case it is set that way on purpose.

My previous comment about the capacity still stands, but I think this should be allowed.

thedadams (approved):

pkg/gateway/client/client.go
g-linville: I looked at this more deeply, and I am just going to revert the changes I made to this back to what it was when you made your first comment. The enforcementBuffer is always allocated with a capacity of 2x the batch size, just like in the main audit logs. So I think we're good.

g-linville: The checking that the cap was greater than 0 was unnecessary in my initial code, so that is still removed.

Issue #7290, opened 2026-07-23
Add community edition licensing and upgrade path


thedadams:
We want to add the ability for users to go from a free open-source version of Obot to a free community edition that unlocks more free features. This process should be seamless and not require any human intervention on Obot's side.

sangee2004:
This issue is addressed when testing with the latest build from `main`

Admin user is able to get a community edition license by providing name and email address.

<img width="1117" height="625" alt="Image" src="https://github.com/user-attachments/assets/2dcfd85c-b11c-48fe-88ba-755bc6e31eb2" />


With this community edition license  , user has access to all Model providers.

<img width="1111" height="691" alt="Image" src="https://github.com/user-attachments/assets/717df9e0-5770-4f7b-aa07-4e266ff14e3b" />

Able to configure `Amazon Bedrock (API Key)` model provider and use model served by this provider as an LLM Gateway for external clients like `Claude Code`. Also able to use the model to chat with obot agents.


When the  community edition license is deleted , we will fall back to  free open-source version. 
If there is any model provider other than OpenAI/Anthropic/Responses Compatible Generic Provider configured , we will be put in `Missing or Invalid License` state

<img width="554" height="471" alt="Image" src="https://github.com/user-attachments/assets/6eb53c52-86f8-4cee-b086-489cbd74a445" />

We will not be able to use `Amazon Bedrock (API Key)` model provider  for chatting with obot agent or as LLM gateway.

Chat from external clients using the provider as LLM gayeway witll fail with - `API Error: 402 configured provider is missing required license entitlements` error.

Obot chat will fail with `tools/call failed: 402 Payment Required: configured provider is missing required license entitlements` error.

Any connection made to MCP servers will also fail - `Error 402 status connecting to http://localhost:8080/mcp-connect/default-outlook-841b850d as SSE: configured provider is missing required license entitlements` 

User can downgrade by deconfiguring the `Amazon Bedrock (API Key)` model provider  in this case or get a comminuty license again. When comminuty license is got again ,  `Amazon Bedrock (API Key)` model provider  can be used for chatting with obot agent and as an LLM gateway successfully. Connection to MCP servers will also start succeeding.


Issues found when testing this feature is tracked in seperate issues:
https://github.com/obot-platform/obot/issues/7320
https://github.com/obot-platform/obot/issues/7321

PR #7303, opened 2026-07-24
feat: add Community Edition enrollment API and UI

thedadams:
This change allows users to request a community edition key from the upgrade server. By submitting their name, email, and, optionally, company name, Obot will make a request to the upgrade server to get a license. If a license is returned succesfully, then Obot will persist it.

Issue: https://github.com/obot-platform/obot/issues/7290

### DIFF
--- a/pkg/upgrade/community.go
+++ b/pkg/upgrade/community.go
@@ -1,0 +1,153 @@
+package upgrade
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"time"
+)
+
+const (
+	communityLicenseEndpoint         = "community-license"
+	communityLicenseRequestTimeout   = 10 * time.Second
+	communityLicenseTokenLifetime    = 5 * time.Minute
+	maxCommunityLicenseResponseBytes = 64 * 1024
+)
+
+var (
+	ErrCommunityLicenseRequest  = errors.New("community license request failed")
+	ErrCommunityLicenseResponse = errors.New("community license response was invalid")
+)
+
+type CommunityLicenseRequest struct {
+	Name           string `json:"name"`
+	Email          string `json:"email"`
+	Company        string `json:"company,omitempty"`
+	InstallationID string `json:"installationId"`
+}
+
+type CommunityLicenseIssuer interface {
+	Issue(context.Context, CommunityLicenseRequest) (string, error)
+}
+
+type CommunityLicenseClient struct {
+	gatewayClient propertyClient
+	httpClient    *http.Client
+	issueURL      string
+}
+
+type communityLicenseResponse struct {
+	LicenseKey string `json:"licenseKey"`
+}
+
+type sanitizedError struct {
+	public error
+	cause  error
+}
+
+func (e *sanitizedError) Error() string {
+	return e.public.Error()
+}
+
+func (e *sanitizedError) Unwrap() []error {
+	if e.cause == nil {
+		return []error{e.public}
+	}
+	return []error{e.public, e.cause}
+}
+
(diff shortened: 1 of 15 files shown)


ivyjeong13 (approved):
UI LGTM

pkg/upgrade/community.go
```diff
@@ -0,0 +1,173 @@
...
+
+	body, err := json.Marshal(request)
+	if err != nil {
+		return "", requestError(err)
+	}
+
+	requestContext, cancel := context.WithTimeout(ctx, communityLicenseRequestTimeout)
+	defer cancel()
+
+	now := time.Now().Add(-time.Second)
```
calvinmclean: curious why we don't just use `time.Now()`?

pkg/upgrade/community.go
```diff
@@ -0,0 +1,173 @@
...
+	"strings"
+	"time"
+
+	"github.com/golang-jwt/jwt/v5"
+	"github.com/obot-platform/obot/pkg/jwt/persistent"
+)
+
+const (
+	communityLicenseEndpoint         = "community-license"
+	communityLicenseRequestTimeout   = 5 * time.Second
```
calvinmclean: I think this timeout is a bit short. Maybe 10s would be safer

pkg/upgrade/community.go
```diff
@@ -0,0 +1,173 @@
...
+	httpRequest.Header.Set("Content-Type", "application/json")
+
+	response, err := c.httpClient.Do(httpRequest)
+	if err != nil {
+		return "", requestError(err)
+	}
+	defer response.Body.Close()
+
+	if response.StatusCode != http.StatusOK {
+		return "", requestError(nil)
```
calvinmclean: nit: can we include the StatusCode in this error?

pkg/upgrade/community.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	limitedBody := io.LimitReader(response.Body, maxCommunityLicenseResponseBytes+1)
+	responseBody, err := io.ReadAll(limitedBody)
+	if err != nil {
+		return "", responseError(err)
+	}
+	if len(responseBody) > maxCommunityLicenseResponseBytes {
+		return "", responseError(nil)
```
calvinmclean: we should use a non-nil error here

pkg/upgrade/community.go
```diff
@@ -0,0 +1,173 @@
...
+	if err := decoder.Decode(&result); err != nil {
+		return "", responseError(err)
+	}
+	if err := ensureJSONEnd(decoder); err != nil {
+		return "", responseError(err)
+	}
+
+	result.LicenseKey = strings.TrimSpace(result.LicenseKey)
+	if result.LicenseKey == "" {
+		return "", responseError(nil)
```
calvinmclean: also include an error here

g-linville (approved):

calvinmclean (commented):
I left some comments about improving errors, then saw that it wouldn't make a difference since we don't use the error anywhere. I think it's still worth adding it in case it's useful later. It might be worth logging the error in the `CreateCommunityLicense` handler since we don't return the actual error anywhere after that.

pkg/upgrade/community.go
thedadams: There is a second rounding issue that subtracting a second fixes.

pkg/upgrade/community.go
thedadams: Done.

pkg/upgrade/community.go
thedadams: Done.

pkg/upgrade/community.go
thedadams: Done.

pkg/upgrade/community.go
thedadams: Done.

calvinmclean (approved):

PR #7329, opened 2026-07-28
chore: add logging for oauth errors

thedadams:
This is supposed to help with identifying when errors are from Obot's OAuth process versus the second-level. To that end, we also add a prefix to the OAuth errors so that it is easier to identify the errors on the client side.

### DIFF
--- a/pkg/api/handlers/mcpgateway/oauth/authorize.go
+++ b/pkg/api/handlers/mcpgateway/oauth/authorize.go
@@ -36,14 +36,23 @@ const (
 	ErrInvalidClientMetadata   ErrorCode = "invalid_client_metadata"
 )
 
-// Error represents an OAuth 2.0 error response.
-type Error struct {
+// oauthError represents an OAuth 2.0 error response.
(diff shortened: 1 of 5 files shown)


pkg/api/handlers/mcpgateway/oauth/authorize.go
```diff
@@ -36,33 +36,39 @@ const (
...
+// oauthError represents an OAuth 2.0 error response.
+type oauthError struct {
 	Code        ErrorCode `json:"error"`
 	Description string    `json:"error_description,omitempty"`
 	State       string    `json:"state,omitempty"`
 }
 
-func (e Error) Error() string {
+func newOAuthError(code ErrorCode, description, state string) oauthError {
+	log.Infof("OAuth error: code=%s description=%s state=%s", code, description, state)
```
thedadams, line 47: I don't really like this, but this was the most straightforward way I could come up with to always log such errors without adding a bunch of lines of code.

calvinmclean (approved):

Issue #7154, opened 2026-07-14
Limit usage shown in the Models page to LLM capability


sangee2004:
**Describe the bug**
The Models page under `LLM Gateway` → `Model Providers` displays a `Usage` column even though the only supported usage today is Language Model (Chat). At the same time, models that do not support any LLM usage (for example, models with an empty usage field such as babbage-002) are also listed, which can be confusing. Models without LLM capabilities should also not appear in the `Add Models` modal when creating model access policies. The Usage column should either be removed or repurposed to indicate LLM support, making it clear which models can actually be used as language models.

Pre-req - Have any model provider configured.
**To Reproduce**
1. Log in as admin user 
2. Go to LLM Gateway-> Model Providers page
3. Click on the models icon on the configured model provider to launch the model page

Notice that there is a `Usage` field presented for all the models in the list with the following usage options:

<img width="921" height="350" alt="Image" src="https://github.com/user-attachments/assets/2a140170-5490-4518-bcdf-f8129f6566fc" />

The only Usage we currently use is the `Language Model (Chat)`. `Usage` field can be entirely dropped from this page, or made into an indicator for LLM support.

Also note that models that do not support any `Usage` are also listed. 

<img width="986" height="467" alt="Image" src="https://github.com/user-attachments/assets/58424899-4dfd-4972-b357-a49077a83cdb" />

```
       {
            "id": "m1-openai-model-provider-980145897e59b51a41a8255648e9982a-5bba5",
            "created": "2026-07-13T17:47:52Z",
            "type": "model",
            "name": "babbage-002",
            "displayName": "babbage-002",
            "targetModel": "babbage-002",
            "modelProvider": "openai-model-provider",
            "active": true,
            "usage": "",
            "dialect": "OpenAIResponses",
            "aliasAssigned": false,
            "modelProviderName": "openai-model-provider",
            "icon": "https://cdn.jsdelivr.net/npm/@phosphor-icons/core@2/assets/duotone/open-ai-logo-duotone.svg"
        },
```

These models, which do not have LLM capability, also appear when creating model access policies.

<img width="1105" height="858" alt="Image" src="https://github.com/user-attachments/assets/7e2dfa7b-3cf2-4ce9-9b9c-758057e6ada9" />

**Expected behavior**
Either drop the `Usage` field entirely from the models page, or make it into an indicator for LLM support.

Models without LLM capabilities should not appear in the `Add Models` modal when creating model access policies.

[...]

PR #7341, opened 2026-07-29
fix: limit models & model access policy to llm/llm-mini

ivyjeong13:
This addresses limiting models to LLM-based models by removing the usage dropdown from the List Models in Model Providers as well as enforcing, for model access policies, that only LLM-based models can be added moving forward by pruning non-LLM models from the default model access policy, enforcing Create/Update for a model access policy & warning users to remove existing models from their custom model access policy if they've added a non-LLM based model. 

Addresses #7154 

custom access policy w/ invalid models:
Screenshot 2026-07-29 at 10 37 32 AM

removed usage dropdown and adjusted sorting by moving LLM-based models to top (then sort alphabetically)
Screenshot 2026-07-29 at 10 37 41 AM

### DIFF
--- a/pkg/controller/handlers/modelaccesspolicy/modelaccesspolicy.go
+++ b/pkg/controller/handlers/modelaccesspolicy/modelaccesspolicy.go
@@ -9,7 +9,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
-// PruneModels ensures invalid and ineffectual model resources are removed from
+// PruneDefaultPolicy ensures invalid and ineffectual model resources are removed from
 // the default ModelAccessPolicy. Custom policies are intentionally left intact
 // so their legacy resources remain visible for users to remediate.
 // This handler removes:
@@ -20,7 +20,7 @@ import (
 //
 // Wildcard suffix patterns (e.g. "claude-haiku-4-5*") are always kept, even when
 // they currently match no models, since they apply to future models as well.
-func PruneModels(req router.Request, _ router.Response) error {
+func PruneDefaultPolicy(req router.Request, _ router.Response) error {
 	policy := req.Object.(*v1.ModelAccessPolicy)
 	if policy.Namespace != system.DefaultNamespace ||
 		policy.Name != system.ModelAccessPolicyPrefix+"-default" {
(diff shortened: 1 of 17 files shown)


g-linville (approved):
backend lgtm

pkg/modelaccesspolicy/validation.go
```diff
@@ -0,0 +1,106 @@
...
+// IsInvalidModelResource reports whether err identifies a model resource that
+// cannot be used in a model access policy.
+func IsInvalidModelResource(err error) bool {
+	var target *invalidModelResourceError
+	return errors.As(err, &target)
+}
+
+// IsAllowedModelUsage reports whether usage can be granted by a model access policy.
+func IsAllowedModelUsage(usage types.ModelUsage) bool {
+	return usage == types.ModelUsageLLM || usage == types.ModelUsage(types.DefaultModelAliasTypeLLMMini)
```
thedadams: I don't think we need the cast here.
```suggestion
	return usage == types.ModelUsageLLM || usage == types.DefaultModelAliasTypeLLMMini
```

thedadams (approved):

pkg/modelaccesspolicy/validation.go
ivyjeong13: I believe they're mismatched types and without it causes a compiling error. But gonna make a change, since I have isAllowedDefaultModelAlias, prolly don't need the checking for an llm-mini in here!

pkg/controller/handlers/modelaccesspolicy/modelaccesspolicy.go
```diff
@@ -3,21 +3,29 @@ package modelaccesspolicy
...
 	"github.com/obot-platform/nah/pkg/router"
 	"github.com/obot-platform/obot/apiclient/types"
+	accesspolicy "github.com/obot-platform/obot/pkg/modelaccesspolicy"
 	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
 	"github.com/obot-platform/obot/pkg/system"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
-// PruneModels ensures invalid and ineffectual model resources are removed from ModelAccessPolicies.
+// PruneModels ensures invalid and ineffectual model resources are removed from
```
njhale: Why did this get changed to only affect the default model access policy?

Issue #7331, opened 2026-07-28
Disconnected tunnel state is not reflected for remote MCP servers in the UI


sangee2004:
**Describe the bug**
When the tunnel backing a remote MCP server disconnects after the server has been connected, external client connections fail as expected. Still, the Obot UI continues to indicate that the server is connected. The `MCP Servers`, `MCP Catalog`, and `MCP Deployments` pages do not reflect the disconnected tunnel state, making it difficult to identify why the remote MCP server is unavailable. The UI should clearly indicate when a remote MCP server's associated tunnel is disconnected.

**To Reproduce**
1. Log in as admin and create an MCP Tunnel entry.
2. Start the tunnel (using docker command/CLI)
3. Create a remote MCP server entry using this tunnel
4. Connect to this remote MCP server from any external client (VS Code in my case)
5. Stop the tunnel process that was started in step2.

Connection to this remote MCP server from VS Code fails, and the server gets to the error state as expected.

There is no indication in the ObotServer UI that MCP server connections are failing for the MCP server.

`MCP Servers` page will show the status as `connected` with no indication that the tunnel associated with this MCP server is in a disconnected state.

<img width="1317" height="338" alt="Image" src="https://github.com/user-attachments/assets/40205830-f6f3-4020-9b30-030ba2ec25c2" />

`MCP Catalog` and `MCP Deployment` pages also do not show any indication that the tunnel associated with this MCP server is in a disconnected state.

`Server Details` tab for this remote server entry fromthe  `MCP Catalog` catalog page:

<img width="1285" height="485" alt="Image" src="https://github.com/user-attachments/assets/df482a27-7f82-410b-902d-a42a4c063f80" />

`MCP Deployment` page

<img width="1266" height="261" alt="Image" src="https://github.com/user-attachments/assets/eabeb512-5a4b-486d-a717-66bd1a32e3ef" />

**Expected behavior**
Provide a clear indication on the `MCP Servers`, `MCP Catalog`, and `MCP Deployments` pages when the tunnel associated with an MCP server is disconnected.

sangee2004:
This issue is addressed when testing with the latest build from `main`.

`Tunnel Disconnected` state is now shown in `MCP Deployment`, `MCP Servers` and `Server Details` page.

PR #7343, opened 2026-07-29
fix(ui): show disconnected MCP tunnel status

thedadams:
Poll live tunnel connections on MCP management routes and surface when a remote server's configured tunnel is unavailable.

This prevents configured servers from appearing healthy after their tunnel client disconnects.

Issue: https://github.com/obot-platform/obot/issues/7331


ui/user/src/lib/stores/mcpTunnelConnections.svelte.ts
```diff
@@ -0,0 +1,89 @@
...
+	current: {
+		connections: undefined,
+		loading: false
+	},
+	refresh,
+	startPolling
+});
+
+function canListTunnelConnections() {
+	return profile.current.hasAdminAccess?.() || profile.current.groups.includes(Group.POWERUSER);
```
ivyjeong13: Randomly noticed, I think `/api/tunnels` might only be in adminAndOwner rules so it might not be available to poweruser/auditors?

thedadams: Oh, yep! Great catch. I'll fix that.

thedadams: Done.

ivyjeong13 (approved):

Issue #7320, opened 2026-07-28
Missing community license option in license invalid resolve dialog.


sangee2004:
**Describe the bug**
After deleting either a Community Edition or Enterprise Edition license while using a model provider that requires a license, the `Resolve` action in the License Invalid banner only offers options to obtain an Enterprise license or downgrade the provider. The dialog should also provide an option to obtain a Community Edition license, allowing users to restore a valid Community license directly.

**To Reproduce**
Scenario 1:
1. Get a Community Edition License Key by going to the `License` page and using the `Get Community License` option.
2. Configure any model provider other than `OpenAI` or `Anthropic`
3. Go to the `License` page and delete the license.
4. Click on `Resolve` in the License Invalid banner.

Notice that the user is presented with an option to get an Enterprise Key or downgrade the provider.

Scenario 2:
1. Get an Enterprise Edition License Key and use the`Add License Key` option.
2. Configure any model provider other than `OpenAI` or `Anthropic`
3. Go to the `License` page and delete the license.
4. Click on `Resolve` in the License Invalid banner. 

Notice that the user is presented with an option to get an Enterprise Key or downgrade the provider.

<img width="540" height="481" alt="Image" src="https://github.com/user-attachments/assets/dfb090af-a2e4-4093-8c18-03cf4a7570a8" />

**Expected behavior**
The user should be presented with the option to get a Community Edition License Key as well.

sangee2004:
@thedadams I am testing with latest build from `main`  - [v0.0.0-dev+c842444e](https://github.com/obot-platform/obot/commit/c842444e) and I am still not seeing community license option being presented in the resolve dialog.

1. Get a Community Edition License Key by going to the `License page` and using the `Get Community License` option.
2. Configure Micrsoft Entra auth provider.
3. Go to the License page and delete the license.
4. Click on Resolve in the License Invalid banner.

Community License is still not presented in the resolve dialog

<img width="924" height="790" alt="Image" src="https://github.com/user-attachments/assets/3a7c7c5f-d95e-442d-9a16-c47b917a409e" />


Reopening for follow up

thedadams:
@sangee2004 This was done prior to us talking in standup that Community license should get access to auth providers.

Ivy is working on making all of those changes.

sangee2004:
This issue is addressed when testing with the latest build from `main`
Community license option is presented with the Resolve dialog.

[...]

PR #7345, opened 2026-07-29
fix(ui): offer community license from resolve dialog

thedadams:
Add the Community Edition enrollment path to the invalid-license dialog so administrators can restore licensed providers without first hunting for the License page. Keep enterprise support and provider downgrade as alternatives.

Issue: https://github.com/obot-platform/obot/issues/7320

### DIFF
--- a/ui/user/src/lib/components/admin/license/LicenseDowngradeDialog.svelte
+++ b/ui/user/src/lib/components/admin/license/LicenseDowngradeDialog.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+	import { resolve } from '$app/paths';
 	import ResponsiveDialog from '$lib/components/ResponsiveDialog.svelte';
 	import ProviderDeconfigureConfirm from '$lib/components/admin/ProviderDeconfigureConfirm.svelte';
 	import {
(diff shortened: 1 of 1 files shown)


ui/user/tests/license-downgrade-dialog.test.mjs
```diff
@@ -0,0 +1,20 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
+import { createServer } from 'vite';
+
+test('invalid license dialog offers Community Edition enrollment', async (t) => {
```
ivyjeong13: This test might not be too valuable, it's affirms that there is href="/admin/license" somewhere in the ssr rendered html and somewhere in it is also the text "Get Community License". 

Wondering if we could add vitest-browser-svelte (edit: or testing-library svelte equivalent!) to do some integration testing and quick tests like these!

For this, it'd be a quick test like
```
  const screen = await render(LicenseDowngradeDialog);

  const link = screen.getByRole('link', {
    name: 'Get Community License'
  });

  await expect.element(link).toHaveAttribute('href', '/admin/license');
```

ivyjeong13 (approved):

PR #7357, opened 2026-07-30
fix: default MCP catalog source to accelerate-data fork with multiUser entries

hbanerjee74:
## Summary
- Repoints `OBOT_SERVER_DEFAULT_MCPCATALOG_PATH` in the `Dockerfile` (both occurrences) from `https://github.com/obot-platform/mcp-catalog` to `https://github.com/accelerate-data/mcp-catalog`.
- `OBOT_SERVER_DEFAULT_SYSTEM_MCPCATALOG_PATH` is untouched — that path is unrelated to this change (system catalog metadata only, not the org-level entries list Studio reads).

## Why
The upstream `obot-platform/mcp-catalog` ships every entry as `serverUserType: singleUser`, so Studio's org-wide MCP Catalog "Enable" flow (which requires `multiUser`) can never enable anything on a fresh install. Root-caused in [VD-3947](https://linear.app/acceleratedata/issue/VD-3947). This is a one-line-per-occurrence env var change — no Go code changes, no new mechanism. `SetUpDefaultMCPCatalog` (`pkg/controller/handlers/mcpcatalog/mcpcatalog.go:914`) already seeds the default catalog's `SourceURLs` from this env var on first controller boot; this PR just points it at a fork that has 4 entries flipped to `multiUser`.

## Companion changes (land together)
- `accelerate-data/mcp-catalog` — accelerate-data/mcp-catalog#1, the fork with the flipped entries.
- `accelerate-data/studio` — [VD-3959](https://linear.app/acceleratedata/issue/VD-3959): Studio-side change to stop listing catalog entries that can never be enabled.

## Test plan
- [x] Diff reviewed — exactly the 2 intended lines changed, nothing else.
- [ ] Build an image from this branch and confirm a fresh boot seeds its default MCP catalog from the fork (tracked in VD-3959).


hbanerjee74:
Opened against the wrong repo by mistake — this is an accelerate-data-internal fork config change, not intended for upstream. Closing; recreating against accelerate-data/obot.

Issue #7244, opened 2026-07-22
Getting Started Guide: Obot Sentry


ivyjeong13:
Implement a getting started guide for Obot Sentry (Discover Shadow AI & Enforce Policies for Unmanaged MCP Servers) for administrative side.

May need to sync with @njhale to think out this guide.

sangee2004:
The Getting Started guide now includes a `Discover Shadow AI & Enforce Policies for Unmanaged MCP Servers` section that walks the admin through the steps to install Obot Sentry.

<img width="529" height="568" alt="Image" src="https://github.com/user-attachments/assets/dae6d14b-3d57-433c-a27a-ab62d880ac66" />

@ivyjeong13 There are no steps yet for enforcing the policies presented in the guide. Wanted to confirm that we will handle this separately from this issue.

Few changes:
This step shown needs to be corrected: - `Go to or set up the Obot Sentry configuration.`

<img width="1086" height="826" alt="Image" src="https://github.com/user-attachments/assets/186c3370-e040-4b35-af7a-494dc9b1ecbd" />

Microsoft Intune installation allows obot-sentry to be installed on devices owned by multiple users in the organization. Wanted to check if we wanted to call it out.
<img width="1105" height="836" alt="Image" src="https://github.com/user-attachments/assets/f2a75bb4-0ac5-43bf-98ee-d3cbcac226d8" />

ivyjeong13:
> [@ivyjeong13](https://github.com/ivyjeong13) There are no steps yet for enforcing the policies presented in the guide. Wanted to confirm that we will handle this separately from this issue.

I'll go ahead and add those steps as part of this ticket!

sangee2004:
All the issues  from https://github.com/obot-platform/obot/issues/7244#issuecomment-5124102762 is addressed when testing with the latest build from `main`.

PR #7359, opened 2026-07-30
fix: disable interaction w/ "getting started" guides & add steps in obot-sentry guide

ivyjeong13:
This addresses putting users even more "on-rails" for the getting started guide by disabling interaction with the highlighted items. As a result, wording for each guide was updated to avoid confusion that they can interact during the guide. It also adds additional steps to highlight the enforcement tools addition in Obot Sentry/Devices page.


g-linville (approved):

ivyjeong13:
Follow up for #7244

PR #7368, opened 2026-07-30
enhance(devices): add limits similar to users

thedadams:
This change adds the same 100 device limit to free and community Obot instances as the user limits. If a new devices is attempted to be enrolled after the limit is hit, a 403 error will be returned.


thedadams:
I looked at the Copilot review and it would have posted that you can't range over an int as its only comment.

njhale (approved):

calvinmclean (approved):

g-linville (approved):

PR #7431, opened 2026-08-03
feat(auth): add device authorization lifecycle

thedadams:


njhale (approved):

pkg/gateway/client/token.go
```diff
@@ -2,11 +2,260 @@ package client
...
+
+		return tx.Model(tr).Clauses(clause.Returning{}).Updates(map[string]any{"state": "", "error": tr.Error}).Error
+	})
+	log.Infof("Verified OAuth state for token request: tokenRequestID=%s success=%v", tr.ID, err == nil)
+	return tr, err
+}
+
+// runTokenCleanup removes expired token requests, retrieved requests, and expired auth tokens.
+func (c *Client) runTokenCleanup(ctx context.Context) {
+	timer := time.NewTimer(tokenCleanupInterval)
```
g-linville, line 201: nit: I think we could use a Ticker here instead of a Timer, and avoid needing to reset it at the end of each loop.

g-linville (approved):

pkg/gateway/client/token.go
thedadams: This is old code, just moved to a different place.

There is a reason we use a timer instead of a ticker, but it's probably not that big of a deal. I will leave it as-is for now because this code hasn't changed in this PR.

calvinmclean (approved):

PR #7439, opened 2026-08-03
fix: stop dropping skill metadata on the way through the API

ibuildthecloud:
`Skill` embeds both `Metadata` (field tagged `metadata`) and `SkillManifest`, whose `MetadataValues` was tagged `metadata` as well. Two promoted fields at equal depth carrying one JSON name is a conflict, and `encoding/json` resolves it by **dropping both** — so no skill ever serialized its metadata, and the frontmatter a repository supplies reached clients as nothing at all.

The manifest field is tagged `metadataValues` instead, which is the name it now carries on the wire. Nothing can be relying on the old name, since it never appeared in a response.

The UI type follows; it was wrong twice over, declaring the colliding name and typing the values as `unknown` rather than strings.

Two tests pin the behaviour: that metadata serializes, and that it is omitted rather than sent as `null` when empty.

Found while working on hosted agents, but unrelated to it — sending separately.

### DIFF
--- a/apiclient/types/skill.go
+++ b/apiclient/types/skill.go
@@ -15,13 +15,16 @@ type Skill struct {
 }
 
 type SkillManifest struct {
-	Name           string            `json:"name,omitempty"`
-	Description    string            `json:"description,omitempty"`
-	DisplayName    string            `json:"displayName,omitempty"`
-	License        string            `json:"license,omitempty"`
-	Compatibility  string            `json:"compatibility,omitempty"`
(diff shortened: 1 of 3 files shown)


ibuildthecloud:
Superseded by a fork-based PR. The Depot `Test Docker Build` check only authenticates via its open-source-pull-request flow, which requires the PR to come from a fork; a branch on obot-platform itself falls back to a plain OIDC exchange that the workflow grants no `id-token: write` for. Reopened as the same commit from ibuildthecloud.

PR #7442, opened 2026-08-03
fix: let websocket handlers under /api hijack the connection

ibuildthecloud:
Requests under `/api` are wrapped so the audit log can record the status and size of the response. That wrapper implements neither `Hijacker` nor `Unwrap`, so `http.ResponseController` cannot reach the writer underneath and **any websocket handler mounted under `/api` fails its upgrade** — the connection cannot be taken over through a wrapper that hides it.

It now implements `Unwrap`, the documented way to expose a capability a wrapper does not itself implement. That is why nothing here enumerates `Hijacker`, `Flusher` and the deadline setters one by one: the controller finds them. The header wrapper alongside it already did this; only the comment saying why is new.

Test covers an `/api` request being hijacked through the wrapper.

Found while working on hosted agents (whose terminal is a websocket under `/api`), but it affects any such handler — sending separately.

### DIFF
--- a/pkg/api/server/server.go
+++ b/pkg/api/server/server.go
@@ -342,3 +342,10 @@ func (rw *responseWriter) Flush() {
 		f.Flush()
 	}
 }
+
+// Unwrap exposes the writer underneath to http.ResponseController. Without it a
+// websocket handler under /api/ cannot hijack the connection, because this
+// wrapper hides the Hijacker the server provides.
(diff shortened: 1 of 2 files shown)


pkg/api/server/server.go
ibuildthecloud, line 253: Just pointing out that Unwrap already existed here and I think we just forgot it on the responseWriter object

thedadams (approved):

Issue #7092, opened 2026-07-01
Git-synced catalogs: mark as "detached" when a catalog entry is removed upstream


g-linville:
Right now, if a catalog entry is removed from the git repo, on the next sync, Obot will delete it too, which then cascades to MCPServers created from it.

We don't want this behavior anymore. Instead, we want to mark it as "detached" and then make it editable, as though it were created from the Obot UI in the first place.

cjellick:
we'll need to do this as part of deprecating and then removing the legacy catalog entries

cjellick:
my motivation for this is that when we eventually delete an item from the catalog, if someone has that deployed, i dont their server to disappear. i just want it to turn into one you can manage in the UI

calvinmclean:
@sangee2004 here is the intended behavior for this feature:

| Action | Result |
| --- | --- |
| Remove an entry with a matching `MCPServer` from a configured Git source | Convert it to an ordinary editable entry and preserve the server |
| Remove an entry with no matching `MCPServer` | Delete it |
| Remove an entire Git Source URL | Delete its still-managed entries and dependent deployments |

sangee2004:
This issue is addressed when testing with the latest build from `main`.

1. Log in as admin and add a catalog source.
2. Log in as any user and connect to any MCP catalog, say `mcp1`, that was synced from the catalog source from step 1.
3. Delete the `mcp1` catalog entry from the GitHub catalog source.
4. As admin, sync the MCP catalog changes.
This results in the `mcp1` catalog entry becoming an `Obot Admin Console` source type, and admin is able to manage this catalog entry.
There is no indication that this catalog entry is in a `Detached` state, and admin cannot tell that this entry was previously associated with a repo.

<img width="1040" height="288" alt="Image" src="https://github.com/user-attachments/assets/a8c2ff4c-0b75-4cc6-82ae-d29187f54590" />

Users connected to this mcp server are able to continue to connect and make tool calls successfully to this mcp server as expected.

5. Delete any mcp catalog entry that does not have connected MCP server instances from the GitHub catalog source.
6. As admin, sync the MCP catalog changes.

The deleted mcp catalog entry is no longer shown in the `MCP catalog` view.

@calvinmclean, wanted to follow up on providing an indication in the `MCP Catalog` view for catalogs in `Detached` state from step 4.

Also wanted to get confirmation on the following workflow:

1. Log in as admin and add a catalog source with source URL pointing to, say, url1.
[...]

PR #7445, opened 2026-08-03
feat: detach removed catalog entries with active servers

calvinmclean:
Addresses #7092

## Summary

- Preserve a Git catalog entry removed upstream only when an `MCPServer` still references it.
- Convert preserved entries into ordinary Obot-managed entries by making them editable, clearing their source URL and entry key, and removing Git apply ownership metadata.
- Delete removed entries that have no matching server.
- Keep whole-source removal destructive: removing a Git Source URL deletes its still-managed entries and dependent deployments, matching `main`.
- Skip all missing-entry reconciliation while any catalog source has a sync error so a partial desired set cannot delete valid entries.
- Keep apply pruning disabled after explicit reconciliation to prevent stale ownership metadata from deleting a freshly converted entry.

There is no separate detached state, API field, or UI. A preserved entry appears and behaves like any other editable entry created in Obot.

## Behavior

| Action | Result |
| --- | --- |
| Remove an entry with a matching `MCPServer` from a configured Git source | Convert it to an ordinary editable entry and preserve the server |
| Remove an entry with no matching `MCPServer` | Delete it |
| Remove an entire Git Source URL | Delete its still-managed entries and dependent deployments |
| Sync while any source has an error | Do not reconcile missing entries |

The server reference check and entry mutation are not transactional. A server created concurrently with removal can race with entry deletion.

## Testing

- `make test`
- `cd ui/user && pnpm run check`
- `cd ui/user && pnpm run lint`
- `git diff --check`


calvinmclean:
oops added reviewers on the wrong PR

pkg/controller/handlers/mcpcatalog/catalog_removal_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+func testCatalog() *v1.MCPCatalog {
+	return &v1.MCPCatalog{
+		TypeMeta: metav1.TypeMeta{APIVersion: v1.SchemeGroupVersion.String(), Kind: "MCPCatalog"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "default",
+			Namespace: "default",
+			UID:       ktypes.UID("catalog-uid"),
+		},
+		Spec: v1.MCPCatalogSpec{SourceURLs: []string{"github.com/obot/catalog"}},
```
g-linville: nit: just to make sure that future AI doesn't think that our GH org is `obot`


```suggestion
		Spec: v1.MCPCatalogSpec{SourceURLs: []string{"github.com/obot-platform/catalog"}},
```

pkg/controller/handlers/mcpcatalog/catalog_removal_test.go
```diff
@@ -0,0 +1,218 @@
...
+			OwnerReferences: []metav1.OwnerReference{{
+				APIVersion: v1.SchemeGroupVersion.String(),
+				Kind:       "MCPCatalog",
+				Name:       catalog.Name,
+				UID:        catalog.UID,
+			}},
+		},
+		Spec: v1.MCPServerCatalogEntrySpec{
+			MCPCatalogName: catalog.Name,
+			SourceURL:      "github.com/obot/catalog",
```
g-linville:
```suggestion
			SourceURL:      "github.com/obot-platform/catalog",
```

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -200,6 +205,121 @@ func addSyncError(syncErrors map[string]string, sourceURL, errMsg string) {
...
+
+	var missingManagedEntries []*v1.MCPServerCatalogEntry
+	for i := range entries.Items {
+		entry := &entries.Items[i]
+		if _, ok := desiredNames[entry.Name]; ok {
+			continue
+		}
+		if entry.Labels[apply.LabelHash] != labels[apply.LabelHash] {
+			continue
+		}
```
thedadams: [NOTE] Do we need this? Would it be enough to say that if the `entry.Spec.SourceURL == ""` then it was created from the UI?

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -200,6 +205,121 @@ func addSyncError(syncErrors map[string]string, sourceURL, errMsg string) {
...
+			if _, configured := configuredSources[mcp.SourceIDForURL(entry.Spec.SourceURL)]; !configured {
+				if err := c.Delete(ctx, entry); err != nil && !apierrors.IsNotFound(err) {
+					return fmt.Errorf("failed to delete catalog entry %q from removed source: %w", entry.Name, err)
+				}
+				log.Infof("Deleted MCP catalog entry from removed source: catalog=%s entry=%s source=%s", catalog.Name, entry.Name, entry.Spec.SourceURL)
+				continue
+			}
+		}
+
+		missingManagedEntries = append(missingManagedEntries, entry)
```
thedadams: [SUGGESTION] It seems that we only ever use the entry's name. So can we construct the `missingNames` set directly here (possibly with a different variable name)?

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -200,6 +205,121 @@ func addSyncError(syncErrors map[string]string, sourceURL, errMsg string) {
...
+	missingNames := make(map[string]struct{}, len(missingManagedEntries))
+	for _, entry := range missingManagedEntries {
+		missingNames[entry.Name] = struct{}{}
+	}
+
+	var servers v1.MCPServerList
+	if err := c.List(ctx, &servers, client.InNamespace(catalog.Namespace)); err != nil {
+		return fmt.Errorf("failed to list servers for removed catalog entries: %w", err)
+	}
+	serverCounts := make(map[string]int, len(missingManagedEntries))
```
thedadams: [SUGGESTION] Instead of tracking server counts, can we delete from the `missingNames` set when we find a server exists?

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -200,6 +205,121 @@ func addSyncError(syncErrors map[string]string, sourceURL, errMsg string) {
...
+		var entry v1.MCPServerCatalogEntry
+		if err := c.Get(ctx, client.ObjectKey{Namespace: catalog.Namespace, Name: entryName}, &entry); err != nil {
+			if apierrors.IsNotFound(err) {
+				return nil
+			}
+			return err
+		}
+
+		entry.Spec.Editable = true
+		entry.Spec.SourceURL = ""
```
thedadams, line 292: Yeah, seems like we can just clue off of the `SourceURL` above and not worry about checking the labels.

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -200,6 +205,121 @@ func addSyncError(syncErrors map[string]string, sourceURL, errMsg string) {
...
+		for key := range entry.Annotations {
+			if strings.HasPrefix(key, apply.LabelPrefix) {
+				delete(entry.Annotations, key)
+			}
+		}
+		for key := range entry.Labels {
+			if strings.HasPrefix(key, apply.LabelPrefix) {
+				delete(entry.Labels, key)
+			}
+		}
```
thedadams, line 303: I do think that removing these annotations and labels is the correct thing to do.

calvinmclean: @thedadams are you agreeing that it's correct to delete/remove the annotations here, or are you saying that this piece should be removed?

thedadams: Sorry, that was confusing. I was saying that even though I think we should remove the label detection above, I think removing these here is still correct.

calvinmclean: Agree. Thanks!

thedadams (approved):

PR #7448, opened 2026-08-03
feat: add the hosted agent resource model (1/11)

ibuildthecloud:
Hosted agents are agents Obot runs on the user's behalf, in a sandbox it
manages. Five resources describe them, and the split is what keeps a template
reusable across installations:

  Harness             a runtime image an agent runs on
  HostedAgent         a template: the harness, the models, MCP servers and
                      skills it may use, and what the user is asked
  HostedAgentInstance one user's running copy of a template
  HostedAgentPool     the compute a set of instances shares
  AgentCatalog        a git source that supplies harnesses and templates

A template is written once and synced everywhere, so it names models by
provider and MCP servers and skills by a portable reference rather than by an
ID the local installation generated. An instance is the only resource with a
user on it; everything else is installation-wide.

HostedAgentAccessRule follows the shape of the existing MCP access rules, so
granting access to an agent is the same idea administrators already know.

---

Part 1 of an 11-PR stack. Base: `main`.

## Stack

**→ 1. #7448 — Resource model** (this PR)
2. #7449 — Sandbox backend interface
3. #7450 — Kubernetes sandbox backend
4. #7451 — Sandbox reachability (models + MCP)
5. #7452 — Terminal + HTTP publish
6. #7453 — Server wiring + Helm
7. #7454 — Controllers / reconcilers
8. #7455 — REST API
9. #7456 — Admin UI
10. #7457 — User UI
11. #7458 — Seed default config source

Merge bottom-up; GitHub retargets each child when its parent merges.

PR #7449, opened 2026-08-03
feat: add an interface for running agent sandboxes (2/11)

ibuildthecloud:
The rest of Obot asks for a sandbox and gets back its state; where it actually
runs is behind this interface. That keeps Kubernetes types out of the
controllers and the API, so a second backend is a new implementation rather
than a change to everything that calls one.

Three implementations ship: disabled, which is the default and refuses
politely; fake, which transitions on a timer so the UI and the controllers can
be developed and tested without a cluster; and the Kubernetes one that follows.

Sizing is here rather than in a backend because it is policy, not mechanism: a
pool's capacity and an instance's share of it mean the same thing wherever the
sandbox runs.

---

Part 2 of an 11-PR stack. Base: `darren/hosted-agents-01-resource-model`.

## Stack

1. #7448 — Resource model
**→ 2. #7449 — Sandbox backend interface** (this PR)
8. #7455 — REST API
9. #7456 — Admin UI
10. #7457 — User UI

### DIFF
--- a/pkg/agentbackend/backend.go
+++ b/pkg/agentbackend/backend.go
@@ -1,0 +1,264 @@
+// Package agentbackend defines the implementation-neutral contract between
+// Obot's hosted-agent controllers and an agent runtime provider.
+package agentbackend
+
+import (
+	"context"
+	"time"
+)
+
+// Backend is the complete desired-state and observation contract required by
+// hosted agents. Implementations must make every operation idempotent.
+type Backend interface {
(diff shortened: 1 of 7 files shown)


ibuildthecloud:
Superseded by https://github.com/obot-platform/obot/pull/7460 — the branch was renamed `darren/*` → `feat/*`, which closed this PR. Same commit, no content change.

PR #7450, opened 2026-08-03
feat: run agent sandboxes on Kubernetes (3/11)

ibuildthecloud:
An instance becomes a Deployment, and a pool becomes the volume and priority
class its instances share. Instances are separated by a subPath on the pool's
volume rather than by a volume each, so a pool is one claim however many
instances it holds.

A sandbox is told nothing about Kubernetes: it reads its configuration from
files and serves on a port. What is here is the translation -- desired state to
objects, object status back to instance status, and utilization from the
metrics API when it is available, degrading to requested resources when it is
not.

Two details exist because Obot usually runs outside the cluster in development.
ReachableServerURL rewrites a loopback address to the node's, since a sandbox
resolving "localhost" would find itself; and the dev proxy reaches a sandbox
through the API server, which is deliberately confined to development because
it grants access to every Service in the cluster.

---

Part 3 of an 11-PR stack. Base: `darren/hosted-agents-02-backend-interface`.

## Stack

1. #7448 — Resource model
**→ 3. #7450 — Kubernetes sandbox backend** (this PR)
8. #7455 — REST API
9. #7456 — Admin UI
10. #7457 — User UI

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -233,6 +233,7 @@ require (
 	github.com/mitchellh/reflectwalk v1.0.2 // indirect
 	github.com/moby/docker-image-spec v1.3.1 // indirect
 	github.com/moby/locker v1.0.1 // indirect
+	github.com/moby/spdystream v0.5.1 // indirect
 	github.com/moby/term v0.5.2 // indirect
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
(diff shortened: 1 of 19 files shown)


ibuildthecloud:
Superseded by https://github.com/obot-platform/obot/pull/7461 — the branch was renamed `darren/*` → `feat/*`, which closed this PR. Same commit, no content change.

PR #7451, opened 2026-08-03
feat: let a sandbox reach the models and MCP servers it was granted (4/11)

ibuildthecloud:
A sandbox authenticates as itself. Its credential carries a group that marks it
as a hosted agent rather than a person, and that group reaches the MCP gateway
and the model proxy and nothing else -- so a credential taken out of a sandbox
cannot enumerate or modify Obot resources.

An agent is authorized against the servers its template was granted rather than
against a user's access, because the agent is not acting as the user: the
template says what it may use, and that list is the whole of it.

Model resolution is shared between the API, which decides whether an agent can
be launched, and the controller, which writes the endpoints into a sandbox's
configuration. If those two disagreed a user would be offered an agent that
then resolves no models. Portable references resolve a template's "<source>::
<key>" naming to whatever local ID an installation generated, which is what
lets one template work everywhere.

---

Part 4 of an 11-PR stack. Base: `darren/hosted-agents-03-kubernetes-backend`.

## Stack

1. #7448 — Resource model
**→ 4. #7451 — Sandbox reachability (models + MCP)** (this PR)
8. #7455 — REST API
9. #7456 — Admin UI
10. #7457 — User UI

### DIFF
--- a/pkg/api/authz/agentconnect_test.go
+++ b/pkg/api/authz/agentconnect_test.go
@@ -1,0 +1,65 @@
+package authz
+
+import (
+	"net/http"
+	"testing"
+
+	types "github.com/obot-platform/obot/apiclient/types"
+	"k8s.io/apiserver/pkg/authentication/user"
+)
+
+func agentConnectRequest(path string) *http.Request {
+	req, _ := http.NewRequest(http.MethodGet, path, nil)
+	return req
+}
+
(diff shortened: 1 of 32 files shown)


ibuildthecloud:
Superseded by https://github.com/obot-platform/obot/pull/7462 — the branch was renamed `darren/*` → `feat/*`, which closed this PR. Same commit, no content change.

PR #7452, opened 2026-08-03
feat: attach a terminal to a sandbox and publish agents that serve HTTP (5/11)

ibuildthecloud:
An interactive agent gets a terminal over a websocket: framed channels carry
input, output and resize, and session errors travel on a control channel rather
than on stderr, which is reserved as a stream a non-TTY session can use for
what the program actually wrote there.

An agent that declares a port is published under /agent-connect/{instance id}
instead. Obot strips that prefix before forwarding, so an agent that assumed it
was at the root would build links landing outside its own prefix -- a
single-page UI would call Obot instead of itself. The prefix is sent per
request as X-Forwarded-Prefix and is also in the sandbox's configuration, for
servers that need it at start-up rather than per request, and relative
redirects are rewritten back inside the prefix idempotently so an agent that
already accounts for it is not prefixed twice.

---

Part 5 of an 11-PR stack. Base: `darren/hosted-agents-04-reachability`.

## Stack

1. #7448 — Resource model
**→ 5. #7452 — Terminal + HTTP publish** (this PR)
8. #7455 — REST API
9. #7456 — Admin UI
10. #7457 — User UI

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -90,6 +90,8 @@ require (
 	sigs.k8s.io/yaml v1.6.0
 )
 
+require github.com/moby/spdystream v0.5.1 // indirect
+
 require (
 	cel.dev/expr v0.25.2 // indirect
 	cloud.google.com/go v0.123.0 // indirect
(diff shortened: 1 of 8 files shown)


ibuildthecloud:
Superseded by https://github.com/obot-platform/obot/pull/7463 — the branch was renamed `darren/*` → `feat/*`, which closed this PR. Same commit, no content change.

PR #7453, opened 2026-08-03
feat: wire the agent backend into the server (6/11)

ibuildthecloud:
The backend is selected by configuration: disabled unless asked for, fake in
development, and Kubernetes when a cluster is available.

The Kubernetes backend needs a client for the local cluster even when MCP
servers do not run there, which is the case a developer hits first: hosted
agents on a cluster, MCP servers on Docker. It builds its own client from the
same config rather than borrowing the MCP one, so enabling hosted agents does
not switch on MCP machinery a deployment has no reason to configure.

---

Part 6 of an 11-PR stack. Base: `darren/hosted-agents-05-terminal-connect`.

## Stack

1. #7448 — Resource model
2. #7449 — Sandbox backend interface
3. #7450 — Kubernetes sandbox backend
4. #7451 — Sandbox reachability (models + MCP)
5. #7452 — Terminal + HTTP publish
**→ 6. #7453 — Server wiring + Helm** (this PR)
7. #7454 — Controllers / reconcilers
8. #7455 — REST API
9. #7456 — Admin UI
10. #7457 — User UI
11. #7458 — Seed default config source

Merge bottom-up; GitHub retargets each child when its parent merges.

PR #7454, opened 2026-08-03
feat: reconcile hosted agents, pools and config sources (7/11)

ibuildthecloud:
The controllers turn the resources into running sandboxes: an instance is
placed in a pool, given a credential and a configuration, and handed to the
backend; a pool is created before anything needs it and removed with its last
instance; a user's instances go when the user does.

The configuration a sandbox receives is built here, and it is the contract the
images depend on. Every endpoint in it is absolute and every skill is already
on disk, so an image boots by reading a file and needs to know nothing about
Obot -- not its address, not its routes. Credentials are written to a second,
tighter-permissioned file so the configuration itself can be world-readable and
carries no secret.

A config source is a git repository of harness and template definitions, synced
on an interval. Sync does not rewrite the references a template makes: a
template naming a server that is not installed still syncs and is reported
unavailable, and installing the server later makes it work with no resync.

---

Part 7 of an 11-PR stack. Base: `darren/hosted-agents-06-server-wiring`.

## Stack

1. #7448 — Resource model
**→ 7. #7454 — Controllers / reconcilers** (this PR)
8. #7455 — REST API
9. #7456 — Admin UI
10. #7457 — User UI

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -73,6 +73,10 @@ func (c *Controller) PreStart(ctx context.Context) error {
 		return fmt.Errorf("failed to ensure default user role setting: %w", err)
 	}
 
+	if err := ensureHostedAgentPoolDefaults(ctx, c.services.StorageClient); err != nil {
+		return fmt.Errorf("failed to ensure hosted agent pool defaults: %w", err)
+	}
+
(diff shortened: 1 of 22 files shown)


ibuildthecloud:
Superseded by https://github.com/obot-platform/obot/pull/7465 — the branch was renamed `darren/*` → `feat/*`, which closed this PR. Same commit, no content change.

PR #7455, opened 2026-08-03
feat: serve the hosted agent API (8/11)

ibuildthecloud:
Endpoints for harnesses, templates, instances, pools, config sources and access
rules, authorized the same way the rest of Obot is: templates and pools are
administrator-owned, an instance belongs to its user, and access rules decide
who may launch what.

Two things are computed on read rather than stored. Availability reports why a
template cannot be launched here -- an unconfigured model provider, an MCP
server that was never installed -- so an administrator sees the whole list
rather than fixing one and being told about the next; it is also enforced when
an instance is created, so the answer cannot be bypassed. The icon shown for an
instance resolves through the template to the harness. Storing either would
mean rewriting every row when a harness changes its icon or a provider is
configured, and leave the stored copy wrong until something did.

---

Part 8 of an 11-PR stack. Base: `darren/hosted-agents-07-controllers`.

## Stack

1. #7448 — Resource model
**→ 8. #7455 — REST API** (this PR)
9. #7456 — Admin UI
10. #7457 — User UI

### DIFF
--- a/pkg/agentcatalog/validate.go
+++ b/pkg/agentcatalog/validate.go
@@ -1,0 +1,73 @@
+// Package agentcatalog holds the server-side rules for an agent config source.
+//
+// The manifest validates its own shape, in apiclient/types, where the UI can
+// apply the same rules before submitting. What lives here is the part that
+// depends on how this server is running rather than on the data: a repository
+// on the local filesystem is a reasonable source when a developer is iterating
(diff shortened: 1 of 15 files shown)


ibuildthecloud:
Superseded by https://github.com/obot-platform/obot/pull/7466 — the branch was renamed `darren/*` → `feat/*`, which closed this PR. Same commit, no content change.

PR #7456, opened 2026-08-03
feat: add the hosted agents admin UI (9/11)

ibuildthecloud:
An admin section for templates, harnesses, pools and config sources, plus the
API client the whole feature is built on.

Rows carry the icon the server resolved, so the UI does not walk the
instance-template-harness chain itself and cannot disagree with the answer the
API gives. A template that cannot run here is shown with the reason and cannot
be launched, which is the same answer the API enforces rather than a second copy
of the rule.

---

Part 9 of an 11-PR stack. Base: `darren/hosted-agents-08-rest-api`.

## Stack

1. #7448 — Resource model
8. #7455 — REST API
**→ 9. #7456 — Admin UI** (this PR)
10. #7457 — User UI

### DIFF
--- a/ui/user/src/lib/components/Layout.svelte
+++ b/ui/user/src/lib/components/Layout.svelte
@@ -20,6 +20,7 @@
 		'agent-management': true,
 		'mcp-server-management': true,
 		'skills-management': true,
+		'hosted-agent-management': true,
 		'device-management': true,
 		'user-management': true,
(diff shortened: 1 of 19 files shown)


ibuildthecloud:
Superseded by https://github.com/obot-platform/obot/pull/7467 — the branch was renamed `darren/*` → `feat/*`, which closed this PR. Same commit, no content change.

PR #7457, opened 2026-08-03
feat: add the hosted agents user UI (10/11)

ibuildthecloud:
A user-facing view for launching an agent and using it.

The terminal is xterm over the same websocket the API serves, and an agent that
serves HTTP is shown in a frame pointed at its published path.

---

Part 10 of an 11-PR stack. Base: `darren/hosted-agents-09-admin-ui`.

## Stack

1. #7448 — Resource model
2. #7449 — Sandbox backend interface
3. #7450 — Kubernetes sandbox backend
4. #7451 — Sandbox reachability (models + MCP)
5. #7452 — Terminal + HTTP publish
6. #7453 — Server wiring + Helm
7. #7454 — Controllers / reconcilers
8. #7455 — REST API
9. #7456 — Admin UI
**→ 10. #7457 — User UI** (this PR)
11. #7458 — Seed default config source

Merge bottom-up; GitHub retargets each child when its parent merges.

PR #7458, opened 2026-08-03
feat: seed a default config source for hosted agents (11/11)

ibuildthecloud:
A fresh installation had no harnesses or templates and no obvious way to get
any: an administrator had to know the catalog repository existed and add it by
hand before anything could be launched. It is now seeded the same way the
default skill repository is, from a flag, so the samples are simply there.

Seeding happens only when the installation has no catalogs at all -- the
existing proxy for "has this server run before" -- so an administrator who
removes it does not find it back on the next restart.

The seed values move into a struct rather than growing the positional argument
list. They are all strings, and a transposed pair would not fail at the call
site; it would surface much later as a repository that does not sync.

A local checkout is accepted as the source only in development mode, where
pointing Obot at a working copy of the catalog is how the samples get iterated
on.

---

Part 11 of an 11-PR stack. Base: `darren/hosted-agents-10-user-ui`.

## Stack

1. #7448 — Resource model
8. #7455 — REST API
9. #7456 — Admin UI
10. #7457 — User UI
**→ 11. #7458 — Seed default config source** (this PR)


ibuildthecloud:
Superseded by https://github.com/obot-platform/obot/pull/7469 — the branch was renamed `darren/*` → `feat/*`, which closed this PR. Same commit, no content change.

PR #7459, opened 2026-08-03
feat: add the hosted agent resource model (1/11)

ibuildthecloud:
Hosted agents are agents Obot runs on the user's behalf, in a sandbox it
manages. Five resources describe them, and the split is what keeps a template
reusable across installations:

  Harness             a runtime image an agent runs on
  HostedAgent         a template: the harness, the models, MCP servers and
                      skills it may use, and what the user is asked
  HostedAgentInstance one user's running copy of a template
  HostedAgentPool     the compute a set of instances shares
  AgentCatalog        a git source that supplies harnesses and templates

A template is written once and synced everywhere, so it names models by
provider and MCP servers and skills by a portable reference rather than by an
ID the local installation generated. An instance is the only resource with a
user on it; everything else is installation-wide.

HostedAgentAccessRule follows the shape of the existing MCP access rules, so
granting access to an agent is the same idea administrators already know.

---

Part 1 of an 11-PR stack. Base: `main`.

## Stack

**→ 1. #7459 — Resource model** (this PR)
8. #7466 — REST API
9. #7467 — Admin UI
10. #7468 — User UI


apiclient/types/agentcatalog.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (m AgentCatalogManifest) Validate() error {
+	if strings.TrimSpace(m.DisplayName) == "" {
+		return fmt.Errorf("displayName is required")
+	}
+	if strings.TrimSpace(m.RepoURL) == "" {
+		return fmt.Errorf("repoURL is required")
+	}
+	return ValidateAgentCatalogURL(m.RepoURL)
+}
```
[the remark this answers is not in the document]
ibuildthecloud: Fixed in `fa6255ba`. You're right, and the gap was live rather than theoretical.

`readAndValidateAgentCatalogManifest` trims `Ref` and then never validates it, and the value flows straight through to `fetcher.Fetch(req.Ctx, source.Spec.RepoURL, source.Spec.Ref)`. The only caller that validated a ref was the seeded default in `data.go`, which called `ValidateGitRef` separately — so an admin-supplied ref through `POST /api/agent-catalogs` reached git unchecked.

What made it clearly wrong rather than merely inconsistent: `HostedAgentManifest` and `HostedAgentInstanceManifest` both validate their own `GitRef` inside `Validate()`. `AgentCatalogManifest` was the only one of the three that didn't, so callers had to remember a rule its siblings enforce for them.

`Validate()` now ends with `ValidateGitRef(m.Ref)`. That covers whitespace-only refs too, since `ValidateGitRef` rejects any character outside `[A-Za-z0-9._/+-]` — a ref of `"   "` now fails rather than being silently accepted.

Two knock-on changes:

- `pkg/agentcatalog.Validate` relaxes the *URL* rule in development mode by bypassing `manifest.Validate()`, which would have bypassed the ref check with it. It now validates the ref explicitly on that path, so accepting a local checkout doesn't also accept an arbitrary ref alongside it. (`b2876d38`)
- The separate `ValidateGitRef` call in `createDefaultAgentCatalog` is now unreachable, since `agentcatalog.Validate` rejects a bad ref first. Removed it rather than leaving dead code. That surfaced as a real test failure — `TestCreateDefaultAgentCatalog/invalid_ref_returns_error` asserted on the old `"invalid default agent catalog ref"` prefix — so the assertion was updated to match the error that is now actually produced. (`b18159a6`)

Test cases added for a ref beginning with `-` (`--upload-pack=evil`), a whitespace-only ref, a ref containing `..`, and a valid `release/v1.2`, plus the development-mode path.

apiclient/types/agentcatalog.go
```diff
@@ -0,0 +1,101 @@
...
+		return fmt.Errorf("%s must be an https URL", field)
+	}
+	if u.Host == "" {
+		return fmt.Errorf("%s must include a host", field)
+	}
+	if strings.Trim(u.Path, "/") == "" {
+		return fmt.Errorf("%s must include a repository path", field)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
ibuildthecloud: Fixed in `fa6255ba`. Both halves were valid, and the credentials one matters more than it first looks.

**Embedded credentials.** `validateRepoURL` backs both `ValidateAgentCatalogURL` and `ValidateGitRepoURL`, and the latter validates `gitRepo` on `HostedAgentInstanceManifest` — a *user*-supplied field. A URL like `https://user:token@host/repo.git` would have been persisted verbatim on the resource and repeated in any error mentioning it. `validateRepoURL` now rejects `u.User != nil`. There's nothing lost: repository credentials are configured separately, so nothing expressible this way was only expressible this way.

**`.git` boundary.** Added, mirroring `NormalizeRepositoryURL`: `github.com` and `gitlab.com` require `owner/repo`, and any other host requires a `.git` suffix or a `.git/` path segment. Factored into `validateRepoBoundary` so both entry points get it.

One thing worth recording, since it's the reason the duplication you spotted exists at all: `apiclient` is a separate Go module and cannot import `pkg/git`, where `IsGitRepoURL` lives. So this can't call the existing helper — it has to restate the rule. I've said so in a comment on `validateRepoBoundary` and noted the two must be kept in step. If that drift risk is a concern, the alternative is moving the rule into `apiclient/types` and having `pkg/git` call *it*, which would make `apiclient` the single source of truth. Happy to do that in a follow-up if you'd prefer; it touches skill repositories, so it felt out of scope for this PR.

I checked the stricter rule against every existing fixture before applying it — each `repoURL`/`gitRepo` literal in the tree is either `.git`-suffixed or a `github.com/owner/repo`, so nothing that used to validate now fails. New test cases cover embedded credentials, a non-`.git` path on an arbitrary host, and `github.com` with an owner but no repository.

The `.git` requirement is the more debatable half — it rejects `https://git.example.com/team/project` for a self-hosted server that serves it without the suffix. That constraint already applies to skill repositories, so this makes the two consistent, but it's a one-line relaxation if it turns out to be too strict in practice.

calvinmclean (approved):

apiclient/types/agentcatalog.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+		return nil
+	}
+	if !strings.HasSuffix(path, ".git") && !strings.Contains(path, ".git/") {
+		return fmt.Errorf("%s path must include a .git repository boundary", field)
+	}
+	return nil
+}
+
+type AgentCatalogList List[AgentCatalog]
```
thedadams, line 138: I'm surprised no new line here passes `go fmt`.

apiclient/types/gitref_test.go
```diff
@@ -0,0 +1,66 @@
...
+import "testing"
+
+// The ref reaches git as an argument, so the validator is what stands between a
+// user-supplied string and git's option parser.
+func TestValidateGitRef(t *testing.T) {
+	for _, tt := range []struct {
+		ref  string
+		want bool
+	}{
+		{"", true},
```
thedadams: We really need to add something to the CLAUDE.md file to put struct field names here and put each field on its own line.

apiclient/types/hostedagent.go
```diff
@@ -0,0 +1,331 @@
...
+	HostedAgentStateError   HostedAgentState = "error"
+)
+
+func (m HostedAgentManifest) Validate() error {
+	if m.Name == "" {
+		return fmt.Errorf("name is required")
+	}
+	if m.HarnessID == "" {
+		return fmt.Errorf("harnessID is required")
+	}
```
thedadams: We are inconsistent here with trimming space (we do it in other files). Not sure if it matters.

apiclient/types/user.go
```diff
@@ -49,6 +49,11 @@ const (
 	GroupTunnel             = "obot-tunnel"
 	GroupTunnelBridge       = "obot-tunnel-bridge"
 	GroupTunnelPeer         = "obot-tunnel-peer"
+	// GroupHostedAgent marks a principal that is a hosted agent sandbox rather
+	// than a person. It never accompanies GroupAPI: an agent reaches the MCP
+	// gateway and the model proxy and nothing else, so a stolen credential
+	// cannot enumerate or modify Obot resources.
+	GroupHostedAgent = "hosted-agent"
```
thedadams: Sounds like you can use `[]string{GroupMCP, GroupLLM}` for this instead of having a new group.

thedadams (approved):

apiclient/types/agentcatalog.go
ibuildthecloud: No change needed — `gofmt` is satisfied here.

There is a blank line between the closing `}` of `validateRepoBoundary` and `type AgentCatalogList`, and the file ends with a newline (`…AgentCatalog]` + `0x0a`). `gofmt -l` on this file is silent, and the `gofmt`/`goimports` formatters are enabled in `.golangci.yml` and pass on every rung of the stack.

My guess is the diff view collapsed the blank line. If you are seeing something concrete that `gofmt` disagrees with, point me at it and I will fix it.

apiclient/types/hostedagent.go
ibuildthecloud: Fixed in `fb16ecd9`. `HostedAgentManifest.Validate` now trims before the emptiness check on both fields:

```go
if strings.TrimSpace(m.Name) == "" { … }
if strings.TrimSpace(m.HarnessID) == "" { … }
```

matching what `AgentCatalogManifest.Validate` already did for `displayName` and `repoURL`. It does matter here: a name of `"   "` was previously accepted and then used to build resource names.

apiclient/types/gitref_test.go
ibuildthecloud: Fixed in `fb16ecd9` — all 15 cases now name their fields:

```go
{ref: "main", want: true},
{ref: "--upload-pack=touch /tmp/pwned", want: false},
```

I read "each field on its own line" as one case per line with named fields rather than fully expanding each literal to four lines; with 15 cases the expanded form pushed the table past a screen and made the pattern harder to scan. Say the word and I will expand it.

On the CLAUDE.md convention: agreed, but I have deliberately left it out of this stack. A repo-wide convention change belongs in its own PR rather than riding along with a feature — same reasoning as a lint-rule change. Happy to open one.

apiclient/types/user.go
ibuildthecloud: Done in `fb16ecd9` — `GroupHostedAgent` is gone. It needed one adjustment to keep working, worth recording.

The group had a second job beyond access. `principal.IsHostedAgent` read it, and two call sites branch on that to take a *different authorization path*:

- `mcpid.go` — a sandbox's MCP access is decided by its grant list, because access-control rules describe people and would never match it
- `llmproxy.go` — same for models

So dropping the group and keeping only `GroupMCP`/`GroupLLM` would have given the right access but lost the marker, silently collapsing those two paths into the user-shaped ones.

The marker is now the `HostedAgentOwnerExtra` the principal already carries — set in `authenticateHostedAgent`, and already what `ResourceOwnerID` depends on. Only a sandbox is issued on behalf of someone else, so it identifies one exactly, and it is a fact we were already recording rather than a new one:

```go
func IsHostedAgent(user kuser.Info) bool {
	if user == nil {
		return false
	}
	owner := user.GetExtra()[HostedAgentOwnerExtra]
	return len(owner) > 0 && owner[0] != ""
}
```

`hostedAgentGroups` is now exactly the access set: `GroupAuthenticated`, `GroupLLM`, `GroupSkills`, and `GroupMCP` when the agent has servers. `GroupAPI` stays absent, and the test asserting that still passes.

Added a case that a person holding `GroupMCP` + `GroupLLM` does **not** read as an agent — that one fails under the literal reading of the suggestion, which is what convinced me the marker had to move rather than disappear.

PR #7460, opened 2026-08-03
feat: add an interface for running agent sandboxes (2/11)

ibuildthecloud:
The rest of Obot asks for a sandbox and gets back its state; where it actually
runs is behind this interface. That keeps Kubernetes types out of the
controllers and the API, so a second backend is a new implementation rather
than a change to everything that calls one.

Three implementations ship: disabled, which is the default and refuses
politely; fake, which transitions on a timer so the UI and the controllers can
be developed and tested without a cluster; and the Kubernetes one that follows.

Sizing is here rather than in a backend because it is policy, not mechanism: a
pool's capacity and an instance's share of it mean the same thing wherever the
sandbox runs.

---

Part 2 of an 11-PR stack. Base: `feat/hosted-agents-01-resource-model`.

## Stack

1. #7459 — Resource model
**→ 2. #7460 — Sandbox backend interface** (this PR)
8. #7466 — REST API
9. #7467 — Admin UI
10. #7468 — User UI


pkg/agentbackend/fake/backend.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+func cloneInstance(value agentbackend.DesiredInstance) agentbackend.DesiredInstance {
+	value.Prompt = append([]string(nil), value.Prompt...)
+	value.Files = append([]agentbackend.File(nil), value.Files...)
+	for i := range value.Files {
+		value.Files[i].Content = append([]byte(nil), value.Files[i].Content...)
+	}
+	value.Secrets = append([]agentbackend.SecretRef(nil), value.Secrets...)
+	if value.Env != nil {
```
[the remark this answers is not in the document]
ibuildthecloud: Fixed in `1ac4aa05`. Correct on both counts — the values were being retained, and the fake has no use for them.

The fix isn't in `cloneInstance` though, and the reason is worth recording because doing it there would have introduced a worse bug.

`cloneInstance`'s result is stored in `current.desired`, and the line above it decides whether anything changed with `!reflect.DeepEqual(current.desired, desired)`. Clearing `Value` only on the way into the store would leave a redacted copy being compared against an unredacted argument, so **every** reconcile would look like a change: `generation++`, state back to `Pending`, and the sandbox restarting forever on a loop.

So the redaction happens at the top of `ReconcileInstance` instead, against the incoming value, which makes the comparison and the store consistent:

```go
desired = desired.Redacted()
```

Nothing is lost by comparing redacted forms. `SecretRef.Version` changes whenever the value behind it changes — that is exactly what it's documented to be for — so a rotated secret is still seen as a change and still restarts the sandbox. Reusing the existing `Redacted()` helper also means the fake and the real contract stay in step rather than having a second hand-rolled clearing loop.

One thing you didn't flag but which made this worse: `TestSecretReferencesAreCopiedWithoutValues` **never set `Value`**. It only asserted that the secret slice wasn't aliased to the caller's, so its name promised coverage it didn't have — which is presumably why this survived. It now sets a real value and asserts three things: the value is cleared, the version is preserved, and `generation` is still 1 after reconciling identical desired state twice. That last assertion is the regression test for the restart loop described above, and it fails against the naive fix.

Added `TestRotatedSecretIsSeenAsAChange` alongside it, asserting that bumping `Version` from `v1` to `v2` does bump `generation` — so the redaction can't later be "optimised" into ignoring rotations.

calvinmclean (approved):

pkg/agentbackend/backend.go
```diff
@@ -0,0 +1,263 @@
...
+	// revision by DesiredInstance.Redacted, and must never be persisted to
+	// instance spec or status, or logged.
+	Value string
+}
+
+// Redacted returns a copy with every secret value cleared, for hashing into the
+// desired revision or for logging. Version still varies with the value, so a
+// rotation changes the revision without the revision ever containing a secret.
+func (d DesiredInstance) Redacted() DesiredInstance {
+	d.Secrets = append([]SecretRef(nil), d.Secrets...)
```
thedadams: `slices.Clone`

thedadams (approved):

pkg/agentbackend/backend.go
ibuildthecloud: I'm surprised linting didn't catch this. Do you not have this rule on? There's something like "modernize" that will catch these.

ibuildthecloud: Answering the linting question, since I looked into it and never reported back here.

**No linter available in golangci-lint v2.12.2 catches this.** I ran `--default=all` — every linter it ships — against a file containing the pattern, and got nothing but revive doc-comment noise. I also tested three spellings specifically under `modernize`, all clean:

```go
append([]S(nil), s...)   // ours
append(s[:0:0], s...)    // the "clipped" form appendclipped documents
append([]S{}, s...)      // empty-literal form
```

The modernize suite bundled in that version (from `golang.org/x/tools@v0.44.0`) is: `appendclipped, bloop, fmtappendf, forvar, mapsloop, minmax, newexpr, omitzero, rangeint, slicescontains, sortslice, stringsbuilder, stringscutprefix, testingcontext, waitgroup`. There is no `slicesclone`, and `appendclipped` — the one that ought to cover it — does not fire through golangci-lint in any of my three tests. gopls may still surface it in an editor; that I could not verify here.

So this was a genuine human catch, not a gap we can close by flipping a switch.

That said, `modernize` would earn its keep for other reasons: it finds 6 real issues in `pkg/agentbackend` alone (`mapsloop` → `maps.Copy`, plus several `newexpr`), and it is what would have caught the `ptr` → `new` cleanup on #7461. It is not enabled — `.golangci.yml` uses `default: none` with an explicit list. I have deliberately kept lint config out of this stack, since enabling a rule reformats or flags code repo-wide and belongs in its own PR.

The fix itself is in `39ab4390`, applied to all four instances of the pattern rather than just the flagged line.

PR #7461, opened 2026-08-03
feat: run agent sandboxes on Kubernetes (3/11)

ibuildthecloud:
An instance becomes a Deployment, and a pool becomes the volume and priority
class its instances share. Instances are separated by a subPath on the pool's
volume rather than by a volume each, so a pool is one claim however many
instances it holds.

A sandbox is told nothing about Kubernetes: it reads its configuration from
files and serves on a port. What is here is the translation -- desired state to
objects, object status back to instance status, and utilization from the
metrics API when it is available, degrading to requested resources when it is
not.

Two details exist because Obot usually runs outside the cluster in development.
ReachableServerURL rewrites a loopback address to the node's, since a sandbox
resolving "localhost" would find itself; and the dev proxy reaches a sandbox
through the API server, which is deliberately confined to development because
it grants access to every Service in the cluster.

---

Part 3 of an 11-PR stack. Base: `feat/hosted-agents-02-backend-interface`.

## Stack

1. #7459 — Resource model
**→ 3. #7461 — Kubernetes sandbox backend** (this PR)
8. #7466 — REST API
9. #7467 — Admin UI
10. #7468 — User UI


pkg/agentbackend/kubernetes/backend.go
```diff
@@ -0,0 +1,301 @@
...
+// DeletePool.
+func (b *Backend) namespaceOwner(ctx context.Context) (*metav1.OwnerReference, error) {
+	b.namespaceOnce.Do(func() {
+		var namespace corev1.Namespace
+		err := b.client.Get(ctx, kclient.ObjectKey{Name: b.opts.Namespace}, &namespace)
+		if apierrors.IsNotFound(err) {
+			// Hosted agents can be the only Kubernetes feature enabled, in which
+			// case nothing else has created this namespace.
+			created := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: b.opts.Namespace}}
+			if createErr := b.client.Create(ctx, created); createErr != nil && !apierrors.IsAlreadyExists(createErr) {
```
thedadams, line 260: We need to make sure there is a role/role binding for this.

pkg/agentbackend/kubernetes/backend.go
```diff
@@ -0,0 +1,301 @@
...
+}
+
+func poolLabels(poolID string) map[string]string {
+	return map[string]string{
+		managedLabel: managedValue,
+		poolLabel:    sanitize(poolID),
+	}
+}
+
+func ignoreNotFound(err error) error {
```
thedadams: Not used. And such a thing already exists in the controller runtime client package.

pkg/agentbackend/kubernetes/podsecurity.go
```diff
@@ -0,0 +1,100 @@
...
+// podSecurityContext returns the pod-level context for the configured level.
+//
+// FSGroup and FSGroupChangePolicy are set at every level, including
+// privileged: they are not admission requirements but are what gives a sandbox
+// write access to its own subdirectory of the shared pool volume without
+// recursively chowning the whole volume on each start.
+func (b *Backend) podSecurityContext() *corev1.PodSecurityContext {
+	securityContext := &corev1.PodSecurityContext{
+		FSGroup:             ptr(b.opts.FSGroup),
+		FSGroupChangePolicy: ptr(corev1.FSGroupChangeOnRootMismatch),
```
thedadams: We live in a modern world.
```suggestion
		FSGroup:             new(b.opts.FSGroup),
		FSGroupChangePolicy: new(corev1.FSGroupChangeOnRootMismatch),
```

pkg/agentbackend/kubernetes/pool.go
```diff
@@ -0,0 +1,248 @@
...
+	if err := b.cachedClient.Get(ctx, kclient.ObjectKey{Name: poolName(ref.ID), Namespace: b.opts.Namespace}, &quota); apierrors.IsNotFound(err) {
+		return observation, nil
+	} else if err != nil {
+		return agentbackend.PoolObservation{}, fmt.Errorf("failed to read pool quota for pool %s: %w", ref.ID, err)
+	}
+
+	observation.Ref.BackendID = poolName(ref.ID)
+	observation.Exists = true
+	observation.ObservedRevision = quota.Annotations[revisionAnnotation]
+	observation.BackendGeneration = fmt.Sprintf("%d", quota.Generation)
```
thedadams: Why not just keep this as an `int64`?

pkg/agentbackend/kubernetes/pool.go
```diff
@@ -0,0 +1,248 @@
...
+	memory := quota.Spec.Hard[corev1.ResourceRequestsMemory]
+	return agentbackend.ResourceQuantity{
+		CPUVCPUs:    float64(cpu.MilliValue()) / 1000,
+		MemoryBytes: memory.Value(),
+	}
+}
+
+func ptr[T any](value T) *T {
+	return &value
+}
```
thedadams:
```suggestion
```

pkg/agentbackend/kubernetes/reachable.go
```diff
@@ -0,0 +1,81 @@
...
+	return parsed.String(), nil
+}
+
+func isLoopback(host string) bool {
+	if host == "localhost" {
+		return true
+	}
+	ip := net.ParseIP(host)
+	return ip != nil && ip.IsLoopback()
+}
```
thedadams, line 64: We have a similar function some place else. Hopefully I'll remember to consolidate them.

pkg/agentbackend/kubernetes/reachable.go
```diff
@@ -0,0 +1,81 @@
...
+
+// anyNodeAddress returns a node's internal address. Any node will do: the
+// developer's machine is the cluster, so every node is the same host.
+func anyNodeAddress(ctx context.Context, client kclient.Client) (string, error) {
+	var nodes corev1.NodeList
+	if err := client.List(ctx, &nodes); err != nil {
```
thedadams, line 70: Cluster Role and Binding needed.

pkg/agentbackend/kubernetes/status.go
```diff
@@ -0,0 +1,127 @@
...
+	for i := range pods {
+		if !pods[i].DeletionTimestamp.IsZero() {
+			continue
+		}
+		if newest == nil || pods[i].CreationTimestamp.After(newest.CreationTimestamp.Time) {
+			newest = &pods[i]
+		}
+	}
+	return newest
+}
```
thedadams, line 113: Same comment about this being someplace else.

pkg/agentbackend/kubernetes/utilization.go
```diff
@@ -0,0 +1,185 @@
...
+		measured, err = b.usage.PodUsage(ctx, b.opts.Namespace, poolLabel+"="+sanitize(ref.ID))
+		if err != nil {
+			return agentbackend.UtilizationSnapshot{}, fmt.Errorf("failed to read pool usage for %s: %w", ref.ID, err)
+		}
+	}
+
+	// instanceID -> measured usage, summed over that instance's running pods.
+	byInstance := make(map[string]agentbackend.ResourceUtilization, len(pods.Items))
+	for i := range pods.Items {
+		pod := &pods.Items[i]
```
thedadams: Weird. And Claude did the same for the deployment loop.

pkg/agentbackend/kubernetes/backend.go
ibuildthecloud: oh this, isn't really true anymore. I changed the helm chart so that agents backend will follow mcp backend. So you can't have just one of them turned on k8s. And I believe the helm chart creates the namespace.

ibuildthecloud: Already covered, but not in this PR — which is a fair thing to miss from here.

The ClusterRole is in `chart/templates/mcp.yaml`, added by rung 6 (#7464):

```yaml
# Needed to read the sandbox namespace's UID when building those owner
# references, and to create it when hosted agents are the only Kubernetes
# feature enabled.
- apiGroups: [""]
  resources: ["namespaces"]
  verbs: ["create", "get", "list", "watch"]
```

The Go code lands here and its RBAC lands with the rest of the chart changes five rungs later, so reviewing this PR alone genuinely looks like a missing grant. That is a real cost of the split; I did not want chart edits scattered across every rung that happens to call the API.

pkg/agentbackend/kubernetes/reachable.go
ibuildthecloud: Same as the namespace one — it exists in rung 6 (#7464), `chart/templates/mcp.yaml`:

```yaml
# Only read when Obot's own hostname is a loopback address, which is a
# developer running outside the cluster: a sandbox cannot reach 127.0.0.1, so
# a node address is substituted. Harmless to omit in production, where the
# configured hostname already resolves.
- apiGroups: [""]
  resources: ["nodes"]
  verbs: ["get", "list"]
```

Worth stressing the second half: this grant is only exercised when Obot's configured hostname is a loopback address. A production install never reaches `anyNodeAddress`, so omitting the grant degrades development convenience rather than breaking anything.

pkg/agentbackend/kubernetes/backend.go
ibuildthecloud: Fixed in `0f911716`. Removed, and the call site now uses `kclient.IgnoreNotFound` from controller-runtime.

Small correction: it was used, once — `instance.go`, on the cleanup-job delete. So it was not dead code, just a needless local reimplementation of something controller-runtime already exports. Your substantive point stands either way.

pkg/agentbackend/kubernetes/utilization.go
ibuildthecloud: Fixed in `0f911716`. Both loops are now `for _, pod := range pods.Items` and `for _, deployment := range deployments.Items`.

The index-and-take-a-pointer form is a pre-Go-1.22 habit: it existed because the loop variable used to be reused across iterations, so `&item` aliased. Since 1.22 the variable is per-iteration and the workaround is just noise. The two helpers that want a pointer take `&deployment` directly now.

And yes — both loops in the same file, written the same way, which is exactly the tell.

pkg/agentbackend/kubernetes/podsecurity.go
ibuildthecloud: Applied in `0f911716`, and taken further than the two lines — all **15** `ptr(…)` call sites across `instance.go`, `podsecurity.go` and `pool.go` are now `new(…)`.

Verified `new(expr)` compiles against the module's `go 1.26.5` before committing to it, since it is new enough to be worth checking rather than assuming.

pkg/agentbackend/kubernetes/pool.go
ibuildthecloud: Deleted in `0f911716`, once all 15 call sites moved to `new(…)`. Nothing references it now.

Incidentally this is what golangci-lint's `modernize` linter flags as `newexpr`, and it is not enabled here — `.golangci.yml` uses `default: none` with an explicit list. Enabling it finds 6 more in `pkg/agentbackend` alone (`mapsloop` → `maps.Copy`, plus more `newexpr`). I have not touched the lint config in this stack on purpose; happy to open that as its own PR.

pkg/agentbackend/kubernetes/pool.go
ibuildthecloud: Good catch — fixed in `fb16ecd9`, `39ab4390`, `0f911716` and `248dc4ba`.

`BackendGeneration` is `int64` end to end now. It was being stringified with `fmt.Sprintf("%d", …)` and then never parsed back, so the string form bought nothing and lost the type.

It reached further than this file, which is why it spans four rungs:

| Rung | Change |
|---|---|
| 1 | `int64` on the API type and the CRD status, plus openapi regen |
| 2 | Both observation structs; dropped the fake's `generation()` conversion |
| 3 | `observation.BackendGeneration = deployment.Generation` / `quota.Generation` |
| 7 | A test literal `"3"` → `3` |

On the generated code: I edited the two openapi schema entries to `integer`/`int64` and then ran `go generate ./...` to check — it produced exactly the same four lines, so what is committed is what the generator emits rather than a hand-patch that would drift.

pkg/agentbackend/kubernetes/reachable.go
ibuildthecloud: Found it, so it is written down rather than depending on either of us remembering.

There is no exact twin of `anyNodeAddress` — nothing else in the repo reads `NodeInternalIP`. The related function is `transformObotHostname` in `pkg/mcp/kubernetes.go`, which solves the same problem (make Obot's address reachable from inside the cluster) by a different route: it swaps in the Service address, where this substitutes a node address for a loopback host.

They are close enough to belong together and different enough that merging them is a real refactor, not a move. I have left it alone deliberately: consolidating means editing `pkg/mcp` from a hosted-agents rung, which puts an unrelated subsystem into this diff. Happy to do it as a follow-up once the stack lands — say the word and I will open it.

Worth noting `services/config.go` in rung 6 now calls both: `ReachableServerURL` first, falling back to `TransformObotHostname`. So they are already composed at the call site, which is probably the shape a consolidation should take.

pkg/agentbackend/kubernetes/status.go
ibuildthecloud: This one has an exact counterpart: `getNewestPod` in `pkg/mcp/kubernetes.go`.

Not identical, and the differences matter if they are merged:

| | `newestPod` (here) | `getNewestPod` (`pkg/mcp`) |
|---|---|---|
| Terminating pods | skipped | included |
| Not found | returns `nil` | returns an error |

The skip is load-bearing here — a terminating pod is exactly what a sandbox mid-restart has, and attaching a terminal to it or reading status from it gives the wrong answer.

Same as the other one: leaving it, because consolidating means editing `pkg/mcp` from this stack. Ready to do it as a follow-up.

`reasonOr`/`messageOr` just below have no counterpart anywhere — the nearest relative is `firstNonEmpty`, which exists twice already (`pkg/api/handlers/imagepullsecrets.go` and `pkg/controller/handlers/imagepullsecret/imagepullsecret.go`), so that is a third candidate for the same cleanup.

thedadams (approved):

pkg/agentbackend/kubernetes/backend.go
ibuildthecloud: Following up on my own earlier reply — I answered where the RBAC lives and never engaged with the point above it, which was the substantive one.

Checked, and it is right: `chart/templates/mcp.yaml` is gated in its entirety on `obot.mcpOnKubernetes`, namespace and RBAC alike, and the chart exposes no agent-backend value at all. So agents on Kubernetes implies MCP on Kubernetes implies the chart already created the namespace. The comment justifying the create-if-missing path with "hosted agents can be the only Kubernetes feature enabled" describes a configuration the chart can no longer produce.

I have left the create itself in place, because `pkg/mcp/manager.go` does the same thing for the same namespace — so removing it here would make hosted agents the odd one out, and it still matters for an install that does not use the chart. What I have corrected is the stale justification.

The `namespaces: create` grant stays for that reason too: the MCP manager needs it regardless of what hosted agents do.

PR #7462, opened 2026-08-03
feat: let a sandbox reach the models and MCP servers it was granted (4/11)

ibuildthecloud:
A sandbox authenticates as itself. Its credential carries a group that marks it
as a hosted agent rather than a person, and that group reaches the MCP gateway
and the model proxy and nothing else -- so a credential taken out of a sandbox
cannot enumerate or modify Obot resources.

An agent is authorized against the servers its template was granted rather than
against a user's access, because the agent is not acting as the user: the
template says what it may use, and that list is the whole of it.

Model resolution is shared between the API, which decides whether an agent can
be launched, and the controller, which writes the endpoints into a sandbox's
configuration. If those two disagreed a user would be offered an agent that
then resolves no models. Portable references resolve a template's "<source>::
<key>" naming to whatever local ID an installation generated, which is what
lets one template work everywhere.

---

Part 4 of an 11-PR stack. Base: `feat/hosted-agents-03-kubernetes-backend`.

## Stack

1. #7459 — Resource model
**→ 4. #7462 — Sandbox reachability (models + MCP)** (this PR)
8. #7466 — REST API
9. #7467 — Admin UI
10. #7468 — User UI


pkg/api/authz/hostedagent.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"net/http"
+
+	"github.com/obot-platform/nah/pkg/router"
+	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
+	"github.com/obot-platform/obot/pkg/system"
+)
+
+func (a *Authorizer) checkHostedAgent(req *http.Request, resources *Resources, u User) (bool, error) {
+	if resources.HostedAgentID == "" || u.IsAdmin || u.IsAuditor {
```
thedadams: In general, admins/auditors shouldn't be able to actually use the hosted agent. Only be able to get/list/update in the API.

pkg/api/authz/hostedagent.go
```diff
@@ -0,0 +1,52 @@
...
+
+	var instance v1.HostedAgentInstance
+	if err := a.get(req.Context(), router.Key(system.DefaultNamespace, resources.HostedAgentInstanceID), &instance); err != nil {
+		return false, err
+	}
+
+	if u.IsAdmin || u.IsAuditor {
+		resources.Authorizated.HostedAgentInstance = &instance
+		return true, nil
+	}
```
thedadams: Same comment here: admins and auditors should be able to see these instances, but shouldn't be able to actually interact with the agent.

pkg/gateway/server/apikey_auth.go
```diff
@@ -17,11 +25,98 @@ const apiKeyAuthPrefix = "ok1-"
...
+//
+// GroupAPI is deliberately absent and must stay absent: it is the only thing
+// keeping an exfiltrated agent credential away from the Obot API, since
+// authorization elsewhere decides which servers and models an agent may use but
+// nothing else restricts which endpoints accept it.
+func hostedAgentGroups(hasMCPServers bool) []string {
+	groups := []string{
+		types2.GroupAuthenticated,
+		types2.GroupLLM,
+		types2.GroupSkills,
```
ibuildthecloud: **[P1] Restrict sandbox credentials to their configured skills.** Granting `GroupSkills` lets this credential call the skills API, but the principal carries no configured skill IDs and `checkSkill` consequently applies the ordinary wildcard access rule. With the default `sar1-everything` rule, an exfiltrated sandbox key can enumerate, preview, and download every skill, including content that the handler materializes using server-side private-repository credentials. This also contradicts the stated boundary that the credential reaches the MCP gateway and model proxy and nothing else. Please remove this group or add an `authorized_skill_ids` check for hosted-agent principals.

pkg/gateway/server/apikey_auth.go
```diff
@@ -17,11 +25,98 @@ const apiKeyAuthPrefix = "ok1-"
...
+		Namespace: system.DefaultNamespace,
+	}, &agent); err != nil {
+		return nil, false, nil
+	}
+
+	// Template resources are granted by the administrator who published the
+	// agent; instance resources were checked against the owner when they were
+	// attached. Neither is re-checked against the owner here, by design.
+	mcpIDs := slices.Concat(agent.Spec.Manifest.MCPServers, instance.Spec.Manifest.MCPServers)
+	modelIDs := a.resolveAgentModelIDs(req.Context(),
```
ibuildthecloud: **[P1] Apply `ModelProviders` when constructing the authorization grant.** This resolver receives only the combined model selections, so it disagrees with `hostedagentmodels.Resolve`, which filters them through `agent.Spec.Manifest.ModelProviders`. In particular, `"*"` is preserved as an unrestricted wildcard, and explicitly selected models from excluded providers are also retained. The generated sandbox configuration may omit those models, but the credential can still call them directly through the proxy. Please reuse the shared resolver here (including provider restrictions) and store only its concrete authorized IDs.

pkg/gateway/server/apikey_auth.go
```diff
@@ -17,11 +25,98 @@ const apiKeyAuthPrefix = "ok1-"
...
+		Name:      instance.Spec.HostedAgentName,
+		Namespace: system.DefaultNamespace,
+	}, &agent); err != nil {
+		return nil, false, nil
+	}
+
+	// Template resources are granted by the administrator who published the
+	// agent; instance resources were checked against the owner when they were
+	// attached. Neither is re-checked against the owner here, by design.
+	mcpIDs := slices.Concat(agent.Spec.Manifest.MCPServers, instance.Spec.Manifest.MCPServers)
```
ibuildthecloud: **[P1] Resolve portable MCP references before placing them in the principal.** Templates intentionally keep `<source>::<key>` references, while the sandbox controller resolves those to installation-local IDs before generating `/mcp-connect/<id>` URLs. Copying the raw template values here means authorization compares the local request ID against an unresolved portable reference and denies the configured server. Please run the combined list through `hostedagentrefs` here so the credential grant and sandbox configuration use the same IDs.

ibuildthecloud (commented):
Three authorization issues found in a fresh pass:

pkg/gateway/server/apikey_auth.go
ibuildthecloud: Fixed in `da6f13b6` — removed, rather than adding a skill-ID check.

A sandbox never calls the skills API. Its skills are written to `/etc/obot/skills/<name>/` and the configuration hands it a `path`, not a URL — `agentConfigSkill` has no URL field at all. So the group granted a capability nothing uses, while `checkSkill` fell through to `UserHasAccessToSkill`, which under the seeded `sar1-everything` rule matches the wildcard subject and returns every skill.

You are also right that it contradicted the comment four lines above it. `hostedAgentGroups` now says why the absence is deliberate, alongside the existing note about `GroupAPI`, and the test asserts both are absent rather than only `GroupAPI`.

pkg/gateway/server/apikey_auth.go
ibuildthecloud: Fixed in `da6f13b6`. The grant now runs through the same resolver the controller uses:

```go
mcpIDs, unresolved := hostedagentrefs.New(a.storage, system.DefaultNamespace).
    MCPServers(req.Context(), slices.Concat(agent.Spec.Manifest.MCPServers, instance.Spec.Manifest.MCPServers))
```

Confirmed the failure you describe: `MCPIDIsAuthorized` compares with `slices.Contains`, so a `<source>::<key>` reference can never equal the local ID the request carries, and the agent is denied the server it was configured with.

Worth adding that it was masked whenever the combined list contained `"*"`, since that short-circuits the check — which is probably why it had not been seen. An unresolved reference names nothing installed here and so grants nothing, which is the same conclusion the controller reaches when it logs and drops one.

pkg/gateway/server/apikey_auth.go
ibuildthecloud: Fixed in `da6f13b6`. The bespoke resolver is gone -- 38 lines -- and the grant is now computed by the shared one, provider restrictions included:

```go
models, err := hostedagentmodels.Resolve(req.Context(), a.storage, system.DefaultNamespace,
    slices.Concat(agent.Spec.Manifest.Models, instance.Spec.Manifest.Models),
    agent.Spec.Manifest.ModelProviders)
```

`Model.ID` is `model.Name`, which is exactly what `modelAllowedForAgent` compares against, so the grant is concrete IDs and no wildcard survives into the principal.

Provider restrictions come from the template alone, matching `modelConfigs`: an instance selects among what its template allows rather than beyond it.

Both halves of what you flagged were real. `"*"` reaching the principal verbatim is the sharper one, because `modelAllowedForAgent` treats it as allow-all — so a template restricted to one provider still yielded a credential that could call every model in the installation, and the configuration omitting them is no restriction when the credential can call the proxy directly.

Two tests pin it shut, both of which fail against the old code:

- `TestWildcardGrantIsConcreteAndProviderRestricted` — `"*"` with providers restricted to Anthropic must not reach the grant verbatim, must include the allowed model, must exclude the other provider's, and `modelAllowedForAgent` must reject it
- `TestExplicitModelFromExcludedProviderIsNotGranted` — naming a model is a selection, not an override of the provider restriction

One behaviour change beyond what you asked for, worth calling out: the old resolver passed non-alias entries through untouched, so a typo or a deleted model stayed in the grant. `Resolve` drops anything that resolves to nothing, so the grant now names only models that exist. I think that is right -- a grant should not contain IDs that match nothing -- but it is a change, not just a move.

This is the third of the three findings, and all three had the same root cause: the credential grant was computed by a second implementation that disagreed with the one building the sandbox config. Disagreeing upward is an escalation, downward is a denial, and we had one of each. Both now go through `hostedagentmodels` and `hostedagentrefs` -- which is what the former's package doc always said it was for.

pkg/api/authz/hostedagent.go
ibuildthecloud: Partly fixed in `da6f13b6`, and partly already the case — worth separating the two.

**Launching.** An administrator already cannot launch an agent no access rule grants them. `Create` calls `accessRuleHelper.UserHasAccessToHostedAgent` unconditionally, and the helper is purely rule-based with no admin or auditor exemption. So the behaviour you are asking for is what the code does; I checked before changing anything rather than adding a redundant guard.

**Revealing.** `POST /{id}/reveal` hands back the configured secret values, and the short-circuit here let an auditor call it. Now gated:

```go
if u.IsAdmin || (u.IsAuditor && !revealsSecrets(req)) {
    return true, nil
}
```

An administrator manages what an installation offers, so reading and editing one stays theirs. An auditor's business is what happened, not what the credentials are.

The gate keys on `req.Pattern`, following `pkg/api/authz/ui.go` which already does this. There is otherwise no "use" surface on the template itself — the routes are list/create/get/update/delete/reveal, and using an agent means launching an instance, which is the check above.

pkg/api/authz/hostedagent.go
ibuildthecloud: Fixed in `da6f13b6`, and I think it was worse than "use".

Two of the routes reaching this check are not management at all:

- `GET /api/hosted-agent-instances/{id}/terminal` — attaches to a live console
- `/agent-connect/{id}` and `/{rest...}` — proxies straight through to whatever the agent serves

So an **auditor**, a deliberately read-only role, could open an interactive shell inside another user's sandbox and type into it. That is not read-only, and it is not really "using the agent" either — it is someone else's working session.

Both now fall through to the ownership check:

```go
if (u.IsAdmin || u.IsAuditor) && !entersSandbox(req) {
    resources.Authorizated.HostedAgentInstance = &instance
    return true, nil
}
```

Seeing that an instance exists, and deleting one, stays administration. `hostedagent_test.go` covers which patterns count, including that ordinary management routes and lookalikes such as `/utilization` do not.

thedadams (approved):

PR #7463, opened 2026-08-03
feat: attach a terminal to a sandbox and publish agents that serve HTTP (5/11)

ibuildthecloud:
An interactive agent gets a terminal over a websocket: framed channels carry
input, output and resize, and session errors travel on a control channel rather
than on stderr, which is reserved as a stream a non-TTY session can use for
what the program actually wrote there.

An agent that declares a port is published under /agent-connect/{instance id}
instead. Obot strips that prefix before forwarding, so an agent that assumed it
was at the root would build links landing outside its own prefix -- a
single-page UI would call Obot instead of itself. The prefix is sent per
request as X-Forwarded-Prefix and is also in the sandbox's configuration, for
servers that need it at start-up rather than per request, and relative
redirects are rewritten back inside the prefix idempotently so an agent that
already accounts for it is not prefixed twice.

---

Part 5 of an 11-PR stack. Base: `feat/hosted-agents-04-reachability`.

## Stack

1. #7459 — Resource model
**→ 5. #7463 — Terminal + HTTP publish** (this PR)
8. #7466 — REST API
9. #7467 — Admin UI
10. #7468 — User UI


go.mod
```diff
@@ -91,6 +91,8 @@ require (
 	sigs.k8s.io/yaml v1.6.0
 )
 
+require github.com/moby/spdystream v0.5.1 // indirect
```
thedadams: Combine this with the require block below.

go.mod
```diff
@@ -142,6 +144,7 @@ require (
 	github.com/chai2010/gettext-go v1.0.2 // indirect
 	github.com/cloudflare/circl v1.6.3 // indirect
 	github.com/cncf/xds/go v0.0.0-20260202195803-dba9d589def2 // indirect
+	github.com/coder/websocket v1.8.15
```
thedadams: This should be in the first require block because it is a direct dependency.

pkg/api/handlers/agentconnect/agentconnect.go
```diff
@@ -0,0 +1,200 @@
...
+
+// Proxy forwards a request to the instance's sandbox.
+//
+// The authorizer has already established that the caller owns this instance
+// and still has access to the agent behind it; an unauthenticated caller never
+// reaches here, because the UI route handling redirects them to sign in first.
+func (h *Handler) Proxy(req api.Context) error {
+	// Anonymous callers reach this handler so that a browser lands on sign-in
+	// rather than a bare 403. Unlike the MCP gateway, which challenges an OAuth
+	// client, this route is opened by a person following a link.
```
thedadams: These two comments seem to contradict each other.

thedadams (approved):

go.mod
ibuildthecloud: Fixed in `4f22fba9`. Merged into the block below, alphabetically among the other indirects.

It was standalone because `go mod tidy` put it there when I added it: the k8s backend attaches a terminal over SPDY, and `k8s.io/streaming` pulls `moby/spdystream` in as an indirect. Tidy appends a new indirect requirement as its own block rather than sorting it into an existing one, and I did not tidy the layout afterwards. `go mod tidy` is happy with it in the merged position.

go.mod
ibuildthecloud: Fixed in `4f22fba9`. `github.com/coder/websocket` is now in the first require block.

It is a direct dependency — the terminal handler uses it for the browser side of the session — which is why it carried no `// indirect` marker while sitting among the indirects. Same cause as the one above: tidy placed it where the dependency was first observed and I did not correct the grouping.

pkg/api/handlers/agentconnect/agentconnect.go
ibuildthecloud: Good catch — they did contradict each other, and the doc comment was the wrong one. Fixed in `4f22fba9`.

The doc said an unauthenticated caller never reaches `Proxy`. The body immediately handles exactly that case, and the body is right: `allowAgentConnectSignIn` in the authorizer admits an unauthenticated caller to this route deliberately, and there is a test asserting it.

The doc now says so, and explains why it is safe:

> An unauthenticated one reaches here deliberately: this route is opened by a person following a link, so the authorizer admits them precisely so that the redirect below can send them to sign in. Nothing about the instance is read before that happens, so it discloses nothing -- not even whether the instance exists.

The inline comment lost the half that duplicated it and keeps only the part the doc does not cover: why this redirects rather than challenging an OAuth client the way the MCP gateway does.

PR #7464, opened 2026-08-03
feat: wire the agent backend into the server (6/11)

ibuildthecloud:
The backend is selected by configuration: disabled unless asked for, fake in
development, and Kubernetes when a cluster is available.

The Kubernetes backend needs a client for the local cluster even when MCP
servers do not run there, which is the case a developer hits first: hosted
agents on a cluster, MCP servers on Docker. It builds its own client from the
same config rather than borrowing the MCP one, so enabling hosted agents does
not switch on MCP machinery a deployment has no reason to configure.

---

Part 6 of an 11-PR stack. Base: `feat/hosted-agents-05-terminal-connect`.

## Stack

1. #7459 — Resource model
**→ 6. #7464 — Server wiring + Helm** (this PR)
8. #7466 — REST API
9. #7467 — Admin UI
10. #7468 — User UI

### DIFF
--- a/pkg/services/config.go
+++ b/pkg/services/config.go
@@ -20,9 +20,13 @@ import (
 	apiclienttypes "github.com/obot-platform/obot/apiclient/types"
 	"github.com/obot-platform/obot/logger"
 	"github.com/obot-platform/obot/pkg/accesscontrolrule"
+	"github.com/obot-platform/obot/pkg/agentbackend"
+	agentbackendfake "github.com/obot-platform/obot/pkg/agentbackend/fake"
+	agentbackendkubernetes "github.com/obot-platform/obot/pkg/agentbackend/kubernetes"
 	"github.com/obot-platform/obot/pkg/api/authn"
 	"github.com/obot-platform/obot/pkg/api/authz"
 	"github.com/obot-platform/obot/pkg/api/handlers"
+	"github.com/obot-platform/obot/pkg/api/handlers/agentconnect"
 	"github.com/obot-platform/obot/pkg/api/handlers/mcpgateway"
 	"github.com/obot-platform/obot/pkg/api/server"
 	"github.com/obot-platform/obot/pkg/api/server/audit"
@@ -80,6 +84,8 @@ import (
 
 var pkgLog = logger.Package()
 
+const leaderElectionRequestTimeout = 15 * time.Second
+
 type (
 	GatewayConfig     gserver.Options
 	AuditConfig       audit.Options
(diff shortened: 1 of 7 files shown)


pkg/services/config.go
```diff
@@ -167,29 +180,35 @@ type Services struct {
 	AuditLogger           audit.Logger
 	DSN                   string
 	ServerURL             string
-	MCPSessionManager     *mcp.SessionManager
-	TunnelManager         *tunnel.Manager
-	OAuthServerConfig     handlers.OAuthAuthorizationServerConfig
+	// AgentServerURL is Obot's address as a sandbox can reach it, which differs
+	// from ServerURL when Obot runs outside the cluster its sandboxes run in.
+	AgentServerURL    string
```
thedadams, line 185: The MCP backends have methods for this purpose. Can we use those instead?

pkg/services/config.go
```diff
@@ -1061,6 +1095,75 @@ func New(ctx context.Context, config Config) (*Services, error) {
...
+	// is not enough: it has to resolve to somewhere a sandbox is permitted to
+	// reach. The egress policy shipped with the chart excludes private ranges,
+	// so a hostname pointing at an internal load balancer or at a node is
+	// refused even though it resolves. MCP servers have always answered this by
+	// addressing Obot's Service instead, and a sandbox reaches Obot for the
+	// same things -- its models, its MCP servers -- so it is given the same
+	// address rather than a second mechanism for the same problem.
+	agentServerURL := config.Hostname
+	if agentBackendKind == "kubernetes" && agentLocalK8sClient != nil {
+		reachable, err := agentbackendkubernetes.ReachableServerURL(ctx, agentLocalK8sClient, config.Hostname)
```
thedadams, line 1138: The MCPSessionManager has a `TransformObotHostname` method for this purpose.

ibuildthecloud: It is used — it is the second of two steps, and I have made the code say why in `fa1181f7` rather than changing the behaviour.

`TransformObotHostname` substitutes Obot's Service, and `serviceFQDN` is only set when `ServiceName` and `ServiceNamespace` are both configured:

```go
if opts.ServiceName != "" && opts.ServiceNamespace != "" {
    serviceFQDN = fmt.Sprintf("%s.%s.svc.%s", ...)
}
```

The chart sets both, so in-cluster it answers and that is the only step that runs. A developer running Obot on their own machine against a local cluster sets neither, so it returns the hostname untouched — and that hostname is a loopback address, which inside a sandbox means the sandbox itself.

`ReachableServerURL` covers only that case: it substitutes a node address, since Obot listens on all of the developer machine's interfaces, and it leaves any non-loopback hostname alone so the in-cluster path falls straight through to `TransformObotHostname`.

So they are complementary rather than duplicative, and using the MCP method alone would break the developer case. The comment now states both cases and why one cannot answer for the other — previously it described the loopback case first and read as though it were the whole story.

pkg/services/config.go
ibuildthecloud: Same answer as the one below, on the call site — the field holds the result of that resolution rather than doing any of its own.

`AgentServerURL` exists because a sandbox and a browser reach Obot differently: `ServerURL` is the public address a published agent builds links from, and this is the address a sandbox dials for its models and MCP servers. In-cluster they are resolved by `TransformObotHostname`, which is the MCP method you are pointing at; outside the cluster that method cannot answer, so `ReachableServerURL` handles the loopback case.

Both are computed once at start-up and stored here rather than resolved per request, so the two consumers -- the config the controller writes, and the credential grant -- cannot disagree about where Obot is.

thedadams (approved):

PR #7465, opened 2026-08-03
feat: reconcile hosted agents, pools and config sources (7/11)

ibuildthecloud:
The controllers turn the resources into running sandboxes: an instance is
placed in a pool, given a credential and a configuration, and handed to the
backend; a pool is created before anything needs it and removed with its last
instance; a user's instances go when the user does.

The configuration a sandbox receives is built here, and it is the contract the
images depend on. Every endpoint in it is absolute and every skill is already
on disk, so an image boots by reading a file and needs to know nothing about
Obot -- not its address, not its routes. Credentials are written to a second,
tighter-permissioned file so the configuration itself can be world-readable and
carries no secret.

A config source is a git repository of harness and template definitions, synced
on an interval. Sync does not rewrite the references a template makes: a
template naming a server that is not installed still syncs and is reported
unavailable, and installing the server later makes it work with no resync.

---

Part 7 of an 11-PR stack. Base: `feat/hosted-agents-06-server-wiring`.

## Stack

1. #7459 — Resource model
**→ 7. #7465 — Controllers / reconcilers** (this PR)
8. #7466 — REST API
9. #7467 — Admin UI
10. #7468 — User UI

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -73,6 +73,10 @@ func (c *Controller) PreStart(ctx context.Context) error {
 		return fmt.Errorf("failed to ensure default user role setting: %w", err)
 	}
 
+	if err := ensureHostedAgentPoolDefaults(ctx, c.services.StorageClient); err != nil {
+		return fmt.Errorf("failed to ensure hosted agent pool defaults: %w", err)
+	}
+
 	resourceMaximums := c.services.MCPSessionManager.KubernetesResourceMaximums()
 	if err := ensureK8sSettings(ctx, c.services.StorageClient, c.services.PodSchedulingSettingsFromHelm, c.services.PSASettingsFromHelm, resourceMaximums); err != nil {
 		return fmt.Errorf("failed to ensure K8s settings: %w", err)
@@ -386,6 +390,40 @@ func ensureDefaultUserRoleSetting(ctx context.Context, client kclient.Client) er
 	return client.Update(ctx, &defaultRoleSetting)
 }
 
+func ensureHostedAgentPoolDefaults(ctx context.Context, client kclient.Client) error {
+	const gibibyte = int64(1024 * 1024 * 1024)
+
+	var defaults v1.HostedAgentPoolDefaults
+	key := kclient.ObjectKey{Namespace: system.DefaultNamespace, Name: "default"}
+	if err := client.Get(ctx, key, &defaults); err == nil {
+		// Defaults are administrator-owned after creation.
+		return nil
+	} else if !apierrors.IsNotFound(err) {
+		return err
+	}
+
+	return client.Create(ctx, &v1.HostedAgentPoolDefaults{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      key.Name,
+			Namespace: key.Namespace,
+		},
+		Spec: v1.HostedAgentPoolDefaultsSpec{
+			Manifest: types.HostedAgentPoolDefaultsManifest{
+				Capacity: types.HostedAgentResourceQuantity{
+					CPUVCPUs:     1,
+					MemoryBytes:  4 * gibibyte,
+					StorageBytes: 20 * gibibyte,
+				},
+				// Seeded explicitly rather than left to the fallback, so an
+				// administrator opening the defaults sees the number that is
+				// actually in force. With the capacity above this gives each
+				// sandbox 250m CPU and 1GiB guaranteed.
+				MaxSandboxes: 4,
+			},
+		},
+	})
+}
+
 // ensureK8sSettings ensures the K8sSettings resource exists with proper configuration.
 // podSchedulingSettings: affinity, tolerations, resources, runtimeClassName - can be managed via Helm OR UI.
 //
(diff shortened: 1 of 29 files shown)


pkg/controller/handlers/cleanup/user.go
```diff
@@ -174,3 +191,127 @@ func (u *UserCleanup) Cleanup(req router.Request, _ router.Response) error {
...
+	}); err != nil {
+		return false, err
+	}
+	if len(instances.Items) > 0 {
+		for i := range instances.Items {
+			if err := req.Delete(&instances.Items[i]); err != nil && !apierrors.IsNotFound(err) {
+				return false, err
+			}
+		}
+		resp.RetryAfter(hostedAgentCleanupRetryInterval)
```
thedadams: Don't need to set a `RetryAfter` because the framework will retry the object when any of the listed items changes.

The same comment holds for all the `RetryAfter` calls in this function.

pkg/controller/controller.go
```diff
@@ -584,6 +622,12 @@ func (c *Controller) setupLocalK8sRoutes() {
 		// instead of waiting for the periodic service-account key rotation loop.
 		c.services.LocalRouter.Type(&corev1.Secret{}).Namespace(c.services.ServiceNamespace).Name(serviceaccounts.NetworkPolicySecretName).IncludeRemoved().HandlerFunc(c.reconcileServiceAccountSecretChange)
 	}
+	// Every handler below reconciles MCP state from cluster objects. The local
+	// router also exists when only hosted agents run on Kubernetes, and in that
+	// case none of this applies.
+	if !mcp.IsKubernetesBackend(c.services.MCPRuntimeBackend) {
+		return
+	}
```
thedadams: Don't do this. Just let the nil-ness of the `EveryReplicaRouter` gate this.

pkg/controller/handlers/hostedagent/credentials/credentials.go
```diff
@@ -0,0 +1,117 @@
...
+	if i == nil || i.gatewayClient == nil {
+		return nil
+	}
+
+	var errs []error
+	if err := i.gatewayClient.DeleteHostedAgentAPIKeys(ctx, instanceID); err != nil {
+		errs = append(errs, err)
+	}
+	if _, err := i.gatewayClient.DeleteCredential(ctx, credentialContext, instanceID); err != nil {
+		if _, ok := errors.AsType[client.CredentialNotFoundError](err); !ok {
```
thedadams: `DeleteCredential` doesn't return a `client.CredentialNotFoundError`. That method is idempotent.

pkg/controller/handlers/hostedagent/hostedagent.go
```diff
@@ -0,0 +1,678 @@
...
+}
+
+// EnsurePool resolves the instance's pool before backend
+// reconciliation. The deterministic names make concurrent first-instance
+// reconciles for one user converge on the same pool and assignment.
+func (h *Handler) EnsurePool(req router.Request, _ router.Response) error {
+	instance := req.Object.(*v1.HostedAgentInstance)
+	if !instance.DeletionTimestamp.IsZero() {
+		return nil
+	}
```
thedadams: If you didn't add `IncludeRemoved()` or `IncludeFinalizing()` then this handler won't be called for instances that are being deleted.

pkg/controller/handlers/cleanup/user.go
ibuildthecloud: Fixed in `9ac18dc9`. All four `RetryAfter` calls in that function are gone, along with the `hostedAgentCleanupRetryInterval` constant and the now-unused `resp` parameters.

You are right that the framework re-reconciles when the listed objects change. Worse than redundant, the fixed 10s interval was a slower answer than the event: deleting the instances is what makes the next pass succeed, and that arrives as a change rather than after a timer.

pkg/controller/controller.go
ibuildthecloud: Fixed in `9ac18dc9` — and this was a regression I introduced, not just an awkward shape.

`setupLocalK8sRoutes` is called unconditionally, on `main` as well. My early return therefore did not only skip what I intended: it stopped the **tunnel peer** handler from registering whenever MCP was not on Kubernetes, which had nothing to do with hosted agents and worked before this stack.

It was also in the wrong place for its stated purpose. The MCP handlers it claimed to guard are inside the `LocalRouter != nil` block *above* it, so they were never gated at all.

The gate now sits on the block it was meant for, and the tunnel peer handler is back to being gated only by nil-ness as you said:

```go
if c.services.LocalRouter != nil && mcp.IsKubernetesBackend(c.services.MCPRuntimeBackend) {
    ... MCP handlers ...
}

if c.services.EveryReplicaRouter != nil {
    ... tunnel peer ...
}
```

The MCP handlers still need a gate of some kind, because the local router now also exists when only hosted agents run on Kubernetes -- that part of the intent was real, just implemented in the wrong spot.

pkg/controller/handlers/hostedagent/credentials/credentials.go
ibuildthecloud: Fixed in `9ac18dc9`. The branch is gone; the error is appended unconditionally now, with a note that the method is idempotent so nobody re-adds the filter.

Worth flagging that I left the other one alone, a few lines up: `RevealCredential` does return `CredentialNotFoundError`, on the first issue for an instance that has no credential yet. That one is load-bearing.

pkg/controller/handlers/hostedagent/hostedagent.go
ibuildthecloud: You are right, and I removed the guard rather than adding the option — fixed in `9ac18dc9`.

`EnsurePool` is registered as a plain `HandlerFunc`, so it is never called for an instance being deleted, and the `DeletionTimestamp` check at the top could not fire. Deletion is already handled where it belongs: `RemoveInstance` is registered with `FinalizeFunc`, and `cleanup.Cleanup` handles the rest.

Adding `IncludeRemoved()` here would have been the wrong direction — pool assignment has nothing to do for an instance on its way out.

thedadams (approved):

PR #7466, opened 2026-08-03
feat: serve the hosted agent API (8/11)

ibuildthecloud:
Endpoints for harnesses, templates, instances, pools, config sources and access
rules, authorized the same way the rest of Obot is: templates and pools are
administrator-owned, an instance belongs to its user, and access rules decide
who may launch what.

Two things are computed on read rather than stored. Availability reports why a
template cannot be launched here -- an unconfigured model provider, an MCP
server that was never installed -- so an administrator sees the whole list
rather than fixing one and being told about the next; it is also enforced when
an instance is created, so the answer cannot be bypassed. The icon shown for an
instance resolves through the template to the harness. Storing either would
mean rewriting every row when a harness changes its icon or a provider is
configured, and leave the stored copy wrong until something did.

---

Part 8 of an 11-PR stack. Base: `feat/hosted-agents-07-controllers`.

## Stack

1. #7459 — Resource model
**→ 8. #7466 — REST API** (this PR)
9. #7467 — Admin UI
10. #7468 — User UI

### DIFF
--- a/pkg/api/handlers/hostedagentinstances.go
+++ b/pkg/api/handlers/hostedagentinstances.go
@@ -1,0 +1,455 @@
+package handlers
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/adhocore/gronx"
+	"github.com/obot-platform/obot/apiclient/types"
+	"github.com/obot-platform/obot/pkg/accesscontrolrule"
+	"github.com/obot-platform/obot/pkg/alias"
+	"github.com/obot-platform/obot/pkg/api"
+	"github.com/obot-platform/obot/pkg/api/authz"
+	"github.com/obot-platform/obot/pkg/hostedagentaccessrule"
+	"github.com/obot-platform/obot/pkg/modelaccesspolicy"
+	"github.com/obot-platform/obot/pkg/skillaccessrule"
+	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
+	"github.com/obot-platform/obot/pkg/system"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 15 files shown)


pkg/api/handlers/hostedagentinstances.go
```diff
@@ -0,0 +1,445 @@
...
+
+func (h *HostedAgentInstanceHandler) Create(req api.Context) error {
+	var body hostedAgentInstanceRequest
+	if err := req.Read(&body); err != nil {
+		return types.NewErrBadRequest("failed to read hosted agent instance manifest: %v", err)
+	}
+
+	if body.HostedAgentID == "" {
+		return types.NewErrBadRequest("hostedAgentID is required")
+	}
```
thedadams: Why not put this in the validate method?

thedadams (approved):

pkg/api/handlers/hostedagentinstances.go
ibuildthecloud: Fixed in `2bd9a353`.

`hostedAgentInstanceRequest` now has its own `Validate`, which checks `hostedAgentID` and then delegates to the embedded manifest's:

```go
func (r hostedAgentInstanceRequest) Validate() error {
	if r.HostedAgentID == "" {
		return fmt.Errorf("hostedAgentID is required")
	}
	return r.HostedAgentInstanceManifest.Validate()
}
```

It goes on the request type rather than into `HostedAgentInstanceManifest.Validate` because `hostedAgentID` is not part of the manifest: an instance names the agent it is created from once, at creation, and never carries it afterwards. Putting it in the manifest's own rules would fail every update, where the field is absent by design.

The wrapper message became "invalid hosted agent instance request" rather than "manifest", since it now covers both.

PR #7469, opened 2026-08-03
feat: seed a default config source for hosted agents (11/11)

ibuildthecloud:
A fresh installation had no harnesses or templates and no obvious way to get
any: an administrator had to know the catalog repository existed and add it by
hand before anything could be launched. It is now seeded the same way the
default skill repository is, from a flag, so the samples are simply there.

Seeding happens only when the installation has no catalogs at all -- the
existing proxy for "has this server run before" -- so an administrator who
removes it does not find it back on the next restart.

The seed values move into a struct rather than growing the positional argument
list. They are all strings, and a transposed pair would not fail at the call
site; it would surface much later as a repository that does not sync.

A local checkout is accepted as the source only in development mode, where
pointing Obot at a working copy of the catalog is how the samples get iterated
on.

---

Part 11 of an 11-PR stack. Base: `feat/hosted-agents-10-user-ui`.

## Stack

1. #7459 — Resource model
8. #7466 — REST API
9. #7467 — Admin UI
10. #7468 — User UI
**→ 11. #7469 — Seed default config source** (this PR)


chart/values.yaml
```diff
@@ -258,6 +265,11 @@ config:
   # config.OBOT_DEFAULT_SKILL_REPO_REF -- The ref (branch, tag, or commit SHA) for the default skill repository. If empty, the repository's default branch is used. Only used on first-time setup.
   # OBOT_DEFAULT_SKILL_REPO_REF: ""
 
+  # config.OBOT_DEFAULT_HOSTED_AGENTS_CATALOG_URL -- The default hosted agent catalog repository URL. Must be a full HTTPS URL. Defaults to https://github.com/obot-platform/hosted-agents-catalog. Only used on first-time setup (before the first owner user is created). Set to an empty string to disable.
+  # OBOT_DEFAULT_HOSTED_AGENTS_CATALOG_URL: ""
+  # config.OBOT_DEFAULT_HOSTED_AGENTS_CATALOG_REF -- The ref (branch, tag, or commit SHA) for the default hosted agent catalog repository. If empty, the repository's default branch is used. Only used on first-time setup.
+  # OBOT_DEFAULT_HOSTED_AGENTS_CATALOG_REF: ""
```
thedadams: Should we uncomment these?

ibuildthecloud: Great question. I would assume yes. I'll look at why it was commented out.

ibuildthecloud: Yes — uncommented in `4c86b2c0`, and confirmed it changes nothing.

`internal-configmap.yaml` skips empty values:

```gotemplate
{{- if (toString $value) }}
{{ $key }}: {{ toString $value | quote }}
{{- end }}
```

so commented-out and `""` are indistinguishable. Verified with `helm template`:

| values.yaml | ConfigMap |
|---|---|
| absent | key not rendered |
| `""` | key not rendered |
| `"https://example.com/mine.git"` | rendered |

`helm lint` passes and no key is emitted with the defaults, so this is purely discoverability — which is the point, since `values.yaml` is what people read.

**One thing this exposes, worth its own issue.** Both entries document "Set to an empty string to disable", and through the chart that is impossible: an empty string is exactly what the template drops, so you get the Go default back rather than "disabled". The same wording sits on `OBOT_DEFAULT_SKILL_REPO_URL` just above, so it predates this stack. Either the template needs to distinguish absent from empty, or the docs should describe how disabling actually works.

thedadams (approved):

PR #7447, opened 2026-08-03
feat: hosted agents

ibuildthecloud:
Runs agents on the user's behalf in a sandbox Obot manages: a template says what an agent is, an instance is one user's running copy of it, and a pool is the compute they share.

Supersedes #7213 — same work, history reorganized from 163 development commits into 10 logical ones and rebased onto current main.

## The commits

| | |
|---|---|
| `add the hosted agent resource model` | Harness, HostedAgent, HostedAgentInstance, HostedAgentPool, AgentCatalog, plus access rules |
| `add an interface for running agent sandboxes` | Backend SPI, with disabled and fake implementations; keeps Kubernetes types out of controllers and the API |
| `run agent sandboxes on Kubernetes` | Deployments per instance, one volume per pool separated by subPath, Pod Security Admission |
| `reconcile hosted agents, pools and config sources` | The controllers, and the configuration contract a sandbox reads |
| `serve the hosted agent API` | Endpoints, plus availability and icon resolution computed on read |
| `let a sandbox reach the models and MCP servers it was granted` | Sandbox identity, grant-list authorization, portable references |
| `attach a terminal to a sandbox and publish agents that serve HTTP` | Terminal over websocket; HTTP agents published under `/agent-connect/{id}` |
| `wire the agent backend into the server` | Configuration, flags, chart |
| `add the hosted agents UI` | Admin screens and the user-facing agent view |
| `seed a default config source for hosted agents` | So a fresh install has samples without hand-adding a repository |

## Design notes

**A sandbox knows nothing about Obot.** It reads `/etc/obot/agent.json` — every endpoint already an absolute URL, every skill already on disk — and boots. Credentials live in a second file (`0440`) so the configuration itself (`0444`) carries no secret and can be read or logged freely.

**A template is portable.** It names models by provider and MCP servers and skills by `<source>::<key>`, resolved against whatever IDs the local installation generated, so one template works everywhere. A template naming something not installed still syncs and is reported unavailable with the reason; installing it later just works.

**A sandbox is not the user.** Its credential carries a group that reaches the MCP gateway and the model proxy and nothing else, and it is authorized against the servers its template was granted rather than against a user's access.

## Related

- Sample images: `obot-platform/hosted-agents-images` · catalog: `obot-platform/hosted-agents-catalog` · example agent: `obot-platform/obot-adk-chatbot`
- The websocket hijack fix this depends on merged separately as #7442
- The dev-server (watchnbuild) work is deliberately **not** here; it is on its own branch
[...]

PR #7470, opened 2026-08-03
fix(ci): authenticate Depot on same-repo PRs and run CI on feat/* branches

ibuildthecloud:
The `Test Docker Build` job fails on any pull request whose branch lives in this repository rather than in a fork:

```
Unable to exchange GitHub OIDC token for temporary Depot token:
  Unable to get ACTIONS_ID_TOKEN_REQUEST_URL env variable
unauthenticated: Invalid token
```

`test-docker-build.yml` declares no `permissions:` block, and GitHub never grants `id-token` implicitly — not even when the default workflow permissions are read-write, as they are here. The failing job's own permission list shows `Contents: write`, `PullRequests: write`, and no `id-token` at all.

This went unnoticed because every pull request so far has come from a fork, and `depot/build-push-action` takes a different path for those:

```
Attempting to acquire open-source pull request OIDC token
Using open-source pull request OIDC token for Depot authentication
```

That challenge flow needs no GitHub OIDC, so fork PRs never exercised the broken branch. The two sibling workflows that build with Depot outside of pull requests — `docker-build-and-push.yml` and `weekly-base-image.yml` — both already request `id-token: write`. This one was simply missing it.

### Also: CI on stacked PRs

Every `pull_request` trigger was filtered to `branches: [main]`, so a PR based on another branch gets **no CI at all** — only the bottom of a stack is tested. Widened to `[main, 'feat/*']` across `go.yml`, `test.yaml`, `test-docker-build.yml`, `user-ui.yml`, `docs.yml`, and `helm.yml`. Existing `paths`/`paths-ignore` filters are unchanged, as are all `push:` triggers.

Found while splitting #7447 into the stack at #7459–#7469, which is the first set of same-repo PRs this repo has had.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches:
       - main
+      - 'feat/*'
     paths:
       - "docs/**"
 
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -7,6 +7,7 @@ on:
   pull_request:
     branches:
       - main
+      - 'feat/*'
     paths-ignore:
       - chart/**
       - ui/user/**
--- a/.github/workflows/helm.yml
+++ b/.github/workflows/helm.yml
@@ -10,6 +10,7 @@ on:
   pull_request:
     branches:
       - main
+      - 'feat/*'
     paths:
       - .github/workflows/helm.yml
       - chart/**
--- a/.github/workflows/test-docker-build.yml
+++ b/.github/workflows/test-docker-build.yml
@@ -1,8 +1,14 @@
 name: Test Docker Build
+
+permissions:
+  contents: read
+  id-token: write
+
 on:
   pull_request:
     branches:
       - main
+      - 'feat/*'
     paths-ignore:
       - docs/**
       - chart/**
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches:
       - main
+      - 'feat/*'
 
 jobs:
   test:
--- a/.github/workflows/user-ui.yml
+++ b/.github/workflows/user-ui.yml
@@ -7,6 +7,7 @@ on:
   pull_request:
     branches:
       - main
+      - 'feat/*'
     paths:
       - ui/user/**
 


ibuildthecloud:
@copilot can you ensure you review the security implication of this change. What does id-token: write do?

ibuildthecloud:
## Security rationale

Adding `id-token: write` deserves scrutiny, so here is the reasoning, with the parts I verified separated from the parts a reviewer should confirm independently.

### This change *reduces* the job's GitHub privileges

`test-docker-build.yml` previously declared no `permissions:` block, so the job inherited the repository default. From the failing run's `Set up job` log, that default is read-write across the board:

```
Actions: write        Contents: write       Deployments: write
Checks: write         Issues: write         Packages: write
PullRequests: write   Pages: write          SecurityEvents: write
...
```

After this PR the same job gets exactly:

```yaml
permissions:
  contents: read
  id-token: write
```

So a job that could previously write repository contents, packages, issues and deployments can now only read the checkout. In GitHub-permission terms this is a strict tightening; `id-token` is the single capability added, and it is the one capability that is never granted implicitly no matter how permissive the default is.

### What `id-token: write` actually grants

It grants no GitHub API access at all. It only lets the job ask GitHub's OIDC provider for a signed JWT identifying this repository and workflow. The blast radius is therefore not "GitHub" but "whichever relying parties trust this repo's OIDC" — here, the Depot project (`bbqjs4tj1g`) that this workflow already builds against and has always built against. The change does not add a relying party; it restores the intended authentication path to the one the workflow was already using.

### The job cannot publish anything

From the run log, the action's resolved inputs are `push: false`, `load: false`, `save: false`, and the resulting command is:

```
depot build --build-arg BASE_IMAGE=... --platform linux/amd64,linux/arm64 --project bbqjs4tj1g .
```

No `--push`, no registry login, and no `secrets.` reference anywhere in the file. The job builds an image and discards it. Image publishing lives in `docker-build-and-push.yml`, which is untouched here and already declares `id-token: write`.

### The branch widening does not change the trust model

The trigger stays `pull_request` — not `pull_request_target` and not `workflow_run`. Untrusted head code is never combined with elevated trust; no such pattern is introduced. Widening `branches:` changes *which base branches* cause CI to run, not *what CI is allowed to do*.

Of the six workflows touched, exactly one references a secret: `test.yaml` passes `secrets.GITHUB_TOKEN`, which is the automatic per-run token subject to the same `permissions` rules, not a stored credential. The other five reference no secrets. All `push:` triggers and every existing `paths` / `paths-ignore` filter are unchanged.

### Who gains capability

Opening a same-repo pull request on a `feat/*` branch requires write access to this repository. External contributors work from forks and gain nothing from this change. So the set of people who can cause this job to run is unchanged: repository collaborators, who already have far broader capability by other means.

### Residual risk worth a reviewer's judgment

1. **Scope of Depot's OIDC trust.** A collaborator can now cause a PR-triggered job to mint an OIDC token for this repo. That is fine if Depot's trust policy is scoped to expected claims, and worth checking if it is broadly scoped to `repo:obot-platform/obot:*`. Narrowing it by `sub` claim would be a good defense-in-depth follow-up, independent of this PR.
2. **Fork behavior.** I did not find explicit GitHub documentation stating that OIDC tokens are withheld for fork pull requests, so I am not asserting it. What the logs do show is that `depot/build-push-action` takes a distinct path for fork PRs ("Attempting to acquire open-source pull request OIDC token"), which is why this bug stayed hidden. Fork behavior is unchanged by this PR either way.
3. **`feat/*` is a broad pattern.** It matches any current or future `feat/*` branch, which is the intent, but a reviewer preferring a narrower blast radius could scope it to `feat/hosted-agents-*`. Note `*` does not match `/`, so `feat/a/b` would not match.

### Verification

This PR is itself a same-repo branch, so it exercises the previously broken path using its own patched workflow. All checks pass, and the Depot build now logs:

```
Exchanged GitHub Actions OIDC token for temporary Depot token
```

against the identical commit content that produced `unauthenticated: Invalid token` on #7459 before the fix.

thedadams (approved):

njhale (approved):

Issue #7434, opened 2026-08-03
Large scan payloads fail to be submitted


njhale:
**Describe the bug**
When an `obot-sentry` scan result payload is too large, scan results can fail to be submitted.

This can be triggered by payload limits at both the obot server level and via proxies like nginx.

e.g.

```
ubmit scan: error code 413 (Request Entity Too Large): <html>
<head><title>413 Request Entity Too Large</title></head>
<body>
<center><h1>413 Request Entity Too Large</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

**Expected behavior**
Large scan results can be submitted without hitting default limits on most proxies or obot.

**Additional context**
There are a few options here, but I think it's best to start with enabling request compression for scans.

Another option is to split up scan submission into a series of smaller submission requests.

njhale:
@sangee2004 This can also wait for a release before testing.

sangee2004:
Able to reproduce this issue with older version of obot-sentry that does not have the fix in a setup having nginx configured in front of obot server.

```
% obot-sentry version
Version:  v0.0.0-dev+0f0f1bec
```
```
obot-sentry scan --force --submit   
Device:    Sangeethas-MacBook-Pro.local (darwin/arm64)
User:      sangeethahariharan
Scanned:   2026-08-06T18:45:18Z
Found:     6 clients, 25 MCP servers, 102 skills, 10 plugins, 118 files

CLIENT          MCP SERVERS  SKILLS  PLUGINS  CONFIG PATH
claude_code     10           7       0        /Users/sangeethahariharan/.claude
claude_desktop  0            27      1        /Users/sangeethahariharan/Library/Application Support/Claude
codex           5            39      9        /Users/sangeethahariharan/.codex
cursor          9            40      0        /Users/sangeethahariharan/.cursor
opencode        0            18      0        -
vscode          1            18      0        /Users/sangeethahariharan/.vscode

7 skills found outside any client's discovery paths
2026/08/06 11:45:18 submit scan: error code 413 (Request Entity Too Large): <html>
<head><title>413 Request Entity Too Large</title></head>
<body>
<center><h1>413 Request Entity Too Large</h1></center>
<hr><center>nginx/1.31.3</center>
</body>
</html>
```

This issue is addressed when testing with the latest build of obot-sentry from `main`
```
% obot-sentry version
Version:  v0.0.0-dev+3921a93d
```

```
 % obot-sentry scan --submit
[...]

PR #7480, opened 2026-08-04
fix: compress device scan submissions on the wire

njhale:
Scan manifests carry the raw text of every captured config file and were
posted uncompressed, so large ones were rejected outright. SubmitDeviceScan
now zstd-compresses them -- a real 201-file scan drops from 2,005,646 to
553,094 bytes, 72% smaller. Other endpoints are unchanged.

api.Context.Body decodes Content-Encoding: zstd, capping the compressed and
decoded streams alike. A server without this parses the compressed bytes as
JSON and fails, so obot-sentry requires a server new enough to decode them.

Part of https://github.com/obot-platform/obot/issues/7434

### DIFF
--- a/pkg/api/request.go
+++ b/pkg/api/request.go
@@ -4,11 +4,14 @@ package api
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 	"slices"
 	"strconv"
+	"strings"
 
+	"github.com/klauspost/compress/zstd"
 	"github.com/obot-platform/obot/apiclient/types"
 	"github.com/obot-platform/obot/pkg/auth"
 	gclient "github.com/obot-platform/obot/pkg/gateway/client"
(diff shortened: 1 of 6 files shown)


pkg/api/request.go
```diff
@@ -80,9 +91,43 @@ func (r *Context) Body(opts ...BodyOptions) (_ []byte, err error) {
...
+		defer zstdReader.Close()
+		compressed, body = true, io.LimitReader(zstdReader, opt.MaxBytes+1)
+	}
+
+	data, err := io.ReadAll(body)
+	if err != nil {
+		// zstd validates nothing until the first read, so a malformed body lands
+		// here and would otherwise become a 500. A MaxBytesError is the cap, which
+		// the deferred mapping turns into a 413.
+		if maxErr := (*http.MaxBytesError)(nil); compressed && !errors.As(err, &maxErr) {
```
thedadams: Should we use `errors.AsType` here instead?

thedadams (approved):

pkg/api/request.go
njhale: yes, done

PR #7481, opened 2026-08-04
fix: missing UI assets served as HTML, and hosted agent seeding on upgrades

ibuildthecloud:
Two independent fixes, one commit each. Both were found deploying hosted agents
to a real environment.

## 1. A missing build asset was served as the page shell

`conf-demo` came back from a deploy with a blank page and a wall of:

> Failed to load module script: Expected a JavaScript-or-Wasm module script but
> the server responded with a MIME type of "text/html".

A request for an asset that is not present fell through to `fallback.html` —
`200`, `text/html`, and no cache policy of its own. Browsers refuse that under
strict MIME checking, and because it is a `200` **every cache in the path
stores it**: the CDN edge, and then each visitor's browser.

A rolling deploy is enough to produce the miss: for a few seconds one replica
serves a shell naming hashes another replica does not have yet. The window is
seconds; what it leaves behind is not. Recovering took a full CDN purge *and* a
browser cache clear — the origin had been healthy the whole time, which is
exactly what the MIME error does not tell you.

Content-hashed names made the diagnosis slower: chunks whose contents had not
changed kept the same name and loaded fine, so only *some* assets failed, which
reads like corruption rather than absence.

**What changed**

- Assets under `/_app/` return `404` when absent. A name derived from a file's
  contents is never a route, so falling back could only ever be wrong there.
  Everything else still falls back — a client-side route is what the shell
  exists to resolve.
- Obot states its own cache policy. It set **none** for the UI (`server.go`
  covers `/api/` only), so whichever CDN sat in front applied its default —
  4 hours, in the case this fixes:

  | | |
  |---|---|
  | `/_app/immutable/…` | `public, max-age=31536000, immutable` |
  | shell, fallback, `/_app/version.json` | `no-cache` |
  | missing asset (404) | `no-store` |

  `no-cache` on the shell matters: it names the assets of the build it came
  from, so a cached copy goes on requesting files a later build no longer has —
  the same failure, self-inflicted. `no-store` on the 404 because an edge
  caches 404s briefly by default, which is long enough to outlive the deploy.

## 2. Hosted agents were never seeded on an upgraded installation

The agent catalog and the access rule permitting anyone to use it were seeded
alongside rules gated on there being no `MCPCatalog`s — commented as a proxy
for *"has this server been started previously."* True of a new installation;
false of every installation that upgrades, which has had catalogs since its own
first boot.

So the seeds ran only where the feature had shipped from the beginning. **An
upgrade arrived with the feature visible, no harnesses, no templates, and
nobody permitted to use them.**

That proxy cannot answer this: "has this server run before" and "has this seed
run before" are different questions, and they diverge exactly at an upgrade.
Seeding now records its own name via `RunOnce`, built on the migration table
that already did this for schema changes. The record outlives what it created —
preserving the property the proxy was protecting, that an administrator who
deletes the catalog does not find it back — and seed and record are written in
one transaction, so a failed seed leaves no record and retries.

## Tests

`ui` had none; it does now. Both suites were verified non-vacuous: reverting
the fixes fails **13** tests, restoring them passes.

The embedded UI build is reached through a variable so tests can supply one.
The real build exists only after `make ui`, so the first version of these tests
passed on my machine and failed in a clean checkout — they were reporting on
whether someone had run make. They now run against an `fstest.MapFS` and pass
with no build present, which is the CI condition.

`TestDataSeedsDefaultAgentCatalogOnFirstBoot` is replaced: its second subtest
asserted *"does not seed when catalogs already exist"*, encoding the bug as
intended behaviour. The replacement covers new installs, upgrades, and
not-seeding-twice.

`go build ./...`, `go vet ./...`, `golangci-lint` (0 issues) and the full test
suite pass. Each commit builds and tests independently.

## Verified live

Against a running deployment:

| Path | Before | After |
|---|---|---|
| `/_app/immutable/entry/app.NOTREAL.js` | `200 text/html` | `404`, `no-store` |
| a real hashed asset | `200`, no cache policy | `200`, `immutable` |
| `/some/deep/client/route` | `200 text/html` | unchanged, `no-cache` |

Draft because the seeding fix deserves a look from whoever owns the upgrade
path — in particular whether `RunOnce` on the migration table is where these
records belong.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -61,7 +61,7 @@ func New(services *services.Services) (*Controller, error) {
 }
 
 func (c *Controller) PreStart(ctx context.Context) error {
-	if err := data.Data(ctx, c.services.StorageClient, data.Defaults{
+	if err := data.Data(ctx, c.services.StorageClient, c.services.GatewayClient, data.Defaults{
 		SkillRepoURL:           c.services.DefaultSkillRepoURL,
 		SkillRepoRef:           c.services.DefaultSkillRepoRef,
 		HostedAgentsCatalogURL: c.services.DefaultHostedAgentsCatalogURL,
--- a/pkg/controller/data/data.go
+++ b/pkg/controller/data/data.go
@@ -42,7 +42,18 @@ type Defaults struct {
 	AllowLocalRepos        bool
 }
 
-func Data(ctx context.Context, c kclient.Client, defaults Defaults) error {
+// seedHostedAgents names the one-time record for the hosted agent seed. It is
+// this seed's own name, so a later seed takes a new one rather than changing
+// what this record means.
+const seedHostedAgents = "seed_hosted_agents"
+
+// OnceRunner runs a named piece of work a single time per installation. It is
+// an interface so that seeding can be tested without a database.
+type OnceRunner interface {
+	RunOnce(ctx context.Context, name string, f func(context.Context) error) error
+}
+
+func Data(ctx context.Context, c kclient.Client, runner OnceRunner, defaults Defaults) error {
 	var defaultModelAliases v1.DefaultModelAliasList
 	if err := yaml.Unmarshal(defaultModelAliasesData, &defaultModelAliases); err != nil {
 		return fmt.Errorf("failed to unmarshal default model aliases: %w", err)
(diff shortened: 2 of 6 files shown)


ibuildthecloud:
Splitting into two PRs — the UI asset/caching fix and the hosted agent seeding fix are unrelated and should be reviewed separately.

Issue #7135, opened 2026-07-09
Support for setting/displaying max caps for user-defined memory and cpu requests in `Server Scheduling` page.


sangee2004:
We now have the ability to support setting max caps for user-defined memory and CPU requests and limits from Helm charts 
```
mcpServerDefaults:
 maxCPURequest: ""
 maxCPULimit: ""
 maxMemoryRequest: "350Mi"
 maxMemoryLimit: ""
```

Provide support for setting/displaying max caps for user-defined memory and CPU requests on the `Server Scheduling` page, similar to how the defaults for user-defined memory and CPU requests and limits are presented.

<img width="881" height="383" alt="Image" src="https://github.com/user-attachments/assets/0081918d-3031-4510-a6ac-09a348884d66" />

hiSandog:
The UI should distinguish a missing cap from a cap explicitly set to zero and preserve the units used by the scheduling configuration. Validation could also reject a maximum below the current request while explaining which value must change. A round-trip test for fractional CPU and memory-unit conversion would help prevent display values from drifting after saving.

cjellick:
Agree with this with the caveat that if theyre set via helm/env var, the must be read-only in the UI.

cjellick:
Closing as completed by #7486. The Server Scheduling UI now supports CPU and memory maximums and treats maximums supplied through Helm as read-only.

PR #7486, opened 2026-08-04
enhance: add setting max caps for request/limits cpu/memory in server-scheduling in ui

ivyjeong13:
This addresses allowing a user to set maximums for request/limits in cpu/memory if maximums haven't been set via helm chart. 

`MaximumsSetViaHelm` was created as a separate check for the maximums to avoid regression of existing behavior -- maximums could be set via helm chart but not trigger setViaHelm and allow users to continue modifying, via UI, other configurations such as resources, affinity, runtimeClass, etc. 

Addresses #7135 

Screenshot 2026-08-10 at 11 06 36 AM

### DIFF
--- a/apiclient/types/k8ssettings.go
+++ b/apiclient/types/k8ssettings.go
@@ -11,6 +11,18 @@ type K8sSettings struct {
 	// Resources configuration (JSON/YAML blob)
 	Resources string `json:"resources,omitempty"`
 
+	// MaxCPURequest is the startup-configured maximum CPU request for MCP server pods.
+	MaxCPURequest string `json:"maxCpuRequest,omitempty"`
+
+	// MaxCPULimit is the startup-configured maximum CPU limit for MCP server pods.
+	MaxCPULimit string `json:"maxCpuLimit,omitempty"`
+
+	// MaxMemoryRequest is the startup-configured maximum memory request for MCP server pods.
+	MaxMemoryRequest string `json:"maxMemoryRequest,omitempty"`
+
+	// MaxMemoryLimit is the startup-configured maximum memory limit for MCP server pods.
+	MaxMemoryLimit string `json:"maxMemoryLimit,omitempty"`
+
 	// RuntimeClassName specifies the RuntimeClass for MCP server pods
 	// This allows running MCP servers with specific container runtimes (e.g., gVisor, Kata)
 	RuntimeClassName string `json:"runtimeClassName,omitempty"`
@@ -29,6 +41,9 @@ type K8sSettings struct {
 
 	// SetViaHelm indicates settings are from Helm (cannot be updated via API)
 	SetViaHelm bool `json:"setViaHelm,omitempty"`
+
+	// MaximumsSetViaHelm indicates resource maximums are from Helm (cannot be updated via API)
+	MaximumsSetViaHelm bool `json:"maximumsSetViaHelm,omitempty"`
 
 	Metadata Metadata `json:"metadata,omitempty"`
 }
--- a/apiclient/types/k8ssettings.go
+++ b/apiclient/types/k8ssettings.go
@@ -11,16 +11,16 @@ type K8sSettings struct {
 	// Resources configuration (JSON/YAML blob)
 	Resources string `json:"resources,omitempty"`
 
-	// MaxCPURequest is the startup-configured maximum CPU request for MCP server pods.
+	// MaxCPURequest is the configured maximum CPU request for MCP server pods.
 	MaxCPURequest string `json:"maxCpuRequest,omitempty"`
 
-	// MaxCPULimit is the startup-configured maximum CPU limit for MCP server pods.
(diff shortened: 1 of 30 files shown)


apiclient/types/k8ssettings.go
```diff
@@ -30,6 +42,9 @@ type K8sSettings struct {
 	// SetViaHelm indicates settings are from Helm (cannot be updated via API)
 	SetViaHelm bool `json:"setViaHelm,omitempty"`
 
+	// MaximumsSetViaHelm indicates resource maximums are from Helm (cannot be updated via API)
+	MaximumsSetViaHelm bool `json:"maximumsSetViaHelm,omitempty"`
```
ivyjeong13, line 46: `MaximumsSetViaHelm` was created as a separate check for the maximums to avoid regression of existing behavior -- maximums could be set via helm chart but not trigger setViaHelm and allow users to continue modifying, via UI, other configurations such as resources, affinity, runtimeClass, etc. 

But wanted to check if this behavior should still be the case? Was this to avoid causing issues for existing deployments but maximums needed to be set?

pkg/mcp/resource_maximums.go
```diff
@@ -117,6 +136,7 @@ func ValidateK8sSettingsResourceMaximums(k8sSettings v1.K8sSettingsSpec, maximum
 }
 
 func ValidateConfiguredK8sSettingsResourceMaximums(k8sSettings v1.K8sSettingsSpec, maximums ResourceMaximums) error {
+	maximums = ResourceMaximumsFromK8sSettings(k8sSettings, maximums)
```
thedadams: The `maximums` that is passed into this function comes from the sessionManager's `KubernetesResourceMaximums` method. That method should get the maximums from the k8s settings, if they are there.

pkg/mcp/manager.go
```diff
@@ -230,6 +232,23 @@ func (sm *SessionManager) KubernetesResourceMaximums() ResourceMaximums {
 	return sm.resourceMaximums
 }
 
+func (sm *SessionManager) EffectiveKubernetesResourceMaximums(ctx context.Context, storageClient kclient.Client) (ResourceMaximums, error) {
```
thedadams: Oh, yeah. We should use this method and not the `KubernetesResourceMaximums`. In fact, we should probably get rid of `KubernetesResourceMaximums` and only use this method.

Then we don't have to recalculate the maximums from my previous comment.

pkg/mcp/kubernetes.go
```diff
@@ -988,6 +988,7 @@ func mcpContainerResources(serverSpecificResources *corev1.ResourceRequirements,
 }
 
 func mcpContainerResourcesWithMaximums(serverSpecificResources *corev1.ResourceRequirements, runtime types.Runtime, nanobotAgent bool, k8sSettings v1.K8sSettingsSpec, maximums ResourceMaximums) corev1.ResourceRequirements {
+	maximums = ResourceMaximumsFromK8sSettings(k8sSettings, maximums)
```
thedadams: Same comment here: the `maximums` passed in should be the "effective maximums" and we don't have to recalculate here.

thedadams (changes requested):

pkg/mcp/manager.go
ivyjeong13: Thanks Donnie! Will get these switched over to effectiveMaximums : )

pkg/mcp/resource_maximums.go
```diff
@@ -22,6 +22,24 @@ type ResourceMaximumExceededError struct {
 	Maximum resource.Quantity
 }
 
+func ResourceMaximumsFromK8sSettings(settings v1.K8sSettingsSpec, fallback ResourceMaximums) ResourceMaximums {
```
thedadams:
```suggestion
func EffectiveResourceMaximums(settings v1.K8sSettingsSpec, fallback ResourceMaximums) ResourceMaximums {
```

pkg/mcp/kubernetes.go
```diff
@@ -486,7 +486,8 @@ func (k *kubernetesBackend) k8sObjects(ctx context.Context, server ServerConfig)
 	// Fetch K8s settings
 	k8sSettings := k.getK8sSettings(ctx)
 
-	mcpResources := mcpContainerResourcesWithMaximums(server.Resources, server.Runtime, server.NanobotAgentName != "", k8sSettings, k.resourceMaximums)
+	maximums := ResourceMaximumsFromK8sSettings(k8sSettings, k.resourceMaximums)
```
thedadams: Do you think it's OK to assume that the `k8sSettings` has the effective resource maximums here? In other words, can we get rid of `k.resourceMaximums` and just take `k8sSettings` and expect it to be right?

ivyjeong13: Thanks Donnie! Dropped `k.resourceMaximums` cause it would be what gets supplied in `k8sSettings` :+1:

thedadams (approved):

PR #7492, opened 2026-08-05
feat: add CLI command to validate catalog YAML

calvinmclean:
While working on our MCP catalog overhaul, I have observed the agent using various Ruby and Python scripts to attempt to validate the YAML entries. This change introduces a `obot mcp validate` command which can validate entries using our actual backend validation code. In addition to the new command, some of the code for parsing and validating had to be moved to be accessible here and I refactored so it works more generally to parse and validate using an iterator.


thedadams:
It's not clear to me that `obot mcp validate` is supposed to validate the YAML for an MCP catalog entry. Would you be opposed to `obot mcp validate-catalog-yaml` or something similar?

calvinmclean:
> It's not clear to me that `obot mcp validate` is supposed to validate the YAML for an MCP catalog entry. Would you be opposed to `obot mcp validate-catalog-yaml` or something similar?

@thedadams that sounds good to me

thedadams (approved):

PR #7505, opened 2026-08-06
feat: explicit detach state and "accept ownership" action

calvinmclean:
Addresses https://github.com/obot-platform/obot/issues/7092

Delete warning:
Screenshot 2026-08-06 at 12 23 05

Detach warning + button:
Screenshot 2026-08-06 at 12 47 56

Sync error if re-added to catalog:
Screenshot 2026-08-06 at 12 49 16

Note: some small UI details changed after recording this, but the overall flow is still accurate

https://github.com/user-attachments/assets/13a88117-e7d1-49c5-a521-34c7aaa883f2

### DIFF
--- a/pkg/controller/handlers/mcpcatalog/mcpcatalog.go
+++ b/pkg/controller/handlers/mcpcatalog/mcpcatalog.go
@@ -236,10 +236,10 @@ func filterConflictingCatalogEntries(ctx context.Context, c client.Client, names
 		key := client.ObjectKeyFromObject(entry)
 		existing, found := existingByName[key]
 		if !found {
-			if err := c.Get(ctx, key, &existing); err != nil && !apierrors.IsNotFound(err) {
-				return nil, nil, err
-			} else if err == nil {
+			if err := c.Get(ctx, key, &existing); err == nil {
 				found = true
+			} else if !apierrors.IsNotFound(err) {
+				return nil, nil, err
 			}
 		}
 		if !found {
--- a/pkg/controller/handlers/mcpcatalog/mcpcatalog.go
+++ b/pkg/controller/handlers/mcpcatalog/mcpcatalog.go
@@ -246,13 +246,7 @@ func filterConflictingCatalogEntries(ctx context.Context, c client.Client, names
 			result = append(result, obj)
 			continue
 		}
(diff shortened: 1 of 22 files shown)


pkg/storage/apis/obot.obot.ai/v1/mcpservercatalogentry.go
```diff
@@ -68,17 +68,19 @@ func (in *MCPServerCatalogEntry) DeleteRefs() []Ref {
...
-	return !in.Spec.Editable && in.Spec.SourceURL != ""
+	return !in.Spec.Detached && !in.Spec.Editable && in.Spec.SourceURL != ""
 }
 
 type MCPServerCatalogEntrySpec struct {
 	Manifest         types.MCPServerCatalogEntryManifest `json:"manifest"`
 	UnsupportedTools []string                            `json:"unsupportedTools,omitempty"`
 	MCPCatalogName   string                              `json:"mcpCatalogName,omitempty"`
 	Editable         bool                                `json:"editable,omitempty"`
+	// +optional
```
thedadams: nit: I assume this affects the code generation somehow, but we don't do this anywhere else.

```suggestion
```

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -205,6 +213,53 @@ func addSyncError(syncErrors map[string]string, sourceURL, errMsg string) {
...
+		}
+
+		key := client.ObjectKeyFromObject(entry)
+		existing, found := existingByName[key]
+		if !found {
+			if err := c.Get(ctx, key, &existing); err != nil && !apierrors.IsNotFound(err) {
+				return nil, nil, err
+			} else if err == nil {
+				found = true
+			}
```
thedadams: nit: I personally like to do this the other way around because it seems cleaner. Feel free to ignore if you don't like it.
```suggestion
			if err := c.Get(ctx, key, &existing); err == nil {
				found = true
			} else if !apierrors.IsNotFound(err) {
				return nil, nil, err
			}
```

thedadams (approved):

ivyjeong13 (approved):

PR #7510, opened 2026-08-07
enhance: use go-sdk client

thedadams:
The Go SDK client does proper protocol version handshake and discovery.


pkg/api/handlers/mcpgateway/oauth/mcpoauthhandler.go
```diff
@@ -122,21 +121,11 @@ func (f *MCPOAuthHandlerFactory) CheckForMCPAuth(req api.Context, mcpServer v1.M
...
-		httpClientOptions := nmcp.HTTPClientOptions{
-			OAuthRedirectURL:              system.MCPOAuthCallbackURL(f.baseURL),
-			OAuthClientName:               "Obot MCP Gateway",
-			OAuthClientIDMetadataDocument: f.cimdDocumentURL,
-			CallbackHandler:               oauthHandler,
-			ClientCredLookup:              oauthHandler,
-			TokenStorage:                  f.tokenStore.ForUserAndMCP(oauthHandler.userID, oauthHandler.mcpID),
-			BlockLoopback:                 !blockingConfig.AllowLocalhostMCP,
-			BlockPrivateIP:                !blockingConfig.AllowPrivateIPMCP,
-			BlockLinkLocal:                !blockingConfig.AllowLinkLocalMCP,
```
calvinmclean, line 135: do we need to use `safehttp` here as well to preserve these configs?

pkg/mcp/oauth.go
```diff
@@ -105,3 +110,1061 @@ func (t *assumeOAuthRequiredTransport) RoundTrip(req *http.Request) (*http.Respo
 	t.mu.Unlock()
 	return t.base.RoundTrip(req)
 }
+
+// storageBackedTokenSource implements the oauth2.TokenSource interface to store new tokens in the TokenStorage.
```
thedadams, line 114: All the code from here down was pulled in from nanobot (in hopes that we can get rid of it as a dependency).

calvinmclean (approved):

pkg/api/handlers/mcpgateway/oauth/mcpoauthhandler.go
thedadams: Good question! It's built into the client now, so we construct the `httpsafe` client and pass it to the SDK MCP client.

calvinmclean: Awesome. Thanks for the clarification :+1:

g-linville (approved):

PR #7519, opened 2026-08-10
Fix editing configuration for remote MCP deployments

calvinmclean:
## Summary

- restore the Edit Configuration action for Remote MCP deployments
- keep existing owner and writable-admin authorization
- continue excluding Composite deployments while preserving Hosted behavior
- reuse the existing deployment configuration dialog and save APIs

## Testing

- `pnpm run check`
- `pnpm run lint`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deployment configuration editing controls.
  * Editing is now available for eligible non-composite deployments while remaining unavailable for composite deployments and runtimes.

### DIFF
--- a/ui/user/src/lib/components/mcp/DeploymentsView.svelte
+++ b/ui/user/src/lib/components/mcp/DeploymentsView.svelte
@@ -23,6 +23,7 @@
 		getMcpServerDeploymentStatus,
 		getServerTypeLabel,
 		getServerUrl,
+		hasEditableConfiguration,
 		hasMissingSecretBindingConfig,
 		isMultiUserServer,
 		supportsMCPBackendDetails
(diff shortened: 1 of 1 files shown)


ui/user/src/lib/components/mcp/DeploymentsView.svelte
```diff
@@ -581,6 +581,10 @@
 	function isRestartableServer(d: MCPCatalogServer) {
 		return d.manifest.runtime !== 'remote' && d.manifest.runtime !== 'composite';
 	}
+
+	function canEditDeploymentConfiguration(d: MCPCatalogServer) {
+		return !d.compositeName && d.manifest.runtime !== 'composite';
```
ivyjeong13, line 587: Wondering if a `hasEditableConfiguration(d)` check should be in here too, in case the catalog entry the deployment is related to has no modifiable values.

ivyjeong13 (approved):
Just the one thought otherwise lgtm

ui/user/src/lib/components/mcp/DeploymentsView.svelte
calvinmclean: Sounds good, thanks! I'll add it and then merge

PR #7530, opened 2026-08-11
enhance: add retention-based cleanup for submitted device scans

njhale:
Submitted device scans and their child rows accumulated indefinitely with
no way to reclaim the space. Add a periodic cleanup loop that deletes
scans older than a configurable retention window, mirroring the existing
MCP and LLM audit log cleanup in the gateway client.

OBOT_SERVER_DEVICE_SCAN_RETENTION_DAYS defaults to 90 days; set it to 0
to disable cleanup, matching the convention used by the audit log
retention settings. Deletion runs once a day in batches so a large
backlog doesn't become a single oversized transaction. Child rows are
removed by the ON DELETE CASCADE constraints on
device_scan_*.device_scan_id.

### DIFF
--- a/pkg/gateway/client/devicescancleanup.go
+++ b/pkg/gateway/client/devicescancleanup.go
@@ -1,0 +1,63 @@
+package client
+
+import (
+	"context"
+	"errors"
+	"time"
+)
+
+func (c *Client) runDeviceScanCleanup(ctx context.Context, retentionDays int) {
+	if retentionDays <= 0 {
+		return
+	}
+
+	err := c.deleteOldDeviceScans(ctx, time.Now().UTC(), retentionDays)
+	if err != nil && !errors.Is(err, context.Canceled) {
+		log.Errorf("Failed to delete old device scans: %v", err)
+	}
+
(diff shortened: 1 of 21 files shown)


pkg/gateway/client/devicescancleanup.go
```diff
@@ -0,0 +1,63 @@
...
+
+// The scans' child rows go with them via the ON DELETE CASCADE
+// constraints on device_scan_*.device_scan_id.
+func (c *Client) deleteOldDeviceScans(ctx context.Context, now time.Time, retentionDays int) error {
+	if retentionDays <= 0 {
+		return nil
+	}
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
```
thedadams: I would guess that these checks aren't needed because we will detect this and return the (wrapped) error later.

thedadams (approved):

pkg/gateway/client/devicescancleanup.go
njhale: yup, removing

Issue #7538, opened 2026-08-12
Filter support for local agents


g-linville:
I am working on extending our existing Filter feature (`MCPWebhookValidation` in the backend) to support local agent tool calls and user prompts.

g-linville:
Moved to Stalled Out due to shifting priorities.

PR #7540, opened 2026-08-12
feat: introduce new v1 filter hook contract

g-linville:
part of https://github.com/obot-platform/obot/issues/7538

This introduces the new v1 API contract for Filters (aka `MCPWebhookValidators`), to allow them to explicitly support tool calls and user prompts coming from local devices.

This PR just introduces the types and the migration for existing MCPWebhookValidators to explicitly have the legacy API contract. After merging this, existing Filters in Main will continue to work just fine, though new ones added won't work until my later PRs are merged. I'm trying to do this one in phases to make it easier to review.


pkg/mcp/webhookhelper.go
```diff
@@ -10,6 +10,24 @@ import (
 	"k8s.io/client-go/tools/cache"
 )
 
+const DeviceFilterSurfaceIndex = "device-filter-surfaces"
+
+// DeviceFilterSurfaceIndexKeys returns applicable surfaces only for enabled, configured Device
+// Filters. Keeping this separate from MCP resource indexes prevents device selection from being
+// inferred from MCP applicability.
+func DeviceFilterSurfaceIndexKeys(obj any) ([]string, error) {
+	filter, ok := obj.(*v1.MCPWebhookValidation)
```
njhale, line 19: I think we should probably use an unchecked assertion or return an error when it's not the right type here, since if this assertion fails, we're not setting the indexer up correctly in the first place.

njhale (approved):

pkg/mcp/webhookhelper.go
g-linville: thanks, fixed (returning an error)

Issue #7537, opened 2026-08-12
enhance: allow admins to filter/search audit logs (LLM and MCP) by API key


calvinmclean:
## Description

Admins can currently filter MCP and LLM audit logs by User. We also want to be able to filter/search by an Agent Auth Scope API key.

## Requirements
- Replace hard deletion of Agent Auth Scope API keys with immediate revocation and retain the details for auditing
- Update MCP and LLM audit logs to record integer ID foreign key to the API key
- Add a new filter to MCP and LLM audit logs to filter by API key
- Never show actual secret API key
- Since we cannot show actual API keys in the UI, we need another way to display. The keys usually have names, but they are often "CLI token" which is not very useful to search. Instead, we can display as a label/badge within the current User column
    - Keys also have stable prefixes using the User ID and Key ID: `ok1-2-32-*****` which we could use
    - Optional: since API Keys might be difficult to search/filter in the filter selection menu, consider allowing clicking the API key label/badge in the table to add that filter
- In the filter selection UI, we can still have a separate dropdown to search API keys with a tooltip to suggest filtering by user first. Explore the possibility to include user's name when no user is shown "CLI token (Calvin McLean)"

## Open Questions
- Should we show key IDs to help disambiguate since many names are duplicated? I generally prefer to keep the integer IDs from the DB internal, but it might be useful to show here. We already do this indirectly in the Agent Auth Scopes API (`ok1-2-32-*****`)
- Some keys do not have names. We could show the key prefix here or use "Unnamed API key"
- Should we use this stable key prefix (`ok1-2-32-*****`) in all cases instead of names? Or only when missing?

cjellick:
> Should we show key IDs to help disambiguate since many names are duplicated? I generally prefer to keep the integer IDs from the DB internal, but it might be useful to show here. We already do this indirectly in the Agent Auth Scopes API (ok1-2-32-*****)

I think we should keep the db id internal and i think we SHOULD show `ok1-2-32-*****`

> Some keys do not have names. We could show the key prefix here or use "Unnamed API key"

if the key is unamed, just show the `ok1-2-32-*****` bit

> Should we use this stable key prefix (ok1-2-32-*****) in all cases instead of names? Or only when missing?
 
If we have the name, we should show it and let users filter by it as well. it will likely be easier for them to reason about.

will-chan:
Just a reminder to extend this concept to LLM token usage too.

calvinmclean:
[...]

PR #7541, opened 2026-08-12
feat: revoke API keys instead of delete

calvinmclean:
This will allow audit tables to reference "deleted" keys and still retain necessary info

Related: #7537

### DIFF
--- a/pkg/gateway/client/apikey.go
+++ b/pkg/gateway/client/apikey.go
@@ -473,12 +473,6 @@ func (c *Client) revokeAPIKey(ctx context.Context, keyID uint, userID *uint) err
 	return nil
 }
 
-// DeleteAPIKeyByID is retained for callers that still use the old lifecycle
-// name. It revokes the key and does not delete retained metadata.
-func (c *Client) DeleteAPIKeyByID(ctx context.Context, keyID uint) error {
-	return c.RevokeAPIKeyByID(ctx, keyID)
-}
-
 // UpdateAPIKeyLastUsed updates the last_used_at timestamp for an API key
 // if more than a minute has elapsed since the previous timestamp.
 func (c *Client) UpdateAPIKeyLastUsed(ctx context.Context, key *types.APIKey) error {
--- a/pkg/gateway/client/apikey.go
+++ b/pkg/gateway/client/apikey.go
@@ -41,6 +41,9 @@ func cloneAPIKey(apiKey types.APIKey) types.APIKey {
 	if apiKey.ExpiresAt != nil {
 		cloned.ExpiresAt = new(*apiKey.ExpiresAt)
 	}
+	if apiKey.RevokedAt != nil {
+		cloned.RevokedAt = new(*apiKey.RevokedAt)
+	}
 	return cloned
 }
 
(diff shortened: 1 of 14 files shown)


calvinmclean:
Requested review too early. Moving to draft

thedadams (commented):
I suggest we do delete keys after some period of time. Perhaps with the same retention period as the corresponding logs we are trying to display?

pkg/gateway/client/apikey.go
```diff
@@ -198,16 +201,17 @@ func (c *Client) CreateHostedAgentAPIKey(ctx context.Context, instanceID string,
...
 		return nil
 	}
-	if err := c.db.WithContext(ctx).Where("hosted_agent_instance_id = ?", instanceID).
-		Delete(&types.APIKey{}).Error; err != nil {
-		return fmt.Errorf("failed to delete hosted agent API keys: %w", err)
+	if err := c.db.WithContext(ctx).Model(&types.APIKey{}).
+		Where("hosted_agent_instance_id = ?", instanceID).
+		Where("revoked_at IS NULL").
+		Update("revoked_at", time.Now().UTC()).Error; err != nil {
+		return fmt.Errorf("failed to revoke hosted agent API keys: %w", err)
```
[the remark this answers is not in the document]
calvinmclean: Cleaning up the API keys requires an additional query and TX to get the list of IDs for the hosted_agent_instance_id. The value is minimal since we still have to validate cache hits by checking the revocation.

pkg/gateway/client/retention.go
```diff
@@ -0,0 +1,110 @@
...
+	}()
+	waitGroup.Wait()
+	return errors.Join(mcpErr, llmErr)
+}
+
+func (c *Client) deleteOldRevokedAPIKeys(ctx context.Context, now time.Time, retentionDays int) error {
+	if retentionDays <= 0 {
+		return nil
+	}
+	if ctx.Err() != nil {
```
njhale: I recently added another `deleteOld*` function for device scans, and @thedadams pointed out that we don't need to check for this since Gorm will exit and return the wrapped context cancellation error.

njhale (approved):

pkg/gateway/client/retention.go
calvinmclean: Cool, thanks!

pkg/gateway/client/apikey.go
```diff
@@ -411,16 +454,31 @@ func (c *Client) GetAPIKeyByID(ctx context.Context, keyID uint) (*types.APIKey,
...
 	}
 	c.invalidateValidatedAPIKeysByID(keyID)
 	return nil
 }
 
+// DeleteAPIKeyByID is retained for callers that still use the old lifecycle
+// name. It revokes the key and does not delete retained metadata.
+func (c *Client) DeleteAPIKeyByID(ctx context.Context, keyID uint) error {
+	return c.RevokeAPIKeyByID(ctx, keyID)
+}
```
g-linville: Can this just be removed, and any callers that use it can use the new name? I think it would make the code in those places cleaner.

calvinmclean: Dang I thought for sure I removed this. It looks like it got re-added when I had the agent rebase and fix conflicts. Thanks!

pkg/gateway/client/retention.go
```diff
@@ -0,0 +1,103 @@
...
+	)
+	waitGroup.Add(2)
+	go func() {
+		defer waitGroup.Done()
+		mcpErr = mcpCleanup()
+	}()
+	go func() {
+		defer waitGroup.Done()
+		llmErr = llmCleanup()
+	}()
```
thedadams: I'm surprised the modernize linter didn't catch this.
```suggestion
	waitGroup.Go(func() {
		mcpErr = mcpCleanup()
	})
	waitGroup.Go(func() {
		llmErr = llmCleanup()
	})
```

thedadams (approved):

g-linville (approved):

pkg/gateway/client/retention.go
calvinmclean: Oh yeah, good idea. I think it's because it used `waitGroup.Add(2)`.

PR #7545, opened 2026-08-12
chore: remove enforcement stuff

g-linville:
part of https://github.com/obot-platform/obot/issues/7538

This removes the old experimental enforcement feature from Obot.
I am keeping its old API route around so that any obot-sentry installations can hit it and get a fake approval, in order to avoid breaking any users who have it installed.


g-linville:
Gonna close and reopen this later as a second PR, to get better AI reviews on fewer changes.

Issue #7543, opened 2026-08-12
enhance: filter token usage by Agent Auth Scope API key


calvinmclean:
## Parent

#7537

## What to build

Attribute LLM token-usage accounting records to the Agent Auth Scope API key that authenticated the request and let administrators filter and compare Token Usage by one or more keys. Keep `run_token_activities` as the authoritative accounting source.

This issue does not change the LLM audit-log token or spend schema and does not introduce joins, foreign keys, or request correlation between token usage and LLM audit logs.

## Acceptance criteria

- [x] New token-usage rows snapshot the authenticating API key ID and its non-secret event-time display name; requests not authenticated by an API key remain unattributed.
- [x] Token-usage API responses expose nullable API-key attribution without exposing API-key secrets.
- [x] The Token Usage page has a searchable, URL-backed, multi-select API-key filter that composes with the existing user and model filters.
- [x] Filter options come from fetched usage records so revoked or retained keys with usage remain selectable. Keys use the shared audit-log presentation: `<key name> (<masked key>) • <user name>`; unnamed keys omit the empty name and use `<masked key> • <user name>`.
- [x] An API Keys subview appears alongside Models, Users, and Spend and renders one input/output usage timeline per attributed key from the currently filtered dataset using the same key presentation.
- [x] API-key charts support the existing name/token sorting and search controls; unattributed usage is omitted from this subview.
- [x] Existing unattributed rows remain visible when no API-key filter is active and are excluded when a key filter is selected; no historical backfill is attempted.
- [x] Existing date-range-only Overall Stats behavior and token quota calculations remain unchanged.
- [x] LLM audit-log persistence, responses, exports, and UI remain unchanged by this issue.
- [x] Backend and UI tests cover attribution, historical rows, multi-key filtering, per-key charts, and no-secret guarantees.

## Blocked by

- #7537

sangee2004:
This issue is addressed when testing with the latest build from `main`

`Token Usage` page now provides the ability to filter by `API Keys`
<img width="1166" height="730" alt="Image" src="https://github.com/user-attachments/assets/74c1e4e6-212b-4aa5-aca3-a7c6a984f59f" />

Graphs specific to `API Key` are presented with the ability to search for a specific key:
<img width="1137" height="773" alt="Image" src="https://github.com/user-attachments/assets/12a4d89d-f3fd-437c-b319-eb3e2a97081b" />

Issues found when testing this feature are tracked in separate issues:
[...]

PR #7542, opened 2026-08-12
enhance: add APIKeyID and APIKeyName to LLM and MCP audit logs

calvinmclean:
Addresses https://github.com/obot-platform/obot/issues/7537

## Summary

- Add `APIKeyID` and `APIKeyName` to LLM and MCP audit-log records and API responses.
- Propagate non-secret API-key attribution through authenticated LLM proxy and MCP gateway requests.
- Snapshot MCP attribution during ingestion, including a masked identifier for unnamed or no-longer-present keys.
- Add database columns and indexes for API-key attribution while retaining `RevokedAt` model compatibility.
- Handle missing and transient API-key lookup failures separately without an unbounded cache.
- Preserve internally collected MCP audit events when attribution lookup fails.

PR #7546, opened 2026-08-12
enhance: filter token usage/spend on API keys

calvinmclean:
Addresses https://github.com/obot-platform/obot/issues/7543
Depends on: https://github.com/obot-platform/obot/pull/7542

This adds new columns to the token usage/spend tracking table to associate with an API key.

Screenshot 2026-08-13 at 08 45 34
Screenshot 2026-08-13 at 08 45 44


g-linville (approved):

njhale (approved):

ui/user/src/routes/admin/token-usage/+page.svelte
```diff
@@ -275,14 +293,42 @@
 			});
 		}
 
+		function computePerAPIKey(): PerAPIKeyRow[] {
+			if (!filtered.length) return [];
+			const byAPIKey = new SvelteMap<string, TokenUsage[]>();
```
ivyjeong13: if just for calculations and doesn't cause reactive rendering, this should just be `Map` instead of `SvelteMap`. I think there is an annoying linter rule that says prefers the sveltemap incorrectly and at this point, I'm tempted to just turn that rule off until they fix it. 😅

ivyjeong13 (approved):
just one minor thing, otherwise UI lgtm!

PR #7551, opened 2026-08-13
enhance: add API key filtering to audit logs

calvinmclean:
## Summary

- add API key filtering to MCP, local-agent, and LLM audit logs
- expose API key filter options using event-time names plus retained non-secret key and owner context
- display API key attribution in audit log views and support filtered exports and schedules
- centralize API key prefix and masked-name construction
- preserve historical filter options when API key metadata is unavailable

Part of #7537.


ui/user/src/lib/components/admin/audit-log-exports/CreateScheduleForm.svelte
```diff
@@ -421,10 +440,12 @@
 					label: 'Message Policy Action',
 					description: 'Filter by whether a message policy was triggered',
 					options:
-						filtersOptions['message_policy_triggered']?.map?.((value) => ({
-							id: value,
-							label: value === 'true' ? 'Triggered' : 'Not triggered'
-						})) ?? []
+						filtersOptions['message_policy_triggered']
```
ivyjeong13: nit: Seeing this logic repeated (string arr enforcement and map with label) and thinking it could be a shared helper

Something like:
```
// utils.ts
export const toStringOptsArray = (
	opts?: readonly unknown[],
	getOptionLabel?: (opt: string) => string
) =>
	(opts ?? [])
		.filter((d): d is string => typeof d === 'string')
		.map((d) => ({ id: d, label: getOptionLabel?.(d) ?? d }));
```

```
const getUserDisplayName = (d) => usersMap.get(d)?.displayName ?? d;

// message_policy_triggered
options: toStringOptsArray(
	filtersOptions['message_policy_triggered'], 
	(d) => d === 'true' ? 'Triggered' : 'Not triggered'
)

// actor
options: toStringOptsArray(filtersOptions['actor'], getUserDisplayName)

// user_id
options: toStringOptsArray(filtersOptions['user_id'], getUserDisplayName)

g-linville (approved):
backend lgtm

ui/user/src/lib/components/admin/audit-log-exports/CreateScheduleForm.svelte
calvinmclean: Thanks! added

ivyjeong13 (approved):

thedadams (approved):

PR #7552, opened 2026-08-13
chore: support scheduling settings for hosted agents

drpebcak:


mevip61-lang (approved):

pkg/tunnel/tunnel_test.go
thedadams, line 1: Seems like these changes are unrelated.

drpebcak: They are, but this test was failing in ci: https://github.com/obot-platform/obot/actions/runs/31727788508?pr=7552

thedadams (approved):

Issue #7563, opened 2026-08-14
Tool Preview population fails for remotes with static OAuth configuration


njhale:
**Describe the bug**
Clicking the `Populate Tool Preview` button for any remote catalog entry with static OAuth configured immediately fails without redirecting the user through the auth flow.

**To Reproduce**
1. Log in as admin and go to MCP Catalog
2. Create a remote MCP server catalog entry and configure static OAuth credentials on it
3. Generate tool previews for the entry

**Expected behavior**
The user is redirected after clicking `Populate Tool Preview`, can complete the auth flow, and tool previews for the MCP server are generated.

**Screenshots**
<img width="1221" height="653" alt="Image" src="https://github.com/user-attachments/assets/877dd7f3-8092-4bff-9ab3-6eaa6a2f53e6" />

**Additional context**
Obot server logs contain the following error log:

```
failed to check for MCP auth: failed to get client for server tool-preview-f2d3c61243c685fa: failed to connect MCP client: calling "initialize": sending "initialize": rejected by transport: unexpected status registering client (404): <!DOCTYPE html><html lang="en-US" class="" 
...
static OAuth client lookup also failed: no credentials found for MCP server tool-preview-f2d3c61243c685fa
```

sangee2004:
This issue is addressed when testing with the latest build from `main`.

Able to generate a tool preview for remote MCP servers with a static OAuth configuration. Also able to configure tools for a remote MCP server with a static OAuth configuration that is added as part of a composite MCP server.

PR #7564, opened 2026-08-14
fix: use static OAuth credentials when generating tool previews

njhale:
The OAuth handler needs a server's catalog entry name to resolve
static OAuth credentials. It does this by fetching the `v1.MCPServer`
from storage and reading its spec. This doesn't work for the ephemeral
MCP servers used to generate tool previews, since they are never
persisted to storage.

To fix this, add a `catalogEntryName` field to the OAuth handler so that
it can resolve credentials without fetching the server from storage.

Addresses https://github.com/obot-platform/obot/issues/7563

### DIFF
--- a/pkg/api/handlers/mcpcatalogs.go
+++ b/pkg/api/handlers/mcpcatalogs.go
@@ -1038,12 +1038,7 @@ func (h *MCPCatalogHandler) generateCompositeToolPreviews(req api.Context, entry
 		}
 
 		// CatalogEntry-based components still use a temporary server with the supplied config.
-		componentID := componentEntry.CatalogEntryID
-		if componentID == "" {
-			componentID = componentEntry.MCPServerID
-		}
-
-		config, ok := configRequest.ComponentConfigs[componentID]
+		config, ok := configRequest.ComponentConfigs[componentEntry.ComponentID()]
 		if !ok {
 			// No config provided for this component, skip it
 			continue
@@ -1062,7 +1057,7 @@ func (h *MCPCatalogHandler) generateCompositeToolPreviews(req api.Context, entry
 			req.ObotNamespace,
 			h.secretBindingAllowedLabel,
 			entry.Namespace,
-			entry.Name,
+			componentEntry.CatalogEntryID,
 			catalogName,
 			componentEntry.Manifest,
 			config.Config,
@@ -1289,7 +1284,7 @@ func (h *MCPCatalogHandler) GenerateComponentToolPreviews(req api.Context) error
 		req.ObotNamespace,
 		h.secretBindingAllowedLabel,
 		composite.Namespace,
-		composite.Name,
+		component.CatalogEntryID,
 		catalogName,
 		component.Manifest,
 		configRequest.Config,
@@ -1420,7 +1415,7 @@ func (h *MCPCatalogHandler) GenerateComponentToolPreviewsOAuthURL(req api.Contex
 		req.ObotNamespace,
 		h.secretBindingAllowedLabel,
 		composite.Namespace,
-		composite.Name,
+		component.CatalogEntryID,
 		catalogName,
 		component.Manifest,
 		configRequest.Config,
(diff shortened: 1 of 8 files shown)


pkg/api/handlers/mcpgateway/oauth/mcpoauthhandler.go
```diff
@@ -284,20 +305,22 @@ func (m *mcpOAuthHandler) NewState(ctx context.Context, conf *oauth2.Config, ver
 }
 
 func (m *mcpOAuthHandler) Lookup(ctx context.Context, _ string) (string, string, error) {
-	if m.mcpID != "" {
+	catalogEntryName := m.catalogEntryName
+	if catalogEntryName == "" && m.mcpID != "" {
 		var server v1.MCPServer
 		if err := m.client.Get(ctx, kclient.ObjectKey{Namespace: system.DefaultNamespace, Name: m.mcpID}, &server); err == nil {
```
njhale: I kept the `mcpID` -> `catalogEntryName` lookup for real servers to preserve the existing behavior -- returning an error -- when the server doesn't exist.

Not sure it's actually necessary though. If not, I can drop it along with the `Ephemeral` field on `ServerConfig`.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -1506,7 +1506,8 @@ func tempServerAndConfig(ctx context.Context, gatewayClient *gclient.Client, cli
 			Name: tempName,
 		},
 		Spec: v1.MCPServerSpec{
-			Manifest: serverManifest,
+			Manifest:                  serverManifest,
+			MCPServerCatalogEntryName: entryName,
```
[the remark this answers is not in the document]
njhale: Good catch! That's correct.

fixed!

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -1549,7 +1545,7 @@ func tempServerAndConfig(ctx context.Context, gatewayClient *gclient.Client, cli
 		return v1.MCPServer{}, mcp.ServerConfig{}, fmt.Errorf("failed to create OAuth client: %w", err)
 	}
 
-	staticOAuthCred, err := gatewayClient.RevealCredential(ctx, []string{system.MCPOAuthCredentialName(entryName)}, entryName)
+	staticOAuthCred, err := gatewayClient.RevealCredential(ctx, []string{system.MCPOAuthCredentialName(entryName)}, "oauth")
```
calvinmclean: nit: can we introduce a constant for "oauth"? Looks like it's used in a few places

calvinmclean (approved):

pkg/api/handlers/mcpcatalogs.go
njhale: That's reasonable. Something like `StaticOAuthCredentialName`?

calvinmclean: Yeah that sounds good

pkg/api/handlers/mcpgateway/oauth/mcpoauthhandler.go
```diff
@@ -242,10 +250,14 @@ type mcpOAuthHandler struct {
 	userID             string
 	oauthAuthRequestID string
 	urlChan            chan string
+
+	// catalogEntryName is the name of the catalog entry to fetch static OAuth credentials for.
+	// If zero, the name is resolved from mcpID at credential lookup time.
+	catalogEntryName string
```
thedadams, line 246: Can we always set this so we don't have to look it up when the static credentials are fetched?

pkg/api/handlers/mcpgateway/oauth/mcpoauthhandler.go
```diff
@@ -234,25 +234,29 @@ func staticOAuthAuthStyle(method string) oauth2.AuthStyle {
...
 	gatewayClient      *client.Client
 	stateMgr           *stateManager
 	mcpID              string
 	mcpURL             string
 	userID             string
 	oauthAuthRequestID string
 	urlChan            chan string
+
+	// catalogEntryName is the name of the catalog entry to fetch static OAuth credentials for.
+	// It's carried on the handler because the ephemeral servers used to generate tool previews
```
thedadams: Does this comment need some updating?

thedadams (approved):

pkg/api/handlers/mcpgateway/oauth/mcpoauthhandler.go
njhale: I removed the second line here to trim it down.

PR #7568, opened 2026-08-14
fix(oauth): preserve MCP resource during token exchange

calvinmclean:
## Summary

This change is required for compatibility with AWS remote MCP server.

- Preserve the protected-resource identifier discovered from MCP OAuth metadata instead of substituting the MCP transport URL.
- Carry that resource through dynamic, static, and OAuth Debugger flows and include it in the authorization-code token exchange alongside the PKCE verifier.
- Continue omitting the `resource` parameter for Microsoft Entra authorization servers, including the documented U.S. Government and China endpoints.

## Why

AWS Managed MCP advertises a protected-resource identifier that can differ from the server connection URL. Obot previously discarded or replaced that value before the token exchange, causing AWS to reject an otherwise valid authorization code with `oauth2: "INVALID_REQUEST"`.

The resource is now discovered once, persisted with the pending OAuth state, and reused consistently for authorization and token requests. No MCP catalog or AWS DCR allowlist changes are required.

## Testing

- Added a regression test asserting the token request contains the matching `resource`, `client_id`, `redirect_uri`, and PKCE `code_verifier`.
- Added coverage for metadata fallback, static OAuth, and OAuth Debugger resource propagation.
- Added authorization and token request coverage for global and national-cloud Entra endpoints, while preserving RFC 8707 behavior for other providers.
- Verified the full Go test suite with `make test`.

### DIFF
--- a/pkg/api/handlers/mcp_oauth_debugger.go
+++ b/pkg/api/handlers/mcp_oauth_debugger.go
@@ -84,6 +84,7 @@ func (m *MCPHandler) RegisterOAuthDebuggerClient(req api.Context) error {
 	state := strings.ToLower(rand.Text())
 
 	conf := oauthDebuggerConfig(clientID, clientSecret, authServer.AuthorizationEndpoint, authServer.TokenEndpoint, registration.TokenEndpointAuthMethod, firstString(registration.RedirectURIs), registration.Scope)
+	resourceURL := mcp.OAuthResourceURL(authServer.AuthorizationEndpoint, serverConfig.URL)
 	if err := req.GatewayClient.CreateMCPOAuthPendingState(
 		req.Context(),
 		req.User.GetUID(),
@@ -92,6 +93,7 @@ func (m *MCPHandler) RegisterOAuthDebuggerClient(req api.Context) error {
 		OAuthDebuggerPendingStateMarker,
 		state,
 		oauth2.GenerateVerifier(),
+		resourceURL,
 		conf,
 	); err != nil {
 		return err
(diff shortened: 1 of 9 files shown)


pkg/api/handlers/mcp_oauth_debugger.go
```diff
@@ -175,7 +177,11 @@ func (m *MCPHandler) ExchangeOAuthDebuggerToken(req api.Context) error {
 		return err
 	}
 	exchangeContext := context.WithValue(req.Context(), oauth2.HTTPClient, httpClient)
-	token, err := conf.Exchange(exchangeContext, input.Code, oauth2.VerifierOption(pendingState.Verifier))
+	exchangeOptions := []oauth2.AuthCodeOption{oauth2.VerifierOption(pendingState.Verifier)}
+	if pendingState.ResourceURL != "" {
+		exchangeOptions = append(exchangeOptions, oauth2.SetAuthURLParam("resource", pendingState.ResourceURL))
```
thedadams: This causes an issued with some auth providers. There is likely a hint in nanobot as to which ones don't like the resource parameter.

pkg/api/handlers/mcpgateway/oauth/statemanager.go
```diff
@@ -49,7 +49,11 @@ func (sm *stateManager) createToken(ctx context.Context, state, code, errorStr,
 		conf.Scopes = strings.Split(ps.Scopes, " ")
 	}
 
-	token, err := conf.Exchange(ctx, code, oauth2.SetAuthURLParam("code_verifier", ps.Verifier))
+	exchangeOptions := []oauth2.AuthCodeOption{oauth2.VerifierOption(ps.Verifier)}
+	if ps.ResourceURL != "" {
+		exchangeOptions = append(exchangeOptions, oauth2.SetAuthURLParam("resource", ps.ResourceURL))
```
thedadams: Same comment here.

pkg/api/handlers/mcp_oauth_debugger.go
calvinmclean: Thank you. I updated these to use `mcp.ExchangeOAuthToken` which handles removing the resourceURL for Microsoft stuff

calvinmclean: The latest commit is pushed but is not loading here. I will keep an eye on it and re-request review when it's updated

thedadams (approved):

njhale (approved):

Issue #7309, opened 2026-07-24
Support 2026-07-28 MCP spec


cjellick:
https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/

In particular, we want to support the stateless mode, but really we want to support the whole thing.

thedadams:
I made a quick video to describing the testing approach. Let me know if you have any questions.
https://caps.thedadams.com/s/r23wqjn1rbbfymq

sangee2004:
We now have support for  2026-07-28 MCP spec.

- Able to connect successfully from MCP Inspector using 2026-07-28 MCP protocol to MCP servers that suppport this spec like vercel mcp server


1. Create a remote MPC server catalog entry for Vercel using the URL: https://mcp.vercel.com
2. Connect to this MCP server from MCP Inspector (npx @modelcontextprotocol/inspector@latest) using the Modern (2026-07-28, sessionless) protocol.
<img width="606" height="338" alt="Image" src="https://github.com/user-attachments/assets/83ff1e58-d92c-4df9-8056-679c60d9c8eb" />

Able to connect successfully and make tool calls as expected.

- Able to connect and make successful tool calls to hosted MCP servers using 2026-07-28 MCP protocol  and the legacy MCP protocol which using `Initialize` method.

1. Create a hosted MCP entry - (npx @modelcontextprotocol/server-everything)
2. Connect to this MCP server from MCP Inspector (npx @modelcontextprotocol/inspector@latest) using the Modern (2026-07-28, sessionless) protocol.
Able to connect successfully and make tool calls as expected.
3. Connect to the same MCP server from VS code client which still uses legacy MCP protocol with `Initialize` method.
Able to connect successfully and make tool calls as expected.
5. Connect to any hosted  MCP server entry like `Github Enterpise` from MCP Inspector (npx @modelcontextprotocol/inspector@latest) using the Modern (2026-07-28, sessionless) protocol.
Able to connect successfully and make tool calls as expected.
6. Connect to the same  hosted MCP server -`Github Enterpise`  from VS code client which still uses legacy MCP protocol with `Initialize` method.
Able to connect successfully and make tool calls as expected.

- Obot server uses  2026-07-28 MCP protocol to connect with MCP servers which can be seen when listing tools for connected MCP servers. If  the server does not support  2026-07-28 MCP protocol , then we fall back to using `Initialize` method.

[...]

PR #7569, opened 2026-08-14
enhance: use MCP Go SDK for gateway instead of nanobot

thedadams:
https://github.com/obot-platform/obot/issues/7309


pkg/api/handlers/mcpgateway/handler.go
```diff
@@ -208,21 +270,16 @@ func (h *Handler) Proxy(req api.Context) error {
...
-			return mcp.ServerNanobotConfig(serverConfig, isCompositeLoopback), nil
+			return mcp.ServerNanobotConfig(serverConfig), nil
 		},
 	}
 
 	ctx := req.Context()
 	ctx = ntypes.WithNanobotContext(ctx, nanobotCtx)
-	if isCompositeLoopback {
-		// Don't audit log composite loopback requests, they are internal to the MCP gateway
-		ctx = nmcp.WithAuditLogMetadata(ctx, map[string]string{mcp.AuditLogIgnore: "true"})
```
g-linville, line 222: It looks like we are now unconditionally ignoring audit? At the very least, the comment is outdated. Not sure what the intended behavior is here, but something looks off to me.

g-linville: If you got rid of the `mcp.AuditLogIgnore` line, then just resolve this comment. It looks like github is still having issues. Your diff still looks like it has it in the diff viewer for me, but here in my comment, the whole thing is red, indicating it was removed.

thedadams: This code only handles the composite loopback (/mcp-connect-composite/...) and we don't audit log at that level.

There is an if-statement on line 142 above that handles the `/mcp-connect/...` endpoints.

g-linville: Oh okay, the removal of the `if` statement made me think that this code was more general than that but I guess not. Cool.

g-linville (approved):

njhale (approved):
Lgtm.

Sorry about the merge conflict. Seems to have been caused by the logs change merging.

njhale (approved):

PR #7573, opened 2026-08-16
fix(auth): retry auth provider sync conflicts

hbanerjee74:
## Summary

- retry optimistic-concurrency conflicts while updating an environment-backed auth provider's sync annotation
- re-read the latest `AuthProvider` resource before each update attempt
- add a regression test that injects a conflict on the first update and verifies the retry succeeds

## Root cause

During Studio's data-encryption-key continuity test, Obot's startup hook raced another update to the same `AuthProvider`. A normal Kubernetes resource-version conflict escaped the hook and caused a process panic and container restart. Later restarts encountered temporary OIDC discovery failures while the Studio API was intentionally being recreated, but those were downstream effects of the initial conflict-driven restart.

This change handles the transient conflict at the update boundary so Obot remains running.

## Verification

- `go test ./pkg/controller/handlers/provider -run 'TestEnsureAuthProviderEnvCredential(RetriesSyncAnnotationConflict|CreatesCredentialAndSyncsProvider)' -count=1`
- `make test`
- `make build`
- `PATH="$(go env GOPATH)/bin:$PATH" make validate-go-code`
- unchanged Studio regression spec with the locally built fixed Obot image: `OBOT_IMAGE=obot-vibedata:auth-provider-conflict-retry npx playwright test --config tests/containers-e2e/playwright.config.ts tests/containers-e2e/data-encryption-key-continuity.spec.ts --workers=1 --retries=0` (6 passed; Obot restart count remained 0)

## Related failure

- Studio Rolling regression run `31959736997`, job `95196291080`


hbanerjee74:
Opened against the wrong base repository; closing in favor of the accelerate-data/obot fork PR.

PR #7603, opened 2026-08-19
enhance: paginate groups

g-linville:
This adds pagination for groups coming from the auth provider. This is cursor-based pagination, not limit+offset.

One thing of note: with these changes, we start saving groups in the `groups` table if they are assigned a GroupRoleAssignment or if they are part of some access policy (i.e. AccessControlRules, SkillAccessRules). Previously, this table only held info about groups that at least one user belongs to. (It still holds those groups as well, with these changes.)

### DIFF
--- a/pkg/gateway/client/group.go
+++ b/pkg/gateway/client/group.go
@@ -2,9 +2,11 @@ package client
 
 import (
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"hash/fnv"
 	"io"
 	"log/slog"
 	"net/http"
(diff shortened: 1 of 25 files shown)


pkg/gateway/types/group.go
```diff
@@ -7,23 +7,57 @@ import "time"
 type Group struct {
 	// ID is the globally unique identifier for the group.
 	// Each auth provider should use a different prefix for their groups to avoid collisions with other providers.
-	ID string `json:"id" gorm:"primaryKey;unique"`
+	ID string `json:"id" gorm:"primaryKey;unique;index:idx_group_auth_provider_name,priority:4"`
```
thedadams, line 10: I'm not going to block this PR here, but I wish that we had concrete evidence that we need them and how much they help.

pkg/gateway/server/user.go
```diff
@@ -342,41 +344,126 @@ func (s *Server) deleteUser(apiContext api.Context) (err error) {
...
+		if _, ok := seen[id]; ok {
+			continue
 		}
-		return apiContext.Write(trimmedGroups)
+
+		seen[id] = struct{}{}
+		ids = append(ids, id)
+	}
+
+	return ids, nil
```
thedadams: You could get rid of `ids` and use `seen` and then 
```suggestion
	return slices.Collect(maps.Keys(seen)), nil
```

pkg/gateway/client/group.go
```diff
@@ -37,96 +54,458 @@ func (e *FetchUserGroupsError) Error() string {
...
+	// paging the cached listing out of the database.
+	LastName string `json:"n,omitempty"`
+	LastID   string `json:"i,omitempty"`
+}
+
+const groupCursorVersion = 1
+
+// encodeGroupCursor renders a position as an opaque token. An empty position stays empty, which is
+// what tells a caller it has reached the end.
+func encodeGroupCursor(cursor groupCursor) (string, error) {
```
thedadams, line 109: Seems like these three functions (here down) share functionality between Obot and the auth providers. We may want to export them from Obot and import them into the auth providers if they change.

thedadams (approved):

pkg/gateway/types/group.go
g-linville: This is just to reduce latency when we query for specific information about groups (i.e. groups with a GroupRoleAssignment). We check the DB first, based on the group ID and auth provider. Without an index, if there are a lot of groups in the DB, it would be a table scan for each one.

PR #7610, opened 2026-08-20
fix(mcp): strip gateway authorization upstream

calvinmclean:
## Summary

- Remove the inbound Obot MCP gateway bearer token
- Continue applying explicitly configured upstream Authorization headers and upstream OAuth tokens through the existing transport.
- Add focused regression coverage for API-key-only, configured bearer, and OAuth credential paths.

### DIFF
--- a/pkg/api/handlers/mcpgateway/handler.go
+++ b/pkg/api/handlers/mcpgateway/handler.go
@@ -230,43 +230,7 @@ func (h *Handler) Proxy(req api.Context) error {
 		(&httputil.ReverseProxy{
 			Transport: client.Transport,
 			Rewrite: func(r *httputil.ProxyRequest) {
-				// SetXForwarded preserves the X-Forwarded-For handling that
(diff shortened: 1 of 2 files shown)


pkg/api/handlers/mcpgateway/handler.go
```diff
@@ -313,6 +277,50 @@ func (h *Handler) Proxy(req api.Context) error {
 	return nil
 }
 
+func rewriteProxyRequest(r *httputil.ProxyRequest, upstreamURL *url.URL) {
+	// Authorization authenticates the client to Obot and must not be forwarded
+	// to the upstream MCP server. The transport adds any configured or OAuth
+	// Authorization header after this rewrite.
+	r.Out.Header.Del("Authorization")
```
calvinmclean, line 284: this function is extracted for testing. Here is the real change

thedadams (approved):

Issue #7607, opened 2026-08-19
Revoked API keys are not identified in the LLM Gateway and MCP audit logs list view


sangee2004:
**Describe the bug**
After an API key is deleted, existing `LLM Gateway` → `Audit Logs` entries associated with that key continue to display the Actor without indicating that the API key has been revoked. The same key is correctly marked as Revoked in the `BY API Key` filter dropdown. The `Audit Logs` list view should consistently identify revoked API keys so their status is clear when reviewing historical entries.

Pre-req - Have any model provider configured.
Have a model from this provider set as the model provider for Claude Code.
Chat in Claude Code.

**To Reproduce**
1. Log in as admin and delete the API key that is being used for chatting with Claude Code.
2. Go to `LLM Gateway`-> `Audit Logs`.

Notice that the `Actor` value for all audit logs related to this deleted API key does not indicate that the API key is revoked in the Audit Logs list view. We do see it marked as being `Revoked` in the `BY API Key` filter dropdown.

<img width="1526" height="805" alt="Image" src="https://github.com/user-attachments/assets/b3fbbf0c-d828-43e5-9c23-2b42e1127310" />

**Note** - The same behavior is also noted for entries in MCP Audit logs

<img width="1413" height="930" alt="Image" src="https://github.com/user-attachments/assets/d658054d-8d2c-4770-b7c4-5851785eddfe" />

**Expected behavior**
Show the API key as revoked in the `Audit Logs` list view, similar to how it appears in the `BY API Key` filter dropdown list.

sangee2004:
This issue is addressed when testing with the latest buiild from `main`.
Revoked API keys are now identified in the LLM Gateway and MCP audit logs list view.

<img width="1394" height="819" alt="Image" src="https://github.com/user-attachments/assets/e0efb1ba-16f3-40ef-84dd-772e9f40e757" />

<img width="1101" height="607" alt="Image" src="https://github.com/user-attachments/assets/3685c815-8779-447a-80d2-381547d92f9d" />

PR #7627, opened 2026-08-21
fix(ui): identify revoked keys in audit logs

calvinmclean:
Addresses https://github.com/obot-platform/obot/issues/7607 by showing revoked status of keys in the column

### DIFF
--- a/pkg/auditlog/presenter.go
+++ b/pkg/auditlog/presenter.go
@@ -70,7 +70,7 @@ func presentMCP(event *api.AuditLogEvent, log gatewaytypes.MCPAuditLog, opts Pre
 	event.EventType = api.AuditLogEventTypeMCPCall
 	// MCPFields.APIKey is a legacy field that may contain the bearer token. Only expose the
 	// event-time, non-secret display snapshot through the normalized audit-log API.
-	event.Actor = mcpActor(log.UserID, apiKeyDisplayName(log.UserID, log.APIKeyID, log.APIKeyName), apiKeyID(log.APIKeyID))
+	event.Actor = mcpActor(log.UserID, apiKeyDisplayName(log.UserID, log.APIKeyID, log.APIKeyName), log.APIKeyRevoked)
 	event.Action = api.AuditLogAction{
 		Operation: mcp.CallType,
 		Name:      mcp.CallIdentifier,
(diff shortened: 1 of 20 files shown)


pkg/auditlog/presenter.go
```diff
@@ -145,18 +145,27 @@ func apiKeyDisplayName(userID string, apiKeyID *uint, name string) string {
 	return fmt.Sprintf("%s (%s)", name, maskedKey)
 }
 
-func mcpActor(userID, credentialID string) api.AuditLogActor {
+func apiKeyID(id *uint) uint {
```
njhale: whenever I need to do this I'm always tempted to:

```go
func valueOrZero[T any](p *T) (zero T) {
	if p != nil {
		return *p
	}

	return
}
```

njhale (approved):

pkg/auditlog/presenter.go
calvinmclean: It's the new `new()`:
```
func ptr[T any](v T) *T {
        return &v
}
```
maybe Go will add a built-in eventually 😆 I think it's fine as-is since it's the only use in this package

ui/user/src/lib/components/admin/audit-logs/AuditLogsPageContent.svelte
```diff
@@ -365,6 +348,36 @@
...
+		const apiKeyIDBatches = batchAuditLogAPIKeyIDs(
+			remoteAuditLogs.map((auditLog) => auditLog.actor.apiKeyID)
+		);
+		if (apiKeyIDBatches.length === 0) {
+			apiKeyFilterOptions.clear();
+			return;
+		}
+		Promise.all(
+			apiKeyIDBatches.map((apiKeyID) =>
+				UserService.listAuditLogFilterOptions('api_key_id', {
```
ivyjeong13: I'm wondering if it's possible to make this a bulk request endpoint, if this is making a request per individual apiKeyID? Unless we're pretty sure the amount of unique apiKeyIDs will be pretty low!

ui/user/src/lib/components/admin/audit-logs/AuditLogsPageContent.svelte
```diff
@@ -365,6 +348,36 @@
...
+					...allFilters,
+					api_key_id: apiKeyID,
+					offset: null,
+					signal: controller.signal
+				})
+			)
+		)
+			.then((responses) => {
```
ivyjeong13: If any of the calls could be slow, might be worth considering changing to handle each response as it arrives instead of waiting for them all to complete with Promise.all

ui/user/src/lib/auditlogs.ts
```diff
@@ -13,37 +13,47 @@ export function isAuditLogAPIKeyFilterOption(
...
 
 export function formatAuditLogAPIKeyName(name: string, maskedKey: string): string {
 	return name && maskedKey && name !== maskedKey ? `${name} (${maskedKey})` : maskedKey || name;
 }
 
+export function formatAuditLogCredentialLabel(credential: string, revoked: boolean): string {
+	return [credential, revoked ? 'Revoked' : ''].filter(Boolean).join(' · ');
+}
+
+export function batchAuditLogAPIKeyIDs(
```
ivyjeong13: Ah, sorry didn't see this till the end! Can ignore my first comment! (though now I'm curious on the max case, if 10 calls (1000 api key ids batched each) plus the filter calls for each filter, how does the filter drawer in the UI handle?)

calvinmclean:
@ivyjeong13 thank you for the comments! I took the suggestions and ended up refactoring to do more of this on the backend in the original responses.
@njhale I dismissed your review since I made some more backend changes

thedadams (approved):

ivyjeong13 (approved):

PR #7629, opened 2026-08-21
enhance: replace nanobot with mmmcp completely

thedadams:
We replace nanobot completely in this change; it is no longer a dependency. This means that we have added some code from nanobot that we need. Also, since we are no longer using nanobot, we don't need token exchange nor external audit log submitting. This functionality is removed.

Issue: https://github.com/obot-platform/obot/issues/7309


pkg/api/handlers/mcpgateway/handler.go
```diff
@@ -12,15 +13,11 @@ import (
...
 
-	"github.com/obot-platform/nanobot/pkg/llm"
-	nmcp "github.com/obot-platform/nanobot/pkg/mcp"
-	"github.com/obot-platform/nanobot/pkg/runtime"
-	"github.com/obot-platform/nanobot/pkg/server"
-	"github.com/obot-platform/nanobot/pkg/session"
-	ntypes "github.com/obot-platform/nanobot/pkg/types"
+	"github.com/modelcontextprotocol/go-sdk/jsonrpc"
+	"github.com/obot-platform/mmmcp"
+	mmmcpconfig "github.com/obot-platform/mmmcp/config"
```
calvinmclean, line 18: legendary import

pkg/api/handlers/mcpgateway/handler.go
```diff
@@ -65,72 +68,87 @@ func auditLogMetadataForPrincipal(metadata map[string]string, user user.Info) ma
...
-		Store:                 sessionStore,
-		AuditLogCollector:     auditLogCollector,
-	})
+	msg, err := jsonrpc.DecodeMessage(body)
 	if err != nil {
-		return nil, err
+		return false
+	}
+	call, ok := msg.(*jsonrpc.Request)
+	if !ok || !call.IsCall() {
```
calvinmclean: Should we check that `call != nil`? It seems like the implementation of DecodeMessage will not return a nil Request but still makes me nervous 🤷

calvinmclean (approved):

pkg/api/handlers/mcpgateway/handler.go
thedadams: Done

pkg/api/handlers/mcpgateway/oauth/token.go
```diff
@@ -346,275 +330,3 @@ func (h *handler) doRefreshToken(req api.Context, oauthClient v1.OAuthClient, re
...
-			return fmt.Errorf("failed to generate token: %w", err)
-		}
-
-		return req.Write(TokenExchangeResponse{
-			AccessToken:     token,
-			IssuedTokenType: tokenTypeAccessToken,
-			TokenType:       "Bearer",
-			ExpiresIn:       max(int(time.Until(expiresAt).Seconds()), 0),
-		})
-	}
```
g-linville, line 538: So I _think_ (but am not entirely sure) that removing this (and other parts of this file that were removed) will break access to the Obot MCP Server for NanobotAgents. I know they are deprecated anyway, but calling this out in case it matters.

g-linville (approved):

pkg/api/handlers/mcpgateway/oauth/token.go
thedadams: It doesn't break it because I've handled it in the gateway. If the request is for the Obot MCP server, then we pass a token for the user the same as what we did here. It just doesn't happen through token exchange.

PR #7638, opened 2026-08-24
chore: remove unused functions and parameters

thedadams:
This removes a bunch of unused input and return parameters mostly. A thorough review is needed to ensure there aren't any bugs (like: "Oh, we're supposed to be using that, but we aren't").


calvinmclean (approved):

g-linville (approved):
Other than the test helpers, I thought about these and nothing came to mind as something that looks like we should have been using. So lgtm

PR #7641, opened 2026-08-24
fix(ui): prevent MCP search input resets while typing

calvinmclean:
## Bug

Addresses https://github.com/obot-platform/obot/issues/7642

Typing in the MCP Servers search bar could feel laggy, and characters could intermittently disappear or reset before a full search term was entered.

## Cause

The shared `Search` component already debounces changes by 300 ms. The MCP Servers page wrapped that callback in another 100 ms debounce, adding unnecessary delay before updating the URL-derived query and extending the window in which typing could feel unresponsive.

## Fix

Remove the redundant page-level debounce and pass search changes directly to the URL update. The shared `Search` component remains responsible for debouncing input.

## Verification

- Manually verified that full search terms can be entered without the observed lag or reset.
- `pnpm run check` passes with zero errors and warnings.
- Prettier and `git diff --check` pass.


calvinmclean:
@claude review

ui/user/src/routes/mcp-servers/+page.svelte
```diff
@@ -19,9 +18,9 @@
 
 	let query = $derived(page.url.searchParams.get('query') || '');
 
-	const updateSearchQuery = debounce((value: string) => {
+	const updateSearchQuery = (value: string) => {
```
ivyjeong13, line 21: Hey Calvin! Just wanted to check, if you remove the rest of the changes except for this, does it resolve the issue for you? I'm thinking this might be the root cause (the 300ms debounce happens for Search which should immediately trigger the update to `query` but instead it has this potential additional 100ms debounce that a user could potentially type in)

calvinmclean: Yeah that seems to do it! Thanks

ivyjeong13 (approved):

PR #7651, opened 2026-08-25
enhance: gate catalog entry deletion on composites and add force delete

njhale:
Deleting a multi-user MCP server referenced by a composite already returned 409
with its dependent composites. Catalog entries had no such check, and neither
path had a way through except editing each composite by hand.

- Apply the same gate to DeleteEntry, matching components by CatalogEntryID.
  Workspace-scoped entries are skipped, since workspaces have no composites.
- Accept ?force=true on both delete endpoints to skip the check. Force does not
  bypass the 403 on component servers or the editable check on Git entries.
- Add a "Delete Anyway" action to the blocked-deletion dialog, wire the conflict
  handler into entry deletes, and handle the conflict on the entry paths in
  EntriesView and McpServerEntryForm, which previously threw unhandled.

### DIFF
--- a/pkg/api/handlers/mcp.go
+++ b/pkg/api/handlers/mcp.go
@@ -66,16 +66,18 @@ type MCPHandler struct {
 	shutdownMCPServer func(string) error
 }
 
-// compositeDeletionDependency represents a composite MCP server or catalog entry that depends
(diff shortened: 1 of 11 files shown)


njhale:
@claude review

PR #7658, opened 2026-08-25
enhance: skip catalog sync if commit sha is unchanged

calvinmclean:
## Summary

Improve Git-backed MCP catalog, system catalog, and skill repository sync performance by resolving the remote commit SHA before cloning.

When a source commit is unchanged, sync skips cloning and parsing that source. Manual Sync remains a force refresh.


calvinmclean:
@claude review

calvinmclean:
@claude review

calvinmclean:
@claude review

pkg/controller/handlers/mcpcatalog/github.go
calvinmclean, line 1: this was simplified significantly and just moved to `mcpcatalog.go` since it was just a single small function

calvinmclean:
@claude review

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
```diff
@@ -140,29 +145,66 @@ func (h *Handler) Sync(req router.Request, resp router.Response) error {
...
+		existingEntries, err := listCatalogEntries(req.Ctx, req.Client, mcpCatalog.Namespace)
+		if err != nil {
+			return fmt.Errorf("failed to list existing catalog entries: %w", err)
+		}
+		persistedEntries := persistedEntriesFromSources(existingEntries, mcpCatalog.Name, skippedSourceIDs)
+		for _, entry := range persistedEntries {
+			if entry.Spec.Manifest.Runtime == types.RuntimeComposite {
+				toAdd = append(toAdd, entry)
+			}
+		}
```
thedadams, line 185: I don't understand what is happening here. Why are we filtering out everything except composites here?

thedadams (commented):
I understand what this change is trying to address. Is this causing problems in some way? How do those problems manifest?

pkg/controller/handlers/mcpcatalog/mcpcatalog.go
calvinmclean: Since existing and unchanged composites can reference newly-synced entries by entryKey, we add them back in here in case they need to be updated

calvinmclean:
> I understand what this change is trying to address. Is this causing problems in some way? How do those problems manifest?

@thedadams I haven't observed any problems, but wanted to get ahead of the constant cloning of our growing catalog. It seemed like it would be a pretty easy thing but ended up being a bit larger

thedadams:
My opinion is that this change is too complicated for something that isn't an issue yet.

Issue #7653, opened 2026-08-25
`roots/list` audit entries show Unknown status and missing response details


sangee2004:
**Describe the bug**
When an external client connects to a hosted MCP server, audit entries generated for `roots/list` operations are reported with an Unknown status. The detail view also does not include the response or response header information. `roots/list` audit entries should report the correct status and display the associated response details.

**To Reproduce**
1. Create a hosted MCP Server for `server-everything` - `npx @modelcontextprotocol/server-everything`
2. Connect to this MCP server from any external client like VS Code and make a few tool calls
3. Log in as admin user and go to `MCP Management`->`Audit Logs` page

Notice that there is an audit entry relating to  `roots/list` that is reported, which hasa  status of `Unknown`
<img width="1245" height="435" alt="Image" src="https://github.com/user-attachments/assets/c8a7ff05-31b7-4d3e-8930-8449b5b94b0a" />

Details of the entry show no response/reponse headers details as well.
<img width="565" height="936" alt="Image" src="https://github.com/user-attachments/assets/aa3f5514-2807-4d60-9539-d892aa5b41f0" />

**Expected behavior**
`Status` for  `roots/list` entries should not report `Unknown` status. Response and Response headers details should be presented in the detail view.

sangee2004:
This issue is addressed when testing with latest build from `main`
Status for `roots/list` entries reports success status. 
<img width="1168" height="143" alt="Image" src="https://github.com/user-attachments/assets/57419873-33a4-48bc-9dba-3b586485e0c8" />
Response and Response headers details are presented in the detail view.

<img width="530" height="763" alt="Image" src="https://github.com/user-attachments/assets/2b41a9e2-77a0-42cc-946c-961201e033da" />

PR #7661, opened 2026-08-26
fix: properly align responses for server requests

thedadams:
Issue: https://github.com/obot-platform/obot/issues/7653


pkg/api/handlers/mcpgateway/proxy_hooks.go
```diff
@@ -108,7 +108,7 @@ func newHookProcessor(req *http.Request, runner mcp.HookRunner, hooks mcp.Hooks,
 		sessionID: mcpSessionID(req.Header, req.URL),
 		disabled:  runner == nil || len(hooks) == 0,
 	}
-	if processor.disabled || req.Method != http.MethodPost || req.Body == nil {
+	if req.Method != http.MethodPost || req.Body == nil || (processor.disabled && processor.audit == nil) {
```
thedadams, line 111: This was the issue. If there were no hooks to run, then we would exit here. However, the client response recording is done below.

njhale (approved):

PR #7664, opened 2026-08-26
enhance: provide config for client native overrides

thedadams:
When using a local agent that supports CIMD, like claude and goose, their redirect URI often contains a port and the CIMD does not. The fix is for the CIMD document to state that the application is "native" but they don't do that. For this reason, we created a "native override" that contains the CIMD URLs for claude code and goose.

This change also provides configuration so that users can add additional CIMD URLs that will be treated the same way.


thedadams:
@claude review

calvinmclean (approved):

PR #7666, opened 2026-08-26
fix: restart auth providers during authentication if configuration changed

g-linville:
for https://github.com/obot-platform/obot/issues/7667

With these changes, during the authentication procedure, we compare the hash of the config of the currently running auth provider daemon to the hash of the config in the database. If they differ, we stop and restart the auth provider daemon with the new configuration.

An unfortunate side-effect of this is that we have to have the controller reconcile the auth providers every 30s. There is an explanation in the code comments (written by me, not an LLM!) to explain why.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,6 +34,7 @@ require (
 	github.com/go-logr/logr v1.4.3
 	github.com/golang-jwt/jwt/v5 v5.3.1
 	github.com/google/jsonschema-go v0.4.3
+	github.com/google/uuid v1.6.0
 	github.com/gorilla/websocket v1.5.4-0.20250319132907-e064f32e3674
 	github.com/jackc/pgx/v5 v5.10.0
 	github.com/keygen-sh/keygen-go/v3 v3.3.0
@@ -190,7 +191,6 @@ require (
 	github.com/google/gnostic-models v0.7.1 // indirect
 	github.com/google/go-querystring v1.1.0 // indirect
 	github.com/google/s2a-go v0.1.9 // indirect
-	github.com/google/uuid v1.6.0 // indirect
 	github.com/googleapis/enterprise-certificate-proxy v0.3.6 // indirect
 	github.com/googleapis/gax-go/v2 v2.15.0 // indirect
 	github.com/gosuri/uitable v0.0.4 // indirect
--- a/pkg/authprovider/configuration.go
+++ b/pkg/authprovider/configuration.go
@@ -1,0 +1,37 @@
+package authprovider
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+
+	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
+)
+
+type daemonConfiguration struct {
+	UID                   string              `json:"uid"`
+	Spec                  v1.AuthProviderSpec `json:"spec"`
+	CredentialEnvironment map[string]string   `json:"credentialEnvironment"`
+}
+
+// DaemonConfigurationHash returns the canonical hash of all persisted inputs
+// that affect an auth provider daemon. Process-level environment is excluded
+// because changing it requires restarting Obot itself.
+func DaemonConfigurationHash(authProvider v1.AuthProvider, credentialEnvironment map[string]string) (string, error) {
+	if credentialEnvironment == nil {
+		credentialEnvironment = map[string]string{}
+	}
+
+	data, err := json.Marshal(daemonConfiguration{
+		UID:                   string(authProvider.UID),
+		Spec:                  authProvider.Spec,
+		CredentialEnvironment: credentialEnvironment,
+	})
+	if err != nil {
+		return "", fmt.Errorf("marshal auth provider daemon configuration: %w", err)
+	}
+
+	digest := sha256.Sum256(data)
+	return hex.EncodeToString(digest[:]), nil
+}
--- a/pkg/controller/authprovider.go
+++ b/pkg/controller/authprovider.go
@@ -26,7 +26,9 @@ import (
 // transaction. This is not trivial, since the credential is outside of kinm and the
 // AuthProvider is inside.
 
-const authProviderReconciliationInterval = 30 * time.Second
+const (
+	authProviderReconciliationInterval = 30 * time.Second
+)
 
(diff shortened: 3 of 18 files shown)


g-linville:
@claude review

pkg/gateway/server/dispatcher/keyedmutex.go
```diff
@@ -0,0 +1,25 @@
...
+	mutex := k.locks[key]
+	if mutex == nil {
+		mutex = &sync.Mutex{}
+		k.locks[key] = mutex
+	}
+	k.lock.Unlock()
+
+	mutex.Lock()
+	return mutex.Unlock
+}
```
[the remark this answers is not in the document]
g-linville: not a concern. there are not that many providers

pkg/gateway/server/dispatcher/dispatcher.go
```diff
@@ -93,37 +96,127 @@ func (d *Dispatcher) URLForAuthProvider(ctx context.Context, namespace, authProv
...
+		u, command, err := d.startDaemon(daemonEnvironment, key, configurationHash, authProvider.Spec.Command, authProvider.Spec.Args...)
+		if err != nil {
+			return url.URL{}, err
+		}
+
+		var current v1.AuthProvider
+		if err := d.client.Get(ctx, kclient.ObjectKey{Namespace: namespace, Name: authProviderName}, &current); err != nil {
+			d.stopDaemonCommand(key, command)
+			return url.URL{}, fmt.Errorf("failed to revalidate provider: %w", err)
+		}
```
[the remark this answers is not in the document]
g-linville: This race is acceptable because it means that just a single request could be processed under the old auth provider config. It doesn't mean that the old config would continue being used indefinitely. So no fix is needed here. It is eventually consistent, and in a short time frame at that.

go.mod
```diff
@@ -34,6 +34,7 @@ require (
 	github.com/go-logr/logr v1.4.3
 	github.com/golang-jwt/jwt/v5 v5.3.1
 	github.com/google/jsonschema-go v0.4.3
+	github.com/google/uuid v1.6.0
```
thedadams, line 37: Use the builtin `uuid` package instead.

pkg/gateway/server/dispatcher/keyedmutex.go
```diff
@@ -0,0 +1,27 @@
...
+
+func (k *keyedMutex) Lock(key string) func() {
+	k.lock.Lock()
+	mutex := k.locks[key]
+	if mutex == nil {
+		mutex = &sync.Mutex{}
+		k.locks[key] = mutex
+	}
+	k.lock.Unlock()
```
thedadams, line 23: We would have better parallelization if we used an `RWMutex` here and only locked when we knew we had to write to the map. Since this is used in a hot path on every request, I think it is worth doing.

pkg/controller/authprovider.go
thedadams, line 1: Instead of doing this, we should just use `resp.RetryAfter` in the controller.

pkg/authprovider/configuration.go
```diff
@@ -0,0 +1,37 @@
...
+	CredentialEnvironment map[string]string   `json:"credentialEnvironment"`
+}
+
+// DaemonConfigurationHash returns the canonical hash of all persisted inputs
+// that affect an auth provider daemon. Process-level environment is excluded
+// because changing it requires restarting Obot itself.
+func DaemonConfigurationHash(authProvider v1.AuthProvider, credentialEnvironment map[string]string) (string, error) {
+	if credentialEnvironment == nil {
+		credentialEnvironment = map[string]string{}
+	}
```
thedadams, line 24: This seems unnecessary because we aren't writing to this map anywhere here.

pkg/controller/handlers/provider/provider.go
```diff
@@ -409,33 +410,30 @@ func (h *Handler) SetAuthProviderConfiguredStatus(req router.Request, _ router.R
...
-			cred.Secrets = make(map[string]string)
-		}
+	cred, err := gatewayClient.RevealCredential(ctx, []string{authProvider.Name, system.GenericAuthProviderCredentialContext}, authProvider.Name)
+	if err != nil && !errors.As(err, &gateway.CredentialNotFoundError{}) {
+		return fmt.Errorf("failed to reveal credential for auth provider %q: %w", authProvider.Name, err)
+	}
+	if cred.Secrets == nil {
+		// Don't pass a nil map so that the provider status functions can distinguish between "no credential found" and "credential found but has no secrets".
+		cred.Secrets = make(map[string]string)
+	}
```
thedadams, line 420: This is confusing because we will pass a non-nil regardless if we find a credential or not.

pkg/gateway/server/dispatcher/daemon.go
```diff
@@ -56,15 +61,38 @@ func (d *Dispatcher) closeDaemons() {
 
 func (d *Dispatcher) stopDaemon(id string) {
 	d.ports.daemonLock.Lock()
-	defer d.ports.daemonLock.Unlock()
+	state, ok := d.ports.daemons[id]
```
thedadams, line 64: Can we change the name of the `d.ports` field to `d.states`?

pkg/gateway/server/dispatcher/daemon.go
```diff
@@ -56,15 +61,38 @@ func (d *Dispatcher) closeDaemons() {
...
+	if ok {
+		delete(d.ports.daemons, id)
+	}
```
thedadams, line 67: We don't need to gate this.

pkg/gateway/server/dispatcher/dispatcher.go
```diff
@@ -93,37 +97,136 @@ func (d *Dispatcher) URLForAuthProvider(ctx context.Context, namespace, authProv
...
 	}
 
-	d.ports.daemonLock.RLock()
-	if port := d.ports.daemonPorts[key]; port != 0 {
-		d.ports.daemonLock.RUnlock()
-		return url.URL{Scheme: "http", Host: fmt.Sprintf("127.0.0.1:%d", port)}, nil
+	unlock := d.providerLocks.Lock(key)
+	defer unlock()
+
+	for {
```
thedadams, line 103: Why do we need a for-loop here?

g-linville:
abandoning this for a controller-based approach

PR #7674, opened 2026-08-27
enhance: optimize MCP servers page load performance

calvinmclean:
## Summary

Addresses https://github.com/obot-platform/obot/issues/7732

- parallelize global layout requests and remove the duplicate profile request: version, license, app preferences, and profile now load concurrently, followed by a concurrent authenticated-data wave for model aliases, models, and notifications
- start the MCP workspace request alongside the parent layout instead of waiting for the parent data first, allowing both request paths to make progress at the same time
- limit MCP catalog loading on the MCP servers page to user-scoped data and avoid duplicate admin/user catalog requests
- render bounded short descriptions as inline Markdown without loading images or full description content
- add `minimal=true` catalog list responses that omit heavyweight description, tool preview, and repository fields while preserving configuration and upgrade notes required by Connect and deployment workflows

## Performance

Production-build HAR comparison using a warm reload baseline:

| Metric | Before | After | Change |
| --- | ---: | ---: | ---: |
| Requests | 365 | 149 | 59% fewer |
| Response bodies | 13.09 MB | 496 KB | 96.2% smaller |
| JSON responses | 2.25 MB | 291 KB | 87% smaller |
| Catalog response | 1.06 MB | 204 KB | 80.8% smaller |
| Catalog data complete | 1,915 ms | 936 ms | 979 ms earlier |

The global layout previously loaded most of its startup data sequentially and requested the profile twice. It now uses two parallel request waves: public/bootstrap data first, then profile-dependent authenticated data. The MCP servers route also begins loading workspaces concurrently with its parent layout, rather than adding that request after the layout's full latency.

The optimized page no longer downloads images embedded in catalog descriptions. It also removes broad admin and power-user-workspace catalog requests from the MCP servers route while keeping admin-created servers visible and inline Connect behavior request-free.

## Testing

- `go test ./pkg/api/handlers`
- `go vet ./pkg/api/handlers`
- `pnpm run check`
- `pnpm run lint`
- `pnpm run test` — 116 tests passed
- production frontend HAR comparison with Vite Preview


calvinmclean:
@claude review

calvinmclean:
@claude review

calvinmclean:
@claude review

calvinmclean:
@claude review

g-linville (approved):

thedadams (approved):
Looks good, but we'll still need Ivy to review.

ivyjeong13 (approved):

Issue #7667, opened 2026-08-26
Auth provider configuration changes are not synchronized across Obot replicas


sangee2004:
**Describe the bug**
In a multi-replica Obot deployment, changes made to an auth provider on one replica are not reflected on the other replica. For example, adding `outlook.com` to `Allowed E-Mail Domains` on pod2 allows logins through that replica, while pod1 continues to reject requests from that domain with a 403 error. Auth provider configuration should be shared and applied consistently across all Obot replicas.

Pre-req - Have Obot server set up with 2 replicas

**To Reproduce**
1. Set up a port forward to one of the Obot replicas, say `pod1` - (`kubectl port-forward pod/<pod1> 8080:8080`)
2. Log in to the obot server and enable the Entra auth provider with the `Allowed E-Mail Domains` set to `obot.ai`
3. Set up the owner account by logging in with an email with domain `obot.ai` 
4. Delete the port forward created in step1. Set up port forwarding to another obot replica, say pod2 -  (`kubectl port-forward pod/<pod1> 8080:8080`)
5. Log in to the Obot server and modify the `Allowed E-Mail Domains` to also allow another domain, say -`outlook.com`
6. Delete the port forward created in step4. Set up port forwarding to the replica from step1  - `pod1` in my case
7. Try to log in with an email with the `outlook.com` domain.

Notice that the user will not be allowed to log in and will be presented with a 403 forbidden error page.

8. Set up port forwarding to the replica from step 4  - `pod2` in my case

Logging in with an email address from the `outlook.com` domain will succeed.

**Expected behavior**
Configuration changes made to the Auth provider should be available to both replicas, and logins attempted in step 7 should succeed without reporting 403 errors.

sangee2004:
This issue is addressed when testing with the latest build from `main` - [v0.0.0-dev+735aa5be](https://github.com/obot-platform/obot/commit/735aa5be)

When following the same steps mentioned in the issue, able to login successfully with an email  from `outlook.com` domain in step 7.

PR #7685, opened 2026-08-28
fix: synchronize auth provider configuration

g-linville:
for https://github.com/obot-platform/obot/issues/7667

This sets up proper synchronizing of provider configuration changes across multiple replicas. I added a new controller that runs in each replica and watches a new singleton called `ProviderDaemonSync`. This contains a mapping of provider names to generations, and each time it increases, the providers are stopped across all replicas so that they restart on the next request with the new configuration.

Actual provider configuration/deconfiguration is now handled by the main leader-elected controller. Previously, we handled deleting/editing/creating the credential in the API, but now we stage a credential (if necessary) and create a ProviderConfigurationChange object which gets processed by the controller, which does the actual configuration change and augments the generation on the ProviderDaemonSync singleton.

For Donnie: in our design discussion, I initially said that the ProviderDaemonSync would be just a timestamp. There were two issues with this: clock skew, and overly impactful daemon restarts, especially for model providers which could have long-lived connections and shouldn't have to restart just because a different provider was [de]configured. So that's why the final design here is a mapping of provider names to generations.


pkg/controller/handlers/providerdaemonsync/providerdaemonsync.go
```diff
@@ -0,0 +1,48 @@
...
+	daemonSync := req.Object.(*v1.ProviderDaemonSync)
+	timestamp := daemonSync.Spec.Timestamp.Time
+	if timestamp.IsZero() {
+		return nil
+	}
+
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	if !timestamp.After(h.lastDaemonRestart) {
+		return nil
```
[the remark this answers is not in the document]
g-linville: We're gonna do this

g-linville:
@claude review

pkg/api/handlers/providerconfigurationchange.go
```diff
@@ -0,0 +1,118 @@
...
+	"k8s.io/apimachinery/pkg/watch"
+	kclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func modelProviderConfigurationChangeName(providerName string) string {
+	return name.SafeConcatName(system.ProviderChangePrefix, providerName)
+}
+
+func stageProviderCredential(req api.Context, secrets map[string]string) (string, error) {
+	stagedName := strings.ReplaceAll(uuid.NewV4().String(), "-", "")
```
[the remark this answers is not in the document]
g-linville: nah we might as well start using the new standard library package

pkg/controller/controller.go
```diff
@@ -254,6 +255,18 @@ func reconcileObotMCPServer(ctx context.Context, storageClient kclient.Client, a
...
 func (c *Controller) PostStart(ctx context.Context, client kclient.Client) {
+	if err := providerconfigurationchange.CleanupOrphanedStagedCredentials(
+		ctx,
+		client,
+		c.services.GatewayClient,
+		time.Now(),
+		providerconfigurationchange.OrphanedStagedCredentialGracePeriod,
+	); err != nil {
+		panic(fmt.Errorf("cleanup orphaned staged provider credentials: %w", err))
+	}
```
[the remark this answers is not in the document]
g-linville: honestly it's not a big deal. in the k8s world, the pods will eventually restart. and it's not like we're gonna accrue a ton of these. I'd rather keep it simple and only do this once poststart

pkg/controller/handlers/providerconfigurationchange/providerconfigurationchange.go
```diff
@@ -0,0 +1,393 @@
...
+	} else {
+		if err := h.deconfigureAuthProvider(ctx, client, authProvider); err != nil {
+			return err
+		}
+	}
+
+	if err := provider.SetAuthProviderConfiguredStatus(ctx, h.gatewayClient, h.licenseProvider, &authProvider); err != nil {
+		return fmt.Errorf("recompute auth provider %q status: %w", authProvider.Name, err)
+	}
+	if err := client.Status().Update(ctx, &authProvider); err != nil {
```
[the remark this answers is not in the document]
g-linville: fixed

njhale (approved):

pkg/api/handlers/providerconfigurationchange.go
```diff
@@ -0,0 +1,121 @@
...
+				cleanupErr = fmt.Errorf("remove staged provider credential %q: %w", change.Spec.StagedCredentialName, cleanupErr)
+			}
+		}
+		if apierrors.IsAlreadyExists(err) {
+			return errors.Join(types.NewErrHTTP(http.StatusConflict, "another provider configuration change is already in progress"), cleanupErr)
+		}
+		return errors.Join(fmt.Errorf("create provider configuration change: %w", err), cleanupErr)
+	}
+
+	if err := waitForProviderConfigurationChangeDeletion(req.Context(), req.Storage, change); err != nil {
```
thedadams: Can we use `wait.For` instead of rolling our own implementation here?

pkg/controller/handlers/provider/provider.go
```diff
@@ -414,7 +414,7 @@ func SetAuthProviderConfiguredStatus(ctx context.Context, gatewayClient *gateway
 		missingConfigParams []string
 	)
 	if len(authProvider.Spec.RequiredConfigurationParameters) > 0 {
-		cred, err := gatewayClient.RevealCredential(ctx, []string{authProvider.Name, system.GenericModelProviderCredentialContext}, authProvider.Name)
+		cred, err := gatewayClient.RevealCredential(ctx, []string{authProvider.Name, system.GenericAuthProviderCredentialContext}, authProvider.Name)
```
thedadams, line 417: 🎉

pkg/controller/handlers/providerconfigurationchange/providerconfigurationchange.go
```diff
@@ -0,0 +1,412 @@
...
+
+	if err := h.advanceDaemonSync(req.Ctx, req.Client, change); err != nil {
+		return err
+	}
+
+	change.Status.Applied = true
+	return nil
+}
+
+func validateChange(change *v1.ProviderConfigurationChange) error {
```
thedadams, line 107: Instead of erring here, should we just delete the configuration change object? This isn't recoverable.

pkg/controller/handlers/providerconfigurationchange/providerconfigurationchange.go
```diff
@@ -0,0 +1,412 @@
...
+	var daemonSync v1.ProviderDaemonSync
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		if err := client.Get(ctx, kclient.ObjectKey{Namespace: system.DefaultNamespace, Name: system.ProviderDaemonSyncName}, &daemonSync); err != nil {
+			if apierrors.IsNotFound(err) {
+				if err := EnsureDaemonSync(ctx, client); err != nil {
+					return err
+				}
+				if err := client.Get(ctx, kclient.ObjectKey{Namespace: system.DefaultNamespace, Name: system.ProviderDaemonSyncName}, &daemonSync); err != nil {
+					return err
+				}
```
thedadams: Can we just do this?
```suggestion
				return EnsureDaemonSync(ctx, client)
```

pkg/controller/routes.go
```diff
@@ -98,6 +101,19 @@ func (c *Controller) setupRoutes() {
...
+
+	// This router intentionally has no leader election. Every replica observes
+	// the singleton and clears only its own process-local provider daemons.
+	if c.services.ProviderDaemonRouter != nil {
+		providerDaemonSync := providerdaemonsync.New(c.services.ProviderDispatcher)
+		c.services.ProviderDaemonRouter.Type(&v1.ProviderDaemonSync{}).
+			Namespace(system.DefaultNamespace).
+			Name(system.ProviderDaemonSyncName).
+			HandlerFunc(providerDaemonSync.Reconcile)
+	}
```
thedadams: I think this should be more separate. Maybe even in another file.

pkg/services/config.go
```diff
@@ -233,6 +233,7 @@ type Services struct {
 	LocalK8sClient            kclient.Client
 	LocalRouter               *router.Router
 	EveryReplicaRouter        *router.Router
+	ProviderDaemonRouter      *router.Router
```
thedadams: Let's give this a more generic name. It may have more functionality added to it later. Something like `EveryReplicaRouter` (and rename the currently named `EveryReplicaRouter` to `K8SEveryReplicaRouter`.

pkg/services/config.go
```diff
@@ -604,6 +605,16 @@ func New(ctx context.Context, config Config) (*Services, error) {
 	if err != nil {
 		return nil, err
 	}
+	providerDaemonRouter, err := nah.NewRouter("obot-provider-daemon-sync", &nah.Options{
```
thedadams: Same comment here about generic name.

pkg/api/handlers/providerconfigurationchange.go
g-linville: Yep, done

pkg/controller/handlers/providerconfigurationchange/providerconfigurationchange.go
g-linville: Yep, fixed

pkg/controller/routes.go
g-linville: Done

pkg/services/config.go
g-linville: Done

pkg/services/config.go
g-linville: Done

pkg/controller/handlers/providerconfigurationchange/providerconfigurationchange.go
g-linville: So I did clean this up some, but your suggestion was not entirely correct. Let me know what you think now.

thedadams (approved):

PR #7692, opened 2026-08-28
fix: show reauthenticate only if oauthmetadata and no configurations

ivyjeong13:
This addresses only showing Reauthenticate/Debug Oauthmeta IFF oauthmetadata is supplied and no editable configurations.


cjellick (approved):

ivyjeong13:
@claude review

PR #7698, opened 2026-08-31
fix: ensure composite manifest populated on update

thedadams:
Prior to this change, it was possible for an update of a composite to completely change its manifest. After this change, we populate the manifest to ensure it is properly updated.


thedadams:
@claude review

njhale (approved):

calvinmclean (approved):

PR #7699, opened 2026-08-31
fix: encrypt sensitive static config

calvinmclean:
## Summary

- store sensitive static MCP configuration in encrypted gateway credentials instead of resource manifests
- redact secrets from API responses while preserving configured-state hints for environment variables, headers, and files
- apply static configuration consistently across catalog entries, user and system servers, composite components, registry responses, webhook validation servers, and MCP connect flows

## Details

- use scoped, collision-safe credential keys and batch credential reads where multiple resources are returned
- migrate legacy inline configuration before normal controller reconciliation
- preserve credential precedence and restore static credentials when dependent resource updates fail
- restart affected servers and composite components when static configuration changes
- copy catalog-managed static credentials when OAuth or another connect flow automatically creates a single-user server
- preserve per-user server scope during automatic creation so the server is not mistaken for a multi-user deployment
- recognize redacted configured static fields in the UI without exposing them as user-supplied configuration
- keep genuinely user-configurable fields visible in launch, edit, composite, filter, and OAuth consent forms
- show static configuration separately from user-supplied configuration in catalog details

## Upgrade behavior

Existing inline static values are extracted into encrypted credentials by the migration included in this PR.


calvinmclean:
@claude review

PR #7700, opened 2026-08-31
refactor: move version checker to new package

calvinmclean:
This change is not strictly necessary, but fixes an organization issue I saw while investigating implementation for a new feature. The background polling job for new Obot versions is not related to the `/version` route handler, so it's moved to a new package and tested more thoroughly.

### DIFF
--- a/pkg/services/config.go
+++ b/pkg/services/config.go
@@ -61,6 +61,7 @@ import (
 	storageservices "github.com/obot-platform/obot/pkg/storage/services"
 	"github.com/obot-platform/obot/pkg/system"
 	"github.com/obot-platform/obot/pkg/tunnel"
+	"github.com/obot-platform/obot/pkg/upgrade"
 	"gorm.io/gorm"
 	gormlogger "gorm.io/gorm/logger"
 	coordinationv1 "k8s.io/api/coordination/v1"
@@ -258,7 +259,6 @@ type Services struct {
 	// environment/Helm config and not modifiable via UI.
 	PSASettingsFromHelm *v1.PodSecurityAdmissionSettings
 
-	DisableUpdateCheck      bool
 	HideK8sDetails          bool
 	MCPRuntimeBackend       string
 	MCPImagePullSecrets     []string
@@ -291,6 +291,7 @@ type Services struct {
 
 	// License provider
 	LicenseProvider *license.Provider
+	VersionChecker  *upgrade.VersionChecker
 }
 
 type hostedAgentPodSchedulingSettings struct {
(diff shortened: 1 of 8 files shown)


calvinmclean:
@claude review

pkg/services/config.go
```diff
@@ -1396,6 +1382,17 @@ func New(ctx context.Context, config Config) (*Services, error) {
...
+	versionChecker, err := upgrade.NewVersionChecker(ctx, upgrade.VersionCheckerOptions{
+		GatewayClient:      gatewayClient,
+		LicenseProvider:    licenseProvider,
+		Engine:             config.MCPRuntimeBackend,
+		DisableUpdateCheck: config.DisableUpdateCheck,
+	})
+	if err != nil {
+		return nil, err
+	}
+	svcs.VersionChecker = versionChecker
```
thedadams: Move this above so that the `VersionChecker` can be instantiated with the rest of the fields.

pkg/upgrade/version_checker.go
```diff
@@ -0,0 +1,206 @@
...
+
+type timer interface {
+	C() <-chan time.Time
+	Reset(time.Duration)
+	Stop()
+}
+
+type realTimer struct {
+	*time.Timer
+}
```
thedadams: Do we need to wrap this? Can with just use the `time.Timer` directly?

pkg/upgrade/version_checker.go
```diff
@@ -0,0 +1,206 @@
...
+	timer := c.newTimer(c.checkInterval)
+	defer timer.Stop()
+
+	for {
+		distribution := "oss"
+		hasValidLicense, err := c.licenseProvider.HasValidLicense(ctx)
+		if err != nil {
+			slog.Debug("failed to refresh license state for upgrade check", "error", err)
+		} else if hasValidLicense {
+			distribution = "enterprise"
```
thedadams: This is no longer true because of the community license. Should we update it?

pkg/upgrade/version_checker.go
calvinmclean: Yeah we ca use time.Timer directly. I kept it this way since it made the tests run without any timers, but using short timers works well enough.

pkg/upgrade/version_checker.go
```diff
@@ -0,0 +1,187 @@
...
+	timer := time.NewTimer(c.checkInterval)
+	defer timer.Stop()
+
+	for {
+		distribution := clienttypes.ProductTelemetryDistributionUnregistered
+		entitlements, err := c.licenseProvider.Entitlements(ctx)
+		switch {
+		case err != nil:
+			slog.Debug("failed to refresh license state for upgrade check", "error", err)
+		case slices.Contains(entitlements, license.CloudEntitlement):
```
calvinmclean, line 126: @thedadams this is all I need for the new entitlement?

thedadams: Yep

thedadams (approved):

PR #7702, opened 2026-09-01
fix(license): ensure all replicas have current entitlements

thedadams:
If the entitlements for a license are rechecked on one replica of Obot, the others won't see those changes until the pod restarts or when the normal license check happens. This causes irregularities with the /api/version endpoint.

This change will ensure the other replicas see the entitlement change after a recheck.


thedadams:
@claude review

g-linville (approved):

njhale (approved):

PR #7725, opened 2026-09-02
fix: avoid catalog status fanout to children

cjellick:
`sql slow query` is our number one log statement. im on a mission to eliminate it!

## Summary

- stop catalog status updates from re-enqueuing every catalog entry and catalog-scoped MCP server
- preserve catalog deletion cleanup with parent-owned finalizers
- retain scheduled source reads, apply, drift repair, and pruning
- document the ownership decision in an ADR

## Why

Catalog children used `DeleteRefs` to reference their parent catalog. Those references also became dependency watches, so routine `MCPCatalog` status writes during an unchanged refresh reconciled hundreds of children at once. The resulting credential queries saturated the database connection pool and appeared as slow SQL despite taking very little time inside PostgreSQL.

## Validation

- Local A/B test with 241 catalog entries, PostgreSQL, a five-connection pool, and 5 ms database latency
- Before: 1,328 credential deletes in a 16-second unchanged-refresh window; 13.21 ms total PostgreSQL execution time, with 200+ ms application timings from pool waiting
- After: zero credential deletes in the equivalent unchanged-refresh window
- Verified a real source change synchronized successfully
- Verified out-of-band drift was repaired
- Verified removing a source entry pruned it
- `go test ./pkg/controller/... ./pkg/storage/...`
- `make lint` (0 issues)

### DIFF
--- a/adr/2026-09-01-catalog-parent-owned-cleanup.md
+++ b/adr/2026-09-01-catalog-parent-owned-cleanup.md
@@ -29,8 +29,9 @@ pruning orphaned entries.
 
 Catalog children no longer register generic deletion references to their parent catalog. MCPCatalog
 and SystemMCPCatalog finalizers instead list and delete their children when the parent itself is
-deleted. Hourly catalog reconciliation, apply, drift repair, and system-catalog pruning remain
-unchanged.
+deleted. The finalizers remain until a follow-up pass confirms that no children remain, and entry
+or catalog-scoped server creation is rejected once a catalog begins terminating. Hourly catalog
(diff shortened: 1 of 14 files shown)


adr/2026-09-01-catalog-parent-owned-cleanup.md
```diff
@@ -0,0 +1,49 @@
+# 2026-09-01: Clean up catalog children from the parent
```
cjellick, line 1: is this ADR useful or just noise?

PR #7726, opened 2026-09-02
fix: reconcile OAuth credentials for catalog entries in one pass

cjellick:
## Why

These handlers are the largest single source of slow query log lines, and the point of this change is to eliminate that noise.

kinm logs any statement over 200ms as `sql query slow`, and the time it measures includes waiting for a connection from a pool of five. The shared environment logs 59,057 of those lines in 24 hours. In the main deployment, which is quieter, the same log holds 2,860 lines, and 801 of them are the single delete this change removes, at a median of 355ms across 194 distinct catalog entries.

None of that is database work. Query Insights puts the mean execution time of the statement itself at about 50 microseconds, and the entries carry `affected=0`, so these are cheap queries that matched nothing and spent their time queueing for a connection.

I want to also explicitly note that I'm trying to keep this PR as small as possible to just get the win we want. There are some pre-existing edge case bugs mentioned at the bottom that I am intentionally not addressing. This pr doesnt make any of them worse; it only makes things better. 

## Where the queries come from

Two controller handlers each queried the credentials table on every reconcile of every catalog entry, and neither checked first whether a credential could exist for the entry it was looking at.

`CleanupUnusedOAuthCredentials` ran a DELETE for every remote entry that does not require static OAuth. In the default catalog that is 168 of 207 entries, so each sweep sent 168 deletes that matched no rows, one query per entry. `EnsureOAuthCredentialStatus` did the same thing with a RevealCredential for every entry that does require static OAuth.

## What changed

`Status.OAuthCredentialConfigured` already records whether a credential exists for an entry, so both queries can be skipped whenever that record cannot be wrong. The two handlers are now one handler, `ReconcileOAuthCredential`, which reads the credential store in two cases.

- The entry requires static OAuth and is not recorded as configured. Reading the store here is also how a wrong record gets corrected, because `SetOAuthCredentials` writes the credential before it sets the sync annotation, and a failure in between leaves a credential the controller has no record of.
- The entry carries the sync annotation, which the API sets after it writes or deletes a credential.

Every other entry, which is most of a catalog, is left alone and no query is sent for it. An entry already recorded as configured is trusted, because every path that removes a credential either sets the sync annotation or deletes the entry outright.

Measured on postgres with `log_statement=all` against a catalog of 45 entries, the statements against the credentials table per full reconcile pass went from 40 to 4.

## Why the two handlers had to become one

nah runs every handler registered for a type even after an earlier one returns an error, because `handlers.Handle` collects errors and keeps looping. While the delete and the status update lived in two handlers, a failed delete did not stop the status from being cleared on the same pass, and once the status was cleared the new guard would never query that entry again, so the OAuth client id and secret stayed in the table with nothing left to remove them.

As one handler the delete has to succeed before the status is cleared, and the sync annotation is cleared last, so a failure anywhere leaves the recheck pending for the next pass.

I checked that this is real rather than theoretical, by adding a row level `BEFORE DELETE` trigger on `credentials` that raises, so a delete matching no rows still succeeds and a delete that would remove a credential fails. With the two separate handlers, flipping an entry off static OAuth logged `failed to delete OAuth credential` and cleared the status anyway, and after removing the trigger no further delete was ever attempted while the client id and secret were still in the table. With the merged handler the status survives the failed delete and the credential is removed on the next pass.

## Why the runtime checks came out

A static OAuth credential is stored under the entry's name, `mcp-oauth-<entry name>`. The runtime is not part of that key, so checking the runtime before deleting is wrong.

Both the cleanup and the finalizer used to return early for any entry whose runtime is not remote. That means an entry which was remote, had a credential configured, and was later changed to another runtime keeps that credential forever. The cleanup skips it for not being remote, the status handler clears its status for the same reason, and nothing else deletes it. That is a leak today, not one this change introduces.

The cleanup now decides from the status instead of the runtime, so the credential is deleted on the pass where the conversion happens, while the entry still records that it had one. An ordinary containerized entry has a false status and still gets no query.

The finalizer had the same hole on the deletion path. It now sweeps a remote entry always, and an entry of any other runtime only when its status says a credential is configured or a sync annotation is still pending. The annotation is in that condition because `SetOAuthCredentials` writes the credential first and sets the annotation second, so in between the entry holds a credential its status knows nothing about, and if it is deleted in that window the annotation is the only thing left pointing at it.

Neither change adds a query in steady state. An entry of another runtime with a clear status is skipped just as it was before, on both paths. The only entries that newly issue a query are the leaked ones, at one delete each, after which their status is false and they go quiet, so the cost is a single small burst on the first reconcile after this deploys.

## Testing

The reconcile is covered as a table of entry states, asserting the number of reads and deletes issued through a fake credential client rather than inferring it from the absence of a panic. Two separate tests cover the ordering, one that a failed delete leaves the status set and one that a failed recheck leaves the sync annotation in place. The finalizer has its own table over runtime, status and a pending annotation.

I confirmed the tests bite by removing each of the two skip guards and the delete error return in turn, and checking that exactly the intended test failed each time.

## Notes

- An entry converted to another runtime before this ships keeps its orphaned credential, the same as today. Repairing those needs a pass over the credential store, which belongs in its own change.
- `RemoveAuditLogCred` has the same unguarded delete on every `MCPServer` reconcile, and is worth a follow up.

### DIFF
--- a/pkg/api/handlers/mcpcatalogs.go
+++ b/pkg/api/handlers/mcpcatalogs.go
@@ -1979,6 +1979,17 @@ func (h *MCPCatalogHandler) DeleteOAuthCredentials(req api.Context) error {
 		return err
 	}
 
+	// Written before the credential is deleted. The controller decides whether an entry is worth a
+	// credential query from its status and this annotation, so a delete that lands without it
+	// leaves the status reading configured with nothing left to correct it.
+	if entry.Annotations == nil {
+		entry.Annotations = make(map[string]string, 1)
+	}
+	entry.Annotations[v1.MCPServerCatalogEntrySyncAnnotation] = "true"
+	if err := req.Update(entry); err != nil {
+		return fmt.Errorf("failed to trigger reconciliation: %w", err)
+	}
+
 	credName := system.MCPOAuthCredentialName(entry.Name)
 	deleted, err := req.GatewayClient.DeleteCredential(req.Context(), credName, system.StaticOAuthCredentialName)
 	if err != nil {
@@ -1995,15 +2006,6 @@ func (h *MCPCatalogHandler) DeleteOAuthCredentials(req api.Context) error {
 				slog.Warn("failed to delete OAuth tokens for MCP server", "serverName", server.Name, "catalogEntryName", entry.Name, "error", err)
 			}
 		}
-	}
-
-	// Trigger reconciliation to update the status
-	if entry.Annotations == nil {
-		entry.Annotations = make(map[string]string, 1)
-	}
-	entry.Annotations[v1.MCPServerCatalogEntrySyncAnnotation] = "true"
-	if err := req.Update(entry); err != nil {
-		return fmt.Errorf("failed to trigger reconciliation: %w", err)
 	}
 
 	return req.Write(map[string]bool{"deleted": deleted})
--- a/pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
+++ b/pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
@@ -269,11 +269,11 @@ func requiresStaticOAuth(entry *v1.MCPServerCatalogEntry) bool {
 // The delete and the status update must stay in one handler. nah runs every handler for a type
 // even after an earlier one errors, so as two handlers a failed delete did not stop the status
 // being cleared, and a cleared status meant the entry was never queried again.
-func (h *Handler) ReconcileOAuthCredential(req router.Request, resp router.Response) error {
-	return reconcileOAuthCredential(req, resp, h.gatewayClient)
+func (h *Handler) ReconcileOAuthCredential(req router.Request, _ router.Response) error {
+	return reconcileOAuthCredential(req, h.gatewayClient)
 }
 
-func reconcileOAuthCredential(req router.Request, _ router.Response, creds credentialClient) error {
+func reconcileOAuthCredential(req router.Request, creds credentialClient) error {
 	entry := req.Object.(*v1.MCPServerCatalogEntry)
 
 	// Set by the API after it writes or deletes a credential.
--- a/pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
+++ b/pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
@@ -346,17 +346,17 @@ func syncOAuthCredential(ctx context.Context, creds credentialClient, entry *v1.
 }
 
 // RemoveOAuthCredentials removes OAuth credentials when a catalog entry is deleted.
-func (h *Handler) RemoveOAuthCredentials(req router.Request, resp router.Response) error {
-	return removeOAuthCredentials(req, resp, h.gatewayClient)
+func (h *Handler) RemoveOAuthCredentials(req router.Request, _ router.Response) error {
+	return removeOAuthCredentials(req, h.gatewayClient)
 }
 
-func removeOAuthCredentials(req router.Request, _ router.Response, creds credentialClient) error {
+func removeOAuthCredentials(req router.Request, creds credentialClient) error {
 	entry := req.Object.(*v1.MCPServerCatalogEntry)
 
-	// Always sweep a remote entry, since being wrong on the way out strands the credential. Sweep
-	// any other runtime only when the status or a pending annotation says one may still exist.
+	// An entry on its way out is always swept, since being wrong there strands the credential. The
+	// rest of the guard only matters if this is ever called outside the deletion path.
 	_, recheck := entry.Annotations[v1.MCPServerCatalogEntrySyncAnnotation]
-	if entry.Spec.Manifest.Runtime != types.RuntimeRemote && !entry.Status.OAuthCredentialConfigured && !recheck {
+	if entry.Spec.Manifest.Runtime != types.RuntimeRemote && !entry.Status.OAuthCredentialConfigured && !recheck && entry.DeletionTimestamp.IsZero() {
 		return nil
 	}
 
--- a/pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
+++ b/pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
@@ -1,6 +1,7 @@
 package mcpservercatalogentry
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"log/slog"
@@ -10,6 +11,7 @@ import (
 	"github.com/obot-platform/obot/apiclient/types"
 	"github.com/obot-platform/obot/pkg/controller/handlers/mcpserver"
 	gclient "github.com/obot-platform/obot/pkg/gateway/client"
+	gatewaytypes "github.com/obot-platform/obot/pkg/gateway/types"
 	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
 	"github.com/obot-platform/obot/pkg/system"
 	"github.com/obot-platform/obot/pkg/utils"
@@ -18,6 +20,14 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+// credentialClient is the slice of the gateway client the OAuth reconcile uses. Taking it as a
+// parameter rather than storing it keeps the reconcile exercisable without a database behind it,
+// and leaves the Handler holding the concrete client the rest of its methods need.
+type credentialClient interface {
+	RevealCredential(ctx context.Context, contexts []string, name string) (gatewaytypes.Credential, error)
+	DeleteCredential(ctx context.Context, credentialContext, name string) (bool, error)
+}
 
 // Handler handles operations for MCP server catalog entries
 type Handler struct {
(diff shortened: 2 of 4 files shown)


pkg/api/handlers/mcpcatalogs.go
```diff
@@ -1942,7 +1947,7 @@ func (h *MCPCatalogHandler) SetOAuthCredentials(req api.Context) error {
 	}
 	entry.Annotations[v1.MCPServerCatalogEntrySyncAnnotation] = "true"
 	if err := req.Update(entry); err != nil {
-		return fmt.Errorf("failed to trigger reconciliation: %w", err)
+		return rollbackOAuthCredential(req.Context(), req.GatewayClient, credName, err)
```
thedadams: We don't want to do this. A conflict error would be fine. That would mean we are currently processing this entry.

cjellick:
@thedadams - requesting your re-review because i simplified this a good bit from when you reviewed it. I removed that startup query. It was there to correct a pre-existing edge case and I dont care about that. The PR was also originally closing some other edge cases and I dont care about fixing those either. I just wanted the most targetted change possible that cleans up the query storm while not make anything worse (this pr still makes the logic better overall I think).

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
```diff
@@ -19,6 +21,14 @@ import (
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// credentialClient is the slice of the gateway client the OAuth reconcile uses. Taking it as a
+// parameter rather than storing it keeps the reconcile exercisable without a database behind it,
+// and leaves the Handler holding the concrete client the rest of its methods need.
```
g-linville: nit: unnecessary AI comment

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
```diff
@@ -247,93 +257,147 @@ func (*Handler) CleanupNestedCompositeEntries(req router.Request, _ router.Respo
...
+//     credential.
+//   - Everything else is left alone, which is most of a catalog. This used to be the expensive
+//     case: every remote entry that does not require static OAuth issued a DELETE matching zero
+//     rows on every reconcile, 168 of the 207 entries in the default catalog, in a burst of one
+//     query per entry through a connection pool of five.
+//
+// Deleting a credential and clearing the status have to happen in that order and in one handler.
+// nah runs every handler registered for a type even after an earlier one returns an error, so
+// while these were two handlers a failed delete did not stop the status from being cleared, and
+// an entry that lost its status that way would never be looked at again.
```
g-linville: This information can be condensed/some of it is unnecessary.

g-linville (approved):
lgtm. this is a nice optimization

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
```diff
@@ -247,93 +254,116 @@ func (*Handler) CleanupNestedCompositeEntries(req router.Request, _ router.Respo
...
+	return false, nil
 }
 
 // RemoveOAuthCredentials removes OAuth credentials when a catalog entry is deleted.
-func (h *Handler) RemoveOAuthCredentials(req router.Request, _ router.Response) error {
+func (h *Handler) RemoveOAuthCredentials(req router.Request, resp router.Response) error {
+	return removeOAuthCredentials(req, resp, h.gatewayClient)
+}
+
+func removeOAuthCredentials(req router.Request, _ router.Response, creds credentialClient) error {
```
thedadams:
```suggestion
func removeOAuthCredentials(req router.Request, creds credentialClient) error {
```

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
```diff
@@ -247,93 +254,116 @@ func (*Handler) CleanupNestedCompositeEntries(req router.Request, _ router.Respo
...
 	entry := req.Object.(*v1.MCPServerCatalogEntry)
 
-	// Only process remote entries
-	if entry.Spec.Manifest.Runtime != types.RuntimeRemote {
+	// Always sweep a remote entry, since being wrong on the way out strands the credential. Sweep
+	// any other runtime only when the status or a pending annotation says one may still exist.
+	_, recheck := entry.Annotations[v1.MCPServerCatalogEntrySyncAnnotation]
+	if entry.Spec.Manifest.Runtime != types.RuntimeRemote && !entry.Status.OAuthCredentialConfigured && !recheck {
```
thedadams: This only gets called when the catalog entry gets deleted. In this case, we don't need the `recheck` variable. If you want to be very safe (in case this function gets used outside of the deletion process), then you can do the following.
```suggestion
	if entry.Spec.Manifest.Runtime != types.RuntimeRemote && !entry.Status.OAuthCredentialConfigured && !recheck && entry.DeletionTimestamp.IsZero() {
```

thedadams (changes requested):

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
cjellick: done

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
cjellick: done. muich more reasonable comments now.

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
cjellick: ack. done

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
```diff
@@ -247,93 +254,116 @@ func (*Handler) CleanupNestedCompositeEntries(req router.Request, _ router.Respo
...
+	credName := system.MCPOAuthCredentialName(entry.Name)
 
-	// Clear sync annotation if present
-	if _, exists := entry.Annotations[v1.MCPServerCatalogEntrySyncAnnotation]; exists {
-		delete(entry.Annotations, v1.MCPServerCatalogEntrySyncAnnotation)
-		if err := req.Client.Update(req.Ctx, entry); err != nil {
-			return fmt.Errorf("failed to clear sync annotation: %w", err)
+	if requiresStaticOAuth(entry) {
+		if entry.Status.OAuthCredentialConfigured && !recheck {
+			return true, nil
```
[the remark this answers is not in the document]
cjellick: Fixed in c06405f29. `DeleteOAuthCredentials` now writes the sync annotation before deleting the credential, so a failed annotation update leaves the credential in place rather than a status nothing can correct.

One caveat on the framing: this narrows the window rather than closing it. If the controller reconciles and clears the annotation in the gap between the two writes, the delete still lands on a `configured=true` status with no marker. Reaching that now requires a reconcile inside the gap between two consecutive statements, where before it only required the annotation write to fail. Closing it entirely would need a transaction spanning the credential store and the object store.

pkg/controller/handlers/mcpservercatalogentry/mcpservercatalogentry.go
```diff
@@ -247,93 +254,116 @@ func (*Handler) CleanupNestedCompositeEntries(req router.Request, _ router.Respo
...
-	var configured bool
-	if err == nil {
-		configured = true
-	} else if !errors.As(err, &gclient.CredentialNotFoundError{}) {
-		return fmt.Errorf("failed to check credential status: %w", err)
+	// Any credential here is left over from when the entry did use static OAuth. The runtime is
+	// not checked, because an entry changed away from remote keeps its credential under the same
+	// name and nothing else would remove it.
+	if !entry.Status.OAuthCredentialConfigured && !recheck {
+		return false, nil
```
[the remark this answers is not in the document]
cjellick: Not changing this one, deliberately.

The fix you are describing is a rollback in `SetOAuthCredentials`, which is what an earlier revision of this PR did and which @thedadams asked me to remove: a conflict there means the entry is already being reconciled, and that reconcile sees an entry that requires static OAuth with a clear status, reads the store, and records the credential. The rollback also introduced a worse bug, deleting a credential a concurrent successful POST had already reported as saved.

The reorder used for the delete path does not work here either. Writing the annotation first lets the controller reconcile in between, find no credential yet, clear the annotation, and then the credential write lands, which reaches the same state.

So this needs the annotation update to fail and the entry to lose `staticOAuthRequired` before the reconcile that failure queued. It is called out in the PR description under what this does not do.

pkg/api/handlers/mcpcatalogs.go
```diff
@@ -1979,6 +1979,17 @@ func (h *MCPCatalogHandler) DeleteOAuthCredentials(req api.Context) error {
...
 
+	// Written before the credential is deleted. The controller decides whether an entry is worth a
+	// credential query from its status and this annotation, so a delete that lands without it
+	// leaves the status reading configured with nothing left to correct it.
+	if entry.Annotations == nil {
+		entry.Annotations = make(map[string]string, 1)
+	}
+	entry.Annotations[v1.MCPServerCatalogEntrySyncAnnotation] = "true"
+	if err := req.Update(entry); err != nil {
+		return fmt.Errorf("failed to trigger reconciliation: %w", err)
```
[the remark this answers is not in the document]
cjellick: Correct, and this is the known residual rather than something the reorder claimed to fix. I noted the same window when pushing it.

The reorder changes what it takes to reach the bad state, not whether it exists. Before, the annotation write simply had to fail, which is an ordinary conflict. Now the controller has to reconcile, reveal, and clear the annotation inside the gap between two consecutive statements in one request handler.

Closing it properly needs a transaction spanning the credential store and the object store, which we do not have. The alternative, a rollback in the API, was tried in an earlier revision of this PR and removed at @thedadams request.

Leaving it as a narrowing, with the remaining window recorded here and in the PR description.

thedadams (approved):

PR #7738, opened 2026-09-03
Add an ADR for the SQL-backed controller leader lock, and bump nah for its startup logging

cjellick:
Two things, both follow-ups to #7727.

## The ADR

The record that should have accompanied #7727. It covers the decision that shipped there with obot-platform/nah#32: the `obot-controller` election moved from a Lease in kinm's versioned store to one row in a `leader_lock` table, plus the `WithLegacyLeaseLock` option that keeps a single leader through the rolling update that introduces the change.

Written from `adr/template.md` per `adr/README.md`. No related issue exists; the work came out of the shared database CPU investigation. No ODP was written for it either, which the README asks for on significant designs, noted here rather than left implied.

## The nah bump

Picks up obot-platform/nah#33, which fixes what the election logs at startup. The old line said the election was deferring to the legacy Lease whenever the option was set, regardless of whether there was anything to defer to. In the main environment, which had already migrated, both replicas logged this and neither read the Lease even once:

```
leader election obot-controller: lock leader_lock/obot-controller, deferring to the legacy Lease kube-system/obot-controller until the first row exists
```

Now the election logs the lock it uses, and the SQL lock separately reports whether it has a row yet:

```
leader_lock/obot-controller has no row; deferring to the legacy lock kube-system/obot-controller until one exists
leader_lock/obot-controller already has a row; the legacy lock kube-system/obot-controller is configured but will not be read
```

That makes the logs distinguish a migration in progress from one that has finished, which matters while the fleet rolls over.

Note that v0.25.5-rc3 was built before nah#33 merged, so environments upgrading to rc3 still show the misleading line. The corrected logging arrives in the build after this merges.

### DIFF
--- a/adr/2026-09-02-sql-leader-lock.md
+++ b/adr/2026-09-02-sql-leader-lock.md
@@ -1,0 +1,93 @@
+# 2026-09-02: Hold the controller leader lock in a SQL table instead of a Lease
+
+- **Status:** Accepted
+- **Date:** 2026-09-02
+- **Supersedes:** None
+- **Superseded by:** None
+
+## Related issues
+
+None. Implemented in https://github.com/obot-platform/obot/pull/7727 and
+https://github.com/obot-platform/nah/pull/32.
+
+## Related ODPs
+
+None.
+
+## Context
+
+Obot runs its controllers on one replica at a time, chosen by leader election. The
+election is client-go's, and until this change its lock was a `coordination.k8s.io`
+Lease object stored in kinm, Obot's own API server backed by Postgres.
+
+kinm stores objects in versioned, append-only tables. A lock is rewritten every
+renew period, two seconds, so between compactions the Lease accumulated hundreds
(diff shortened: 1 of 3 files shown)


adr/2026-09-02-sql-leader-lock.md
```diff
@@ -0,0 +1,91 @@
...
+replica. The team accepted the window. Running that one deploy with a `Recreate`
+strategy removes it.
+
+Each replica logs at startup which lock its election uses. While it follows a Lease,
+it logs once who holds it, once when the five second wait starts, and once when it
+creates the row.
+
+## References
+
+- nah SQL lock: `pkg/leader/locks/sql.go` and `pkg/leader/leader.go` in
```
njhale, line 81: I'm on the fence about whether we should use permalinks here to avoid reference drift or not.

njhale (approved):

PR #7701, opened 2026-09-01
ui asset 404 user agent gate

cjellick:
- **fix: serve UI assets regardless of User-Agent**
- **fix: set cache headers on the UI's assets and entry points**

PR #7740, opened 2026-09-03
fix: send an ETag for the embedded UI files

cjellick:
Follow-up to #7701, from the review question about why we return neither `ETag` nor `Last-Modified` for the HTML pages that carry `Cache-Control: no-cache`.

## Why there is no validator today

We serve the UI out of an `embed.FS`. Go takes `Last-Modified` from `FileInfo.ModTime`, every file in an `embed.FS` reports the zero time, and `ServeContent` omits the header for a zero time. Go never generates an `ETag` on its own, it only honors one the handler already set, and we did not set one.

## What that costs

`no-cache` tells a client to store the response and revalidate before every use. Revalidation is only cheap when origin can answer `If-None-Match` with a 304, and we gave it nothing to match on, so every navigation and every refresh comes back as a 200 with the whole page.

Checked against obot-qa, which is still on the pre-#7701 build:

| Request | cf-cache-status | Cache-Control | Last-Modified | ETag | Conditional GET |
|---|---|---|---|---|---|
| `/` and `/admin` | DYNAMIC | none | none | none | 200 |
| `/_app/immutable/...js` | MISS then HIT | `max-age=14400` | present | none | 304 |

Origin sends neither header on either path. Both headers on the asset are Cloudflare's, where `max-age=14400` is a 4 hour Browser Cache TTL setting and the `Last-Modified` is stamped when Cloudflare stored the object, not a build time. The asset revalidates against a validator Cloudflare invented at the edge. The HTML is DYNAMIC, so Cloudflare does not cache it and never invents one, which leaves the HTML with no validator anywhere. That is the 200 on refresh the reviewer saw.

## The change

Hash every embedded file once at startup and serve it with that hash as its `ETag`. `ServeContent` reads the `ETag` we set when it checks preconditions, so it answers a matching `If-None-Match` with a 304 itself and nothing else in the handler moves.

A content hash is the right validator here because it is identical on every replica running the same build and changes only when the file changes. `Last-Modified` would have to be invented, since the embedded files carry no timestamp, and process start time would differ per replica and move on every restart even when the file did not change. There is no reason to add both.

Verified on the wire against the real embedded build:

```
/                                      200 cc="no-cache"                              bytes=5140
                                       revalidate -> 304                              bytes=0
/_app/immutable/nodes/30.C7KYbbNh.js   200 cc="public, max-age=31536000, immutable"   bytes=15370
                                       revalidate -> 304                              bytes=0
/favicon.ico                           200 cc=""                                      bytes=15086
                                       revalidate -> 304                              bytes=0
```

## Cost

Hashing the current build is 675 files and about 22MB, measured at 42ms once during startup. If that is not wanted at startup it can be made lazy with a `sync.Map`, but 42ms seemed cheaper than the concurrency.

## Cloudflare

Cloudflare rewrites a strong `ETag` to a weak one when it compresses a response, so the browser revalidates with `W/"..."` against the strong form we set. `If-None-Match` uses weak comparison, so that still returns a 304. There is a test pinning it.

## Two things outside this PR

The Browser Cache TTL of 4 hours is overriding origin. Once #7701 deploys, `public, max-age=31536000, immutable` on the hashed assets gets downgraded to `max-age=14400` before it reaches the browser, so browsers revalidate them every 4 hours for no reason. Switching that setting to "Respect Existing Headers" is what makes the `immutable` header do anything.

Cloudflare is also storing 404s for asset URLs. I poisoned one on QA with a non-browser User-Agent and Cloudflare recorded it, though it expired quickly rather than persisting. The `no-store` on that 404 in #7701 is what stops it, and it is the part of that fix most worth confirming after deploy.

## Tests

- Every served file carries an `ETag`, covering the HTML entry points, the SPA fallback, a hashed asset, and `favicon.ico`.
- Revalidation returns 304 with no body.
- A weakened `W/` form of our own `ETag` still returns 304.
- `Cache-Control` survives the 304, so the client is still told to revalidate next time.
- A stale `ETag` serves the new file, which is the case that matters after a deploy.
- Different files get different `ETag`s, and unchanged content keeps the same one.
- No `ETag` on the 404, which stays `no-store`.

`ui/handler_test.go` also gains `admin.html` in the stub FS, which was missing, so `/admin` was 404ing in tests rather than serving.

### DIFF
--- a/ui/handler.go
+++ b/ui/handler.go
@@ -1,8 +1,11 @@
 package ui
 
 import (
+	"crypto/sha256"
 	"embed"
+	"encoding/base64"
 	"fmt"
+	"io"
 	"io/fs"
 	"net/http"
 	"net/http/httputil"
@@ -28,10 +31,12 @@ type uiServer struct {
 	// fsys holds the built UI. It is a field so tests can supply a stub in place
 	// of the embedded build, which only exists after `make ui`.
 	fsys fs.FS
+	// etags maps a path in fsys to that file's content hash, used as its ETag.
+	etags map[string]string
 }
 
 func Handler(devPort, userOnlyPort int) http.Handler {
-	server := &uiServer{fsys: embedded}
+	server := newUIServer(embedded)
 
 	if userOnlyPort != 0 {
 		server.rp = newUIProxy(userOnlyPort)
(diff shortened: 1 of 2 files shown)


cjellick:
not really worth doing. it doesnt win anything in reality and just adds complexity

PR #7744, opened 2026-09-03
feat: add MCP tester

g-linville:
This adds a new MCP Tester feature to Obot, where users can connect to any server they have access to and chat with the default LLM and make tool calls to test out the MCP server. All of the conversations are ephemeral and are not stored. The LLM gateway handles the chat, so traffic is still captured in the LLM audit logs. There is also an inspector UI for tools, prompts, and resources, where the user can view and use them and make tool calls.

There are tons of UI changes here, but the vast majority of it is new components that I am using for the UI. We are also now using the TypeScript MCP SDK to make tool calls (and these go through Obot's MCP gateway and thus get audit logged properly too).

There aren't any new storage primitives or anything that might need to be migrated if we change fundamental aspects of this feature in the future, so it is a low risk merge despite its large size. (Sorry I was not able to break this down into small chunks.)

The backend changes are mostly to support a common model interface between the UI and the backend, and avoid requiring the UI to be aware of the various LLM API formats (Chat Completions, Responses, Messages, etc.), and instead handle that in the backend.


ui/user/src/lib/components/mcp/tester/json-schema.ts
```diff
@@ -0,0 +1,190 @@
...
+		}
+		if (schema.maxLength !== undefined && value.length > schema.maxLength) {
+			errors.push(`${label} must contain at most ${schema.maxLength} characters`);
+		}
+		if (schema.pattern) {
+			try {
+				if (!new RegExp(schema.pattern).test(value)) errors.push(`${label} has an invalid format`);
+			} catch {
+				// The server remains authoritative when it supplies an invalid pattern.
+			}
```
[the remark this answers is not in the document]
g-linville: we're just gonna not match this regex client side at all

apiclient/types/mcptester.go
```diff
@@ -0,0 +1,128 @@
...
+	MCPTesterErrorProvider            MCPTesterErrorCode = "provider_error"
+	MCPTesterErrorUnsupportedResponse MCPTesterErrorCode = "unsupported_response"
+	MCPTesterErrorCancelled           MCPTesterErrorCode = "cancelled"
+)
+
+// MCPTesterStreamEvent is the only event shape emitted to the browser. Exactly
+// one payload field is populated for each Type.
+type MCPTesterStreamEvent struct {
+	Type   MCPTesterStreamEventType  `json:"type"`
+	Delta  string                    `json:"delta,omitempty"`
```
njhale: nit: This isn't a blocker, but I think we should start applying go-doc comments to all net-new exported types, fields, functions/methods; even if the names are self describing. Right now, it seems like agents don't have great judgement around what's obvious/non-obvious to readers. For example, it would be nice to know that `Delta` is only populated when `Type` is `text_delta`.

I'm curious what other folks think though.

pkg/api/handlers/mcptester.go
```diff
@@ -0,0 +1,191 @@
...
+
+	proxyRequest, err := mcptester.NewLLMProxyRequest(req.Context(), h.serverURL, model, body, req.Request.Header)
+	if err != nil {
+		return writeMCPTesterError(req, http.StatusInternalServerError, types.MCPTesterErrorProvider, "failed to prepare model request", true)
+	}
+	response, err := h.httpClient.Do(proxyRequest)
+	if err != nil {
+		if req.Context().Err() != nil || errors.Is(err, context.Canceled) {
+			return writeMCPTesterError(req, httpStatusClientClosed, types.MCPTesterErrorCancelled, "request cancelled", false)
+		}
```
calvinmclean, line 106: Is this non-standard 499 status necessary? We don't use it anywhere else and, as far as I can tell, it isn't added for a specific communication with the UI.

g-linville: No, I'll replace it. Sorry, I noticed that AI chose that status code and I was unfamiliar with it, but I just looked it up and yeah it's something nginx-specific.

ui/user/src/lib/components/mcp/tester/ChatComposer.svelte
```diff
@@ -0,0 +1,107 @@
...
+		// The box is border-box but scrollHeight excludes the border, so the
+		// borders have to be added back or a single line still overflows by 2px.
+		const style = getComputedStyle(textareaElement);
+		const borders =
+			Number.parseFloat(style.borderTopWidth) + Number.parseFloat(style.borderBottomWidth);
+		const content = textareaElement.scrollHeight + borders;
+		const height = Math.min(Math.max(content, MIN_HEIGHT), MAX_HEIGHT);
+		textareaElement.style.height = `${height}px`;
+		// Only show the scrollbar once the draft actually outgrows the max height.
+		textareaElement.style.overflowY = content > MAX_HEIGHT ? 'auto' : 'hidden';
```
ivyjeong13: overflow-y: 'auto' should do this natively (showing scrollbar once the contents exceeds it container's max height)

edit: Ah adding on: not a blocker! Just the comment seems odd when overflow-y: auto should already be doing that, wondering if something is rendering outside the container causing the scrollbar to always show

pkg/mcptester/normalize.go
```diff
@@ -0,0 +1,427 @@
...
+
+// TODO(g-linville): see if there is an existing function for this
+func firstNonEmpty(values ...string) string {
+	for _, value := range values {
+		if value != "" {
+			return value
+		}
+	}
+	return ""
+}
```
calvinmclean: you can use one of these:
```go
	utils.FirstSet(values...)
	cmp.Or(values...)
```

Maybe `cmp.Or` is preferred since it's built-in?

pkg/mcptester/proxy.go
```diff
@@ -0,0 +1,51 @@
...
+const (
+	anthropicVersion = "2023-06-01"
+)
+
+// NewLLMProxyRequest builds the request the tester handler will send through
+// Obot's existing LLM gateway. Reusing the browser request's authentication
+// headers makes the gateway reconstruct the same user principal, so model
+// access policies, message policies, usage metering, and audit identity remain
+// on the established proxy path. The supplied context propagates cancellation.
+func NewLLMProxyRequest(ctx context.Context, serverURL string, model ResolvedModel, body []byte, inbound http.Header) (*http.Request, error) {
```
calvinmclean, line 24: curious why we make a self-request? Could we just refactor some of the LLM Proxy code so it can be exported and reused?

pkg/mcptester/request.go
```diff
@@ -0,0 +1,428 @@
...
+		Status:  result.Status,
+		Content: result.Content,
+	}
+	body, _ := json.Marshal(value)
+	return string(body)
+}
+
+func toolArguments(arguments json.RawMessage) string {
+	return string(arguments)
+}
```
calvinmclean: nit: can we remove this and just use `string(...)`?

ui/user/src/lib/components/mcp/tester/McpContent.svelte
```diff
@@ -0,0 +1,160 @@
...
+	{/if}
+{:else if type === 'resource_link' && typeof item?.uri === 'string'}
+	<div class={BOX}>
+		{#if externalURL}
+			<!-- eslint-disable svelte/no-navigation-without-resolve -- validated external resource URL -->
+			<a
+				class="link link-primary break-all"
+				href={externalURL}
+				target="_blank"
+				rel="noopener noreferrer">{typeof item.name === 'string' ? item.name : item.uri}</a
```
ivyjeong13: You can remove the eslint-disable and add "external" to `rel` here!

ui/user/src/lib/stores/mcpServersAndEntries.svelte.ts
```diff
@@ -58,9 +58,14 @@ const store = $state<{
 });
 
 function filterOutDuplicateAndDeleted(servers: MCPCatalogServer[]) {
-	return servers.filter(
-		(server, index, self) => index === self.findIndex((t) => t.id === server.id) && !server.deleted
-	);
+	const live = servers.filter((server) => !server.deleted);
```
ivyjeong13: nit: Seeing the extra filters, map, and some, wondering if we could collapse iterations into one, something like:
```
function filterOutDuplicateAndDeleted(servers: MCPCatalogServer[]) {
	const byId = new Map<string, MCPCatalogServer>();
	for (const server of servers) {
		if (server.deleted) continue;
		const existing = byId.get(server.id);
		if (!existing) {
			byId.set(server.id, server);
		} else if (!existing.canConnect && server.canConnect) {
			byId.set(server.id, { ...existing, canConnect: true });
		}
	}
	return [...byId.values()];
}
```

ivyjeong13 (approved):

pkg/mcptester/proxy.go
g-linville: We benefit from all of the middleware for authz, audit logging, etc. which would get basically bypassed if we did it any other way, or made much more complicated at least.

pkg/mcptester/normalize.go
g-linville: Didn't know about this one, I like that

pkg/api/handlers/mcptester.go
```diff
@@ -0,0 +1,190 @@
...
+	if response.StatusCode < http.StatusOK || response.StatusCode >= http.StatusMultipleChoices {
+		return h.writeProxyError(req, response)
+	}
+
+	req.ResponseWriter.Header().Set("Content-Type", "text/event-stream")
+	req.ResponseWriter.Header().Set("X-Accel-Buffering", "no")
+	req.WriteHeader(http.StatusOK)
+	req.Flush()
+
+	_ = mcptester.NormalizeStream(req.Context(), model.Dialect, chatRequest.Tools, response.Body, func(event types.MCPTesterStreamEvent) error {
```
njhale: I could see it being useful to log errors returned by this.

pkg/api/handlers/mcptester.go
```diff
@@ -0,0 +1,190 @@
...
+
+	decoder := json.NewDecoder(bytes.NewReader(body))
+	decoder.DisallowUnknownFields()
+	var request types.MCPTesterChatRequest
+	if err := decoder.Decode(&request); err != nil {
+		return types.MCPTesterChatRequest{}, fmt.Errorf("invalid request body: %w", err)
+	}
+	if err := decoder.Decode(&struct{}{}); !errors.Is(err, io.EOF) {
+		return types.MCPTesterChatRequest{}, fmt.Errorf("request body must contain exactly one JSON value")
+	}
```
njhale:
```suggestion
	if decoder.More() {
		return types.MCPTesterChatRequest{}, fmt.Errorf("request body must contain exactly one JSON value")
	}
```

see https://pkg.go.dev/encoding/json#Decoder.More

calvinmclean (approved):

Issue #7691, opened 2026-08-28
Navigation Updates


ivyjeong13:
Collapse back into one navigation (vs. formerly Obot App & Administration)

## Admin View

- Dashboard
- AI Resources
    - ~~vMCP~~ (Will be added in as a follow-up)
    - MCP Servers [Servers | Sources | Deployments | Filters | Tunnels | Access Policies ]
    - Skills [Skills | Sources | Access Policies ]
    - Models [Providers | Access Policies]
- Operations
    - Audit Logs [MCP | Model]
    - Usage [MCP | Model]
    - Inventory (old devices link)
    - Enforcement Events (old enforcement decisions link)
- Agents [Templates | Harnesses | Pools | Sources]
- Identity & Access  [Users, Agents (Agent Auth Scopes), Groups, Roles, Auth Providers]
- Platform  [License, Branding, Notifications, MCP Config (formerly Server Scheduling), Registry Connections (formerly Image Pull Secrets)]

## Basic User View

- Dashboard
- AI Resources
    - ~~vMCP~~
    - Skills
    - Models
- Policies
- Operations (only see the users’s own resource, nice-to-have)
    - Audit Logs [MCP | Model]
    - Usage [MCP | Model]
    - Inventory (old devices link -- they see their scans)
    - Enforcement Events (old enforcement decisions link) <-- (can basic users see this?)
- Agents (user has to be able to launch agents as it is today)
- Identity & Access [Agents (Agent Auth Scopes)]

Add Obot CLI under Profile dropdown on top right

ivyjeong13:
Will have a follow-up issue to fix Getting Started guides with the navigation updates

sangee2004:
@ivyjeong13 I see some descrapancies in naming and order of the tabs from what is called out in this issue . Wanted to confirm if the ones presented in the latest UI are as expected?

For Admin users:
1. Should the `Deployments` tab appear after `Sources`?

<img width="1467" height="421" alt="Image" src="https://github.com/user-attachments/assets/878ef0bf-2aa1-4cf3-b753-c4ac409aac18" />

2. Should the `LLM` tab be renamed to `Model`?
<img width="1433" height="682" alt="Image" src="https://github.com/user-attachments/assets/d940f093-c546-4683-a275-b0d4480e2b7d" />

<img width="1435" height="663" alt="Image" src="https://github.com/user-attachments/assets/ba99e209-d7dc-45d1-8885-6875f9ffd8c6" />

[...]

Issue #7677, opened 2026-08-27
Revert app scheduling page


cjellick:
=Regarding this issue - https://github.com/obot-platform/obot/issues/7132

I think all I meant was that if im managing the values that appear on the Server Scheduling page via helm, then i should be able to see all those values in a read-only form on that page. 

...i thinjk the server scheduling page probably already behaved this way, so i dont know why i called it out, but i definitely dont want that App Scheduling page. it should be revertered.

sangee2004:
This issue is addressed when testing with the latest build from `main`

When `/api/app-k8s-settings` is accessed, the user is presented with a Forbidden error.

`App Scheduling` page is not shown in UI anymore.

PR #7748, opened 2026-09-03
fix: revert/remove app scheduling endpoint

ivyjeong13:
Addresses #7677 
Note: UI route dropped as part of navigation updates #7691 

/app-scheduling route was removed in navigation changes. This addresses dropping/removing the app k8s settings endpoint.


ivyjeong13:
@claude review

g-linville (approved):

PR #7755, opened 2026-09-04
fix: stop re-sweeping the legacy audit log credential on every reconcile

cjellick:
## Why

`RemoveAuditLogCred` is one of the two largest sources of slow query log lines in the main deployment. Over 24 hours it issues 757 deletes across 141 servers, and every one of them matches no rows.

```
credential DELETEs in 24h: 1,662
  catalog entry OAuth cleanup   905   affected=0: 905   (#7726)
  RemoveAuditLogCred            757   affected=0: 757   (this PR)
```

kinm logs any statement over 200ms as `sql query slow`, and the time it measures includes waiting for a connection from a pool of five. These deletes are individually cheap, around 50 microseconds of actual execution, and there are simply far too many of them.

## What the credential is

An older Obot stored one credential per server holding the token exchange client id and secret and the audit log token, written by `EnsureOAuthClient`:

```go
Context:  server.Name,
ToolName: server.Name,
Env: {TOKEN_EXCHANGE_CLIENT_ID, TOKEN_EXCHANGE_CLIENT_SECRET, AUDIT_LOG_TOKEN}
```

#7629 removed token exchange and external audit log submitting along with the `oauthclients` handler that wrote it, and added `RemoveAuditLogCred` to sweep what was left behind. For a SystemMCPServer the equivalent credential is named `<server>-secret-info`.

Nothing recreates either one, so a server only ever needs sweeping once.

## Why it keeps running

It is registered as a plain handler rather than a finalizer, unlike its two siblings in the same file:

```go
root.Type(&v1.MCPServer{}).HandlerFunc(credentialCleanup.RemoveAuditLogCred)
root.Type(&v1.SystemMCPServer{}).HandlerFunc(credentialCleanup.RemoveAuditLogCred)
```

So instead of running once when a server is deleted, it runs on every reconcile of every server that exists.

## What changed

The handler  records that it has run in an annotation on the server:

```
obot.ai/audit-log-credential-removed
```

The handler returns immediately when that annotation is present, so a server costs one delete and one update, once, and nothing after that.

The annotation is written only after the delete succeeds. A failed delete leaves the sweep pending for the next pass rather than marking a server clean while it still holds the credential.

## Testing

A table covers the MCPServer and SystemMCPServer credential names and the already swept case, which asserts that a seeded credential survives untouched. A separate test closes the gateway client to force the delete to fail and asserts the annotation is not written. These run against a real sqlite backed credential store rather than a fake, using the same helper pattern as the existing tests in this package.

### DIFF
--- a/pkg/controller/handlers/cleanup/credentials.go
+++ b/pkg/controller/handlers/cleanup/credentials.go
@@ -2,6 +2,7 @@ package cleanup
 
 import (
 	"fmt"
+	"log/slog"
 
 	"github.com/obot-platform/nah/pkg/router"
 	"github.com/obot-platform/obot/pkg/api/handlers"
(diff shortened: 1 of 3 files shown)


cjellick:
we have two flakes going in the tests right now
https://github.com/obot-platform/obot/pull/7756
https://github.com/obot-platform/obot/pull/7757

thedadams (approved):

PR #7756, opened 2026-09-04
fix: wait for the default catalog before running integration tests

cjellick:
## Why

`TestMCPServerLifecycle_NPXEverything/create_and_configure` intermittently fails on its first API call:

```
mcp_lifecycle_test.go:44: POST /api/mcp-catalogs/default/entries: 404
  MCPCatalog.obot.obot.ai "default" not found
```

The harness waits for `/api/healthz`, which only means the HTTP server is listening. The default MCPCatalog is created by `SetUpDefaultMCPCatalog` inside the controller's `PostStart` hook, which runs behind leader election with nothing synchronizing it to the HTTP server. So the tests can start before the catalog exists.

When it fails, the stack shows `PostStart` still at `EnsureJWK`, which runs before the catalog setup.

## What changed

Wait for `GET /api/mcp-catalogs/default` to answer before handing the application to the tests, using the same poll and failure handling as `waitForHealth`.

Both waits watch `a.done`, so a server that exits during startup still reports its own error rather than timing out.

### DIFF
--- a/tests/integration/main_test.go
+++ b/tests/integration/main_test.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/obot-platform/obot/pkg/server"
 	"github.com/obot-platform/obot/pkg/services"
+	"github.com/obot-platform/obot/pkg/system"
 )
 
 type obotApplication struct {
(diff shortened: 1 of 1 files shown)


thedadams (approved):

thedadams:
The integration tests are failing. I thought I fixed that flake previously.

cjellick:
> The integration tests are failing. I thought I fixed that flake previously.

well it looks lie the npx flake. 
    --- FAIL: TestMCPServerLifecycle_NPXEverything/launch (60.01s)

cjellick:
merging. @calvinmclean - just tagging you explicitly as an fyi.

PR #7762, opened 2026-09-04
Backport: wake kinm watches with LISTEN/NOTIFY and refresh them on promotion (#7736)

cjellick:
Backport of #7736 to `v0.25-hotfixes`.

This changes how a replica learns about another replica's writes, from polling every 2 seconds to Postgres LISTEN/NOTIFY. Most of the change is in kinm itself, so the bulk of this PR is the pin from `cd2688a2a64e` to `2231aa14b6a7`. On the Obot side, `Factory.Refresh` is called from the controller post start hook so a replica that has just been promoted lists everything again instead of carrying a standby's staleness into the leader.

## Notes on the backport

`go.mod` and `go.sum` were the only conflict. Rather than take main's versions, this keeps the branch's files, bumps only kinm, and runs `go mod tidy`. The result matches main exactly on the lines the original commit changed. The `golang.org/x/crypto`, `term`, `sys` and `text` bumps that came along on main are transitive requirements of the new kinm, so tidy pulled the same ones here. Every other file is byte-identical to the original commit.

The ADR from the original commit is deliberately left on main.

The kinm bump spans six commits: a CI workflow, two dependabot grpc bumps, a dependabot codeowners change, the LISTEN/NOTIFY change ([kinm#29](https://github.com/obot-platform/kinm/pull/29)), and a fix that jitters the initial compaction schedule ([kinm#28](https://github.com/obot-platform/kinm/pull/28)). The compaction jitter fix rides along.

## Testing

- `go build ./...` passes
- `make test` passes, no failures
- `go mod tidy` is stable and the tree is clean
- `make lint` reports 4 staticcheck deprecation issues. All 4 are pre-existing, reproduce on the parent commit, and are in files this change does not touch.


cjellick:
Closing: this was opened from a branch on the upstream repo by mistake. Reopening from a fork instead.

Issue #7768, opened 2026-09-04
Composite catalog entry environment variables cause false deployment drift


calvinmclean:
**Describe the bug**
Composite MCP catalog entries accept top-level `env` configuration even though a composite server has no executable process that can consume those variables. The deployment mapping omits the unsupported field from the running server, and drift detection then compares the running manifest against the catalog entry and immediately marks the deployment as needing an update.

The deployment diff also uses the minimal catalog-entry list response for the new version. Because that response intentionally omits `description` and `toolPreview`, those fields appear to have been removed even though they are unchanged.

**To Reproduce**
1. Create a composite MCP catalog entry with at least one component server and an `env` field on the top-level composite manifest.
2. Deploy or connect to the composite server.
3. Go to `MCP Servers` -> `Deployments`.
4. Notice that the new deployment is immediately marked as needing an update.
5. Select `View Diff` for the deployment.
6. Notice that the top-level environment configuration differs because it was omitted from the running server, while the new version also appears to be missing its description and tool previews.

**Expected behavior**
Top-level environment variables should be rejected when validating composite catalog entries, with guidance to configure them on the applicable component server manifests.

`View Diff` should load the full catalog entry so unchanged descriptions and tool previews are included in the comparison.

**Additional context**
The unsupported composite `env` field is the source of the actual drift. The missing description and tool previews are a display side effect introduced when MCP catalog list responses became minimal.

cjellick:
this should go on honld until the vMCP change stabilitizes. im moving out of v0.26

PR #7769, opened 2026-09-04
fix: prevent false MCP deployment diffs

calvinmclean:
## Summary

- load the full catalog entry before rendering deployment diffs so fields omitted from minimal list responses are not shown as removals
- reject top-level environment variables on composite catalog entries because configuration must belong to an executable component server. This previously would always cause a drift to be detected because these values are not copied to the MCP Server. It is not a valid use-case anyways so it's disabled
- add regression coverage for both the full-detail diff request and the composite deployment drift chain

addresses #7768

### DIFF
--- a/ui/user/src/lib/components/mcp/DeploymentsView.svelte
+++ b/ui/user/src/lib/components/mcp/DeploymentsView.svelte
@@ -399,6 +399,19 @@
 		return prompted;
 	}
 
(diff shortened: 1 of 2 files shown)


ivyjeong13 (approved):

calvinmclean:
moving to draft since this needs to wait on vMCP stuff to wrap up

Issue #7782, opened 2026-09-08
Reusing a valid owner activation link completes password setup but ends with an incomplete setup error


sangee2004:
**Describe the bug**
When the initial owner activation link is opened once and then reused later with the same valid token, setting the password succeeds, but the UI incorrectly ends on a `This setup link is incomplete` error page. The account is actually activated and the new password works, so the backend flow completes while the frontend state is inconsistent. After successful password setup, the user should sign in as the configured initial owner and be taken to the EULA instead of seeing an error.

Pre-req - Helm install of the OBOT server with 5 replicas and having OBOT_SERVER_LOCAL_AUTH_INITIAL_OWNER_EMAIL, OBOT_SERVER_LOCAL_AUTH_INITIAL_OWNER_SETUP_TOKEN_EXPIRATION_HOURS, and OBOT_SERVER_LOCAL_AUTH_INITIAL_OWNER_SETUP_TOKEN configured

**Note**: This issue is also seen in a non-HA Docker setup.

**To Reproduce**
1. Use the activation link with a valid token: `http://localhost:8080/activate#token=<token>`
When the user sees the `Set Your Password` modal, do nothing and close the window.
2. Use the same activation link with a valid token (same as in step 1)
When the `Set Your Password` modal is presented, set a password and click `Set Password and Continue`.

Notice that the user is presented with the following error: `This setup link is incomplete.`

<img width="1217" height="907" alt="Image" src="https://github.com/user-attachments/assets/5248e0ca-ba45-437a-afcf-17491d73ea88" />

The password reset succeeded, and I can log in with the password set in step 3.

**Expected behavior**
The user should not see the `This setup link is incomplete` message; instead, they should log in to obot as the OBOT_SERVER_LOCAL_AUTH_INITIAL_OWNER_EMAIL user and be presented with the EULA.

sangee2004:
This issue is addressed when testing with the latest buidl from `main`.

When following the steps mentioned in the issue , user is taken to obot UI logegd as the OBOT_SERVER_LOCAL_AUTH_INITIAL_OWNER_EMAIL user and is presented with the EULA as expected.

Issue #7775, opened 2026-09-08
`Finish Later` option is incorrectly available during initial owner activation


sangee2004:
**Describe the bug**
When the initial owner accesses the activation link, the `Set Your Password` modal provides a `Finish Later` option that redirects to the Obot login page. Since the owner has not yet created a password, they cannot log in from this page. The `Finish Later` option should not be available during initial owner activation, and the user must complete password setup.

Pre-req - Have the OBOT server started with the following environment variables set: `-e OBOT_SERVER_ENABLE_AUTHENTICATION=true -e OBOT_SERVER_LOCAL_AUTH_INITIAL_OWNER_EMAIL=sangeetha@obot.ai -e OBOT_SERVER_LOCAL_AUTH_INITIAL_OWNER_SETUP_TOKEN_EXPIRATION_HOURS=1 -e OBOT_SERVER_LOCAL_AUTH_INITIAL_OWNER_SETUP_TOKEN=<token>`

**To Reproduce**
1. Use athe activationlink http://localhost:8080/activate#token=<token>

The user sees the `Set Your Password` modal, which includes a `Finish Later` link that takes the user to the obot login page.

<img width="1070" height="897" alt="Image" src="https://github.com/user-attachments/assets/4a0ae531-006e-4125-8650-6a21982962ac" />

Since the user hasn't set a password yet, they can't log in with valid credentials.

**Expected behavior**
Do not provide the `Finish Later` link when the activation link is accessed.

njhale:
@sangee2004 I ended up keeping `Finish later`; Once my changes are in, clicking `Finish later` will still redirect you to the login page, but the login button will remain disabled and the form will contain a message explaining that activation is required before you can log in. As soon as the activation flow is fully completed, the login button will be enabled and you'll no longer see the "activation required" message.

I figured this was a reasonable compromise since `Finish later` is explicitly mentioned in the ODP as a way to abort activation without manually clearing your browser cookies.

sangee2004:
This issue is addressed when testing with the latest build from `main`


When `Finish Later` option is used , user is now presented with the following Obot Login page that contain a message explaining that activation is required before you can log in.

<img width="537" height="421" alt="Image" src="https://github.com/user-attachments/assets/b20785dc-d649-47b3-93e3-e40413c16e32" />

PR #7790, opened 2026-09-09
fix: initial owner activation flow

njhale:
The server used to redirect a setup session's request for `/activate` to `/change-password?rd=/activate` before the page ran. Only the browser ever sees the `#token=...` fragment, so nobody looked at the token: a valid or invalid reopened link both landed on the password form, authorized by the first visit's cookie, and `rd` recorded `/activate` as a place to return to. Setting the password then bounced back to a bare `/activate`, which reported "This setup link is incomplete." Reopening the link after completion showed "Session Expired".

The server now serves `/activate` to setup sessions too, and the page strips the token before deciding what to do:

| Situation | Result |
| --- | --- |
| Open the link with no session | Token exchanged, then the password page, then the auth providers view |
| Open the link while a setup session exists | Token validated and exchanged for a fresh setup session, then the password page |
| Open a bad link, with or without a setup session | "invalid or expired setup link" |
| Open `/activate` with no link | "Obot requires activation. Open the setup link from your provisioning email to continue." |
| `/activate` once the password is set, or when signed in | Redirected to `/` before the page renders |
| Edit the fragment on `/activate` | The page re-runs |

The Local provider in `GET /api/auth-providers` reports `requiresActivation` until the owner has a password. The login page uses it to disable the Local provider and show the same message, which is where **Finish later** now lands instead of a login form that cannot work.

Addresses:
- https://github.com/obot-platform/obot/issues/7775
- https://github.com/obot-platform/obot/issues/7782
- https://github.com/obot-platform/security-issues/issues/76


pkg/api/handlers/authprovider.go
```diff
@@ -89,6 +101,9 @@ func (ap *AuthProviderHandler) List(req api.Context) error {
 		if authProviderStatus.Staged {
 			authProviderStatus.VerifiedEmail = verifiedEmail
 		}
+		if err := setRequiresActivation(req, a, authProviderStatus); err != nil {
+			return err
+		}
```
njhale, line 106: This will only call the gateway db once per list, for a configured local auth provider. One alternative would be to inject something similar to the license provider into `providers.AuthProviderStatus`, but that seemed like overkill.

njhale:
@claude review

calvinmclean (approved):

ivyjeong13 (approved):
LGTM!

Issue #24, opened 2024-09-18
Invoke should error out when invalid agent or thread id is specified even without having to start a chat.


sangee2004:
commit - `e9bb8ffd`
**Steps to reproduce the problem:**
1. Chat with an agent by providing an invalid thread id in `-t` option.
```
%otto invoke -t abc a1wqggd
Input: hello
2024/09/18 15:07:51 error code 404 (Not Found): Thread.otto.gptscript.ai "abc" not found
```
Notice that there is no error message presented and user is promoted for an input.
Once input is provided, then we are presented with an error message

2.  Chat with an invalid agent id using `otto invoke`
```
% otto invoke tttt          
Input: hello
2024/09/18 15:08:07 error code 404 (Not Found): slug tttt not found
```
Notice that there is no error message presented and user is promoted for an input.
Once input is provided, then we are presented with an error message

**Expected Behavior:**
Invoke should error out when invalid agent or thread id is specified even without having to start a chat.

thedadams:
This is still happening with the latest version of Otto8. I just wanted to point out a technical limitation of this request.

The parsing and "tool gathering" happens at a lower level. That lower level requires that the input be provided. So, the CLI will prompt the user for input and pass that on to the "invoker" for processing. If we wanted to change that, it would either create a weird UX (run the agent first with no input) or we would have to change the way the invoker works.

I would also like to point out that this is the way the UI works. You have to chat with an agent (i.e. provide an input) before you would get such an error.

PR #7798, opened 2026-09-09
Coalesce auth provider group refreshes and back off after failures

cjellick:
## Problem

An Okta org started returning 429s for group lookups, which surfaced to users as a failed login rather than as stale group data.

`ensureGroups` runs inside request authentication, through `UserDecorator.AuthenticateRequest`, so it executes on **every authenticated request**. There is no background job involved. The 10 minute check window is the only thing standing between request volume and the auth provider, and `AuthProviderGroupsLastChecked` is only advanced by a refresh that **succeeds**. That leaves two ways to overload a provider:

1. **Bursts.** There was no deduplication, so every request in flight when the window lapsed called the provider on its own. A single page load fires enough parallel API requests to turn one user's group check into a burst of identical calls.
2. **A self-sustaining retry loop.** Once the provider started failing, the window never advanced, so every subsequent request retried immediately and the provider never got a chance to recover. Authentication failed outright throughout, since the error propagates out of `AuthenticateRequest`. There was no fallback to the groups already in the database, even though the throttled path already reads them.

## Change

**Coalescing.** Concurrent refreshes for one identity now share a single call through `singleflight`, keyed on the ID that is actually sent to the provider so the key and the outbound request are one to one.

**Failure cooldown.** An identity whose refresh just failed is served from the groups already in the database until `groupFailureCooldown` passes. That is the same data the request would have been served a moment earlier, so a rate limited provider degrades group freshness instead of blocking sign in.

Both are per-process. Across replicas each one refreshes an identity at most once per check window, which is a fixed multiple of one replica's calls rather than something that grows with traffic. A cross-replica claim on the identity row would collapse that further, but it would change what `AuthProviderGroupsLastChecked` means and introduce a first-login edge case, which did not seem worth it at this ratio.

**Context detachment.** The shared refresh runs on a context detached from the calling request. It is shared, so a leader whose client disconnects must not cancel it for everyone waiting behind it. That makes an explicit bound necessary, and `fetchGroups` was using `http.DefaultClient`, which has no timeout, unlike the two neighboring provider calls.

**A shorter timeout for this path.** `groupProviderTimeout` is 30 seconds, which suits the two admin group endpoints that already used it but is far too long here now that failing is cheap. The Okta SDK's retry budget is up to two waits of 30 seconds, so under a sustained limit it cannot finish inside 30 seconds anyway; the retry only helps when Okta's per-minute bucket happens to reset inside the timeout. Waiting the full 30 seconds first buys a coin flip, and every request coalesced behind the refresh waits along with it. `fetchGroups` now uses its own client at `userGroupRefreshTimeout` (10s), which keeps the case the retry is good at and cuts the bad case to a third. Obot disconnecting cancels the provider's request context, so the SDK's remaining retries stop rather than burning Okta quota nobody is waiting for.

## Effect

| | Before | After |
|---|---|---|
| Normal load | one provider call per API request | one per user per replica per 10 min |
| Rate limited | one call per request, indefinitely | one per user per replica per minute |
| User impact when limited | authentication fails, indefinitely | one 10s failure per user per minute, cached groups otherwise |

See [the comment below](https://github.com/obot-platform/obot/pull/7798#issuecomment-5608300477) for how the cooldown here and the SDK retry config in obot-platform/enterprise-providers#24 divide the work between them.

## Testing

Three tests in `group_refresh_test.go`. Each was checked against a mutation of the code it covers, not only against the fix:

| Test | Mutation | Result without the fix |
|---|---|---|
| `...CoalescesConcurrentRefreshes` | call `refreshGroups` directly | `requests = 20, want 1` |
| `...BacksOffAfterFailedRefresh` | disable the cooldown check | provider error surfaces instead of cached groups |
| `...OutlivesCancelledLeader` | drop `context.WithoutCancel` | `context canceled` |

Full package passes under `-race`. `go build ./...` and `go vet ./pkg/gateway/...` clean.

One caveat worth a reviewer's eye: the coalescing test holds the provider handler open, waits for the leader to arrive, then sleeps 250ms to let the other callers pile up before releasing. It is the conventional way to test `singleflight` and the window is wide, but it is the one timing dependency in the new tests.

## Related

obot-platform/enterprise-providers#24 fixes the Okta SDK side, where the first 429 was fatal because the SDK was created with no rate limit configuration. The two are independent and each is worth having on its own.

### DIFF
--- a/pkg/gateway/client/client.go
+++ b/pkg/gateway/client/client.go
@@ -12,6 +12,7 @@ import (
 	types2 "github.com/obot-platform/obot/apiclient/types"
 	"github.com/obot-platform/obot/pkg/gateway/db"
 	"github.com/obot-platform/obot/pkg/gateway/types"
+	"golang.org/x/sync/singleflight"
 	"k8s.io/apiserver/pkg/server/options/encryptionconfig"
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -58,6 +59,13 @@ type Client struct {
 	deviceScanCleanupInterval time.Duration
 	deviceScanDeleteBatchSize int
 	mcpOAuthTokenTrigger      func(context.Context, string) error
+
+	// groupRefresh collapses concurrent group fetches for one identity into a single call.
+	// groupBackoff holds identities whose last refresh failed, so a failing provider is left
+	// alone for a cooldown. Both are per-process; see ensureGroups.
+	groupRefresh     singleflight.Group
+	groupBackoffLock sync.Mutex
+	groupBackoff     map[string]time.Time
 }
 
 func New(ctx context.Context, db *db.DB, storageClient kclient.Client, encryptionConfig *encryptionconfig.EncryptionConfiguration, mcpOAuthTokenTrigger func(context.Context, string) error, ownerEmails, adminEmails []string, auditLogPersistenceInterval time.Duration, auditLogBatchSize, auditLogRetentionDays, llmAuditLogRetentionDays, deviceScanRetentionDays int, llmAuditEnabled bool) *Client {
@@ -90,6 +98,7 @@ func New(ctx context.Context, db *db.DB, storageClient kclient.Client, encryptio
 		auditLogDeleteBatchSize:   defaultAuditLogDeleteBatchSize,
 		deviceScanCleanupInterval: defaultDeviceScanCleanupInterval,
 		deviceScanDeleteBatchSize: defaultDeviceScanDeleteBatchSize,
+		groupBackoff:              make(map[string]time.Time),
 	}
 
 	go c.runMCPAuditLogPersistenceLoop(ctx, auditLogPersistenceInterval)
--- a/pkg/gateway/client/group.go
+++ b/pkg/gateway/client/group.go
@@ -39,6 +39,13 @@ const (
 	// client gives up.
 	groupProviderTimeout = 30 * time.Second
 
+	// userGroupRefreshTimeout bounds a group refresh on the authentication path. It is much
+	// shorter than groupProviderTimeout because failing there is cheap: the cooldown then serves
+	// groups from the database, at most one check window old. Waiting longer mostly buys a retry
+	// that only lands if Okta's bucket happens to reset inside the timeout, and every request
+	// coalesced behind the refresh waits along with it.
+	userGroupRefreshTimeout = 10 * time.Second
+
 	// groupFailureCooldown is how long to leave the auth provider alone after a failed refresh.
 	// Only a successful refresh advances the check window, so without this a failing provider is
 	// called again by the very next request.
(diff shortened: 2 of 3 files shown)


cjellick:
## How this cooldown and the SDK retry config fit together

The fix here and the one in obot-platform/enterprise-providers#24 both respond to the same 429s, so it is worth writing down which one is doing what. They act at different layers with different budgets:

```
obot request  ──────────────────────────────────────────────┐
  ensureGroups          cooldown 60s per identity           │
                        check window 10 min per identity    │
    fetchGroups         userGroupProviderClient timeout 10s ┤ obot's budget
      │                                                     │
      └─→ okta-auth-provider                                │
            SDK doWithRetries   MaxRetries 2 (3 attempts)   │
                                each wait min(reset+1, 30s) ┘ provider's budget
              └─→ GET /api/v1/users/{id}/groups
```

The retry works in **seconds**, inside one request. The cooldown works in **a minute**, across requests. The check window works in **ten minutes**. Each catches what the layer below could not.

### When the retry wins

Okta's per-minute buckets reset at the top of the next minute, so `Get429BackoffTime` returns 1 to 61 seconds. If the org tips over shortly before a reset, attempt 1 gets a 429, the SDK waits the few seconds to the reset, attempt 2 succeeds. The user sees a slightly slow request, obot gets real groups, `recordGroupRefresh` is called with a nil error, and the cooldown never engages.

That case was impossible before #24: `MaxRetries` defaulted to 0, so `oktaBackoff.NextBackOff` returned `backoff.Stop` on the first 429 and the SDK performed zero retries.

### When the cooldown takes over

Under sustained limiting the retry cannot finish inside obot's budget, so the cooldown is what carries the system:

```
t=0     leader calls the provider, other requests coalesce behind it
t=0     SDK attempt 1 -> 429, wait min(reset+1, 30)
t=10    obot's client times out. Those requests fail authentication.
        The disconnect cancels the provider's r.Context(), so the SDK
        abandons its remaining retries rather than burning them.
t=10    cooldown starts
t=10-70 every request for this user served from the database. Fast, successful.
t=70    next request probes again
```

Before this PR there was no step after `t=10`: the check window is only advanced by a refresh that succeeds, so every subsequent request retried immediately, forever, and authentication failed the whole time.

### Why the timeout is 10s and not 30s

The cooldown period and Okta's bucket period are both about a minute, so **the cooldown already waits out the reset, outside the request, where nobody is blocked.** That makes waiting inside the request the worse of the two. A longer timeout only helps when the reset lands inside it, and `MaxBackoff` clamping means a longer wait can fire the retry early into a bucket that has not reset, which 429s again and adds a wasted request to a limit already exceeded. Ten seconds keeps the imminent-reset case and stops paying for the rest.

### Why the SDK retry config is still worth having

It is not only serving this path. The same client serves three other Okta-hitting endpoints, none of which has a cooldown: `FetchGroupsByIDs` (8 concurrent lookups, can self-inflict a 429), `FetchGroupPage`, and `BuildGroupMigrationMapping` (walks every group in the org). Those callers tolerate latency far better than an auth request and have nothing else protecting them.

### Known rough edge

`groupBackoff` is keyed per identity, but Okta's rate limit is per org. So when the org is limited, each distinct user whose window lapses gets their own failed probe and their own cooldown, on the order of active users divided by ten per minute. That is bounded and far better than per request, but a per-provider circuit breaker would match what is actually being protected. I left it per identity because a per-provider breaker lets one user's failure suppress refreshes for everyone, which is the riskier default. Worth revisiting if the probe volume turns out to matter.

njhale:
Superseded by https://github.com/obot-platform/obot/pull/7928

PR #7806, opened 2026-09-10
feat: integrate MCP Tester with model proxy

g-linville:
This integrates the MCP Tester with the new model proxy that we are going to serve. It is used only when we have a URL for it (which we do, by default) and can be explicitly disabled by setting the environment variable for it to `""`.

There are also backend routes to allow the admin to toggle it on and off, view the configured status, and view remaining tokens for the day. Ivy will handle the UI for this after this is merged.

This should not be merged until the model proxy itself is deployed.


g-linville:
@claude review

calvinmclean (approved):

chart/templates/internal-configmap.yaml
```diff
@@ -12,7 +12,7 @@ data:
     {{- fail (printf "config.%s must be set under .Values.secret (not .Values.config)" $key) -}}
     {{- end }}
   {{- $value := index $cfg $key }}
-  {{- if (toString $value) }}
+  {{- if or (toString $value) (eq $key "OBOT_SERVER_MODEL_PROXY_URL") }}
```
thedadams, line 16: Why do we need this?

pkg/gateway/llmaudit/response_accumulator.go
```diff
@@ -65,7 +65,7 @@ func responseFormatForRequestPath(requestPath string) responseFormat {
 
 func responseFormatForProvider(modelProvider string) responseFormat {
 	switch modelProvider {
-	case system.OpenAIModelProvider:
+	case system.OpenAIModelProvider, "model-proxy":
```
thedadams: Can we add this as a constant next to the other ones?

pkg/services/config.go
```diff
@@ -95,6 +97,7 @@ type (
 )
 
 type Config struct {
+	ModelProxyURL                                  string   `usage:"MCP Tester model proxy base URL; empty disables fallback" default:"https://model-service.obot.ai" env:"OBOT_SERVER_MODEL_PROXY_URL"`
```
thedadams:
```suggestion
	ModelProxyURL                                  string   `usage:"MCP Tester model proxy base URL; empty disables fallback" default:"https://model-service.obot.ai"`
```

pkg/api/handlers/version.go
```diff
@@ -145,6 +150,25 @@ func (v *VersionHandler) getVersionResponse(ctx context.Context) (map[string]any
...
+	hasValidLicense, err := v.LicenseProvider.HasValidLicense(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// Set after OBOT_SERVER_VERSIONS so capability values cannot be overridden.
+	values["hasModelProvider"] = nil
+	if availability.HasModelProvider != nil {
+		values["hasModelProvider"] = *availability.HasModelProvider
+	}
```
thedadams: Does this work?
```suggestion
	values["hasModelProvider"] = availability.HasModelProvider
```

pkg/api/authz/authz.go
```diff
@@ -33,6 +33,9 @@ var (
 	tunnelPeerResources = newPathMatcher("GET /tunnel/peer")
 
 	adminAndOwnerRules = []string{
+		"GET /api/model-proxy",
+		"PUT /api/model-proxy",
+		"GET /api/model-proxy/usage",
```
thedadams: I like the idea of keep the admin paths here as simple as possible.
```suggestion
		"/api/model-proxy",
		"/api/model-proxy/",
```

pkg/api/handlers/mcptester.go
```diff
@@ -40,20 +42,30 @@ type MCPTesterHandler struct {
...
 		serverResolver: serverResolver,
 		accessHelper:   accessHelper,
 		modelResolver:  modelResolver,
 		serverURL:      serverURL,
 		httpClient:     httpClient,
+		fallbackClient: mcptester.NewFallbackHTTPClient(),
+	}
+
+	if len(fallback) > 0 {
+		h.fallback = fallback[0]
```
thedadams: I think we should do one of two things:
1. Add a comment explaining that if any `MCPTesterFallbackOptions` are provided, then only the first one will be used; or
2. Add another function called `NewMCPTesterHandlerWithFallback`, keep the signature of `NewMCPTesterHandler` the same, but have it call` NewMCPTesterHandlerWithFallback` with an empty options struct.

pkg/api/handlers/mcptester_fallback.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/obot-platform/obot/pkg/gateway/client"
+	"github.com/obot-platform/obot/pkg/mcptester"
+	v1 "github.com/obot-platform/obot/pkg/storage/apis/obot.obot.ai/v1"
+)
+
+var (
+	errMCPTesterLicenseRequired = errors.New("register an Obot license to use MCP Tester without a model provider")
+)
+
+type MCPTesterFallbackOptions struct {
```
thedadams: What does "Fallback" mean here? Does it mean the tester is going to use the model proxy?

pkg/api/handlers/mcptester_fallback.go
```diff
@@ -0,0 +1,98 @@
...
+	if h.fallback.License == nil {
+		return nil, nil, types.NewErrHTTP(http.StatusServiceUnavailable, "installation license is unavailable")
+	}
+
+	licenseKey, err := h.fallback.License.LicenseKey(ctx)
+	if err != nil {
+		return nil, nil, types.NewErrHTTP(http.StatusServiceUnavailable, "installation license is unavailable")
+	}
+
+	if strings.TrimSpace(licenseKey) == "" {
```
thedadams: nit: we don't need to trim here because the license would be invalid if there was extra space.

pkg/api/handlers/modelproxy.go
```diff
@@ -0,0 +1,127 @@
...
+		settings:      settings,
+		configuredURL: configuredURL,
+		responsesURL:  responsesURL,
+		license:       license,
+		client:        mcptester.NewModelProxyUsageHTTPClient(),
+	}
+}
+
+func (h *ModelProxyHandler) Get(req api.Context) error {
+	req.ResponseWriter.Header().Set("Cache-Control", "no-store")
```
thedadams: This is already added to every request to an `/api/` endpoint, so I think we can all the instances in this file.

pkg/api/handlers/modelproxy.go
```diff
@@ -0,0 +1,127 @@
...
+	decoder.DisallowUnknownFields()
+
+	var input types.ModelProxySettingsUpdate
+	if err := decoder.Decode(&input); err != nil || input.Enabled == nil {
+		return types.NewErrHTTP(http.StatusBadRequest, "request must contain an enabled boolean")
+	}
+
+	if err := decoder.Decode(&struct{}{}); !errors.Is(err, io.EOF) {
+		return types.NewErrHTTP(http.StatusBadRequest, "request must contain exactly one JSON object")
+	}
```
thedadams: Do we need to be this careful here? As far as I know, we don't do this anywhere else.

pkg/api/handlers/modelproxy.go
```diff
@@ -0,0 +1,127 @@
...
+
+	ctx, cancel := context.WithTimeout(req.Context(), mcptester.ModelProxyUsageTimeout)
+	defer cancel()
+
+	if h.license == nil {
+		return types.NewErrHTTP(http.StatusServiceUnavailable, "installation license unavailable")
+	}
+
+	key, err := h.license.LicenseKey(ctx)
+	if err != nil || strings.TrimSpace(key) == "" {
```
thedadams: nit:
```suggestion
	if err != nil || key == "" {
```

pkg/api/handlers/modelproxy.go
```diff
@@ -0,0 +1,127 @@
...
+		return types.NewErrHTTP(http.StatusServiceUnavailable, "a valid installation license is required to read model proxy usage")
```
thedadams: A "service unavailable" error makes it sound like there is something wrong with the server. I would suggest a Forbidden error in this case.

pkg/gateway/client/modelproxy.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+// ModelProxyEnabled reads shared state on every call. Only an absent property
+// uses the backwards-compatible default.
+func (c *Client) ModelProxyEnabled(ctx context.Context) (bool, error) {
+	p, err := c.GetProperty(ctx, modelProxyEnabledKey)
+	if errors.Is(err, gorm.ErrRecordNotFound) {
+		return true, nil
+	}
+
```
thedadams: nit: check the error as soon as possible (but not sooner)
```suggestion
```

pkg/gateway/client/modelproxy.go
```diff
@@ -0,0 +1,63 @@
...
+	switch p.Value {
+	case "true":
+		return true, nil
+	case "false":
+		return false, nil
+	default:
+		return false, errors.New("invalid model proxy settings")
+	}
```
thedadams: Can we use `strconv.ParseBool` here?

pkg/gateway/client/modelproxy.go
```diff
@@ -0,0 +1,63 @@
...
+
+// SetModelProxyEnabled uses an atomic upsert so concurrent first writes cannot
+// race to create the singleton. Preserve the existing property encryption path.
+func (c *Client) SetModelProxyEnabled(ctx context.Context, enabled bool) error {
+	now := time.Now()
+	p := types.Property{
+		Key:       modelProxyEnabledKey,
+		Value:     strconv.FormatBool(enabled),
+		CreatedAt: now,
+		UpdatedAt: now,
```
thedadams: I think `gorm` sets these automatically.

pkg/gateway/server/dispatcher/modelconfiguration.go
```diff
@@ -0,0 +1,103 @@
...
+
+	var providers v1.ModelProviderList
+	if err := storage.List(ctx, &providers, kclient.InNamespace(system.DefaultNamespace)); err != nil {
+		return false, fmt.Errorf("read model providers: %w", err)
+	}
+
+	var configured bool
+	for _, provider := range providers.Items {
+		if !provider.DeletionTimestamp.IsZero() {
+			return false, fmt.Errorf("model provider deletion is pending")
```
thedadams: Should we error in this case? Can we just ignore it and move on?

pkg/mcptester/fallback.go
```diff
@@ -0,0 +1,211 @@
...
+	"net/url"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/obot-platform/obot/apiclient/types"
+)
+
+const (
+	DefaultModelProxyURL = "https://model-service.obot.ai"
```
thedadams: Since the configuration has a default set, I wouldn't expect this to be used anywhere. A quick search reveal it is only used in tests, but I could be wrong about that.

pkg/mcptester/fallback.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+	}
+
+	if port := parsed.Port(); port != "" {
+		number, err := strconv.Atoi(port)
+		return err == nil && number > 0 && number <= 65535
+	}
+
+	return !strings.HasSuffix(parsed.Host, ":")
+}
```
thedadams: Do we need this? Seems like a request would fail if this failed for some reason.

pkg/mcptester/usage.go
```diff
@@ -0,0 +1,142 @@
...
+
+	return req, nil
+}
+
+func ReadModelProxyUsage(response *http.Response) (types.ModelProxyUsage, error) {
+	failure := &ModelProxyUsageError{Status: http.StatusBadGateway, Message: "The model proxy returned an invalid usage response."}
+	if response.StatusCode != http.StatusOK {
+		switch response.StatusCode {
+		case http.StatusUnauthorized, http.StatusForbidden:
+			failure.Status = http.StatusServiceUnavailable
```
thedadams: Same comment here about status.

chart/templates/internal-configmap.yaml
g-linville: To support setting an explicitly empty string, to disable model proxy altogether

pkg/api/handlers/mcptester_fallback.go
g-linville: Yes. I'll ask AI to make this clearer

g-linville: I removed the use of "fallback" and made it clear throughout the code that this refers to the model proxy

pkg/api/handlers/modelproxy.go
g-linville: No, we didn't need to do this. Fixed

pkg/api/handlers/modelproxy.go
g-linville: Now we distinguish between an error from trying to get the license key, and an empty license key (the latter returns a 403).

pkg/gateway/server/dispatcher/modelconfiguration.go
g-linville: I changed this so that we continue but we do return an error if all we found was one pending deletion

pkg/mcptester/fallback.go
g-linville: You are correct. The const is now removed.

pkg/mcptester/fallback.go
g-linville: This is simplified now to only what we need

g-linville:
Had to rebase on main; sorry. It removed an import that I needed to restore in order for CI to pass.

thedadams (approved):

pkg/mcptester/modelproxy.go
```diff
@@ -0,0 +1,177 @@
...
+		CheckRedirect: func(*http.Request, []*http.Request) error { return http.ErrUseLastResponse },
+	}
+}
+
+// Preserve the existing values without appending or inferring a client address.
+func copyModelProxyIPHeaders(outbound, inbound http.Header) {
+	for _, name := range []string{"X-Forwarded-For", "X-Real-IP"} {
+		for _, value := range inbound.Values(name) {
+			outbound.Add(name, value)
+		}
```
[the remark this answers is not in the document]
g-linville: This is intentional

pkg/api/handlers/mcptester.go
```diff
@@ -69,46 +80,106 @@ func (h *MCPTesterHandler) Chat(req api.Context) error {
...
+				status, code, message, retryable = http.StatusForbidden, types.MCPTesterErrorLicenseRequired, errMCPTesterLicenseRequired.Error(), false
+			} else if httpErr, ok := errors.AsType[*types.ErrHTTP](err); ok && httpErr.Code == http.StatusBadRequest {
+				status, code, message, retryable = http.StatusBadRequest, types.MCPTesterErrorInvalidRequest, httpErr.Message, false
+			}
+
+			return writeMCPTesterError(req, status, code, message, retryable)
+		}
+
+		outboundClient = h.modelProxyClient
+		audit = gserver.NewTesterAudit(h.modelProxy.GatewayClient, req.Request, req.User, body)
```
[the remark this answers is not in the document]
g-linville: This is intentional

g-linville:
To test this one, set up a fresh Obot instance with no model provider, and create a vMCP and chat with it in the MCP Tester. It should work.

sangee2004:
Able to use `Model Proxy` to chat with vMCP servers using `MCP Tester` when no model provider is configured.

There was 1 issue relating to the Model audit logs reporting the browser user agent instead of `obot-mcp-tester` which is tracked in  https://github.com/obot-platform/obot/issues/7914

PR #7817, opened 2026-09-11
enhance: determine multi-user status by component

thedadams:
The initial implementation of vMCP servers determined the multi-user status for the entire vMCP. That means if there was only 1 piece of configuration out of 10 components, then the whole vMCP would be marked as single-user.

After this change, we determine multi-user status at the component level. So in the above example, only the server that had the one piece of configuration would be marked at single-user. The other components would be multi-user.

An admin still has the ability to mark a component as "force single user" if they wish.


g-linville (approved):
Backend lgtm. In the future, I recommend asking Astra to include "adequate whitespace". I've found that this makes the resulting code a bit easier to read.

calvinmclean (approved):

PR #7832, opened 2026-09-11
fix: retry on setting configuration hash on vMCP create

thedadams:


pkg/api/handlers/vmcp_test.go
```diff
@@ -208,8 +208,20 @@ func TestVMCPHandlerCreateStoresStaticConfigurationInCredential(t *testing.T) {
 		}
 	}
 	published := false
+	updateAttempts := 0
```
njhale, line 211: codex/chatgpt really doesn't like to use zero values

njhale (approved):

pkg/api/handlers/vmcp_test.go
thedadams: Yeah, I go back and fix them from time to time.

Issue #7847, opened 2026-09-12
vMCP deployments are not represented clearly on the Deployments page


sangee2004:
**Describe the bug**
When a vMCP is connected, the `Deployments` page lists only the individual MCP server instances that make up the vMCP and does not show the vMCP deployment itself. The individual MCP server entries also don't indicate which vMCP they belong to, making it hard to understand or manage the deployment relationship. The page should list the vMCP entry, identify its member MCP server deployments, and ideally manage deletion at the vMCP level rather than exposing delete actions for individual member servers.

**To Reproduce**
1. Create a vMCP entry and add a few MCP servers to it.
2. Connect to this vMCP entry 

<img width="935" height="640" alt="Image" src="https://github.com/user-attachments/assets/076b300d-6856-49bc-b7b3-3cbddc7953ce" />

3. Go to the `Deployments` page
Note that the vMCP entry itself isn't listed here. However, the individual MCP servers that make up vMCP are listed here. These individual MCP server entries also don't indicate which vMCP server they belong to.

<img width="941" height="768" alt="Image" src="https://github.com/user-attachments/assets/e53bbb8f-96bb-4d3d-8825-a6e1986594d4" />

**Expected behavior**
1. vMCP entries should be listed
2. MCP server entries should indicate which vMCP they belong to
3. Should we not have a delete option for individual MCP server entries and have it only at the vMCP entry level?

ivyjeong13:
VMCP Deployments will be in the VMCPS navigation as a Deployments tab!

ivyjeong13:
One I just realized is the empty deployments is not centered and will address along with batch of VMCP bugfixes! 

<img width="945" height="473" alt="Image" src="https://github.com/user-attachments/assets/9b81bc5d-9f8b-4085-8872-0745d1221ff5" />

sangee2004:
We now have vMCP->Deployments page which has all vMCP entries listed.

<img width="1414" height="681" alt="Image" src="https://github.com/user-attachments/assets/4e90d475-f1cc-4c69-8657-f85448ea3b65" />
<img width="1130" height="547" alt="Image" src="https://github.com/user-attachments/assets/c77499f9-3adf-4169-a5c3-7db120160c8d" />

`Delete` option is not presented for entries in MCP Servers->Deployments page.

Issue #7820, opened 2026-09-11
Admins are shown unauthorized connection and profile management options for user-created vMCPs


sangee2004:
**Describe the bug**
When a Basic user creates a vMCP, an Admin can see a `Connect` option for that vMCP even though attempting to connect correctly results in a forbidden error. Admins can also access the `Profiles` tab, add profiles to the user-owned vMCP, and automatically assign a default profile. For vMCPs created by Basic users, Admins should not be offered Connect or profile-management actions, and the default profile should not be automatically assigned.

**To Reproduce**
1. Log in as a basic user and create a vMCP - `sannew1` in my case.
2. Log in as an admin user and go to the vMCPs page

Notice that `sannew1` is shown with the `Connect` option.

When connecting to this MCP server as an admin user, you will encounter a `forbidden` error, which is expected.

3. Click on this vMCP entry to get to the details page and go to the `Profiles` tab
Notice that there is a ` default` profile already present, and the admin is allowed to add more profiles to this vMCP server.

<img width="1290" height="298" alt="Image" src="https://github.com/user-attachments/assets/6ecd3f90-e599-496b-8e68-f9295e537e84" />

**Expected behavior**
Admins should not see the `Connect` option for vMCP servers created by basic users.
Admin should not be presented with the option to add profiles to vMCP servers created by basic users.
The default profile should not be assigned to vMCP servers created by basic users.

thedadams:
This has both backend (remove profiles for personal servers) and frontend work.

thedadams:
Backend changes have merged.

sangee2004:
This issue is addressed when testing with the latest build from `main`

1. Admins does not see the `Connect` option for vMCP servers created by basic users.
2. Admin is not presented with `Profiles` option anymore for the  vMCP servers created by basic users.

Issue #7597, opened 2026-08-19
vMCP Experience


ivyjeong13:
There is backend work that needs to happen, but this issue is for the UI/UX portion.

Admins can create vMCPs and share via profiles.
Add ability for users to create vMCPs if they have access to catalog entries.
Add ability for users to connect to vMCP if they've been shared a vMCP via an admin.

ivyjeong13:
Gonna go ahead and move this over but any bugs, please continue to create as separate issues. Thank ya very much!

sangee2004:
The `vMCPs` page is available to admins and regular users, allowing them to create and manage vMCPs.

Issues found when testing this feature are tracked as separate issues.

PR #7855, opened 2026-09-12
enhance: vmcps ux adjustments

ivyjeong13:
Follow-on for #7597

This addresses setting the default profile on vmcp creation to "me", adds a profiles hint helper on first vmcp creation, moves the MCP tester to be a third view called "Tester" in the VMcp, adds loader state for oauth/drag & drop in when adding vmcp components, adds Deployments tab for admins in VMcps, and tweaks some ux around policies.

Addresses #7820 
Addresses #7847


ui/user/src/lib/components/vmcps/VMcpCard.svelte
```diff
@@ -40,21 +38,20 @@
...
+		userID
 	}: Props = $props();
 
-	const roughEstimationText =
-		'This is a rough approximation of the number of tools available. The exact number may vary.';
-
-	let hasFooterContent = $derived(note || tools);
+	let isCreator = $derived(Boolean(userID && profile.current.id === userID));
+	let canDelete = $derived(Boolean(profile.current.isAdmin?.() || isCreator));
+	let canConnect = $derived(!userID || isCreator);
```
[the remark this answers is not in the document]
ivyjeong13: going with leave is for now, otherwise we'd have to determine the profiles subjects could not just be user id, could be 3rd party group, agent identity, etc. admin can just try to connect, see forbidden, and go into profiles to update the rule

apiclient/types/mcpserver.go
```diff
@@ -462,6 +462,11 @@ type MCPServer struct {
 
 	// CompositeName is the name of the composite server that this MCP server is a component of, if there is one.
 	CompositeName string `json:"compositeName,omitempty"`
+
+	// VMCPID owns a shared component server, mutually exclusive with VMCPInstanceID.
+	VMCPID string `json:"vmcpID,omitempty"`
+	// VMCPComponentID identifies the VMCP component whose cached catalog entry was used to create this server.
+	VMCPComponentID string `json:"vmcpComponentID,omitempty"`
```
thedadams, line 471: We should expose the `VMCPInstanceID` so we have it if/when we need it.

apiclient/types/vmcp.go
```diff
@@ -207,15 +207,15 @@ type VMCPInstanceStatus struct {
...
+func (m *VMCPManifest) Default(personalServer bool, userID string) {
 	m.DefaultConfigurationPolicies()
 
 	if personalServer {
 		m.Profiles = nil
 	} else if m.Profiles == nil {
 		m.Profiles = []VMCPProfile{{
 			Name:          "default",
-			Subjects:      []Subject{{Type: SubjectTypeSelector, ID: "*"}},
+			Subjects:      []Subject{{Type: SubjectTypeUser, ID: userID}},
```
thedadams, line 218: Just for my own understanding: does this mean that the default profile will now be just the user who created the server?

ivyjeong13: Yup, walking through with Will, he wanted so it doesn't become automatically available on creation and admin needs to intentionally update the profiles to share with everybody/others! Was hoping this change is small to add here, but lemme know and I can revert if it should be separated out!

apiclient/types/mcpserver.go
ivyjeong13: Can do! :+1:

apiclient/types/mcpserver.go
```diff
@@ -465,6 +465,8 @@ type MCPServer struct {
 
 	// VMCPID owns a shared component server, mutually exclusive with VMCPInstanceID.
 	VMCPID string `json:"vmcpID,omitempty"`
+	// VMCPInstanceID owns a single-user component server, mutually exclusive with VMCPID.
+	VMCPInstanceID string `json:"vmcpInstanceID,omitempty"`
```
thedadams, line 469: Let's set its value in the API, too.

thedadams (approved):

apiclient/types/mcpserver.go
ivyjeong13: Thanks Donnie! Sorry I missed that! Updated and also updated DeploymentViews in MCP Servers to have a link for it if supplied. : )

ui/user/src/lib/components/vmcps/VMcpCard.svelte
```diff
@@ -139,30 +136,21 @@
...
 	<div class="pointer-events-auto relative z-10">
-		<VMcpCardActions {id} {connectURL} {connectButtonId} {onConnect} />
+		<VMcpCardActions
+			{id}
+			{connectURL}
+			{connectButtonId}
+			{onConnect}
+			{hideTest}
+			disabled={!canConnect}
+		/>
```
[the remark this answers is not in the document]
ivyjeong13: no, seems would be confusing why the connect wouldn't be there so disabled and explaining why is better

Issue #7861, opened 2026-09-13
Connecting to a vMCP with user-provided configuration intermittently returns a 409 error


sangee2004:
**Describe the bug**
When a Basic user connects to a vMCP containing `Exa Search` with the configuration set to `Provided at connection`, submitting valid configuration values can intermittently result in a 409 error. Retrying the connection with the same configuration succeeds, and we've observed the issue only once so far. Users should be able to configure and connect to the vMCP without encountering transient 409 errors.

Reproduction - I've seen this issue happen only once so far when connecting to the vMCP server.

**To Reproduce**
1. Log in as admin and create a vMCP with Exa Search, choosing `Provided at connection` for all configs.

<img width="576" height="582" alt="Image" src="https://github.com/user-attachments/assets/69312146-bf0e-4a3c-891a-de54a28d900c" />

2. As an admin user, connect to this MCP server.
3. Log in as a basic user and use the `Test vMCP` option to connect to this server

When the user is prompted to enter config values, enter valid values and click on `Configure`

The user is presented with the following 409 error:

<img width="1132" height="647" alt="Image" src="https://github.com/user-attachments/assets/834d10d9-dcef-41c1-8568-08670fa89ab7" />

On retry, the connection succeeds.

Note: When I tried the same steps to connect to this vMCP server as another user, I did not see the 409 error.

**Expected behavior**
User should not encounter 409 errors when connecting to vMCP servers.

sangee2004:
Not able to reproduce this issue when testing with the latest build from `main`.
I have not encountered 409 errors when conncting to vMP configured with `Provided at connection` for configuration parameters.

PR #7875, opened 2026-09-14
fix(vmcp): retry instance configuration writes on conflict

thedadams:
Configuring or deconfiguring a vMCP instance read the instance and wrote the configuration-sync annotation back without tolerating a concurrent write. The VMCPInstance controllers reconcile tool selection and the user configuration hash right after an instance is created, which is exactly when a user submits connection-time configuration, so the write could lose the optimistic-concurrency check and surface as an intermittent 409.

The credential was already stored at that point, so the failure also left the instance without the annotation that triggers reconciliation of the configuration the user had just supplied.

Issue: https://github.com/obot-platform/obot/issues/7861


pkg/api/handlers/vmcpinstance.go
```diff
@@ -173,27 +174,92 @@ func (*VMCPInstanceHandler) Configure(req api.Context) error {
...
+// annotation, and the OAuth consent flow waits for the status to catch up to
+// it, so it must describe the credential that is actually stored.
+//
+// written is the credential this request just stored. The write is retried on
+// conflict because the VMCPInstance controllers (ReconcileToolSelection,
+// SyncUserConfigurationHash) write to the instance concurrently, bumping the
+// ResourceVersion between the read and the write. A concurrent Configure or
+// Deconfigure can also replace the credential in that window, so whenever the
+// annotation has moved since this request read the instance, the credential
+// that actually won is re-read and its digest recorded instead of written.
```
g-linville: something is off about this comment

g-linville (approved):

calvinmclean (approved):

Issue #7715, opened 2026-09-01
product telemetry: implement scheduling, collecting, and sending metrics


calvinmclean:
## Parent

obot-platform/obot#7693

## What to build

Implement the context-bound product-analytics job in Obot. It starts with the service lifecycle, runs once immediately, then schedules each daily run for 00:05 UTC. When stored or force-enabled consent is true, each run sends a report containing installation ID, UTC report time, distribution, normalized engine, current version, and the product-usage metrics defined by the running release. When consent is missing or false, the job does not make any request to `/product-telemetry`; opting out also disables publishing engine and distribution.

Collect the ODP's initial metrics using the documented definitions. Daily activity and event counts cover the previous full UTC day; inventory metrics are snapshots at collection time. Failures are non-fatal and delivery retries use exponential backoff with the same report identity. Duplicate reports are safe because the receiver deduplicates by installation ID and UTC report date.

`/check-upgrade` remains independent of product-telemetry consent. It is controlled by its existing environment-variable toggle and publishes only the current version, not distribution or engine.

## Acceptance criteria

- [x] The job starts and stops with the service context and does not delay or fail startup.
- [x] After the startup run, the scheduler calculates the next 00:05 UTC boundary and repeats at 00:05 UTC each day without drifting based on the previous run's duration.
- [x] Missing or false consent entirely disables requests to `/product-telemetry`; true or force-enabled consent sends metadata and all metrics defined by the running release.
- [x] Distribution reports `Cloud` when explicitly configured and otherwise reports `Unregistered`, `Registered`, `Cloud`, or `Enterprise` from license state and type.
- [x] Metric definitions, exclusions, previous-full-day UTC boundaries, and snapshot timing match the ODP.
- [x] Collection preserves unavailable versus measured-zero values and never includes prohibited content or custom configuration details.
- [x] Collection or delivery failure is logged without payload contents and retried with bounded exponential backoff using the original report identity.
- [x] `/check-upgrade` remains controlled by its existing environment-variable toggle and publishes only the current version; it does not publish distribution or engine.
- [x] Tests cover the startup run, 00:05 UTC scheduling, consent changes between runs, force-enable behavior, no `/product-telemetry` request when opted out, every collector, cancellation, retries, and non-fatal startup.

## Blocked by

- obot-platform/obot#7711
- obot-platform/obot#7713
- obot-platform/obot#7716

calvinmclean:
[...]

PR #7903, opened 2026-09-15
fix: match the built-in MCP catalog source URL by prefix 

njhale:
A catalog source URL names the repository followed by an optional branch, and
the shipped default pins https://github.com/obot-platform/mcp-catalog/v2-schema. The telemetry collector compared the whole source against github.com/obot-platform/mcp-catalog, so nothing matched: every built-in catalog entry was counted toward
customMCPServerEntryCount and builtInMCPServers was always empty.

Match the built-in catalog by prefix so a pinned branch no longer hides it, and cover the branch spellings in the source URL normalization test, which previously exercised only the bare repository URL.

Follows Up: https://github.com/obot-platform/obot/issues/7715


pkg/producttelemetry/request.go
```diff
@@ -201,6 +201,17 @@ func collectMCPEntryMetrics(ctx context.Context, storageClient kclient.Reader, d
 	return builtIns, customCount, nil
 }
 
+func isBuiltInMCPCatalogSource(sourceURL string) bool {
+	sourceID := strings.TrimSuffix(mcp.SourceIDForURL(sourceURL), ".git")
+	switch sourceID {
+	// Match both the top-level repo and the v2-schema branch
+	case builtInMCPCatalogSourceURL, builtInMCPCatalogSourceURL + "/v2-schema":
```
thedadams: Does it make sense to do prefix matching? Anything "official" will come from a branch of the repo most likely.

g-linville (approved):

pkg/producttelemetry/request.go
```diff
@@ -201,6 +201,12 @@ func collectMCPEntryMetrics(ctx context.Context, storageClient kclient.Reader, d
 	return builtIns, customCount, nil
 }
 
+func isBuiltInMCPCatalogSource(sourceURL string) bool {
+	sourceID := strings.TrimSuffix(mcp.SourceIDForURL(sourceURL), "/v2-schema")
+
+	return strings.TrimSuffix(sourceID, ".git") == builtInMCPCatalogSourceURL
```
thedadams: This is what I was thinking so we can change the branch without needing to change this code.
```suggestion
	return string.HasPrefix(strings.TrimSuffix(sourceID, ".git"), builtInMCPCatalogSourceURL)
```

njhale: yeah, we can just do this, but your suggestion as written also matches `github.com/obot-platform/mcp-catalog<garbage>`. I guess it's not really something we should be concerned about since we control everything in the `obot-platform` org anyway

thedadams: If it's valid,then we should support it here. Ideally, we never have to adjust this code again.

calvinmclean (approved):

pkg/producttelemetry/request.go
```diff
@@ -174,7 +174,7 @@ func collectMCPEntryMetrics(ctx context.Context, storageClient kclient.Reader, d
 	builtIns := make([]clienttypes.ProductTelemetryBuiltInMCPServer, 0)
 	var customCount int64
 	for _, entry := range entries.Items {
-		if strings.TrimSuffix(mcp.SourceIDForURL(entry.Spec.SourceURL), ".git") != builtInMCPCatalogSourceURL {
+		if !strings.HasPrefix(mcp.SourceIDForURL(entry.Spec.SourceURL), builtInMCPCatalogSourceURL) {
```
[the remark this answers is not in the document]
njhale: this is known and accepted.

pkg/producttelemetry/request.go
njhale: matching by prefix dropped the need for trimming the `.git` suffix entirely, so I ended up dropping the helper function in favor of inlining `string.HasPrefix(mcp.SourceIDForURL(sourceURL), builtInMCPCatalogSourceURL))` at the call site.

thedadams (approved):

PR #7905, opened 2026-09-15
docs: describe the staged auth provider switch

njhale:
The switching procedure still told owners to force-enable bootstrap
login, deconfigure the active provider, and promote the new owner by
hand from the Users page. None of that works now: the provider serving
logins can no longer be deconfigured, and the replacement grants Owner
on its own.

Document the staged switch instead, along with the verification window
and how to resume or unstage a switch in progress.


docs/docs/configuration/auth-providers.md
```diff
@@ -327,78 +328,43 @@ This section describes the steps involved in switching authentication providers
...
-2. Authenticate using your Microsoft Entra credentials.
-3. After successful authentication, a user with the default role is created.
-4. Log out.
+All users authenticated via GitHub are immediately logged out and must log in again through Entra. You remain logged in as the Entra account you verified with.
 
-### Step 5: Promote the Entra User to Owner
-1. Log in to the admin console (`<obot-server>/admin`).
-2. Click **Sign in with Bootstrap Token**.
+:::note
+User identities are scoped to the authentication provider used during login. Signing in through Entra creates a new user record, so the work and settings belonging to a user's GitHub account are lost.
```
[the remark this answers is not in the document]
njhale: They are partially cleaned up today (group assignments, etc) and I don't think it's worth calling out here until we flesh that out fully.

docs/docs/configuration/auth-providers.md
```diff
@@ -327,78 +328,43 @@ This section describes the steps involved in switching authentication providers
 
 - Authentication is already enabled.
 - GitHub is configured as the active authentication provider.
-- An initial **Owner** user is already set up.
+- You are logged in as an **Owner**.
+- You are running the Obot Community or Enterprise edition, which include Entra. See [Obot Editions](../enterprise/overview.md) to enable one.
 
-### Step 1: Verify Bootstrap Login is Enabled
+### Step 1: Configure Microsoft Entra
```
calvinmclean, line 334: Any reason for this being specific to Entra, or just used as an example?

njhale: This was the pre-existing example; decided to keep it since it covers the common use case of switching to a provider that requires registration

calvinmclean (approved):

thedadams (approved):

PR #7910, opened 2026-09-15
fix: streamline first local user setup process

g-linville:
With these changes, on a fresh Obot instance, there is a more streamlined setup process for the initial local owner (we skip the confirmation step).


g-linville:
The thing that Copilot mentioned in its suppressed comment cannot be hit on a fresh install, so it is not a concern. It could only be hit by a partially set up installation on a previous version that upgrades to the next.

calvinmclean (approved):

ivyjeong13 (approved):

Issue #7911, opened 2026-09-15
MCP server entries show non-functional Audit Logs and Usage options


sangee2004:
**Describe the bug**
MCP server entries expose a `View Audit Log `action that consistently returns no audit log entries. The `Audit Logs` and `Usage` tabs in the MCP server detail view also return no data. These non-functional options should be removed from MCP server entries and their detail pages.

**To Reproduce**
1. Log in as admin
2. Go to the `MCP Servers` page
Notice that we have a `View Audit Log` option, which always returns no audit log entries.

<img width="1046" height="457" alt="Image" src="https://github.com/user-attachments/assets/e2705110-49e7-4007-8817-5f28431a5dcf" />

<img width="1026" height="579" alt="Image" src="https://github.com/user-attachments/assets/3cfba553-6bca-4489-9421-c1af9de16740" />

3. Select any entry and go to the `Audit Logs` or `Usage` tab

Notice that no entries are returned here.

<img width="1033" height="580" alt="Image" src="https://github.com/user-attachments/assets/f54474a9-81e9-446e-ab10-90676f8c1f9e" />
<img width="1026" height="579" alt="Image" src="https://github.com/user-attachments/assets/b4a5a72d-1358-408d-9ea9-5256c106aab0" />

**Expected behavior**

Remove the `View Audit Log` option for MCP server entries.
Remove the `Audit Logs` and `Usage` tabs that are shown when any MCP server entry is selected.

sangee2004:
This issue is addressed when testing with the latest build from `main`

`View Audit Log` option is not present for MCP server entries.
<img width="1112" height="399" alt="Image" src="https://github.com/user-attachments/assets/256a1b92-1f50-4be5-ae60-87472bf30462" />

`Audit Logs` and `Usage` tabs are not shown when any MCP server entry is selected.
<img width="1107" height="374" alt="Image" src="https://github.com/user-attachments/assets/36e2fe69-c65d-4523-a848-5db86664f79a" />

PR #7912, opened 2026-09-15
enhance(ui): remove audit log entry points from MCP server views

thedadams:
The MCP Servers table showed a "View Audit Logs" row action for every entry, and the single server view carried an "Audit Logs" tab. Both are removed. The per-user "View Audit Logs" links in the Connected Users table linked into that tab, so they are removed as well.

With the audit action gone, the row actions menu was empty for most rows, so the menu is removed entirely. Deleting an entry and configuring static OAuth credentials both remain reachable from the entry detail view. Dropping the menu also orphaned the bulk-delete and single-server delete dialogs, which were already unreachable because the table is never given selection props.

Table rows also get a little more vertical breathing room (48px -> 56px).

Issue: https://github.com/obot-platform/obot/issues/7911


ui/user/src/routes/mcp-servers/EntriesView.svelte
```diff
@@ -374,146 +282,7 @@
...
-								{#if canDelete}
-									<button
-										class="menu-button-destructive"
-										onclick={(e) => {
-											e.stopPropagation();
-											deletingEntry = catalogEntry;
-											toggle(false);
-										}}
-									>
-										<Trash2 class="size-4" />
```
ivyjeong13, line 428: Wanted to double check that we also wanted to remove the ability to delete/bulk delete a catalog entries they own/created from the table?

thedadams: It wasn't being used, so Claude removed it.

ivyjeong13: Ohhh I see, the bulk delete wasn't! But the Delete Entry should still be being used

Image

PR #7930, opened 2026-09-16
fix: ignore vMCP servers and instances in catalog cleanup

thedadams:
vMCP servers and instances were getting cleaned up in the catalog controller based on access control rules. Such servers are meant to exist as long as their vMCP or vMCP instance is meant to exist. A personal vMCP server should be cleaned up based on access control rules (which was already happening).

This change also adds a controller to clean up vMCP instances if the corresponding user no longer matches any profiles.


pkg/controller/handlers/cleanup/user.go
```diff
@@ -136,12 +180,20 @@ func (u *UserCleanup) Cleanup(req router.Request, _ router.Response) error {
 		return err
 	}
 
+	var instancesDeleted int
 	for _, instance := range instances.Items {
+		// Don't delete instances associated to vMCP components because they will be
+		// deleted whent the vMCP instance is deleted.
```
njhale, line 186:
```suggestion
		// deleted when the vMCP instance is deleted.
```

njhale (approved):

calvinmclean (approved):

g-linville (approved):

pkg/controller/handlers/cleanup/user.go
thedadams: I'll fix this in the next PR. Too many changes in flight right now.

PR #7947, opened 2026-09-16
enhance: update vmcps default profile created to admin group

ivyjeong13:
This addresses defaulting the profile of a created vmcp, that isn't personal, to the admin group instead of the creator.


ivyjeong13:
Saw Donnie's PR merged in, closing this one out!

PR #7954, opened 2026-09-17
chore: format tests

calvinmclean:
run `go fix -fixtool="$(which tabletestfmt)" ./...`

Tired of selectively committing tests when I run [the formatter](https://github.com/calvinmclean/go-table-test-formatter) in new branches so this gets the whole repo in the right spot.


g-linville (approved):
very nice

thedadams (approved):

